PR #1, opened 2011-01-17
json-message utf8_encode()

01-Scripts:
Using the source-integration by jreese
jreese/source-integration
I discovered a problem with german umlauts (äöüÄÖÜ) in json reply-messages.
First I wanted to fix the problem in the source-integration but jreese thought an integration in mantisbt would be better:
https://github.com/jreese/source-integration/pull/1#issuecomment-675011

### DIFF
--- a/core/json_api.php
+++ b/core/json_api.php
@@ -36,7 +36,7 @@ require_api( 'url_api.php' );
  */
 function json_url( $p_url, $p_member = null ) {
 	$t_data = url_get( $p_url );
-	$t_json = json_decode( $t_data );
+	$t_json = json_decode( utf8_encode($t_data) );
 
 	if( is_null( $p_member ) ) {
 		return $t_json;


dregad:
Hello @01-Scripts 

I came up on this PR of yours following an [issue report in Mantis bug tracker](https://www.mantisbt.org/bugs/view.php?id=20350), from a user who had issues with broken Japanese characters. 

I was wondering about what your original use case was for having issues with accented characters, because by definition [JSON strings must be Unicode-encoded](https://tools.ietf.org/html/rfc7159#section-8.1).

Based on this, I believe this change was incorrect, and should be reverted. 

Let me know your thoughts, here or directly in https://www.mantisbt.org/bugs/view.php?id=20350.

01-Scripts:
Hello @dregad 

I had the problem with german umlauts (äöüÄÖÜ) in the communication between Mantis and the source-integration plugin as [described here](https://github.com/jreese/source-integration/pull/1#issuecomment-675011).
Also after reading the [RFC 7159](https://tools.ietf.org/html/rfc7159#section-8.1) I thought it is the right way to do an UTF8_encode of the $t_data.

Maybe you can have a look at the  [original discussion](https://github.com/jreese/source-integration/pull/1#issuecomment-675011) and clarify whats the real issue and where and how to fix it.

Thanks & Regards,
Michael

tomop-tg:
> Also after reading the RFC 7159 I thought it is the right way to do an UTF8_encode of the $t_data.

 No, it isn't right way. Do you really read RFC 8159?

* JSON strings sent by server shall be Unicode-encoded.
* utf8_encode() must be passed *ISO-8859-1*-encoded strings.

 You say that breaking UTF-8 strings sent by servers with right implementation is right.

 You shall use UTF-8 encoding in your commit log, or server-side (github?) shall be fix to send Unicode-encoded strings if you get broken strings.

 If you want to resolve in Mantis-side, make url_get() in url_api.php:
* detect encodings with "Content-Type" HTTP header.
* use *iconv()* or *mb_convert_encoding()* to encode with UTF-8 from other encodings.

tomop-tg:
I tried on my environment to get commit logs from the his repository on github.

```shell
curl -i 'https://api.github.com/repos/01-Scripts/01-Artikelsystem/commits?per_page=50' > commits.txt
$ file commits.txt
commits.txt: UTF-8 Unicode text
$ php -r 'var_dump(json_decode(file_get_contents("commits.txt")));'
(snip)
      ["message"]=>
      string(66) "#718 Fehlende UTF-8-Unterstützung für title=-Ausgabe hinzufügen"
(snip)
$ php -v
PHP 5.6.16 (cli) (built: Nov 27 2015 12:21:20)
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2015 Zend Technologies
```
github sends UTF-8 strings and messages with umlauts are not broken.

01-Scripts:
Hello @tomop-tg, 

thanks for the clarification. Now I get it.
Please feel free to revert my change. 

Regards,
Michael 

>

dregad:
@01-Scripts thanks for your feedback, will revert soon.

dregad:
For the record, the change was reverted in 47c512fc2fba30f6840c9eba476f9a2c892b5a53 (1.3.x / master) and 7f4ba27f2e46513bcbdcc2efd03e9c36a5a7a0f4 (1.2.x)

PR #43, opened 2012-02-28
#0006939: Number of Private notes visible for reporters

Dentxinho:
Private notes aren't counted anymore if the user is not allowed to see them.


atrol:
Closed in favor of https://github.com/mantisbt/mantisbt/pull/856

PR #44, opened 2012-02-29
When uploading attachments via soap mantis api chdir to mantis base dir,...

arturaz:
... otherwise file uploads fail.


dregad:
Closing this PR, since there was no feedback or follow up on it for several years.

PR #101, opened 2014-01-13
Notifications support with html emails

vboctor:
The goal of this change is to add notifications support where each notification has all necessary information about the change.  The notification handler can then format this into the appropriate format that matches the channel used to dispatch the notification.  For now, this change only supports html emails.  However, the plan is to add support for text emails, twitter, etc whether they are done as plugins or natively.

The change minimized changes to existing MantisBT code for now (to simplify usage with master-1.2.x / master).  However, once we agree on the approach, I expect further changes around email_api.php cleanup and addition of plugin events.  The change doesn't also include localization.

The changes include the following:
- Notification support for issue add, edit, and delete.
- Notification support for comment add, edit, and delete.
- Email api changes to support html emails and embedding of images.
- HTML templates that allow customization of what is sent out for each event.  Ideally, we use some standard templating format instead of the custom / simple one I included here.  We should also think through the localization aspect.
- Include of "Sent by Mantis Bug Tracker" and timestamp at the footer.  The timestamp makes sure that email clients like gmail don't trim the message assuming it is redundant with previous ones.
- Inclusion of user avatars and MantisBT avatar in the message.
- Emails no longer include a dump of the issue with a message at the top.  However, they just focus on the change and provide the user with a link to open the issue.  For example, when a note is added, the email will indicate that user 'X' has commented on issue 'Y', include user avatar, issue id, title, and comment text.  In addition, there will be a "Reply" link that allows user to respond to such comment.


l266509:
**@vboctor @giallu @daryn @rombert @atrol @grangeway @dregad @Dentxinho @bieli @tomcarlson @syncguru @diegovice @caribe @fager @nikoutel @6RUN0 @AgnidykhBayky @sayyidsh @PixelstreamAU @JohnMS @mikeching @g-taki @benv8nam @stew721 @sayazyi @lezouretalice @SuriyaaKudoIsc @solovets @kacmem @xiaojiangang @chenyongze @myu9882 @markmarco16 @mgretler @lincolnwung @krime @scottmatheina @FavioGalvis @vand3rson @like9729 @badfiles @zenithtekla @SureShinde @netsf @csatwood @miao606 @mdj20130907 @bitlin @ioroid @supertassu @TasCraft @fernandodelucca @ulkuderner @vbathke @hersonpc @jrtorres42 @thejuanwhocode @cyuankai @mantianxingxing @shirleycharlin @boby75 @Lordshell @BurningWind @jaguarxii @Praymundo @johnzastrow @ewmartinez @marcioklein @lypzln @nospam1961 @jllano @lincoln56 @modir @xiaocnie @sivpro @planetbrett @teodortalov @ltpgt @pbcreativesolutions @Gantha @Gerolkae @gunsmoker @nikerabbit @translatewiki  Whats status of this PR?

PR #149, opened 2014-03-13
Add support DejaVu Font http://dejavu-fonts.org/ for MantisBT Graph plugin

Kirill:


dregad:
With the revised graphs in 2.0, this is no longer useful.

PR #157, opened 2014-04-06
Commit to ask for the Target version when an issue is set to a specific ...

MrBricodage:
...status (configurable)

see http://www.mantisbt.org/bugs/view.php?id=16774 for historical solution with Mantis 1.2-15

Commit is Adapted to Mantis 1.2.17


dregad:
Replaced by PR #611

PR #172, opened 2014-05-01
Allow choice of email validation method

dregad:
Following implementation of 50d235ad101f61a6c6888316e827fd225ad4b9cd the
validation of email addresses was done by PHPMailer::validateAddress()
instead of filter_var() with FILTER_SANITIZE_EMAIL to allow RFC5322-
compliant emails like 'user@domain' that are rejected by PHP's method
which only accepts 'user@domain.tld'.

However, these 'top-level-domain-only' addresses are often considered as
invalid by SMTP servers but there is a strong use-case for them in
intranet environments.

To enable both scenarios, this commit introduces new constants for
$g_validate_email, allowing the admin to pick whether they want
validation using the PHP method (default) or strict RFC5322.

Backwards compatibility is maintained because EMAIL_VALIDATE_PHP == ON,
and validation can still be disabled with OFF.

Fixes #16894

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -530,10 +530,33 @@ $g_notify_flags['monitor'] = array(
 $g_email_receive_own = OFF;
 
 /**
- * set to OFF to disable email check
+ * Email addresses validation
+ *
+ * Determines if and how email addresses are validated. Possible values for
+ * this config are:
+ * - EMAIL_VALIDATE_PHP (default): validate using PHP's built-in method (see
+ *   {@link http://php.net/filter_var} with FILTER_SANITIZE_EMAIL filter
+ *   {@link http://php.net/filter.filters.validate}
+ *   This should be used with internet-facing installations.
+ *   Emails must have the form 'user@domain.tld'
+ * - EMAIL_VALIDATE_RFC5322: validate using an RFC5322-compliant regex
+ *   (see {@link http://squiloople.com/2009/12/20/email-address-validation/}.
+ *   This is useful for intranet installations, when relying on addresses
+ *   without a top-level domain (e.g. 'user@domain')
+ * - EMAIL_VALIDATE_AUTO: let PHPMailer's
+ *   {@link http://phpmailer.github.io/PHPMailer/classes/PHPMailer.html#method_validateAddress validateAddress()}
+ *   method pick the best validation pattern. On recent systems, this is
+ *   generally the same as EMAIL_VALIDATE_RFC5322.
+ * - OFF: disable email validation
+ *
+ * NOTE: Regardless of how this option is set, validation is not performed
+ * when using LDAP email (i.e. when $g_use_ldap_email = ON), as we assume that
+ * it is handled by the directory.
+ * @see $g_use_ldap_email
+ *
  * @global integer $g_validate_email
  */
-$g_validate_email = ON;
+$g_validate_email = EMAIL_VALIDATE_PHP;
 
 /**
  * set to OFF to disable email check
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -550,7 +550,7 @@ $g_email_receive_own = OFF;
  *   standard-compliant, with this you can.
  * - EMAIL_VALIDATE_AUTO: let PHPMailer's
  *   {@link http://phpmailer.github.io/PHPMailer/classes/PHPMailer.html#method_validateAddress validateAddress()}
- *   method pick the best validation pattern. As of PHPMailer v5.2.8 on recent
+ *   method pick the best validation pattern. As of PHPMailer v5.2.9, on recent
  *   systems, this is generally the same as EMAIL_VALIDATE_RFC5322.
  * - OFF: disable email validation
  *
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -534,29 +534,38 @@ $g_email_receive_own = OFF;
  *
  * Determines if and how email addresses are validated. Possible values for
  * this config are:
- * - EMAIL_VALIDATE_PHP (default): validate using PHP's built-in method (see
+ * - EMAIL_VALIDATE_HTML5 (default): validate using the pattern given by the
+ *   HTML5 specification for 'email' type form input elements
+ *   {@link http://www.w3.org/TR/html5/forms.html#valid-e-mail-address}
+ *   This is the more versatile and standard option, which should be used in
+ *   most situations.
+ * - EMAIL_VALIDATE_PHP: validate using PHP's built-in method (see
  *   {@link http://php.net/filter_var} with FILTER_SANITIZE_EMAIL filter
  *   {@link http://php.net/filter.filters.validate}
- *   This should be used with internet-facing installations.
- *   Emails must have the form 'user@domain.tld'
- * - EMAIL_VALIDATE_RFC5322: validate using an RFC5322-compliant regex
+ *   This considers addresses like 'user@domain' (i.e. without a top-level
+ *   domain) as invalid.
+ * - EMAIL_VALIDATE_RFC5322: validate using a fully RFC5322-compliant regex
  *   (see {@link http://squiloople.com/2009/12/20/email-address-validation/}.
- *   This is useful for intranet installations, when relying on addresses
- *   without a top-level domain (e.g. 'user@domain')
+ *   This probably overkill in most scenarios, but if you need to be fully
+ *   standard-compliant, with this you can.
  * - EMAIL_VALIDATE_AUTO: let PHPMailer's
  *   {@link http://phpmailer.github.io/PHPMailer/classes/PHPMailer.html#method_validateAddress validateAddress()}
- *   method pick the best validation pattern. On recent systems, this is
- *   generally the same as EMAIL_VALIDATE_RFC5322.
+ *   method pick the best validation pattern. As of PHPMailer v5.2.8 on recent
+ *   systems, this is generally the same as EMAIL_VALIDATE_RFC5322.
  * - OFF: disable email validation
  *
- * NOTE: Regardless of how this option is set, validation is not performed
+ * With internet-facing installations, it is strongly advised to use either
+ * the default HTML5 (recommended) or PHP methods (if you don't need to handle
+ * email addresses without a top-level domain).
+ *
+ * NOTE: Regardless of how this option is set, validation is never performed
  * when using LDAP email (i.e. when $g_use_ldap_email = ON), as we assume that
  * it is handled by the directory.
  * @see $g_use_ldap_email
  *
  * @global integer $g_validate_email
  */
-$g_validate_email = EMAIL_VALIDATE_PHP;
+$g_validate_email = EMAIL_VALIDATE_HTML5;
 
 /**
  * set to OFF to disable email check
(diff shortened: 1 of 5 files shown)


dregad:
I rebased this on top of latest master, and added a new commit to use HTML5-compliant regex in email_regex_simple(). This should be pretty much good to go now, but after re-reading [grangeway's argument](https://github.com/mantisbt/mantisbt/pull/172#issuecomment-45272203),

> My point was about simplifying the email validation not complicating it.
> At the moment, users specify whether to do email validation YES or NO.
> *[snip]*
> *[offering]* a choice between EMAIL_VALIDATE_PHP, EMAIL_VALIDATE_RFC5322 and EMAIL_VALIDATE_HTML5 is over-complicating things for end users. *[snip]* hell, I can't even tell what the difference is between RC5322 and HTML5 validation from the description in the config - the code suggests that RFC5322 is needed to allow foo@localhost, but this passes HTML5 validation too.

I'm having second thoughts, and wonder if I'm maybe over-engineering this.

Your comments are welcome.

config_defaults_inc.php
```diff
@@ -512,10 +512,42 @@
...
+ *   This is the more versatile and standard option, which should be used in
+ *   most situations.
+ * - EMAIL_VALIDATE_PHP: validate using PHP's built-in method (see
+ *   {@link http://php.net/filter_var} with FILTER_SANITIZE_EMAIL filter
+ *   {@link http://php.net/filter.filters.validate}
+ *   This considers addresses like 'user@domain' (i.e. without a top-level
+ *   domain) as invalid.
+ * - EMAIL_VALIDATE_RFC5322: validate using a fully RFC5322-compliant regex
+ *   (see {@link http://squiloople.com/2009/12/20/email-address-validation/}.
+ *   This probably overkill in most scenarios, but if you need to be fully
```
vboctor: What would I lose by being fully compliance.  overkill in terms of what?

config_defaults_inc.php
```diff
@@ -512,10 +512,42 @@
...
+ * this config are:
+ * - EMAIL_VALIDATE_HTML5 (default): validate using the pattern given by the
+ *   HTML5 specification for 'email' type form input elements
+ *   {@link http://www.w3.org/TR/html5/forms.html#valid-e-mail-address}
+ *   This is the more versatile and standard option, which should be used in
+ *   most situations.
+ * - EMAIL_VALIDATE_PHP: validate using PHP's built-in method (see
+ *   {@link http://php.net/filter_var} with FILTER_SANITIZE_EMAIL filter
+ *   {@link http://php.net/filter.filters.validate}
+ *   This considers addresses like 'user@domain' (i.e. without a top-level
```
vboctor: I thought I read a comment where you said HTML5 accepts user@domain as well?

config_defaults_inc.php
```diff
@@ -512,10 +512,42 @@
...
+ *   domain) as invalid.
+ * - EMAIL_VALIDATE_RFC5322: validate using a fully RFC5322-compliant regex
+ *   (see {@link http://squiloople.com/2009/12/20/email-address-validation/}.
+ *   This probably overkill in most scenarios, but if you need to be fully
+ *   standard-compliant, with this you can.
+ * - EMAIL_VALIDATE_AUTO: let PHPMailer's
```
vboctor: I'm not sure I see the value of AUTO specially that it is not the default value.

config_defaults_inc.php
```diff
@@ -512,10 +512,42 @@
...
+ *   {@link http://phpmailer.github.io/PHPMailer/classes/PHPMailer.html#method_validateAddress validateAddress()}
+ *   method pick the best validation pattern. As of PHPMailer v5.2.9, on recent
+ *   systems, this is generally the same as EMAIL_VALIDATE_RFC5322.
+ * - OFF: disable email validation
+ *
+ * With internet-facing installations, it is strongly advised to use either
+ * the default HTML5 (recommended) or PHP methods (if you don't need to handle
+ * email addresses without a top-level domain).
```
vboctor: I recall that we use to create administrator account as administrator@localhost email.  I assume this will be fine until the user attempts to update administrator user, right?  Should we default to empty email address since it is not valid anyway?

config_defaults_inc.php
```diff
@@ -512,10 +512,42 @@
...
+ *
+ * NOTE: Regardless of how this option is set, validation is never performed
+ * when using LDAP email (i.e. when $g_use_ldap_email = ON), as we assume that
+ * it is handled by the directory.
+ * @see $g_use_ldap_email
+ *
  * @global integer $g_validate_email
  */
-$g_validate_email = ON;
+$g_validate_email = EMAIL_VALIDATE_HTML5;
```
vboctor: I wonder if we should deprecate and create a new config option or make EMAIL_VALIDATE_HTML5 have the same value as ON which was the previous default.

core/email_api.php
```diff
@@ -84,24 +84,19 @@
...
+ *
+ * Uses the standard HTML5 pattern defined in
+ * {@link http://www.w3.org/TR/html5/forms.html#valid-e-mail-address}
+ * Note: the original regex from the spec has been modified to
+ * - escape the '/' in the first character class definition
+ * - remove the '^' and '$' anchors to allow matching anywhere in a string
+ *
  * @return string
  */
 function email_regex_simple() {
```
vboctor: Is this being used?  Why doesn't its name reflect the validation type?  Aren't we using PHPMailer for validation?

vboctor:
Given that we are not doing rocket since here, and the real goal is some sanity checking on the email address.  Why don't we leave the configuration as is, but for the implementation we say that an email address is valid if any of the validates say it is.

```
$t_valid_email = html5_valid( $t_email ) || rfc_valid( $t_email ) || xyz_valid( $t_email );
```

This way the user doesn't have to make any choices, implementation details are hidden from user, etc.

I think this really can be two checkins:
- Upgrade phpmailer 
- Implement validation changes.

config_defaults_inc.php
dregad: I did this considering the case of instances that may have set $g_validate_email to 'ON' in their config_inc.php, and *may* want to keep the current behavior. Most likely the lack of support of addresses without tld has never been an issue for them.

Those who left things as default will automatically switch to the HTML5 validation.

I don't see the benefit from the switch to a new config. If you believe we should push more strongly for HTML5, I would rather change the values for the constants definition (i.e. have EMAIL_VALIDATE_HTML5 = 1).

core/email_api.php
dregad: PHPmailer is indeed used for validation.

This is used in string API to parse the text (e.g. a bugnote) and replace the e-mail address with a mailto link.

dregad:
It probably should be 2 checkins indeed, but since I went through 3 upgrades of the library in this PR's lifecycle and there are some dependencies between them, I did not want to go through the hassle of the merge conflicts resolution (call me lazy ;-) )

As for your suggestion to check for all of them, consider that they are superseding each other, i.e. `RFC > HTML5 > PHP`. In other words, a PHP-valid address is by definition valid for the other 2 (but the reverse is not true). Therefore, such a test would be useless, we could simply use the RFC method.

I believe it's best to use PHPmailer's HTML5 validation, and use the HTML5 regex internally for parsing strings, so we are consistent in our handling of email addresses - and spare a few CPU cycles too... the HTML5 regex is much more simple, compare it with the RFC one, which is a monster:

```
'/^[a-zA-Z0-9.!#$%&\'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}' .
'[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/sD',
```
vs
```
'/^(?!(?>(?1)"?(?>\\\[ -~]|[^"])"?(?1)){255,})(?!(?>(?1)"?(?>\\\[ -~]|[^"])"?(?1)){65,}@)' .
'((?>(?>(?>((?>(?>(?>\x0D\x0A)?[\t ])+|(?>[\t ]*\x0D\x0A)?[\t ]+)?)(\((?>(?2)' .
'(?>[\x01-\x08\x0B\x0C\x0E-\'*-\[\]-\x7F]|\\\[\x00-\x7F]|(?3)))*(?2)\)))+(?2))|(?2))?)' .
'([!#-\'*+\/-9=?^-~-]+|"(?>(?2)(?>[\x01-\x08\x0B\x0C\x0E-!#-\[\]-\x7F]|\\\[\x00-\x7F]))*' .
'(?2)")(?>(?1)\.(?1)(?4))*(?1)@(?!(?1)[a-z0-9-]{64,})(?1)(?>([a-z0-9](?>[a-z0-9-]*[a-z0-9])?)' .
'(?>(?1)\.(?!(?1)[a-z0-9-]{64,})(?1)(?5)){0,126}|\[(?:(?>IPv6:(?>([a-f0-9]{1,4})(?>:(?6)){7}' .
'|(?!(?:.*[a-f0-9][:\]]){8,})((?6)(?>:(?6)){0,6})?::(?7)?))|(?>(?>IPv6:(?>(?6)(?>:(?6)){5}:' .
'|(?!(?:.*[a-f0-9]:){6,})(?8)?::(?>((?6)(?>:(?6)){0,4}):)?))?(25[0-5]|2[0-4][0-9]|1[0-9]{2}' .
'|[1-9]?[0-9])(?>\.(?9)){3}))\])(?1)$/isD'
```

core/email_api.php
vboctor: Got it, maybe the documentation for the method can clarify that.

vboctor:
I'm trying to avoid the complexity around defining three models and requiring users to understand the differences between them.  How about the following model:

- Email validation ON/OFF remains as is.
- Add email validation regex that is set to html5 and is used in both validation and hyperlinking.
- In the documentation for regex, we can provide links to alternative regex that users can leverage if needed.

Thoughts?

dregad:
Yeah, I guess that's pretty much what I had in the back of my mind when I mentioned my over-engineering this thing. I'll update the branch accordingly (but not tonight, it's time for bed). Thanks for your feedback.

dregad:
With this, we're going back to a simpler approach as per @vboctor's suggestion in https://github.com/mantisbt/mantisbt/pull/172#issuecomment-89863348

core/email_api.php
```diff
@@ -140,6 +147,7 @@ function email_is_valid( $p_email ) {
 					return true; # no need to check mx record details (below) if we've explicity allowed the domain
 				}
 			}
+			log_event( LOG_EMAIL, "failed - not in limited domains list '$t_limit_email_domains'" );
```
vboctor: When we error out for a domain, shouldn't the message include the address that was considered invalid?

core/email_api.php
```diff
@@ -156,11 +164,14 @@ function email_is_valid( $p_email ) {
 				if( checkdnsrr( $t_host, 'ANY' ) ) {
 					return true;
 				}
+				log_event( LOG_EMAIL, "failed - mx/dns record check" );
```
vboctor: Include the failed email address.

vboctor:
:+1:

core/email_api.php
dregad: It's logged right above, see line 138

core/email_api.php
dregad: See https://github.com/mantisbt/mantisbt/pull/172/files#discussion_r28646482

dregad:
Quite happy to finally close the loop on this one :-)

PR #523, opened 2014-10-25
Fix install page when database is not created yet

syncguru:
This issue results in a broken install page when attempting fresh install of mantisbt with these two conditions:

- config_inc.php exists
- Database is not created yet

Fixes #17805


dregad:
@syncguru can we close this ?

syncguru:
Yes.

PR #555, opened 2014-12-04
Fix [ More Events ... ] link in timeline box

syncguru:
The problem was that the timeline entry div is going over the link thus lending it un-clickable. This fix sets a specific entry height and prevent overlapping.

Fixes #17918


dregad:
@syncguru any updates on this ? TIA

syncguru:
@dregad I just updated this PR with a better fix for this issue. Please review.

dregad:
Merged in 9b5fbe57c63c10cf5831dc9ee9e04bf8f6097dd9

PR #560, opened 2014-12-10
[discuss] SOAP API: add method for searching within issues

mhabrnal:
Add method 'mc_search_issues' to SOAP API. Filter is in JSON format.
It is also possible to filter by custom fields.

The example of filter with 'os' == "CentOS" and custom field 'abrt_hash' == "hash_content":

{"abrt_hash":"hash_content", "os":"CenstOS"}

Signed-off-by: Matej Habrnal <mhabrnal@redhat.com>

### DIFF
--- a/api/soap/mantisconnect.wsdl
+++ b/api/soap/mantisconnect.wsdl
@@ -268,12 +268,12 @@
  </xsd:complexType>
  <xsd:complexType name="FilterSearchData">
   <xsd:all>
-   <xsd:element name="project_id" type="tns:IntegerArray" minOccurs="1"/>
+   <xsd:element name="project_id" type="tns:IntegerArray" minOccurs="0"/>
    <xsd:element name="search" type="xsd:string" minOccurs="0"/>
    <xsd:element name="category" type="tns:StringArray" minOccurs="0"/>
    <xsd:element name="severity_id" type="tns:IntegerArray" minOccurs="0"/>
    <xsd:element name="status_id" type="tns:IntegerArray" minOccurs="0"/>
-   <xsd:element name="priority_id" type="tns:StringArray" minOccurs="0"/>
+   <xsd:element name="priority_id" type="tns:IntegerArray" minOccurs="0"/>
    <xsd:element name="reporter_id" type="tns:IntegerArray" minOccurs="0"/>
    <xsd:element name="handler_id" type="tns:IntegerArray" minOccurs="0"/>
    <xsd:element name="note_user_id" type="tns:IntegerArray" minOccurs="0"/>
@@ -299,8 +299,6 @@
    <xsd:element name="tag_string" type="tns:StringArray" minOccurs="0"/>
    <xsd:element name="tag_select" type="tns:IntegerArray" minOccurs="0"/>
    <xsd:element name="custom_fields" type="tns:FilterCustomFieldArray" minOccurs="0"/>
-   <xsd:element name="page_number" type="xsd:integer" minOccurs="1"/>
-   <xsd:element name="issues_per_page" type="xsd:integer" minOccurs="1"/>
   </xsd:all>
  </xsd:complexType>
  <xsd:complexType name="FilterDataArray">
@@ -334,9 +332,8 @@
  </xsd:complexType>
  <xsd:complexType name="FilterCustomField">
   <xsd:all>
-   <xsd:element name="name" type="xsd:string" minOccurs="0"/>
-   <xsd:element name="id" type="xsd:integer" minOccurs="0"/>
-   <xsd:element name="value" type="tns:StringArray" minOccurs="0"/>
+   <xsd:element name="field" type="tns:ObjectRef" minOccurs="0"/>
+   <xsd:element name="value" type="tns:StringArray" minOccurs="1"/>
   </xsd:all>
  </xsd:complexType>
  <xsd:complexType name="FilterCustomFieldArray">
@@ -808,13 +805,17 @@
 <message name="mc_filter_search_issue_headersRequest">
   <part name="username" type="xsd:string" />
   <part name="password" type="xsd:string" />
-  <part name="filter" type="tns:FilterSearchData" /></message>
+  <part name="filter" type="tns:FilterSearchData" />
+  <part name="page_number" type="xsd:integer" />
+  <part name="per_page" type="xsd:integer" /></message>
 <message name="mc_filter_search_issue_headersResponse">
   <part name="return" type="tns:IssueHeaderDataArray" /></message>
 <message name="mc_filter_search_issuesRequest">
   <part name="username" type="xsd:string" />
   <part name="password" type="xsd:string" />
-  <part name="filter" type="tns:FilterSearchData" /></message>
+  <part name="filter" type="tns:FilterSearchData" />
+  <part name="page_number" type="xsd:integer" />
+  <part name="per_page" type="xsd:integer" /></message>
 <message name="mc_filter_search_issuesResponse">
   <part name="return" type="tns:IssueDataArray" /></message>
 <message name="mc_config_get_stringRequest">
--- a/api/soap/mantisconnect.wsdl
+++ b/api/soap/mantisconnect.wsdl
@@ -510,6 +510,18 @@
   <part name="issue_id" type="xsd:integer" /></message>
 <message name="mc_issue_getResponse">
   <part name="return" type="tns:IssueData" /></message>
+<message name="mc_issues_getRequest">
+  <part name="username" type="xsd:string" />
+  <part name="password" type="xsd:string" />
+  <part name="issue_ids" type="tns:IntegerArray" /></message>
+<message name="mc_issues_getResponse">
+  <part name="return" type="tns:IssueDataArray" /></message>
+<message name="mc_issues_get_headerRequest">
+  <part name="username" type="xsd:string" />
+  <part name="password" type="xsd:string" />
+  <part name="issue_ids" type="tns:IntegerArray" /></message>
+<message name="mc_issues_get_headerResponse">
+  <part name="return" type="tns:IssueHeaderDataArray" /></message>
 <message name="mc_issue_get_historyRequest">
   <part name="username" type="xsd:string" />
   <part name="password" type="xsd:string" />
@@ -818,6 +830,14 @@
   <part name="per_page" type="xsd:integer" /></message>
 <message name="mc_filter_search_issuesResponse">
   <part name="return" type="tns:IssueDataArray" /></message>
+<message name="mc_filter_search_issue_idsRequest">
+  <part name="username" type="xsd:string" />
+  <part name="password" type="xsd:string" />
+  <part name="filter" type="tns:FilterSearchData" />
+  <part name="page_number" type="xsd:integer" />
+  <part name="per_page" type="xsd:integer" /></message>
+<message name="mc_filter_search_issue_idsResponse">
+  <part name="return" type="tns:IntegerArray" /></message>
 <message name="mc_config_get_stringRequest">
   <part name="username" type="xsd:string" />
   <part name="password" type="xsd:string" />
@@ -949,6 +969,16 @@
     <documentation>Get the issue with the specified id.</documentation>
     <input message="tns:mc_issue_getRequest"/>
     <output message="tns:mc_issue_getResponse"/>
+  </operation>
+  <operation name="mc_issues_get">
+    <documentation>Get all issues matching the ids.</documentation>
+    <input message="tns:mc_issues_getRequest"/>
+    <output message="tns:mc_issues_getResponse"/>
+  </operation>
+  <operation name="mc_issues_get_header">
+    <documentation>Get all issues header matching the ids.</documentation>
+    <input message="tns:mc_issues_get_headerRequest"/>
+    <output message="tns:mc_issues_get_headerResponse"/>
   </operation>
   <operation name="mc_issue_get_history">
     <documentation>Get the history of the issue with the specified id.</documentation>
@@ -1175,6 +1205,11 @@
     <input message="tns:mc_filter_search_issuesRequest"/>
     <output message="tns:mc_filter_search_issuesResponse"/>
   </operation>
+  <operation name="mc_filter_search_issue_ids">
+    <documentation>Get the issue ids that match the custom filter and paging details.</documentation>
+    <input message="tns:mc_filter_search_issue_idsRequest"/>
+    <output message="tns:mc_filter_search_issue_idsResponse"/>
+  </operation>
   <operation name="mc_config_get_string">
     <documentation>Get the value for the specified configuration variable.</documentation>
     <input message="tns:mc_config_get_stringRequest"/>
@@ -1295,6 +1330,16 @@
   </operation>
   <operation name="mc_issue_get">
     <soap:operation soapAction="http://www.mantisbt.org/bugs/api/soap/mantisconnect.php/mc_issue_get" style="rpc"/>
+    <input><soap:body use="encoded" namespace="http://futureware.biz/mantisconnect" encodingStyle="http://schemas.xmlsoap.org/soap/encoding/"/></input>
+    <output><soap:body use="encoded" namespace="http://futureware.biz/mantisconnect" encodingStyle="http://schemas.xmlsoap.org/soap/encoding/"/></output>
+  </operation>
+  <operation name="mc_issues_get">
+    <soap:operation soapAction="http://www.mantisbt.org/bugs/api/soap/mantisconnect.php/mc_issues_get" style="rpc"/>
+    <input><soap:body use="encoded" namespace="http://futureware.biz/mantisconnect" encodingStyle="http://schemas.xmlsoap.org/soap/encoding/"/></input>
+    <output><soap:body use="encoded" namespace="http://futureware.biz/mantisconnect" encodingStyle="http://schemas.xmlsoap.org/soap/encoding/"/></output>
+  </operation>
+  <operation name="mc_issues_get_header">
+    <soap:operation soapAction="http://www.mantisbt.org/bugs/api/soap/mantisconnect.php/mc_issues_get_header" style="rpc"/>
     <input><soap:body use="encoded" namespace="http://futureware.biz/mantisconnect" encodingStyle="http://schemas.xmlsoap.org/soap/encoding/"/></input>
     <output><soap:body use="encoded" namespace="http://futureware.biz/mantisconnect" encodingStyle="http://schemas.xmlsoap.org/soap/encoding/"/></output>
   </operation>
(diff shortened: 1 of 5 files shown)


mhabrnal:
Hi @vboctor, @rombert, The first version of the abrt mantis reporter is finished ``https://github.com/abrt/libreport/pull/313``, so I had a time to work on the feedback from you and i have created a new version of the searching SOAP API method.

api/soap/mantisconnect.wsdl
```diff
@@ -259,6 +266,43 @@
...
   </xsd:all>
  </xsd:complexType>
+ <xsd:complexType name="FilterSearchData">
+  <xsd:all>
+   <xsd:element name="project_id" type="tns:IntegerArray" minOccurs="1"/>
+   <xsd:element name="search" type="xsd:string" minOccurs="0"/>
+   <xsd:element name="category" type="tns:StringArray" minOccurs="0"/>
+   <xsd:element name="severity_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="status_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="priority_id" type="tns:StringArray" minOccurs="0"/>
```
vboctor: Why is priority_id a StringArray?

api/soap/mantisconnect.wsdl
```diff
@@ -259,6 +266,43 @@
...
+   <xsd:element name="start_day" type="xsd:integer" minOccurs="0"/>
+   <xsd:element name="start_month" type="xsd:integer" minOccurs="0"/>
+   <xsd:element name="start_year" type="xsd:integer" minOccurs="0"/>
+   <xsd:element name="end_day" type="xsd:integer" minOccurs="0"/>
+   <xsd:element name="end_month" type="xsd:integer" minOccurs="0"/>
+   <xsd:element name="end_year" type="xsd:integer" minOccurs="0"/>
+   <xsd:element name="tag_string" type="tns:StringArray" minOccurs="0"/>
+   <xsd:element name="tag_select" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="custom_fields" type="tns:FilterCustomFieldArray" minOccurs="0"/>
+   <xsd:element name="page_number" type="xsd:integer" minOccurs="1"/>
```
vboctor: Technically speaking the page_number and issues_per_page are not part of the filter definition, but are part of a search request.  I wonder if we should have a FilterDefinition and SearchRequest.  A FilterSearchData can reference a filter id or a filter definition + page size / number.  This makes the filter definition re-usable with other methods to add/get filters for example.

api/soap/mantisconnect.wsdl
```diff
@@ -259,6 +266,43 @@
    <xsd:element name="url" type="xsd:string" minOccurs="0"/>
   </xsd:all>
  </xsd:complexType>
+ <xsd:complexType name="FilterSearchData">
+  <xsd:all>
+   <xsd:element name="project_id" type="tns:IntegerArray" minOccurs="1"/>
```
vboctor: I don't think we should make fields like project_id, issues_per_page and page_number required.  Each one has a sensible default value: project_id should default to 0 (for all projects), issues_per_project to the configuration option that determines number of issues on View Issues page, page number to first page (I assume we are 1-based).

api/soap/mantisconnect.wsdl
```diff
@@ -288,6 +332,20 @@
    <xsd:element name="require_closed" type="xsd:boolean" minOccurs="0"/>
   </xsd:all>
  </xsd:complexType>
+ <xsd:complexType name="FilterCustomField">
+  <xsd:all>
+   <xsd:element name="name" type="xsd:string" minOccurs="0"/>
+   <xsd:element name="id" type="xsd:integer" minOccurs="0"/>
+   <xsd:element name="value" type="tns:StringArray" minOccurs="0"/>
```
vboctor: Should we require the value for custom field?  Otherwise, why would the record exist?

api/soap/mc_filter_api.php
```diff
@@ -23,6 +23,50 @@
...
+# Path to MantisBT is assumed to be the grand parent directory.  If this is not
+# the case, then this variable should be set to the MantisBT path.
+# This can not be a configuration option, then MantisConnect configuration
+# needs MantisBT to be included first to make use of the constants and possibly
+# configuration defined in MantisBT.
+
+require_api( 'filter_constants_inc.php' );
+
+// doesn't contain 'custom_fields'
+$_SOAP_API_TO_FILTER_API_NAMES = array (
```
vboctor: Typically we use $g_some_normal_name

api/soap/mc_filter_api.php
```diff
@@ -147,3 +191,165 @@ function mc_filter_get_issue_headers( $p_username, $p_password, $p_project_id, $
...
+    }
+
+    $t_filter = array( '_view_type' => 'advanced' );
+
+    // default fields
+    foreach( $_SOAP_API_TO_FILTER_API_NAMES as $t_soap_name => $t_filter_name ) {
+        if ( isset ( $p_filter_search[$t_soap_name]) ) {
+
+            $t_value = $p_filter_search[$t_soap_name];
+            error_log('Simple extraction, value is ' . print_r($t_value, true) . ' .' );
```
vboctor: I would drop these error_logs in favor of log_event() calls (see core/logging_api.php).  This directs the log to MantisBT log files and have the log entries visibility controlled by $g_logging_level (see config_defaults_inc.php) having LOG_FILTERING set.

api/soap/mc_filter_api.php
```diff
@@ -147,3 +191,165 @@ function mc_filter_get_issue_headers( $p_username, $p_password, $p_project_id, $
...
+
+/**
+ * Get all issues matching the custom filter.
+ *
+ * @param string                $p_username         The name of the user trying to access the filters.
+ * @param string                $p_password         The password of the user.
+ * @param FilterSearchData      $p_filter_search    The custom filter.
+ * @return array that represents an IssueDataArray structure
+ */
+function mc_filter_search_issues( $p_username, $p_password, $p_filter_search ) {
```
vboctor: Looks like we have too much code duplication between the two methods.  How about the following:
- Adding mc_issue_get_many( array(ids) ) that returns an array of the same element type as mc_issue_get().
- Using the output of the issue_search_headers() to provide the input for mc_issue_get_many().

Suggests for better name for mc_issue_get_many() is certainly welcome!

In some cases, this may mean an extra round trip, but it is useful to have our methods usable in more than one scenario.  We can even go to the model where we just return ids from the search and then have methods than return array of headers or array of full issues.

api/soap/mc_filter_api.php
```diff
@@ -147,3 +191,165 @@ function mc_filter_get_issue_headers( $p_username, $p_password, $p_project_id, $
...
+    }
+
+    // object to array
+    if( is_object($p_filter_search) ) {
+        $p_filter_search = get_object_vars($p_filter_search);
+    }
+
+    $t_project_id = $p_filter_search['project_id'];
+
+    if( !mci_has_readonly_access( $t_user_id, $t_project_id  ) ) {
```
vboctor: We are checking access against a single project here.  We are unlikely to disclose projects, but it is likely that APIs will fail with some html that will break soap.  We should run this validation on all the project ids supplied and fail if the list contains any non-accessible projects.  Another option is to filter the project_ids based on access and just pass to the filters the ones that the user has access to.

api/soap/mantisconnect.wsdl
```diff
@@ -288,6 +332,20 @@
    <xsd:element name="require_closed" type="xsd:boolean" minOccurs="0"/>
   </xsd:all>
  </xsd:complexType>
+ <xsd:complexType name="FilterCustomField">
+  <xsd:all>
+   <xsd:element name="name" type="xsd:string" minOccurs="0"/>
```
vboctor: We typically use ObjectRef for cases where a user can supply an id or a name to reference an entity.

vboctor:
@mhabrnal this is so close.  I've done a more thorough and think this is almost ready to go in.  I would love to see this make it into the 1.3 release.  So let's address these comments and get it in.  It will be a great addition.

mhabrnal:
@vboctor I made adjustments according to your comments. I hope it will be according to your suggestions.

vboctor:
@mhabrnal how about the suggestion to have the following methods:

- mc_filter_search() - takes in the filter critera, page, # per page and returns IntegerArray of issue ids.
- mc_issues_get() - given an integer array returns array of issue data.
- mc_issues_get_header() - given an integer array returns array of issue header data.

@rombert , @dregad - any other thoughts?

Thanks.

api/soap/mc_filter_api.php
```diff
@@ -147,3 +190,145 @@ function mc_filter_get_issue_headers( $p_username, $p_password, $p_project_id, $
...
+ * @param integer $p_per_page    Number of issues to display per page.
+ * @return array that represents an IssueDataArray structure
+ */
+function mc_filter_search_issues( $p_username, $p_password, $p_filter_search, $p_page_number, $p_per_page ) {
+
+    $t_rows = mci_filter_search_get_rows( $p_username, $p_password, $p_filter_search, $p_page_number, $p_per_page);
+
+    $t_result = array();
+    foreach( $t_rows as $t_issue_data ) {
+        $t_result[] = mci_issue_data_as_array( $t_issue_data, $t_user_id, $t_lang );
```
atrol: $t_user_id and $t_lang are not set

mhabrnal: Thank you for the comment, you are right.

mhabrnal:
@vboctor for me is better to have more methods which do all i need than call more methods consecutively. So what about add another method mc_filter_search_issue_ids?

vboctor:
So in addition to the ones you have, add the following:

- mc_filter_search_issue_ids() - returns the array of issue ids.
- mc_issues_get() - given an integer array returns array of issue data.
- mc_issues_get_header() - given an integer array returns array of issue header data.

mhabrnal:
@vboctor i added those three methods.

api/soap/mantisconnect.wsdl
```diff
@@ -288,6 +330,19 @@
    <xsd:element name="require_closed" type="xsd:boolean" minOccurs="0"/>
   </xsd:all>
  </xsd:complexType>
+ <xsd:complexType name="FilterCustomField">
+  <xsd:all>
+   <xsd:element name="field" type="tns:ObjectRef" minOccurs="0"/>
```
vboctor: Should this have minOccurs="1"?

api/soap/mc_issue_api.php
```diff
@@ -1639,3 +1639,93 @@ function mci_issue_data_as_header_array( BugData $p_issue_data ) {
...
+        return mci_soap_fault_login_failed();
+    }
+
+    $t_lang = mci_get_user_lang( $t_user_id );
+
+    $t_result = array();
+    foreach( $p_issue_ids as $t_id ) {
+
+        if( !bug_exists( $t_id ) ) {
+            return SoapObjectsFactory::newSoapFault( 'Client', 'Issue ' . $t_id . ' does not exist' );
```
vboctor: I wonder if we should just return the issues that exist and skip the ones that don't.  If the user requests issue X, Y, Z and they get only X,Z then they will know that Y doesn't exist, but still get the others.

api/soap/mc_issue_api.php
```diff
@@ -1639,3 +1639,93 @@ function mci_issue_data_as_header_array( BugData $p_issue_data ) {
...
+        }
+
+        $t_project_id = bug_get_field( $t_id, 'project_id' );
+        $g_project_override = $t_project_id;
+        if( !mci_has_readonly_access( $t_user_id, $t_project_id ) ) {
+            return mci_soap_fault_access_denied( $t_user_id );
```
vboctor: I would do the same here and just skip the ones that the user doesn't have access to.

api/soap/mc_issue_api.php
```diff
@@ -1639,3 +1639,93 @@ function mci_issue_data_as_header_array( BugData $p_issue_data ) {
...
+        }
+
+        if( !access_has_bug_level( config_get( 'view_bug_threshold', null, null, $t_project_id ), $t_id, $t_user_id ) ) {
+            return mci_soap_fault_access_denied( $t_user_id );
```
vboctor: same here.

api/soap/mc_issue_api.php
```diff
@@ -1639,3 +1639,93 @@ function mci_issue_data_as_header_array( BugData $p_issue_data ) {
...
+    if( $t_user_id === false ) {
+        return mci_soap_fault_login_failed();
+    }
+
+    $t_lang = mci_get_user_lang( $t_user_id );
+
+    $t_result = array();
+    foreach( $p_issue_ids as $t_id ) {
+
+        if( !bug_exists( $t_id ) ) {
```
vboctor: Looks like we have duplicated code here.  See if we can refactor, otherwise, apply the same by skipping the issues that don't exist or user have no access it.

vboctor:
It would be great if we can add some tests for these methods since we rely on such tests to validate the SOAP APIs.

api/soap/mantisconnect.wsdl
mhabrnal: You are right.

mhabrnal:
ok, i'll try to write some tests tomorrow.

mhabrnal:
@vboctor i added the tests.

vboctor:
I think this is ready once the two minor comments are addressed.  I will merge it right away then.

mhabrnal:
@vboctor i addressed the comments. i hope it is right now.

vboctor:
@mhabrnal Did you revert to an older snapshot of the code somehow?  I can only see 2 commits.

mhabrnal:
@vboctor i don't think so. I see 5 comment here.

mhabrnal:
@vboctor i reverted commits sometime. Mostly when the tests failed, but never to an older snapshot. On my branch are 5 commits too (see https://github.com/mhabrnal/mantisbt/commits/master).

api/soap/mantisconnect.wsdl
```diff
@@ -259,6 +266,41 @@
    <xsd:element name="url" type="xsd:string" minOccurs="0"/>
   </xsd:all>
  </xsd:complexType>
+ <xsd:complexType name="FilterSearchData">
+  <xsd:all>
+   <xsd:element name="project_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="search" type="xsd:string" minOccurs="0"/>
```
rombert: Do we want a more self-explanatory term than search? E.g. text_search, full_text_search or similar

api/soap/mantisconnect.wsdl
```diff
@@ -259,6 +266,41 @@
...
+   <xsd:element name="category" type="tns:StringArray" minOccurs="0"/>
```
rombert: category_name would be clearer IMO

api/soap/mantisconnect.wsdl
```diff
@@ -259,6 +266,41 @@
...
+   <xsd:element name="status_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="priority_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="reporter_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="handler_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="note_user_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="resolution_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="product_version" type="tns:StringArray" minOccurs="0"/>
+   <xsd:element name="user_monitor_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="hide_status_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="sort" type="xsd:string" minOccurs="0"/>
```
rombert: sort_field ?

api/soap/mantisconnect.wsdl
```diff
@@ -288,6 +330,19 @@
    <xsd:element name="require_closed" type="xsd:boolean" minOccurs="0"/>
   </xsd:all>
  </xsd:complexType>
+ <xsd:complexType name="FilterCustomField">
+  <xsd:all>
+   <xsd:element name="field" type="tns:ObjectRef" minOccurs="1"/>
+   <xsd:element name="value" type="tns:StringArray" minOccurs="1"/>
```
rombert: Is it intentional to use a string array rather than a string here?

api/soap/mantisconnect.wsdl
```diff
@@ -747,6 +814,30 @@
   <part name="per_page" type="xsd:integer" /></message>
 <message name="mc_filter_get_issue_headersResponse">
   <part name="return" type="tns:IssueHeaderDataArray" /></message>
+<message name="mc_filter_search_issue_headersRequest">
+  <part name="username" type="xsd:string" />
+  <part name="password" type="xsd:string" />
+  <part name="filter" type="tns:FilterSearchData" />
+  <part name="page_number" type="xsd:integer" />
```
rombert: Would it make sense to include page_number and per_page in the FilterSearchData? TBH I'm not sold on the idea but putting it up for discussion

rombert:
This looks great! I made a couple of cosmetic comments, but nothing which should block merging this PR in its current state. We can decide to address them in follow-up changes if needed.

Also tested generation of Java stubs from the new WSDL and it looks good.

api/soap/mantisconnect.wsdl
mhabrnal: In the web mantisbt filter UI is it named as search so i used the same name.

api/soap/mantisconnect.wsdl
mhabrnal: IMHO is better to have an opportunity to search by more custom field values at the same time.

api/soap/mantisconnect.wsdl
mhabrnal: In first version of the new searching method it was as a part of a filter. We discussed it with @vboctor and finally we did it the same way as the other filtering method (mc_filter_get_issues etc.).

mhabrnal:
@vboctor @rombert, please let me know, which of the comments i have to addressed. i'll send a patch ASAP :)

rombert:
From my POV there is nothing to address for this PR to be merged

On Mon, Feb 2, 2015 at 3:22 PM, Matěj Habrnál <notifications@github.com>
wrote:

> @vboctor <https://github.com/vboctor> @rombert
> <https://github.com/rombert>, please let me know, which of the comments i
> have to addressed. i'll send a patch ASAP :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mantisbt/mantisbt/pull/560#issuecomment-72456872>.
>



-- 
http://robert.muntea.nu/

vboctor:
Thanks @mhabrnal for your contribution.  It is great to have this feature implemented, we wanted to get to it for a while.  Hopefully we work together further in the future.

PR #564, opened 2014-12-23
Refactor db_helper_compare_days()

dregad:
Proposed fix for https://www.mantisbt.org/bugs/view.php?id=17980

I have not fully tested this.

### DIFF
--- a/admin/schema.php
+++ b/admin/schema.php
@@ -50,11 +50,9 @@ if( !function_exists( 'db_null_date' ) ) {
 function installer_db_now() {
 	global $g_db;
 
-	$t_timezone = @date_default_timezone_get();
-	date_default_timezone_set( 'UTC' );
-	$t_time = $g_db->BindTimeStamp( time() );
-	@date_default_timezone_set( $t_timezone );
-	return $t_time;
+	# Timezone must not be set to UTC prior to calling BindTimestamp(), as
+	# ADOdb assumes a local timestamp and does the UTC conversion itself.
+	return $g_db->BindTimeStamp( time() );
 }
 
 # Special handling for Oracle (oci8):
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -890,26 +890,26 @@ function db_helper_like( $p_field_name, $p_case_sensitive = false ) {
 }
 
 /**
- * A helper function to compare two dates against a certain number of days
- * @param string|integer $p_date1_id_or_column Column or value to compare.
- * @param string|integer $p_date2_id_or_column Column or value to compare.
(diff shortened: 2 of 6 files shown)


dregad:
@vboctor can you please review the updated code taking your comments into consideration, and also check if the time issue you mentioned in the tracker has been addressed by the partial revert of commit b0d0c9dc904fa6ef6e0bc73848f216cd415429b8.

admin/schema.php
```diff
@@ -50,11 +50,7 @@ function db_null_date() {
 function installer_db_now() {
 	global $g_db;
 
-	$t_timezone = @date_default_timezone_get();
-	date_default_timezone_set( 'UTC' );
-	$t_time = $g_db->BindTimeStamp( time() );
-	@date_default_timezone_set( $t_timezone );
-	return $t_time;
+	return $g_db->BindTimeStamp( time() );
```
vboctor: Let's add a code comment in this line to avoid making the same mistake again in the future.

core/database_api.php
```diff
@@ -890,26 +890,26 @@ function db_helper_like( $p_field_name, $p_case_sensitive = false ) {
...
-	$t_db_type = config_get_global( 'db_type' );
-
-	$p_date1 = $p_date1_id_or_column;
-	$p_date2 = $p_date2_id_or_column;
-	if( is_int( $p_date1_id_or_column ) ) {
-		$p_date1 = db_param();
-	}
-	if( is_int( $p_date2_id_or_column ) ) {
-		$p_date2 = db_param();
+function db_helper_compare_days( $p_val_or_col_1, $p_operator, $p_val_or_col_2, $p_num_secs ) {
```
vboctor: Any reason why this is not db_helper_compare_time()?

manage_user_page.php
```diff
@@ -124,8 +124,8 @@
 
 $t_days_old = 7 * SECONDS_PER_DAY;
```
vboctor: OK, let's open a code cleanup bug for this.

vboctor:
The idea of this method in general is to unnatural and complicated.  I guess eventually it may make sense to simplify this somehow, maybe by having simpler methods like field_is_at_least_n_seconds_old( 'date_created' ) or just not having a method at all.  But for now, your fix is probably fine.  I think we should have a cleanup bug out of this PR, but otherwise get it in to fix the main issue.

dregad:
I pushed 2 commits to address your latest comments.

> we should have a cleanup bug out of this PR

Follow up in https://www.mantisbt.org/bugs/view.php?id=19317

dregad:
Branch rebased and commits squashed in preparation for merge

PR #565, opened 2014-12-30
Don't process history entries outside date range

vboctor:
The following queries cut the number of queries in half for a busy week:

- Remove skip() method from events class since history api does the filtering.
- When processing history events for an issue, only process ones within desired
time range.

Before:
- Total queries executed: 1232
- Unique queries executed: 1231
- Total query execution time: 0.7097 seconds

After:
- Total queries executed: 574
- Unique queries executed: 573
- Total query execution time: 0.2581 seconds

Issue #17966


vboctor:
Merged via afacf2ec8669755de62984197c63f3a9f1f8ba31

PR #566, opened 2015-01-03
Fix captcha validation

dregad:
Following 7bb78e4581ff1092c811ea96582fe602624cdcdd, it was no longer
possible to validate the captcha, because the PHP session is destroyed
as part of the logout performed in signup.php.

We now retrieve the captcha key from the session before logging out, and
recreate the session variable after that.

Fixes [#17993](https://www.mantisbt.org/bugs/view.php?id=17993)

### DIFF
--- a/core/session_api.php
+++ b/core/session_api.php
@@ -171,6 +171,7 @@ class MantisPHPSession extends MantisSession {
 		}
 
 		unset( $_SESSION[ $this->key ] );
+		session_write_close();
 	}
 }
 
--- a/signup.php
+++ b/signup.php
@@ -37,6 +37,9 @@
 	$f_email = email_append_domain( trim( $f_email ) );
 	$f_captcha = utf8_strtolower( trim( $f_captcha ) );
 
+	# Retrieve captcha key now, as session might get cleared by logout
+	$t_form_key = session_get_int( CAPTCHA_KEY, null );
+
 	# force logout on the current user if already authenticated
 	if( auth_is_user_authenticated() ) {
 		auth_logout();
@@ -50,8 +53,6 @@
 
 	if( ON == config_get( 'signup_use_captcha' ) && get_gd_version() > 0 	&&
 				helper_call_custom_function( 'auth_can_change_password', array() ) ) {
-		$t_form_key = session_get( CAPTCHA_KEY );
-
 		# captcha image requires GD library and related option to ON
 		$t_key = utf8_strtolower( utf8_substr( md5( config_get( 'password_confirm_hash_magic_string' ) . $t_form_key ), 1, 5) );
 


dregad:
@vboctor feedback on this would be welcome, as I'm not 100% sure of what side-effects closing and reopening the PHP session might have, particularly with neighboring apps.

vboctor:
Wouldn't it be easier if we sign-out the user in the signup_page.php?

dregad:
To avoid issues with signup.php being accessed directly on instances where form_security_validation = OFF, we would have to do a logout in signup.php anyway, so I'm not sure what the benefit would be. Can you please explain ?

vboctor:
The documentation of session_write_close() indicates that once called other requests can write to the session, hence, it tells me that they are not expecting the script to re-init the session again and start accessing it.  So I don't think the current fix is safe.

Would it be possible to not have the call for session_write_close() and the re-initialization of the session and just move the fetching of the captcha from the session before the auth_logout() call, as you have done in your changes?

dregad:
> once called other requests can write to the session

This is already the case today, after a script ends (e.g. signup_page.php) and the next one (signup.php) starts, isn't it ?

> Would it be possible to not have the call for session_write_close() and the re-initialization of the session and just move the fetching of the captcha from the session before the auth_logout() call, as you have done in your changes?

That was my initial approach actually, but then I realized that if signup.php is reloaded then user would get the "session variable not found" error 2701 again, as reported in issue 17993 (hence my comment about _ensuring consistent behavior_ in signup.php) so I tried the session approach.

@nextgens made an [interesting comment](https://www.mantisbt.org/bugs/view.php?id=17984#c42106) regarding the session, maybe you can provide more information.

vboctor:
> This is already the case today, after a script ends (e.g. signup_page.php) and the next one
> (signup.php) starts, isn't it ?

My point is about terminating the session and re-initializing it within signup.php where we call auth_logout() and then session_init().

> nextgens made an interesting comment regarding the session, maybe you can provide more
> information.

Yep, that would be a big changes and will probably mean that we would have to use URL session id based approach rather than a cookie based approach.  I doubt we will ever go that route.

> That was my initial approach actually, but then I realized that if signup.php is reloaded then
> user would get the "session variable not found" error 2701 again, as reported in issue 17993
> (hence my comment about ensuring consistent behavior in signup.php) so I tried the session
> approach.

Why would we want to support signup being called more than once?  Doesn't this allow the OCR guessing and retrying as well?  I think we really should design for signup_page.php then signup.php scenario with a single shot at calling signup.php.  This will enable us to logout in signup page if we want or keep it where it is and just get the captcha id before we logout the user then use that variable.

nextgens:
Keep in mind that the generated (and now cached) captcha might be unreadable...

dregad:
@vboctor 
>  Doesn't this allow the OCR guessing and retrying as well? 

No because the captcha's key is cached, so it would output the same image.

>  I think we really should design for signup_page.php then signup.php scenario with a single shot at calling signup.php

OK I'm fine with that, I'll update the PR with the code to reopen the session removed.
We just need to be aware that users will get error 2710 if the page is reloaded for any reason, and they'll have to go back to *signup_page.php* to obtain a new captcha.

@nextgens 
> Keep in mind that the generated (and now cached) captcha might be unreadable...

*signup_page.php* still generates a new key each time it's loaded, so the user could refresh the page if the first captcha they got were unreadable. The cached image is only used once the form has been submitted.

PR #567, opened 2015-01-19
Timeline fixes

dregad:
This PR fixes a couple of minor timeline-related issues

- [0018034](https://www.mantisbt.org/bugs/view.php?id=18034): Number of events is sometimes less than 50
- [0018035](https://www.mantisbt.org/bugs/view.php?id=18035): "More events" link should only appear when necessary

### DIFF
--- a/core/classes/IssueStatusChangeTimelineEvent.class.php
+++ b/core/classes/IssueStatusChangeTimelineEvent.class.php
@@ -31,6 +31,16 @@ class IssueStatusChangeTimelineEvent extends TimelineEvent {
 	private $issue_id;
 	private $old_status;
 	private $new_status;
+	private $type;
+
+	/**
+	 * Status change types to be displayed in Timeline
+	 * IGNORED = not displayed
+	 */
+	const IGNORED = 0;
+	const RESOLVED = 1;
+	const CLOSED = 2;
+	const REOPENED = 3;
 
 	/**
 	 * @param integer $p_timestamp  Timestamp representing the time the event occurred.
@@ -45,6 +55,35 @@ class IssueStatusChangeTimelineEvent extends TimelineEvent {
 		$this->issue_id = $p_issue_id;
 		$this->old_status = $p_old_status;
 		$this->new_status = $p_new_status;
+		$this->type = $this->change_type();
+	}
+
+	/**
+	 * Return the type of status change
+	 * @return int One of the status change constants defined above
+	 */
+	private function change_type() {
+		$t_resolved = config_get( 'bug_resolved_status_threshold' );
+		$t_closed = config_get( 'bug_closed_status_threshold' );
+
+		if( $this->old_status < $t_closed && $this->new_status >= $t_closed ) {
+			return IssueStatusChangeTimelineEvent::CLOSED;
+		} else if( $this->old_status < $t_resolved && $this->new_status >= $t_resolved ) {
+			return IssueStatusChangeTimelineEvent::RESOLVED;
+		} else if( $this->old_status >= $t_resolved && $this->new_status < $t_resolved ) {
+			return IssueStatusChangeTimelineEvent::REOPENED;
+		} else {
+			return IssueStatusChangeTimelineEvent::IGNORED;
+		}
+	}
+
+	/**
+	 * Whether to skip this timeline event.
+	 * This normally implements access checks for the event.
+	 * @return boolean
+	 */
+	public function skip() {
+		return $this->type == IssueStatusChangeTimelineEvent::IGNORED;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


core/classes/IssueStatusChangeTimelineEvent.class.php
```diff
@@ -45,24 +55,67 @@ public function __construct( $p_timestamp, $p_user_id, $p_issue_id, $p_old_statu
...
+				break;
+			case IssueStatusChangeTimelineEvent::REOPENED:
+				$t_string = sprintf(
+					lang_get( 'timeline_issue_reopened' ),
+					user_get_name( $this->user_id ),
+					string_get_bug_view_link( $this->issue_id )
+				);
+				break;
+			case IssueStatusChangeTimelineEvent::IGNORED:
+			default:
```
vboctor: Doesn't the default here introduces the same problem you are trying to fix?  where we think we displayed N and we actually display less than N because of a set of events that are not skipped yet generate no output?  Should we error out in case of default?

core/classes/IssueStatusChangeTimelineEvent.class.php
```diff
@@ -45,24 +55,67 @@ public function __construct( $p_timestamp, $p_user_id, $p_issue_id, $p_old_statu
...
 		$t_resolved = config_get( 'bug_resolved_status_threshold' );
 		$t_closed = config_get( 'bug_closed_status_threshold' );
 
-		if( $this->old_status < $t_closed && $this->new_status >= $t_closed ) {
-			$t_string = sprintf( lang_get( 'timeline_issue_closed' ), user_get_name( $this->user_id ), string_get_bug_view_link( $this->issue_id ) );
-		} else if( $this->old_status < $t_resolved && $this->new_status >= $t_resolved ) {
-			$t_string = sprintf( lang_get( 'timeline_issue_resolved' ), user_get_name( $this->user_id ), string_get_bug_view_link( $this->issue_id ) );
+		if( $this->old_status < $t_resolved && $this->new_status >= $t_resolved ) {
+			return IssueStatusChangeTimelineEvent::RESOLVED;
+		} else if( $this->old_status < $t_closed && $this->new_status >= $t_closed ) {
```
vboctor: In the case of an issue moving from new/assigned to closed.  Do we want to show it as "resolved" or "closed"?  In my implementation I was going for "closed", with your updates, it will show up as "resolved".

core/classes/IssueStatusChangeTimelineEvent.class.php
```diff
@@ -45,24 +55,67 @@ public function __construct( $p_timestamp, $p_user_id, $p_issue_id, $p_old_statu
...
+			return IssueStatusChangeTimelineEvent::IGNORED;
+		}
+	}
+
+	/**
+	 * Whether to skip this timeline event.
+	 * This normally implements access checks for the event.
+	 * @return boolean
+	 */
+	public function skip() {
```
vboctor: How about instead of re-introducing the concept of skip, we have the method that generates the html return an array() of strings.  Where a string entry == one event.  In this case, array() will be empty and won't be counted, most events will have one entry, but there is also the option of having an event that generates more than one row.  Another advantage for this approach is that we don't need to synchronize the logic in skip() and html(), which was the issue skip checking only for ignore, yet html also handing default.

vboctor: This may also remove the need for the constants and separation of classification of events from generation of strings.

vboctor:
I have suggested some comments on the specific implementation of the fix + suggestion to simplify the implementation.  Thanks.

core/classes/IssueStatusChangeTimelineEvent.class.php
dregad: Looks like I changed the order of the if statements by mistake, will fix.

core/classes/IssueStatusChangeTimelineEvent.class.php
dregad: I don't think so, based on my testing, but maybe I missed something ?

The logic for excluding event (i.e. using the skip() method) is actually in timeline API's timeline_events() function; the array it generates does (or should) not contain any "empty" events (IGNORED type), so in theory we should never reach this default case unless something went wrong. 

It's belt and braces I guess - would you prefer that I remove lines 116-118 ?

core/classes/IssueStatusChangeTimelineEvent.class.php
dregad: My initial approach followed a similar but less elaborated solution; the filtering logic was in timeline_print_events() function, testing an empty string returned by html() within the foreach loop.

Then I thought it was simpler and more elegant to have a "clean" array (i.e. only containing useful data) to begin with.

> option of having an event that generates more than one row

I'm not sure what is the use for this, or if it's even possible. Did you have anything specific in mind ? 
I think that having possible multiple entries would make counting events much more complex, and I'm not sure it's worth it.

> we don't need to synchronize the logic in skip() and html()

They are used separately and don't need to be in sync: 
- skip() is used to exclude events from the array built by timeline_events() and 
- timeline_print_events() / html() just print whatever is in the array.

A possibly simpler alternative to reintroducing "skip" could be to have the constructor throw an exception.

dregad:
Many thanks for your review and feeback.

core/classes/IssueStatusChangeTimelineEvent.class.php
dregad: Looks like I changed the order of the if statements by mistake, will fix.

dregad:
Just pushed a couple of commits that should address your first 2 comments.

Let me know your thoughts on alternative approaches to skip() (see https://github.com/mantisbt/mantisbt/pull/567#discussion_r23210028)

vboctor:
I personally think that it is simpler to return an array() that can have 0 or 1 entries (with possibility to have more than one in the future).  The complexity of counting events when an event can be skipped anyway.  We also will drop skip() and the constants that are created.

Regarding to the scenario of one change causing multiple events, I don't have a strong scenario in mind.  But it gives a choice of making a change reflect as two events like re-assigning an issue can produce person B picked the issue and person A stopped working on issue.  No clear advantage over a single more complex event except if access checks want to hide 1 and keep the other in some cases.

Anyway, your current implementation would work as well.  So feel free to checkin either way.  So consider this a sign-off.

dregad:
Since you're OK with the current implementation, and considering that it does the job, I'll save myself the effort of working on an alternative solution ;-)

PR #568, opened 2015-01-23
fixed false escape

mfoXen:


dregad:
Fixed in 93e0d03960de980b6d3fc807926f7c554a9bf230. Thanks for your contribution.

PR #569, opened 2015-01-24
Bug19258: Fix custom date field in filter

syncguru:
It appears that this was never fully implemented. This fix enables the correct date field based on user selection of comparison operator in the select menu above (see snapshots)

Fixes #19258: Custom fields of type date are disabled in filter

screen shot 2015-01-24 at 10 44 05 am
screen shot 2015-01-24 at 10 44 16 am
screen shot 2015-01-24 at 10 44 26 am

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -4289,21 +4289,12 @@ function print_filter_custom_field_date( $p_field_num, $p_field_id ) {
 			$t_end = $t_end_time;
 			break;
 		case CUSTOM_FIELD_DATE_ONORBEFORE:
-			$t_start_disable = false;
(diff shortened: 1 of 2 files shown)


vboctor:
+1 - make sure your checkins and PR requests reference the bugs as "Fixes # 1234" without the space (I put the space so that github doesn't try to link it.  I've edited the PR description, but you may need to fix the checkin comments to have our bug tracker pickup the checkins.

@dregad lets get this in beta2 - I can merge if you are OK with the change.

dregad:
Fixed in c76a39d96052a3bc7a2c34c81a9339a9bb424f3d

PR #570, opened 2015-01-27
Using $t_config_filename instead of repeating this calc

mfoXen:
This also fixes an incorrect "Please add the following lines to '/opt/lampp/htdocs/mantisbt/config_inc.php' before continuing: " if the file could not be written. The new config-subfolder wasn't mentioned


dregad:
Fixed in 5057e418b76da7bdc915202df3afdb68a7b7e9b5, thank you.

For future contributions, it would be nice of you read and abide by the guidelines in https://www.mantisbt.org/wiki/doku.php/mantisbt:git_commit_messages.

PR #574, opened 2015-02-02
Configure avatar site

andersruneson:
See issue #0019319

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -976,6 +976,18 @@
 	$g_show_avatar_threshold = DEVELOPER;
 
 	/**
+	 * To host your own avatar site change this setting
+	 * @global string $g_show_avatar_url
+	 */
+	$g_show_avatar_url = 'http://www.gravatar.com';
+
+	/**
+	 * To host your own avatar site change this setting
+	 * @global string $g_show_avatar_url_secure
+	 */
+	$g_show_avatar_url_secure = 'https://secure.gravatar.com';
+
+	/**
 	 * Show release dates on changelog
 	 * @global int $g_show_changelog_dates
 	 */
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -976,16 +976,12 @@
 	$g_show_avatar_threshold = DEVELOPER;
 
 	/**
-	 * To host your own avatar site change this setting.
+	 * To host your own avatar site set this setting.
 	 * A https url is needed if MantisBT is being accessed via https
 	 * otherwise there will be warnings of unsecure content.
 	 * @global string $g_show_avatar_url
 	 */
-	if ( http_is_protocol_https() ) {
-		$g_show_avatar_url = 'https://secure.gravatar.com';
-	} else {
-		$g_show_avatar_url = 'http://www.gravatar.com';
-	}
+	$g_show_avatar_url = '';
 
 	/**
 	 * Show release dates on changelog
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -976,16 +976,16 @@
 	$g_show_avatar_threshold = DEVELOPER;
 
 	/**
-	 * To host your own avatar site change this setting
+	 * To host your own avatar site change this setting.
+	 * A https url is needed if MantisBT is being accessed via https
+	 * otherwise there will be warnings of unsecure content.
 	 * @global string $g_show_avatar_url
 	 */
-	$g_show_avatar_url = 'http://www.gravatar.com';
-
-	/**
-	 * To host your own avatar site change this setting
-	 * @global string $g_show_avatar_url_secure
-	 */
-	$g_show_avatar_url_secure = 'https://secure.gravatar.com';
+	if ( http_is_protocol_https() ) {
+		$g_show_avatar_url = 'https://secure.gravatar.com';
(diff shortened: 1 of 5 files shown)


vboctor:
First of all, thanks for your contribution and splitting your avatar related changes into two logical pull requests.  It would be great to have such feature requests against master.

Did you consider the option of fitting this in our plugin model?  Where there can be a gravatar plugin, db-avatar, LDAP/AD avatar, etc?

vboctor:
Here is a related issue in the official bugtracker:
https://www.mantisbt.org/bugs/view.php?id=8257

andersruneson:
Sorry, I'm not that skilled developer.

vboctor:
:)  How are you currently using this change?  What do you point the gravatar url to?

andersruneson:
I have a simple php-script that serves images from disk, so the url points to the same server as mantis.
Optimal would of course be to get images directly from LDAP/AD.

vboctor:
I personally don't mind getting this in.  Would you mind posting your php-script as a github gist or github repository and including a link to it here?  This way we can refer to it from the comments/documentation in the pull request as a sample script for such end-point?

andersruneson:
I couldn't get github to set syntax higlightling.
https://gist.github.com/andersruneson/d14979b1076318b3d2f8

dregad:
>I couldn't get github to set syntax higlightling.

There should be a selection list next to the file name where you can pick the language.

andersruneson:
There is a dropdown but nothing happens when I change to PHP and click update Gist.

core/http_api.php
```diff
@@ -137,9 +137,9 @@ function http_security_headers() {
 		$t_avatar_img_allow = '';
 		if ( config_get_global( 'show_avatar' ) ) {
 			if ( http_is_protocol_https() ) {
-				$t_avatar_img_allow = "; img-src 'self' https://secure.gravatar.com:443";
+				$t_avatar_img_allow = "; img-src 'self' ".config_get( 'show_avatar_url_secure' ).":443";
 			} else {
-				$t_avatar_img_allow = "; img-src 'self' http://www.gravatar.com:80";
+				$t_avatar_img_allow = "; img-src 'self' "config_get( 'show_avatar_url' )":80";
```
vboctor: Are you missing concat dots here?  Also in general there should be spaces around the dot, you please fix the one for the secure case.

config_defaults_inc.php
```diff
@@ -976,6 +976,18 @@
 	$g_show_avatar_threshold = DEVELOPER;
 
 	/**
+	 * To host your own avatar site change this setting
+	 * @global string $g_show_avatar_url
+	 */
+	$g_show_avatar_url = 'http://www.gravatar.com';
```
vboctor: I would rather if we have one config option called $g_show_avatar_url.  We default it in config_defaults_inc.php based on ssl vs. not.  Then everywhere in the code we just get the value of the config and use it.  If an instance wants to override that, they will just have one option to override and they will have to set it to a value that matches they https vs. http scheme.  The documentation of the config option should clarify that https avatar site is needed if MantisBT itself is being accessed via https, otherwise, there will be warnings of unsecure content.

vboctor:
The syntax highlighting seems to be a github bug.  I also tried to do it, it shows up correctly in the editor, but doesn't persist on save.

@andersruneson I've added a couple of comments, once fixed, I don't have a problem with getting this in.

@dregad can also review and comment.

andersruneson:
I hope I haved addressed the comments.

Thanks for feedback.

config_defaults_inc.php
```diff
@@ -976,6 +976,18 @@
...
 	/**
+	 * To host your own avatar site change this setting.
+	 * A https url is needed if MantisBT is being accessed via https
+	 * otherwise there will be warnings of unsecure content.
+	 * @global string $g_show_avatar_url
+	 */
+	if ( http_is_protocol_https() ) {
+		$g_show_avatar_url = 'https://secure.gravatar.com';
+	} else {
+		$g_show_avatar_url = 'http://www.gravatar.com';
```
atrol: What about using local variable $t_protocol for the check?
Should be a bit faster.

dregad: I wouldn't rely on that, I'm planning to get rid of that variable in config_default_inc.php.

dregad:
I need to have a closer look at this. I like the idea, but I'm not convinced by the current implementation (particularly the $g_show_avatar_url).

andersruneson:
I noticed that http_is_protocol_https isn't available in config_defaults_inc.php. I will fix that as soon as I get your other comments.

However, again, I'm not very interested in fixing up the complete avatar implementation.

Let me also state that I've used Jira and other Atlassian products. Those products let you set a parameter called "Gravatar server" with the following help:
*Optional custom Gravatar server (leave blank to use Gravatar.com)*

dregad:
> I noticed that http_is_protocol_https isn't available in config_defaults_inc.php. 

Basically what I don't like is having logic reside in config_defaults_inc.php. IMO it belongs in user API (or ultimately in some gravatar plugin). 

> set a parameter called "Gravatar server"  [...] (leave blank to use Gravatar.com)

This is a much better approach IMO.

vboctor:
Closing in favor of #711

PR #575, opened 2015-02-04
Handling single category case

vboctor:
If there is a single category and it is required to set the category
field, then auto-select such category.

Fixes #19264


dregad:
+1

PR #576, opened 2015-02-09
Timeline tags hyperlink to Tag Details page

dregad:
https://www.mantisbt.org/bugs/view.php?id=19352

### DIFF
--- a/core/classes/IssueTagTimelineEvent.class.php
+++ b/core/classes/IssueTagTimelineEvent.class.php
@@ -53,6 +53,7 @@ class IssueTagTimelineEvent extends TimelineEvent {
 	 */
 	public function html() {
 		$t_string = $this->tag ? lang_get( 'timeline_issue_tagged' ) : lang_get( 'timeline_issue_untagged' );
+		$t_tag_row = tag_get_by_name( $this->tag_name );
 
 		$t_html = $this->html_start();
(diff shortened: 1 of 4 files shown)


core/classes/IssueTagTimelineEvent.class.php
```diff
@@ -45,6 +46,8 @@ public function __construct( $p_timestamp, $p_user_id, $p_issue_id, $p_tag_name,
 		$this->issue_id = $p_issue_id;
 		$this->tag_name = $p_tag_name;
 		$this->tag = $p_tag;
+
+		$this->tag_row = tag_get_by_name( $p_tag_name );
```
vboctor: Why don't you remove the class property and move this to the html() method rather than here?  Technically speaking skip() could return true, though we don't override it here.  Also why does tag_get_by_name() uses like to find the tag rather than equality?

vboctor:
+1

dregad:
> Why don't you remove the class property and move this to the html() method rather than here?

Good point, fixed.

> why does tag_get_by_name() uses like to find the tag rather than equality?

I don't know, really. This dates back to the original implementation of tag API (see https://github.com/mantisbt/mantisbt/commit/9bb8b957504715ba7261d5976d547c8229f52e18#diff-0dd9ea1f6f91718b591f7f3808f988b1R225)

PR #577, opened 2015-02-19
Make Assign work with required custom fields

vboctor:
This bug was introduced in 1.3.x where users are not able to assign issues within projects that have required custom fields on update. This failure happens even independent of whether the fields are populated or not.  The fix is to enabling the Assign-To action to work independent of custom fields values.  This this is an assign-to action rather than a full update.

Fixes #19265

### DIFF
--- a/bug_update.php
+++ b/bug_update.php
@@ -65,6 +65,7 @@ form_security_validate( 'bug_update' );
 
 $f_bug_id = gpc_get_int( 'bug_id' );
 $t_existing_bug = bug_get( $f_bug_id, true );
+$f_update_type = gpc_get_string( 'action_type', BUG_UPDATE_TYPE_NORMAL );
 
 if( helper_get_current_project() !== $t_existing_bug->project_id ) {
 	$g_project_override = $t_existing_bug->project_id;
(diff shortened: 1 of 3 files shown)


bug_update.php
```diff
@@ -65,6 +65,9 @@
 
 $f_bug_id = gpc_get_int( 'bug_id' );
 $t_existing_bug = bug_get( $f_bug_id, true );
+$f_update_type = gpc_get_string( 'action_type', BUG_UPDATE_TYPE_NORMAL );
+
+$t_update_action = $f_update_type == BUG_UPDATE_TYPE_NORMAL;
```
dregad: I'm not sure we actually need this variable, IMO it makes the code more confusing, and I'd rather just have 

```if( $t_cf_def[$t_cf_require_check] && $f_update_type == BUG_UPDATE_TYPE_NORMAL ) {```

dregad:
Except for my above comment, I'm OK with this change.

vboctor:
Merged 899b93cc9f279a9364f62e6e11c5b6b438e9b358

PR #578, opened 2015-02-19
Switching projects should maintain current page

vboctor:
When possible maintain the current page as user switches project.
A bug was introduced in 1.3.x that made switching projects always
redirects to default home page.

Fixes #19351


dregad:
> A bug was introduced in 1.3.x

Suggest to specify the offending change b78f966e013d494ea27d92bba39d797a1266b7b7 in the commit message instead of the vague *1.3.x* reference.

vboctor:
Merged 9ad6013cd225eb690ad943c64048e7337ff59bd6

PR #583, opened 2015-03-06
Time tracking selector doesn't maintain dates

vboctor:
Fixes #19473


vboctor:
Merged via c5f234363e910f49b27d03d1b71da33978da8a68

PR #584, opened 2015-03-08
Reporter can't re-open or close issues even if they have access

vboctor:
See https://www.mantisbt.org/bugs/view.php?id=19271


dregad:
I did not actually test this, but looking at the code it seems OK so +1

bug_update.php
```diff
@@ -243,7 +250,7 @@
 }
 
 # Ensure that the user has permission to change the view status of the issue.
-if( $t_existing_bug->view_state !== $t_updated_bug->view_state ) {
+if( (int)$t_existing_bug->view_state !== $t_updated_bug->view_state ) {
```
atrol: Seems this introduced a regression
https://www.mantisbt.org/bugs/view.php?id=20478

dregad: Why use strict type check ?

PR #585, opened 2015-03-08
Temp fix to inability to disable notifications

vboctor:
The proper fix for this is not trivial involving database schema change, localization, etc.
Until this is done, we will treat uncustomizable notifications the same way as the
default preference where the "email_on_status" preference is used.
The issue will not be marked as resolved until we add the necessary preferences.
However, we should consider a model where adding user preferences doesn't
require a schema change.

Issue #19459


atrol:
+1

dregad:
+1

vboctor:
Merged via 0298961561b51932db0d5dd66c84ec9eb45c5f47

PR #587, opened 2015-03-08
Fix account updated email formatting error

vboctor:
Fixes #19269

### DIFF
--- a/manage_user_update.php
+++ b/manage_user_update.php
@@ -177,20 +177,25 @@ $t_result = db_query( $t_query, $t_query_params );
 if( $f_send_email_notification ) {
 	lang_push( user_pref_get_language( $f_user_id ) );
 	$t_changes = '';
+
 	if( strcmp( $f_username, $t_old_username ) ) {
-		$t_changes .= lang_get( 'username_label' ) . lang_get( 'word_separator' ) . $t_old_username . ' => ' . $f_username . "\n";
+		$t_changes .= lang_get( 'username_label' ) . ' ' . $t_old_username . ' => ' . $f_username . "\n";
 	}
+
 	if( strcmp( $t_realname, $t_old_realname ) ) {
-		$t_changes .= lang_get( 'realname_label' ) . lang_get( 'word_separator' ) . $t_old_realname . ' => ' . $t_realname . "\n";
+		$t_changes .= lang_get( 'realname_label' ) . ' ' . $t_old_realname . ' => ' . $t_realname . "\n";
 	}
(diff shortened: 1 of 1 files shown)


manage_user_update.php
```diff
@@ -177,20 +177,25 @@
 if( $f_send_email_notification ) {
 	lang_push( user_pref_get_language( $f_user_id ) );
 	$t_changes = '';
+
 	if( strcmp( $f_username, $t_old_username ) ) {
-		$t_changes .= lang_get( 'username_label' ) . lang_get( 'word_separator' ) . $t_old_username . ' => ' . $f_username . "\n";
+		$t_changes .= lang_get( 'username_label' ) . ' ' . $t_old_username . ' => ' . $f_username . "\n";
```
atrol: Not sure what's the right solution for it, we had hardcoded ': ' in master-1.2.x
https://github.com/mantisbt/mantisbt/blob/master-1.2.x/manage_user_update.php#L152
Furthermore we had just language string username but not username and username_label (which are the same at the moment in English)
In some languages there is the ':' in language string $s_username_label but not in all languages.
The same applies to other strings (e.g. realname, email)

dregad:
Since the issue seems to be caused by use of an html-encoded space (`&#32;`) as word separator - introduced in 33d66aa4d06ca0323592f4a0f44affe927d83e05 - then the right way to fix this would be to change `$s_word_separator = ' ';` in *lang/strings_english.txt*.

Unless there was a valid reason for html-encoding the space as opposed to using the ascii character directly, but I can't think of any (but check with @siebrand maybe, since he authored that).

vboctor:
I got the fix in, but we can follow up with another if we have better ideas.  We should have a model that works for email and html emails, and I wonder if we really need this in the localization files.

dregad:
>  We should have a model that works for email and html emails

I believe my proposal above would work in both cases (a space is a space after all, and as mentioned I see no difference between `'&#32;'` and `' '`  so we might as well use the latter).

WRT *better ideas*, I can't think of any atm - as for needing this in localization files, it would be best to have a chat with Siebrand.

vboctor:
I'll keep this fix focused on fixing the regression compared to 1.2.x.  My personal opinion is that we don't need word_separator.  But I opened the following bug to continue the discussion:
https://www.mantisbt.org/bugs/view.php?id=19602

Even if we change word_separator to a normal space, the emails will be broken until all languages are localized.  Other usages of word_separator are typically within the context of html and hence works fine.

PR #588, opened 2015-03-10
Add default project view status config

vboctor:
Fixes #3874

### DIFF
--- a/manage_proj_create_page.php
+++ b/manage_proj_create_page.php
@@ -113,7 +113,7 @@ if( project_table_empty() ) {
 				<label for="project-view-state"><span><?php echo lang_get( 'view_status' ) ?></span></label>
 				<span class="select">
 					<select id="project-view-state" name="view_state">
-						<?php print_enum_string_option_list( 'view_state' ) ?>
+						<?php print_enum_string_option_list( 'view_state', config_get( 'default_project_view_status', null, ALL_USERS, ALL_PROJECTS ) ) ?>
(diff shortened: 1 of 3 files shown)


dregad:
+1

manage_proj_create_page.php
```diff
@@ -113,7 +113,7 @@
 				<label for="project-view-state"><span><?php echo lang_get( 'view_status' ) ?></span></label>
 				<span class="select">
 					<select id="project-view-state" name="view_state">
-						<?php print_enum_string_option_list( 'view_state' ) ?>
+						<?php print_enum_string_option_list( 'view_state', config_get( 'default_project_view_status' ) ) ?>
```
atrol: You could/should? use config_get_global

vboctor: If I use config_get_global() then users have to have access to the server and edit the config file to change this setting.  Otherwise, they can just go and add a setting as follows:

All Projects, All Users, default_project_view_status, integer, 50

I will use the appropriate config_get() parameters to make sure we don't pickup setting associated with current project, although that doesn't make sense for this config option.

atrol: +1 for
```php
config_get( 'default_project_view_status', null, ALL_USERS, ALL_PROJECTS);
```

vboctor:
Merged via 131e51da5e323b7d292e0494f29d7ac7891b42c7 after applying @atrol 's comment.

PR #589, opened 2015-04-01
Synchronous email sending via shutdown function

dregad:
In early 2014, an attempt was made to improve performance for synchronous e-mail sending for fastcgi by processing the queue at the bottom of each displayed page (in html_end() function, see cea405cc), which introduced regressions for SOAP API (issue [#17458](https://www.mantisbt.org/bugs/view.php?id=17458)) as well as plugins and other corner cases (issues [#17460](https://www.mantisbt.org/bugs/view.php?id=17460)).

This commit resolves the problem by registering a shutdown function to process the email queue in core.php, which ensures that email gets sent no matter what.

To avoid multiple executions of the shutdown function for a single user request (which may lead to several executions of core.php, e.g. when building dynamic css, javascript translations, etc), an arbitrary 5 seconds delay is observed between each register_shutdown_function() call.

Fixes [#17460](https://www.mantisbt.org/bugs/view.php?id=17460)

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -788,10 +788,9 @@ function email_store( $p_recipient, $p_subject, $p_message, array $p_headers = n
 	$t_email_id = email_queue_add( $t_email_data );
 
 	# Set the email processing flag for the shutdown function
+	$g_email_shutdown_processing |= EMAIL_SHUTDOWN_GENERATED;
 	if( $p_force ) {
 		$g_email_shutdown_processing |= EMAIL_SHUTDOWN_FORCE;
-	} else {
-		$g_email_shutdown_processing |= EMAIL_SHUTDOWN_GENERATED;
 	}
 
 	return $t_email_id;


core/email_api.php
```diff
@@ -1395,3 +1395,51 @@ function email_build_visible_bug_data( $p_user_id, $p_bug_id, $p_message_id ) {
...
+		# Define shutdown function registration time with an arbitrary delay
+		# to prevent concurrent calls when multiple, parallel http requests
+		# are sent by a single user action
+		$t_now = time();
+		$t_registration_time = session_get_int( 'email_shutdown_function_set', $t_now) ;
+
+		if( $t_registration_time <= $t_now  ) {
+			$t_register = true;
+			# 5 seconds delay until the next registration
+			session_set( 'email_shutdown_function_set', $t_now + 5 );
```
dregad: This may not be the best approach, but I couldn't think of any other way to prevent the shutdown function from being called up to 5 times for a single page request. Maybe you have a better idea ?

The 5 seconds delay is arbitrary, maybe it's too much... Informal testing on my dev box shows that 2 seconds is enough in most cases to prevent multiple calls to email_send_all() from a single request.

core/email_api.php
```diff
@@ -1395,3 +1395,51 @@ function email_build_visible_bug_data( $p_user_id, $p_bug_id, $p_message_id ) {
...
+				break;
+			default:
+				$t_register = true;
+		}
+	} else {
+		# Define shutdown function registration time with an arbitrary delay
+		# to prevent concurrent calls when multiple, parallel http requests
+		# are sent by a single user action
+		$t_now = time();
+		$t_registration_time = session_get_int( 'email_shutdown_function_set', $t_now) ;
```
vboctor: Fix spacing at the end of the line.

core/email_api.php
```diff
@@ -1395,3 +1395,51 @@ function email_build_visible_bug_data( $p_user_id, $p_bug_id, $p_message_id ) {
...
+/**
+ * Registers shutdown function to send queued emails after script execution
+ * @return void
+ */
+function email_shutdown_function_register() {
+	$t_register = false;
+	if( php_sapi_name() == 'cli' ) {
+		# Skip processing for predefined scripts
+		switch( basename( $_SERVER['SCRIPT_NAME'] ) ) {
+			case 'send_emails.php':
```
vboctor: How about having the ability for any script page to SKIP_EMAIL_SENDING and having this code just check for that rather than specific script names?  We should also use the same technique to skip generation for scenarios where we generate secondary resources like css, javascript, etc.

core/email_api.php
vboctor: If I run two pages that generate email in 5 seconds, could this result in the second email being queued up?  I would rather use a model where the pages that can trigger in parallel would skip the registration of the shutdown method if that is possible.

core/email_api.php
```diff
@@ -1395,3 +1395,51 @@ function email_build_visible_bug_data( $p_user_id, $p_bug_id, $p_message_id ) {
...
+		register_shutdown_function( 'email_shutdown_function' );
+	}
+}
+
+/**
+ * The email sending shutdown function
+ * @return void
+ */
+function email_shutdown_function() {
+	global $g_email_stored;
```
vboctor: I assume this reflects email_generated_in_current_request rather than email_stored_in_database, correct?  Should we make the variable name clearer either way?

core/email_api.php
```diff
@@ -1395,3 +1395,51 @@ function email_build_visible_bug_data( $p_user_id, $p_bug_id, $p_message_id ) {
...
+		log_event( LOG_EMAIL, 'Registering shutdown function for ' . $_SERVER['SCRIPT_NAME'] );
+		register_shutdown_function( 'email_shutdown_function' );
+	}
+}
+
+/**
+ * The email sending shutdown function
+ * @return void
+ */
+function email_shutdown_function() {
```
vboctor: Would it make sense if we go this route to have a shutdown_method that can do multiple things one of them is email?  Also always call this and then within it decide whether cronjob is on/off.  This way we can eventually add a plugin event, or do other work there?

vboctor:
I think relying on the cronjob (when enabled) for email notifications if fine, but scenarios like signup, user creation, or reset passwords, we would want to be able to trigger that right away as part of the shutdown function.  Maybe it is via a force_send_emails flag.  I'm not sure of the original implementation, but from feedback I got, people did notice the delay, but this didn't like waiting for emails that they needed to continue their workflow.

core/email_api.php
dregad: > If I run two pages that generate email in 5 seconds, could this result in the second email being queued up?

Technically yes

core/email_api.php
dregad: Good idea, I'll look into that

core/email_api.php
dregad: > I assume this reflects email_generated_in_current_request

Correct. I just reused the variable that Paul created... Will rename.

core/email_api.php
dregad: Systematically registering the shutdown function and moving the cronjob on/off logic within it would give the added benefit of allowing some scripts to skip or force email sending. I'll refactor the code accordingly.

dregad:
Thanks for your detailed and constructive feedback. I'm planning to update the PR with your suggestions over the Easter week-end.

vboctor:
Great.  Looking forward to it!

dregad:
This change passes the two test cases described below:

1. update a bug
2. request password reset

Given the status of *$g_email_send_using_cronjob*:

a) *OFF*: both test cases should trigger sending of emails at end of request via shutdown function
b) *ON*: case 1 should queue the email but not send it, while case 2 will send process queued emails

core/email_api.php
```diff
@@ -79,9 +79,16 @@
 # reusable object of class SMTP
 $g_phpMailer = null;
 
-# Indicates whether any emails are currently stored for process during this request.
-# Note: This is only used if not sending emails via cron job
-$g_email_stored = false;
+/**
+ * Indicates how generated emails will be processed by the shutdown function
+ * at the end of the current request's execution; this is a binary flag:
+ * - EMAIL_SHUTDOWN_SKIP       Initial state: do nothing (no generated emails)
```
vboctor: Instead of repeating this list in two places, why don't we document it in the constants file and just refer here to EMAIL_SHUTDOWN_* constants.

core/email_api.php
```diff
@@ -734,14 +741,16 @@ function email_relationship_child_resolved_closed( $p_bug_id, $p_message_id ) {
...
+ * @param string  $p_subject   Subject of email message.
+ * @param string  $p_message   Body text of email message.
+ * @param array   $p_headers   Array of additional headers to send with the email.
+ * @param boolean $p_force     True to force sending of emails in shutdown function,
+ *                             even when using cronjob
  * @return integer|null
  */
-function email_store( $p_recipient, $p_subject, $p_message, array $p_headers = null ) {
-	global $g_email_stored;
+function email_store( $p_recipient, $p_subject, $p_message, array $p_headers = null, $p_force = false ) {
```
vboctor: Wouldn't have the last parameter use EMAIL_SHUTDOWN_* give more flexibility and be more readable.  For example, default value would be EMAIL_SHUTDOWN_GENERATED, while user can set it to EMAIL_SHUTDOWN_FORCE or even EMAIL_SHUTDOWN_GENERATED | EMAIL_SHUTDOWN_FORCE.

vboctor:
:+1:  I added a couple of comments.  Feel free to push.

core/email_api.php
dregad: I actually think providing a boolean flag is clearer and more meaningful from an API user's perspective.

I'm not sure about the flexibility argument, considering that EMAIL_SHUTDOWN_FORCE actually implies EMAIL_SHUTDOWN_GENERATED. 

To be technically correct, I probably should revise the assignment of *$g_email_shutdown_processing* in *email_store()* to always set the EMAIL_SHUTDOWN_GENERATED flag.

core/email_api.php
dregad: Not sure what is the other place you're referring to... in constants_inc.php ? IMO it is better to document the meaning of the constants where they are actually used.

core/email_api.php
vboctor: ok

core/email_api.php
vboctor: How about if they are used in multiple places?  Do we document in each function we pass them to?  It is useful to refer to the documentation whenever they are used.  But documentation with definition is useful.  But no big deal.

vboctor:
Go for it once you are ready.  No blockers from my side.

PR #591, opened 2015-04-05
Edited Issue's handler is set to blank when user has been demoted

dregad:
https://www.mantisbt.org/bugs/view.php?id=19574

### DIFF
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -296,6 +296,11 @@ function print_user_option_list( $p_user_id, $p_project_id = null, $p_access = A
 		$t_users = project_get_all_user_rows( $p_project_id, $p_access );
 	}
 
+	# Ensure the specified user exists in the array
+	if( $p_user_id != NO_USER && !array_key_exists( $p_user_id, $t_users ) ) {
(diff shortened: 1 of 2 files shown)


vboctor:
:+1: finally :) -- there must be very old bugs against this one.  I recall probably even opening one myself.

Another twin of this problem is filtering on users that are disabled, yet still have issues assigned to them or have reported issues.  Though it is a different issue.

project_get_all_user_rows - This is called often, just make sure the change in its return value doesn't cause any issues.

dregad:
> Another twin of this problem is filtering on users that are disabled, yet still have issues assigned to them or have reported issues. Though it is a different issue.

Yes this one has been bugging me as well, but it's not as simple to fix - and probably requires running SELECT DISTINCT statements (or equivalent), against potentially unindexed columns which may have a performance impact. That's something for another day ;-)

> project_get_all_user_rows - This is called often, just make sure the change in its return value doesn't cause any issues.

I grepped the code and ran some tests, didn't notice any side effects.

dregad:
> there must be very old bugs against this one. I recall probably even opening one myself.

I searched the tracker (again), but didn't find any.

PR #592, opened 2015-04-08
add option to configure visibility of timeline

schriftgestalt:
Fixes #0019586: allow to disable the timeline in the config file

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2791,6 +2791,12 @@ $g_view_changelog_threshold = VIEWER;
 $g_roadmap_view_threshold = VIEWER;
 
 /**
+ * threshold for viewing timeline
+ * @global integer $g_view_timeline_threshold
+ */
+$g_view_timeline_threshold = VIEWER;
+
+/**
  * threshold for updating roadmap, target_version, etc
  * @global integer $g_roadmap_update_threshold
  */
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2792,9 +2792,9 @@ $g_roadmap_view_threshold = VIEWER;
 
 /**
  * threshold for viewing timeline
- * @global integer $g_view_timeline_threshold
+ * @global integer $g_timeline_view_threshold
  */
-$g_view_timeline_threshold = VIEWER;
+$g_timeline_view_threshold = VIEWER;
 
 /**
  * threshold for updating roadmap, target_version, etc
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2791,16 +2791,16 @@ $g_view_changelog_threshold = VIEWER;
 $g_roadmap_view_threshold = VIEWER;
 
 /**
- * threshold for viewing timeline
- * @global integer $g_timeline_view_threshold
- */
-$g_timeline_view_threshold = VIEWER;
-
-/**
  * threshold for updating roadmap, target_version, etc
  * @global integer $g_roadmap_update_threshold
  */
 $g_roadmap_update_threshold = DEVELOPER;
+
+/**
+ * threshold for viewing timeline
+ * @global integer $g_timeline_view_threshold
+ */
+$g_timeline_view_threshold = VIEWER;
 
 /**
  * status change thresholds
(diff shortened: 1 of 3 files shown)


atrol:
It would be fine if the setting is also documented in administrators guide.
e.g. have a look at https://github.com/mantisbt/mantisbt/commit/131e51da5e323b7d292e0494f29d7ac7891b42c7 how to implement it.

config_defaults_inc.php
```diff
@@ -2791,6 +2791,12 @@
 $g_roadmap_view_threshold = VIEWER;
 
 /**
+ * threshold for viewing timeline
+ * @global integer $g_view_timeline_threshold
+ */
+$g_view_timeline_threshold = VIEWER;
```
vboctor: I would rename this to timeline_view_threshold (similar to roadmap_view_threshold).

vboctor:
Thanks @schriftgestalt for your contribution.

:+1: once config option is renamed and documented.

schriftgestalt:
but what about the "view_changelog_threshold"?

schriftgestalt:
I added the documentation. I hope it is OK like this.

dregad:
>> I would rename this to timeline_view_threshold (similar to roadmap_view_threshold).

> but what about the "view_changelog_threshold"?

Sorry @vboctor but I think @schriftgestalt is right - in terms of consistency, we actually have more settings named $g_view_xxx_threshold than we have $g_xxx_view_threshold:

```
$ git grep ^\$.*view.*thresh -- config_defaults_inc.php
config_defaults_inc.php:1484:$g_view_summary_threshold = MANAGER;
config_defaults_inc.php:1601:$g_view_sponsorship_total_threshold = VIEWER;
config_defaults_inc.php:1608:$g_view_sponsorship_details_threshold = VIEWER;
config_defaults_inc.php:2102:$g_view_attachments_threshold = VIEWER;
config_defaults_inc.php:2539:$g_view_bug_threshold = VIEWER;
config_defaults_inc.php:2598:$g_view_handler_threshold = VIEWER;
config_defaults_inc.php:2605:$g_view_history_threshold = VIEWER;
config_defaults_inc.php:2655:$g_view_proj_doc_threshold = ANYBODY;
config_defaults_inc.php:2791:$g_roadmap_view_threshold = VIEWER;
config_defaults_inc.php:2803:$g_timeline_view_threshold = VIEWER;
config_defaults_inc.php:3020:$g_view_configuration_threshold = ADMINISTRATOR;
config_defaults_inc.php:3832:$g_tag_view_threshold = VIEWER;
config_defaults_inc.php:3896:$g_time_tracking_view_threshold = DEVELOPER;
config_defaults_inc.php:4014:$g_due_date_view_threshold = NOBODY;
```
(list edited to remove a few configs which match the regex but are actually not view thresholds)

That's 9x *view_xxx* vs 4x *xxx_view* (not counting the new timeline config). Based on that, I would rather have it like @schriftgestalt originally submitted.

Then we can discuss whether we should update $g_roadmap_view_threshold, $g_tag_view_threshold, $g_time_tracking_view_threshold and $g_due_date_view_threshold for consistency.

dregad:
That being said, I don't think this PR ready to merge as it is, because the display does not look good when timeline is switched off: we have this big white area on the right side of the screen.

screenshot-my_view

I think we should fall back to a 2-columns display with standard "my view" boxes "à la 1.2.x" when timeline is hidden.

vboctor:
I don't think the count is the deciding factor here.  It is what we want to consider as the old way of doing it vs. the new way of doing it.

I personally prefer the $g_feature_knob model since it enables nice natural grouping and auto-complete for all configs relating to a feature:

$g_roadmap_enabled
$g_roadmap_view_threshold
$g_roadmap_etc


rather than

$g_view_roadmap_threshold
$g_view_changelog_threshold
$g_etc_roadmap
$g_enabled_changelog
$g_enabled_roadmap

You get my point.

dregad:
OK fair enough, that makes sense - in that case let's do the update the other way around (I was just thinking about minimizing the number of obsoleted configs)

vboctor:
I've opened https://www.mantisbt.org/bugs/view.php?id=19630 to capture the discussion about changing existing names that follow old pattern.  I've also added comments on my perspective.  Short term, I think we should mainly focus on new configs.

atrol:
> Why is this feature not merged?

Because of https://github.com/mantisbt/mantisbt/pull/592#issuecomment-93680455 and because the PR has to be rebased on master-1.3.x branch.

vboctor:
Closing in favor of #976.

PR #594, opened 2015-04-08
Support adding a note + attachment in one step

vboctor:
Fixes #19589

### DIFF
--- a/bugnote_add_inc.php
+++ b/bugnote_add_inc.php
@@ -58,7 +58,7 @@ require_api( 'lang_api.php' );
 <?php
 	collapse_open( 'bugnote_add', '', 'form-container' );
 ?>
-<form id="bugnoteadd" method="post" action="bugnote_add.php">
+<form id="bugnoteadd" method="post" action="bugnote_add.php" enctype="multipart/form-data">
 	<?php echo form_security_field( 'bugnote_add' ) ?>
 	<input type="hidden" name="bug_id" value="<?php echo $f_bug_id ?>" />
 	<table>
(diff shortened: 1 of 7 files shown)


bugnote_add_inc.php
```diff
@@ -127,6 +127,34 @@
...
 
+	if( access_has_bug_level( config_get( 'upload_bug_file_threshold' ), $f_bug_id ) ) {
+		$t_file_upload_max_num = max( 1, config_get( 'file_upload_max_num' ) );
+		$t_max_file_size = (int)min( ini_get_number( 'upload_max_filesize' ), ini_get_number( 'post_max_size' ), config_get( 'max_file_size' ) );
+?>
+			<tr>
+				<th class="category">
+					<?php echo lang_get( $t_file_upload_max_num == 1 ? 'upload_file' : 'upload_files' ) ?>
+					<br />
+					<?php echo print_max_filesize( $t_max_file_size ); ?>
```
atrol: remove echo, print_max_filesize is void and does the echo itself
I am aware you copied the code from bug_file_upload_inc.php

dregad: If the code is indeed duplicated, then I think it should be moved into an API function instead.

vboctor:
I've done some re-factoring to drive consistency for the following:
- Detecting when files are not provided (before we were using null vs. -1).
- Checking for access level, now we always use method to check if user can upload rather than generic access check.
- Added a new method in file_api to process files.
- Removed the echo of the print method (was in the bug uploads, but not project doc uploads).

As for the duplication of the UI logic, I haven't refactored this for the following reasons:
- Modern UI has this changed.
- This code was already duplicated on bug report page, bug view page, and upload for project files.
- Supporting drag and drop attachments will change the html code.
- Labels are not exactly the same, sometime we use "select files" under title "upload files".  And sometimes when there is no time, we just use "upload file".
- The UI code is not exactly the same.

PR #595, opened 2015-04-11
Support multi-line default values for memo fields

vboctor:
- When type is text area enable setting multi-line default.
- When type is text area, don't enable possible values field.

Fixes #19542

### DIFF
--- a/core/custom_field_api.php
+++ b/core/custom_field_api.php
@@ -442,6 +442,11 @@ function custom_field_update( $p_field_id, array $p_def_array ) {
 			case 'possible_values':
 			case 'default_value':
 			case 'valid_regexp':
+				# Possible values doesn't apply to textarea fields
+				if( $p_def_array['type'] == CUSTOM_FIELD_TYPE_TEXTAREA && $t_field == 'possible_values' ) {
+					$t_value = '';
+				}
+
 				$t_update .= $t_field . '=' . db_param() . ', ';
 				$t_params[] = (string)$t_value;
 				break;
--- a/javascript/manage_custom_field_edit_page.js
+++ b/javascript/manage_custom_field_edit_page.js
@@ -1,0 +1,39 @@
+/*
+# Mantis - a php based bugtracking system
+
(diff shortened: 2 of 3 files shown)


core/custom_field_api.php
```diff
@@ -442,6 +442,11 @@ function custom_field_update( $p_field_id, array $p_def_array ) {
 			case 'possible_values':
 			case 'default_value':
 			case 'valid_regexp':
+				# Possible values doesn't apply to textarea fields
+				if ( $p_def_array['type'] == CUSTOM_FIELD_TYPE_TEXTAREA && $t_field == 'possible_values' ) {
```
dregad: No space before `(`

javascript/manage_custom_field_edit_page.js
```diff
@@ -0,0 +1,19 @@
+// Handle switching between textarea custom field type and other types
```
dregad: Add the standard header (see common.js)

Not sure what the strategy is wrt javascript files - in some cases we have dedicated files (e.g. login.js, bugFilter.js) and in others the code just goes into common.js. Maybe we should define that as part of our guidelines ?

In any case, would it not make sense to name the file in a somewhat more generic manner (e.g. custom_field.js) in case we later need to reuse it elsewhere ?
Should the file name be more specific

dregad:
A few minor comments, other than that +1 from me

javascript/manage_custom_field_edit_page.js
vboctor: I've added the header, though it seems like a big header for a javascript file that we want to send over the wire.

I left the name of the file as is, since it only applies to this page.  So I thought it makes more sense to reflect the name.  We can rename it later if we need to.

I think it is OK to have multiple files, but we don't want to have a single page including too many files.  If that is the case, then we may want to consider consolidation + compression + caching strategy.

vboctor:
Merged via 119aca1377a1c9770129f223d44f5e6b4bffc637

PR #596, opened 2015-04-16
Don't create FQDN link when the bug doesn't exist

dregad:
Fixes https://www.mantisbt.org/bugs/view.php?id=19606

Other commits are code cleanup.


vboctor:
:ship: looks good.

syncguru:
+1

PR #597, opened 2015-04-18
Protect against malicious REPORTERs

vboctor:
This is a proposal against enabling spammers to signup and spam
the bug tracker.  The changes limits the number of possible changes
by signed up users to 10 per hour.  Although the feature and limits are configurable.
This check doesn't apply when signup feature is not enabled or when users have access
level above the default was assigned on signup.

TODO: Update manual.
TODO: spam_api/config vs. antispam_api/config?

Issue #19259


vboctor:
@dregad @atrol @rombert This is a change to avoid spam like the one we just got.  We should apply something like this very quickly to avoid another incident.  Let me know your feedback.

core/history_api.php
```diff
@@ -157,6 +157,24 @@ function history_get_events_array( $p_bug_id, $p_user_id = null ) {
...
+ * @param  integer $p_duration_in_seconds The time window in seconds.
+ * @param  [type]  $p_user_id             The user id or null for logged in user.
+ * @return integer The number of changes done by user in the specified time window.
+ */
+function history_count_recent_events( $p_duration_in_seconds, $p_user_id = null ) {
+	$t_user_id = ( ( null === $p_user_id ) ? auth_get_current_user_id() : $p_user_id );
+
+	$t_params = array( db_now() - $p_duration_in_seconds );
+
+	$t_query = 'SELECT count(*) as event_count FROM {bug_history} WHERE date_modified > ' . db_param();
```
atrol: missing WHERE for $t_user_id

config_defaults_inc.php
```diff
@@ -383,6 +383,23 @@
  */
 $g_max_lost_password_in_progress_count = 3;
 
+#############
+# Anti-Spam #
+#############
+
+/**
+ * Max number of events to allow for users with default access level when signup is enabled.
+ * Use 0 for no limit.
```
dregad: Add 
```
 * @see $g_default_new_account_access_level
```

core/history_api.php
```diff
@@ -157,6 +157,24 @@ function history_get_events_array( $p_bug_id, $p_user_id = null ) {
 }
 
 /**
+ * Counts the number of changes done by the specified user within specified time window.
+ * @param  integer $p_duration_in_seconds The time window in seconds.
+ * @param  [type]  $p_user_id             The user id or null for logged in user.
+ * @return integer The number of changes done by user in the specified time window.
+ */
+function history_count_recent_events( $p_duration_in_seconds, $p_user_id = null ) {
```
dregad: Since this retrieves user-specific data, maybe rename it *history_count_user_recent_events* ?

core/spam_api.php
```diff
@@ -0,0 +1,44 @@
...
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+require_api( 'access_api.php' );
+require_api( 'history_api.php' );
+
+/**
+ * Checks if the current user is suspected to be a spammer.  This is useful to run before
+ * actions like adding issues or issue notes.  If the user is determined to demonstrated
+ * spammy behavior, then this method with trigger an error and exit the script.
+ */
```
dregad: Suggest rephase
```
* Triggers an error if the current user is suspected to be a spammer.
* This should be run before actions like adding issues or issue notes. If the
* user is determined to demonstrate spammy behavior, this method will trigger an
* error and exit the script.
```

core/spam_api.php
```diff
@@ -0,0 +1,44 @@
...
+
+	$t_spam_max_count = config_get( 'spam_max_event_count' );
+	if( $t_spam_max_count == 0 ) {
+		return;
+	}
+
+	if( history_count_recent_events( config_get( 'spam_time_windows_in_seconds' ) ) <= $t_spam_max_count ) {
+		return;
+	}
+
```
dregad:
```
	$t_spam_time_window = config_get( 'spam_time_windows_in_seconds' );
	if( history_count_recent_events( $t_spam_time_window ) <= $t_spam_max_count ) {
		return;
	}

	error_parameters( $t_spam_max_count, $t_spam_time_window / 60 )
```

(see revised error message string below)

lang/strings_english.txt
```diff
@@ -1718,3 +1718,5 @@ $MANTIS_ERROR[ERROR_DEPRECATED_SUPERSEDED] = 'Deprecated functionality: "%1$s",
 $MANTIS_ERROR[ERROR_DISPLAY_USER_ERROR_INLINE] = 'Warning: The system is configured to display MantisBT errors (E_USER_ERROR) inline. Program execution will continue; this may lead to system/data integrity issues.';
 $MANTIS_ERROR[ERROR_TYPE_MISMATCH] = 'Data Type mismatch. Enable detailed error messages for further information.';
 $MANTIS_ERROR[ERROR_BUG_CONFLICTING_EDIT] = 'This issue has been updated by another user, please return to the issue and submit your changes again.';
+$MANTIS_ERROR[ERROR_SPAM_SUSPECTED] = 'Exceeded allowed limit, blocked to avoid spam.';
```
dregad: Suggest rewording to make the error message more user friendly.

```
You have reached the allowed activity limit of %d events within the last %d minutes; your action has been blocked to avoid spam, please try again later.
```

I believe it would be even better if we indicated "please try again in xxxx minutes", but calculating that would either require another new API function to get the date of the user's earliest change within the spam window, or renaming  history_count_recent_events() to history_get_user_recent_events() and make it return an array.

dregad:
> TODO: spam_api/config vs. antispam_api/config?

I'd prefer antispam_api personally.

Please see my initial review comments inline, I will now perform some testing locally.

dregad:
This is more a fix for https://www.mantisbt.org/bugs/view.php?id=19261 than 19259 IMO.

core/spam_api.php
```diff
@@ -0,0 +1,44 @@
...
+# (at your option) any later version.
+#
+# MantisBT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
```
dregad: Missing PHPDoc block

/**
 * Antispam API
 *
 * @package CoreAPI
 * @subpackage AntispamAPI
 * @copyright Copyright 2015  MantisBT Team - mantisbt-dev@lists.sourceforge.net
 * @link http://www.mantisbt.org
 *
 * @uses access_api.php
 * @uses history_api.php
 */

dregad:
I think we need to add `spam_check()` calls for all mass-update actions (for example, in my test environment, a reporter could add tags without restrictions).

On the subject of mass-update, the check does not work properly: I set a spam_max_event_count = 3, but was able to add 4 bugnotes in one go. I believe the `<=` operator at line 39 in spam_check() should be changed to `<`.

While testing, I realized that some unique actions (from an end-user perspective) actually result in several history events. For example, copy a bug from the view page = 3 events:

- 1 entry in the original issue (BUG_CLONED_TO event)
- 2 entries in the new issue (NEW_BUG + BUG_CREATED_FROM events)

Maybe we should only consider selected categories of events in the count ?

core/history_api.php
vboctor: Renamed to history_count_user_recent_events.

core/spam_api.php
vboctor: Revised new function documentation as suggested.

core/spam_api.php
vboctor: Revised to pass parameters to include the parameters, but kept it in seconds.  If we allow seconds resolution, then we want to display it correctly in error messages.

lang/strings_english.txt
vboctor: We can add the ability to be more thorough in the error message later.  For now, the important point is to select the right threshold to not trip normal users while stopping spammers with minimal damage.

core/spam_api.php
vboctor: Added missing phpdoc block for new API.

core/history_api.php
vboctor: Good catch fixed.  It seems that I was already asleep while writing this :smile:

config_defaults_inc.php
vboctor: Added the see also reference.

vboctor:
We should think of this as the first checkin in a journey to fight spam.  Overtime we will have more checks.  Checks may be more complex. etc. etc.  But for now, we need at least have a very simple scenario.  Until we have a something we believe works, I would rather reduce the work about communicating how anti-spam works internally, otherwise, we will have to keep worrying about localization, etc.  But eventually we can do.

- See responses to specific comments.  Mostly applied all.
- tag creation now checks for spam.
- renamed api name and configs to antispam.
- updated the manual with a new section for antispam.
- fixed the off by one check for max limit.

As for the case where we are generating multiple event types per user action, we could consider skipping certain event types.  That probably makes the query a bit more expensive, but is doable.  I suspect in the future we can make improvements like that + checking volume of email generated + etc.  For now, I think we need to think about what a typical reporter does in an hour and make sure that doesn't hit our limit.  Whether it is reporting a couple of issues, or commenting on few issues.

I will get this in, and then we can do follow up checkins in our journey against anti-spam.

dregad:
OK, I'm fine with this approach. Don't forget to update the tracker on mantisbt.org after checkin ;-)

dregad:
And also keep that feature in mind for the 1.3 announcement

rombert:
+1 LGTM

PR #598, opened 2015-04-18
Fix required custom field not set on update

vboctor:
Users without write access to a required custom field can't update issue at all.

If there is a custom field that is marked as required-on-update with write access set to MANAGER.
Then developers who can update the issue, can no longer update it, otherwise,
they get error about required custom field not set.

Fixes #19634


dregad:
I'm OK with this on principle, but I think it would be good to inform the user that part of their update was rejected. I haven't given this much thought, and I'm not sure how easy that would be - maybe passing an array of errors to print_successful_redirect_to_bug() / print_successful_redirect().

vboctor:
OK, revise to just skip when user doesn't have write access and field is not set.  Otherwise, the behavior is as before where user will get an error message if they try to set a field that they don't have write access to.

PR #602, opened 2015-04-28
Fix SQL error when custom field is in filter/sort

vboctor:
If the same custom field is used for filtering to a value that is not "any" and is also used for sorting,
the generated query defines the same table alias twice causing a SQL error.
The fix is to use different aliasing prefix for filtering vs. sorting.

Fixes #19670

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -962,7 +962,9 @@ function filter_get_query_sort_data( array &$p_filter, $p_show_sticky, array $p_
 				$t_def = custom_field_get_definition( $t_custom_field_id );
 				$t_value_field = ( $t_def['type'] == CUSTOM_FIELD_TYPE_TEXTAREA ? 'text' : 'value' );
 				$c_cf_alias = 'custom_field_' . $t_custom_field_id;
-				$t_cf_table_alias = '{custom_field_string}_' . $t_custom_field_id;
+
+				# Distinguish filter table aliases from sort table aliases (see #19670)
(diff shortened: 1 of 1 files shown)


core/filter_api.php
```diff
@@ -962,7 +962,7 @@ function filter_get_query_sort_data( array &$p_filter, $p_show_sticky, array $p_
 				$t_def = custom_field_get_definition( $t_custom_field_id );
 				$t_value_field = ( $t_def['type'] == CUSTOM_FIELD_TYPE_TEXTAREA ? 'text' : 'value' );
 				$c_cf_alias = 'custom_field_' . $t_custom_field_id;
-				$t_cf_table_alias = '{custom_field_string}_' . $t_custom_field_id;
+				$t_cf_table_alias = '{custom_field_string}_sort_' . $t_custom_field_id;
```
atrol: Keep in mind that length of database object names are limited to 30 characters in Oracle.

vboctor: I'll replace '_sort_' with '_s'.

dregad:
To be on the safe side (since we don't know how users will set table pre- and suffix), and considering that we don't actually need the alias to contain the full table name, I would recommend `$t_cf_table_alias = 'cf_sort_' . $t_custom_field_id;`

Anyway, good catch, +1 from me.

vboctor:
@dregad I've applied your suggestion.  Thanks.

PR #603, opened 2015-05-01
Support attaching multiple files with same name

vboctor:
Fixes #19683

### DIFF
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -641,8 +641,19 @@ function file_add( $p_bug_id, array $p_file, $p_table = 'bug', $p_title = '', $p
 		trigger_error( ERROR_FILE_NOT_ALLOWED, ERROR );
 	}
 
-	if( !file_is_name_unique( $t_file_name, $p_bug_id ) ) {
-		trigger_error( ERROR_FILE_DUPLICATE, ERROR );
+	$t_org_filename = $t_file_name;
+	$t_suffix_id = 1;
+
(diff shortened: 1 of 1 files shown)


atrol:
Commented at
https://www.mantisbt.org/bugs/view.php?id=19683#c50682

dregad:
I didn't actually test the code, but I'm OK with this on principle; see also my comment in the tracker.

syncguru:
Looks good.
I wonder if the behavior should be consistent between these two cases:
(1) attaching multiple files with same name, 
(2) moving attachments between issues where the destination has a file with the same name

case 1 will rename new attachment (after this PR) while case 2 will error out

vboctor:
Good catch @syncguru, I've noticed this and hence I didn't remove the error message -- we don't currently support moving specific attachments from bug X to bug Y, so this is about the case where a bug is moved from project A to project B and the projects are associated with different upload folders.  For a long time now, we have been using hashes for filenames, so there shouldn't be collisions, but if there is a collision, then it is outside of the scope of this fix.

I would expect that when we have a collision it will be in the case of very old attachments which had names like "readme.txt" rather than a hash.  We should ideally at one point add an upgrade step that renames these to hashes.

As for the case where the hash conflicts, I doubt we need to handle this case, and if so, we should probably handle it in other code paths as well.

I've created the following bug to track this work:
https://www.mantisbt.org/bugs/view.php?id=19686

PR #604, opened 2015-05-02
Remove jquery mobile

dregad:
Just realized, as I was upgrading jQuery/jQueryUI to the latest versions, that we also include (an old version of) jQuery Mobile in 1.3. 

That was introduced in b256b8976d6a05a397bcee428b08bb2d2f168573.

As far as I can see, this is not used anywhere in the code, so unless you guys know what this is all about, I propose to remove the jQuery Mobile javascript and related CSS files.


atrol:
+1
You might also consider to remove library/ezc/Feed and library/ezc/Mail (not used, not in 1.2.x)

syncguru:
+1

rombert:
+1 . Might be worth mentioning in the release notes, in case someone was depending on it.

vboctor:
:+1:

dregad:
I opened https://www.mantisbt.org/bugs/view.php?id=19687 to track this removal. 

The same issue will be used to cover removal of ezc Feed and Mail as suggested by @atrol in https://github.com/mantisbt/mantisbt/pull/604#issuecomment-98379713

PR #605, opened 2015-05-04
Remove html_head_javascript() and relocate js to bottom of the page to speed up page rendering

syncguru:
Fixes #13285: Move script inclusions from HEAD to document footer


core/html_api.php
```diff
@@ -765,10 +746,21 @@ function html_footer() {
...
+	echo "\t" . '<script type="text/javascript" src="' . helper_mantis_url( 'javascript_config.php' ) . '"></script>' . "\n";
+	echo "\t" . '<script type="text/javascript" src="' . helper_mantis_url( 'javascript_translations.php' ) . '"></script>' . "\n";
+	html_javascript_link( 'jquery-1.11.3.min.js' );
+	html_javascript_link( 'jquery-ui-1.11.4.min.js' );
+	html_javascript_link( 'common.js' );
+	foreach ( $g_scripts_included as $t_script_path ) {
+		html_javascript_link( $t_script_path );
+	}
+
 	event_signal( 'EVENT_LAYOUT_BODY_END' );
```
vboctor: Should the body end event proceed the javascript, so the javascript is right before the closing body tag?

core/html_api.php
```diff
@@ -430,25 +427,9 @@ function require_js( $p_script_path ) {
...
-	html_javascript_link( 'jquery-ui-1.11.4.min.js' );
-	html_javascript_link( 'common.js' );
-	foreach ( $g_scripts_included as $t_script_path ) {
-		html_javascript_link( $t_script_path );
-	}
-}
 
 /**
- * (7) End the <head> section
+ * (6) End the <head> section
```
vboctor: I'm not sure these numbers add much value.  I suggest remove them.  We already have a place where we show the order all methods are called.  See line 31.

vboctor:
The title of the commit is too long.  The issue title is more compact and I think conveys the message.  We try to have commit titles be 50 characters or less.

Otherwise :+1: with some minor comments.

dregad:
Are we sure that this really has no side effects ? Are all of the cases indeed covered by $document.ready() ? I'm concerned about the jscalendar (we really need to replace this thing at some point in the near future !), and also what about plugins ? They could be using require_js() too (didn't check)

vboctor:
From [Best Practices for Speeding Up Your Web Site](https://developer.yahoo.com/performance/rules.html)

**Put Scripts at the Bottom**
The problem caused by scripts is that they block parallel downloads. The HTTP/1.1 specification suggests that browsers download no more than two components in parallel per hostname. If you serve your images from multiple hostnames, you can get more than two downloads to occur in parallel. While a script is downloading, however, the browser won't start any other downloads, even on different hostnames.

In some situations it's not easy to move scripts to the bottom. If, for example, the script uses document.write to insert part of the page's content, it can't be moved lower in the page. There might also be scoping issues. In many cases, there are ways to workaround these situations.

An alternative suggestion that often comes up is to use deferred scripts. The DEFER attribute indicates that the script does not contain document.write, and is a clue to browsers that they can continue rendering. Unfortunately, Firefox doesn't support the DEFER attribute. In Internet Explorer, the script may be deferred, but not as much as desired. If a script can be deferred, it can also be moved to the bottom of the page. That will make your web pages load faster.

dregad:
For the record: I'm not saying this shouldn't be done, I just want to make sure we're not introducing any regressions.

vboctor:
I don't have a position either, just posted the reference.  I suggest we come up with the list of features that we want to make sure work properly:

- jscalender for reports / due date picker
- filtering javascript (clicking field, switching filters, ...)
- custom field default value (string vs. text area switching).
- changing current project.

@dregad anything else you suggest checking?

syncguru:
@dregad the issues that result from moving JS files or changing their loading order usually too severe to go unnoticed, that said, a quick testing will reveal problems quite easily. As such I don't consider this change to be risky.

dregad:
Not sure what you did with the pull request here, but I suggest you rebase the branch again on top of latest master and push it again.

syncguru:
Not sure what went down either. Rebased the branch... should be good now.

vboctor:
Any reason why this change is not merged yet?

dregad:
No objections from me, as long as the 

> quick testing will reveal problems quite easily

has actually been performed...

syncguru:
Yes, this was tested. I need to update this PR to resolve conflicts though.

syncguru:
@vboctor @dregad - this PR is updated and tested.

vboctor:
:+1:

dregad:
Merged. Thanks Rafik !

PR #607, opened 2015-05-17
Status legend fixes

dregad:
- https://mantisbt.org/bugs/view.php?id=19573
- https://mantisbt.org/bugs/view.php?id=19725


core/bug_group_action_api.php
```diff
@@ -115,7 +115,7 @@ function bug_group_action_print_bug_list( array $p_bug_ids_array ) {
 	echo '</table>';
 	echo '</div>';
 
-	if( STATUS_LEGEND_POSITION_BOTTOM == $t_legend_position ) {
+	if( ( $t_legend_position & STATUS_LEGEND_POSITION_BOTTOM ) == STATUS_LEGEND_POSITION_BOTTOM) {
 		echo '<br />';
```
vboctor: This new line is not emitted on the other pages.  I wonder if we should just call into html_top_status_legend() and html_bottom_status_legend().  These would do the check and calling into the html_status_legend().

vboctor:
:+1: with the comment about having the logic in APIs rather than in 3 pages.

dregad:
Good point about moving the logic to the API.

But in that case, what about an alternative approach: adding a new "target location" parameter (i.e. top/bottom) and just let html_status_legend() handle all of the work instead of adding 2 new functions ?

dregad:
Added commits taking your comments into consideration. Let me know your feedback, esp. @vboctor regarding the API change.

vboctor:
:+1: looks good.

PR #613, opened 2015-06-24
config_flush_cache() now clears $g_cache_config_eval too

dregad:
The config api uses several cache variables. Currently the config_flush_cache() function only clears $g_cache_config and $g_cache_config_access.

It should also clear $g_cache_config_eval.

https://www.mantisbt.org/bugs/view.php?id=19881


atrol:
Commented at https://www.mantisbt.org/bugs/view.php?id=19881#c50990

vboctor:
:+1: The change looks good.

PR #614, opened 2015-07-06
Optimize timeline features to minimize time and sql load

vboctor:
This change uses now works directly on bug history table rows and only processes the number of rows
necessary to render the number of rows required in the UI.  This significantly reduces the load in the
default view which doesn't show all rows.  The big savings happen in busy weeks, with about parity for
normal weeks in terms of load.

**Date Range based on mantisbt.org dataset**
2014-12-05 .. 2014-12-12 [Prev] [Next]

**After Optimization (collapsed)**
Total queries executed: 54
Unique queries executed: 54
Total query execution time: 0.0746 seconds

**Before Optimization (collapsed)**
Total queries executed: 515
Unique queries executed: 515
Total query execution time: 0.2072 seconds

**After Optimization (all)**
Total queries executed: 514
Unique queries executed: 514
Total query execution time: 0.1747 seconds

**Before Optimization (all)**
Total queries executed: 515
Unique queries executed: 515
Total query execution time: 0.2086 seconds

Fixes [#18015](https://www.mantisbt.org/bugs/view.php?id=18015)

### DIFF
--- a/core/classes/TimelineEvent.class.php
+++ b/core/classes/TimelineEvent.class.php
@@ -30,43 +30,14 @@
 class TimelineEvent {
 	protected $timestamp;
 	protected $user_id;
-	protected $tie_breaker;
 
 	/**
 	 * @param integer $p_timestamp   Timestamp representing the time the event occurred.
 	 * @param integer $p_user_id     An user identifier.
-	 * @param boolean $p_tie_breaker A value to sort events by if timestamp matches (generally issue identifier).
 	 */
-	public function __construct( $p_timestamp, $p_user_id, $p_tie_breaker ) {
+	public function __construct( $p_timestamp, $p_user_id ) {
 		$this->timestamp = $p_timestamp;
 		$this->user_id = $p_user_id;
-		$this->tie_breaker = $p_tie_breaker;
-	}
-
-	/**
-	 * Comparision function for ordering of timeline events.
-	 * We compare first by timestamp, then by the tie_breaker field.
-	 * @param TimelineEvent $p_other An instance of TimelineEvent to compare against.
-	 * @return integer
-	 */
-	public function compare( TimelineEvent $p_other ) {
-		if( $this->timestamp < $p_other->timestamp ) {
-			return -1;
-		}
-
-		if( $this->timestamp > $p_other->timestamp ) {
-			return 1;
-		}
-
-		if( $this->tie_breaker < $p_other->tie_breaker ) {
-			return -1;
-		}
-
-		if( $this->tie_breaker > $p_other->tie_breaker ) {
-			return 1;
-		}
-
-		return 0;
 	}
 
 	/**
--- a/core/history_api.php
+++ b/core/history_api.php
@@ -185,6 +185,8 @@ function history_count_user_recent_events( $p_duration_in_seconds, $p_user_id = 
  * @param integer $p_start_time The start time to filter by, or null for all.
  * @param integer $p_end_time   The end time to filter by, or null for all.
  * @param integer $p_max_events The maximum number of events to return or null/0 for all.
+ * @param integer $p_sort_order The sort order ASC, DESC, or null to use default ordering
+ *                              as per config.
  * @return array
  */
 function history_get_raw_events_array( $p_bug_id, $p_user_id = null, $p_start_time = null, $p_end_time = null, $p_max_events = null, $p_sort_order = null ) {
@@ -327,7 +329,8 @@ function history_get_raw_events_array( $p_bug_id, $p_user_id = null, $p_start_ti
 			}
 		}
 
-		# if history is not scoped to a single bug, then include the bug id in the returned array for each element.
+		# if history is not scoped to a single bug, then include the bug id in the returned array
+		# for each element.
 		if ( $p_bug_id === null ) {
 			$t_raw_history[$j]['bug_id'] = $v_bug_id;
 		}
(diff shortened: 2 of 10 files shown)


core/classes/TimelineEvent.class.php
```diff
@@ -30,43 +30,14 @@
...
 	 */
-	public function __construct( $p_timestamp, $p_user_id, $p_tie_breaker ) {
+	public function __construct( $p_timestamp, $p_user_id ) {
 		$this->timestamp = $p_timestamp;
 		$this->user_id = $p_user_id;
-		$this->tie_breaker = $p_tie_breaker;
-	}
-
-	/**
-	 * Comparision function for ordering of timeline events.
```
dregad: Typo: *comparison*

core/history_api.php
```diff
@@ -305,6 +327,11 @@ function history_get_raw_events_array( $p_bug_id, $p_user_id = null, $p_start_ti
 			}
 		}
 
+		# if history is not scoped to a single bug, then include the bug id in the returned array for each element.
```
dregad: line break

core/timeline_inc.php
```diff
@@ -44,14 +43,11 @@
...
-$t_events = timeline_sort_events( $t_events );
 
-$t_num_events = timeline_print_events( $t_events, ( $f_all ? 0 : MAX_EVENTS ) );
+timeline_print_events( $t_events );
 
-# Don't display "More Events" link if there are no more entries to show
-# Note: as of 2015-01-19, this does not cover the case of entries excluded
-# by filtering (e.g. Status Change not in RESOLVED, CLOSED, REOPENED)
-if( !$f_all && $t_num_events < count( $t_events )) {
+# We will compromise accuracy of more link and showing exactly N without clicking more in favor or reducing load.
```
dregad: line break

core/classes/TimelineEvent.class.php
atrol: The line containing the typo has been removed in this commit.

core/history_api.php
```diff
@@ -179,14 +179,20 @@ function history_count_user_recent_events( $p_duration_in_seconds, $p_user_id =
...
+ *                           then returned array will have a field for bug_id, otherwise it won't.
  * @param integer $p_user_id A valid user identifier.
  * @param integer $p_start_time The start time to filter by, or null for all.
  * @param integer $p_end_time   The end time to filter by, or null for all.
+ * @param integer $p_max_events The maximum number of events to return or null/0 for all.
  * @return array
  */
-function history_get_raw_events_array( $p_bug_id, $p_user_id = null, $p_start_time = null, $p_end_time = null ) {
-	$t_history_order = config_get( 'history_order' );
+function history_get_raw_events_array( $p_bug_id, $p_user_id = null, $p_start_time = null, $p_end_time = null, $p_max_events = null, $p_sort_order = null ) {
```
atrol: Missing @param for $p_sort_order

core/classes/TimelineEvent.class.php
dregad: Duh. Please ignore me ;)

dregad:
Nice work. Review comments:

The change in history_get_raw_events_array() feels a bit like a hack, especially the bit about the returned array's structure changing depending on input parameters. Not a huge deal, but maybe it's a sign that we need a new, dedicated API function ?

I'm also concerned about what would happen for a low-privileged user in a "tight" environment where many history records could get filtered out... In such case, it may be worth having a loop to "fill" the timeline up to *MAX_EVENTS*, what do you think ?

And a few nitpicks:

- Commit d5bf4071b5d878b44a97fbcd668378837bdf14ca: *should*, not "shoudl"
- Commit 8c9114e9f5cf511f646e43fc6326272df74a7800 message: "to work directory on history" --> to work *directly* on history ? Or am I missing something ?
- Commit messages lines are too long... [Should be wrapped at 72 chars](https://www.mantisbt.org/wiki/doku.php/mantisbt:git_commit_messages)

core/history_api.php
vboctor: Not sure what you mean by a line break.  What length do you want to limit the line to?  It is shorter than the line above it.

core/timeline_inc.php
vboctor: Removal of the count here is due to the fact that the returned array me be less than 50, yet there is  more rows.  That is because we limit the number of rows returned from the database, then we do further filtering outside.  Unless we have a streaming implementation, we can't check the count here.  Also count is a best guess that there may be further rows, but is not guaranteed.  The best approach is to fetch count + 1, then check, which i wasn't doing anyway.

vboctor: I added a line break after 100 characters.

core/history_api.php
vboctor: I added a line break after 100 characters.

vboctor:
@dregad I've done further refactoring that includes the cleanup work and guarantees accuracy in terms of number of events in the timeline and when to show the "show more" link.

**After Optimization (default)**
Total queries executed: 151
Unique queries executed: 151
Total query execution time: 0.0988 seconds

**After optimization (all)**
Total queries executed: 565
Unique queries executed: 565
Total query execution time: 0.1901 seconds

I'm not sure of an easy way to go back and revise commits that are not on the top of the branch.  So unless you really care, I won't be revising text in these.

@atrol Your comments are also addressed.

dregad:
@vboctor thanks for the feedback and corrections. 

So it appears that with these additional changes, we're 
- slightly less efficient than the original code in terms of number of queries while performance remains about the same (slightly less in fact)
- 3 times more queries than initial fix with ~25ms longer execution time but I guess that's the expected price to pay for accuracy.

I'll try to have a more detailed look and give this a test run later on.

atrol:
Would be fine to consider not just query execution time but also total page execution time.

vboctor:
@dregad assuming that my comparison is accurate, there may be slight increase in # of queries / execution time as you noted.  Did you get a chance to test?  I would like to get this in.  We can always add further improvements later.

PR #615, opened 2015-07-14
Fix Excel errors for numeric cells with spaces

vboctor:
Fixes #19927

### DIFF
--- a/core/excel_api.php
+++ b/core/excel_api.php
@@ -160,16 +160,21 @@ function excel_get_default_filename() {
 }
 
 /**
- * Escapes the specified column value and includes it in a Cell Xml.
+ * Escapes the specified column value and includes it in a Cell Xml as a string.
  * @param string $p_value The value.
  * @return string The Cell Xml.
  */
 function excel_prepare_string( $p_value ) {
-	$t_type = is_numeric( $p_value ) ? 'Number' : 'String';
+	return excel_get_cell( $p_value, 'String' );
+}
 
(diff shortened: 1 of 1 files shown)


atrol:
Not sure if it's a good idea to change data when exporting.
I would expect that I get **exactly** the same data when exporting/importing. 
Let's also wait for feedback to https://www.mantisbt.org/bugs/view.php?id=19927#c51058

FavioGalvis:
I am not having this issue, tryed to duplicate #19927 unsuccesfully. 
Also tested the code and this not have any repercutions on my reports, so I suppose a trim on report will not do any harm

:+1:

vboctor:
I've updated the fix based on the discussion on the associated MantisBT issue.  Please test and confirm that everything is OK.

core/excel_api.php
```diff
@@ -185,6 +190,12 @@ function excel_prepare_string( $p_value ) {
  * @return string
  */
 function excel_get_cell( $p_value, $p_type, array $p_attributes = array() ) {
+	if ( !is_int( $p_value ) ) {
+		$t_value = str_replace( array( '&', "\n", '<', '>' ), array( '&amp;', '&#10;', '&lt;', '&gt;' ), trim( $p_value ) );
```
atrol: Is there still a need for the trim?
As mentioned before, I would expect that I get exactly the same data when exporting/importing.

vboctor:
> Is there still a need for the trim?

@atrol I removed the trim.

atrol:
now +1

PR #617, opened 2015-07-28
Master 1.2.x - gitignore - htaccess and error_log

holodyn:
For 1.2.x branch

Added .htaccess and error_log to the .gitignore files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Live Environment
 .htaccess
+error_log
 
 # Config files
 config_inc.php
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Live Environment
+.htaccess
+
 # Config files
 config_inc.php
 custom_constants_inc.php
@@ -38,7 +41,7 @@ plugins/*
 *.kdev4
 nbproject/
 
-# IIS7 
+# IIS7
 web.config
 
 #libraries


FavioGalvis:
:+1:

vboctor:
@holodyn Thanks for the contribution.  The changes seem to be mostly environment specific.  Also, in the future, please submit against master branch not 1.2.x.

- .htaccess - I don't see why we should exclude this.
- errorlog - this is environment specific
- space at end of line - not really a big deal.

:-1: for merging.

holodyn:
@vboctor Absolutely right, this is LAMP specific but pretty common in those environments.  Maybe a more appropriate header comment would have been:

````
# LAMP Environment
.htaccess
error_log
````

Allowing .htaccess is a bit of a hack to solve the problem with the `admin` folder.  In the .htaccess a rule can be placed to reject calls to /admin, so that the folder can remain, allowing for a clean `git status`.  You are already excluding things like config_inc.php for that same reason, so it seemed appropriate. 

The personal goal is allowing a repo to be deployed without having to maintain a unique branch.

**Comment by dregad** [comment]
While I understand why you'd want to deploy MantisBT from a git repo, please note for the record that the only officially supported way of installing the software is by using the release tarballs we publish.
Therefore, I'm also opposed to merging this.

**Comment by holodyn** [comment]
Unfortunate decision.  You may be interested in following that logic with the removal of web.config.

**Comment by atrol** [comment]
> following that logic with the removal of web.config.

@dregad, I agree with that.
If we don't allow .htaccess  we should also not allow web.config for IIS.

**Comment by dregad** [comment]
The problem with *.htaccess* is that we bundle a few of those in the repo, so we can't just exclude it globally like that:

```
$ find -name .htaccess
./config/.htaccess
./core/.htaccess
./doc/.htaccess
./lang/.htaccess
./library/.htaccess
./library/securimage/.htaccess
./library/securimage/database/.htaccess
./plugins/.htaccess
./scripts/.htaccess
```

The *web.config* exclusion was added by grangeway in 2008 (see 1968eb5d3f76eedb2d2fcf88b0bf6a33780b0a71), I don't know the rationale for that. I don't use IIS so I can't say for sure, but I am not sure why it would be needed.

**Comment by atrol** [comment]
Have a look at master branch where grangeway had to add
```
!library/securimage/web.config
```

I am fine with removing the web.config stuff in master to be consistent with .htaccess

PR #618, opened 2015-07-28
Master 1.2.x - Changelog obsolete toggle

holodyn:
Added a new configuration value for toggling obsolete version visibility in the changelog.

$g_show_changelog_obsolete = ON | OFF;

### DIFF
--- a/changelog_page.php
+++ b/changelog_page.php
@@ -154,7 +154,7 @@ foreach( $t_project_ids as $t_project_id ) {
 	$t_relation_table = db_get_table( 'mantis_bug_relationship_table' );
 
 	# grab version info for later use
-	$t_version_rows = version_get_all_rows( $t_project_id, /* released */ null, /* obsolete */ false );
+	$t_version_rows = version_get_all_rows( $t_project_id, /* released */ null, /* obsolete */ (config_get( 'show_changelog_obsolete' ) === ON ? null : false) );
 
 	# cache category info, but ignore the results for now
 	category_get_all_rows( $t_project_id );
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -989,6 +989,12 @@
 	$g_show_changelog_dates = ON;
 
 	/**
+	 * Show obsolete versions in the changelog list
+	 * @global int $g_changelog_show_obsolete
+	 */
+	$g_show_changelog_obsolete = OFF;
+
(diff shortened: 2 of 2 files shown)


vboctor:
Thanks @holodyn 

Here is the suggested feedback:
- Create an issue in the bug tracker and reference it from the commit description (e.g. Fixes #....).
- Submit the patch request against master branch.
- Update the manual with the new config documentation.

syncguru:
@holodyn please address @vboctor feedback and update this PR. Otherwise it looks good to me.

+1

dregad:
Do we really need 2 more configuration options (i.e. this one plus the one in #619) ?

While the requested feature makes complete sense and would be quite useful indeed, IMO it would be better to modify the changelog page to allow user to dynamically select what to display - a kind of filter, allowing this as well as other options such as a date or version range, etc.

atrol:
-1 for the PR

+1 for
> IMO it would be better to modify the changelog page to allow user to dynamically select what to display

$g_show_changelog_obsolete = ON would cause performance issues in some cases.
Thus you don't want to have it always on, but sometimes you want.
This can't be achieved by a global setting that can't be easyli changed and can't be changed by all users.

Maybe we would need something like
$g_view_obsolete_versions_threshold = DEVELOPER;
to avoid that each user is able to stress the server by setting the filter to include obsolete versions.

dregad:
I suggest that we park this and review it again after refactoring the Changelog and Roadmap as per https://www.mantisbt.org/bugs/view.php?id=9488. 

For the record, I have a work-in-progress branch for that, but it has been put on hold until 1.3 is released.

holodyn:
> IMO it would be better to modify the changelog page to allow user to dynamically select what to display

Having a filter would solve our issue, especially if we could configure the default settings for new visitors. Ultimately this would provide even more flexibility so a definite +1.

> $g_view_obsolete_versions_threshold = DEVELOPER;

While I can see a place for this, it would not work for our particular needs.  We are attempting to show the entire history of change, not just those releases that we still consider active / supported.

Pagination on the changelog seems like something worth consideration.  Regardless of how one organization may use the obsolete flag, the log could become very large over time.

holodyn:
Update - this feature has been moved into a plugin found here:
https://github.com/WebuddhaInc/mantisbt-plugin-wbquickreport

PR #620, opened 2015-07-29
parameter 'redirect_url' support

QuestorX:
Support parameter 'redirect_url' to divert to an alternate page after
deleting a relationship. Needed for plug-ins that manage and view
relationships. (e.g. RelationshipColumnView)

### DIFF
--- a/bug_relationship_delete.php
+++ b/bug_relationship_delete.php
@@ -125,8 +125,8 @@ if( bug_exists( $t_dest_bug_id ) ) {
 form_security_purge( 'bug_relationship_delete' );
 
 $f_redirect_url = gpc_get_string( 'redirect_url', '' );
-if( '' != $f_redirect_url ) {
-   print_header_redirect( $f_redirect_url );
+if( !is_blank( $f_redirect_url ) ) {
+	print_header_redirect( $f_redirect_url );
 } else {
-print_header_redirect_view( $f_bug_id );
+	print_header_redirect_view( $f_bug_id );
 }
--- a/bug_relationship_delete.php
+++ b/bug_relationship_delete.php
@@ -124,4 +124,9 @@ if( bug_exists( $t_dest_bug_id ) ) {
 
 form_security_purge( 'bug_relationship_delete' );
 
+$f_redirect_url = gpc_get_string( 'redirect_url', '' );
+if( '' != $f_redirect_url ) {
+   print_header_redirect( $f_redirect_url );
+} else {
 print_header_redirect_view( $f_bug_id );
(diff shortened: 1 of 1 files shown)


bug_relationship_delete.php
```diff
@@ -124,4 +124,9 @@
 
 form_security_purge( 'bug_relationship_delete' );
 
+$f_redirect_url = gpc_get_string( 'redirect_url', '' );
+if( '' != $f_redirect_url ) {
```
dregad: `is_blank( $f_redirect_url )`

dregad:
I'm fine with this, but please file an issue at http://mantisbt.org/bugs/ and reference it in your commit message (i.e. `Fixes #xxxx`).

QuestorX:
Fixes #0020009

QuestorX:
Fixes #0020009

grangeway:
@dregad: form security?

QuestorX:
The changes were made outside of the scope of the "Form security". Is this a problem?

===>
form_security_validate( 'bug_relationship_delete' );
   ...
form_security_purge( 'bug_relationship_delete' );
*/changes here/*
<===

atrol:
The problem is that $f_redirect_url is not sanitized.
So it can be used to redirect to unwanted pages and/or to inject JavaScript code.

No time to have a deeper look.
```php
print_header_redirect( $f_redirect_url, true, true );
```
might do the job.

atrol:
I thought about the PR in general.

a) Isn't the right way to implement events for plugins?

b) If someone has a look at the code and checks if $f_redirect_url is used in Mantis he will find no usage.
Thus, he might cleanup this as dead code.
Writing a comment is certainly no option
```php
# don't remove this as some of our users might use plugin RelationshipColumnView
```

QuestorX:
OK, you're right. The implementation here is currently only for my plugin. That makes absolutely no sense.
With my plugin deleting relationships from the 'View Issues' page must be working. Otherwise, the page 'View Issue Details' opens. Is there an alternative way for this? After deleting a relationship the 'View Issues' page must be opened again.

atrol:
> Is there an alternative way for this?

The only short term solution would be to copy most of the code from bug_relationship_delete.php to your plugin and to adjust it to your needs.

QuestorX:
Thank you for your support. I will follow the advice of atrol and implement my own Delete function. Please close the Issue with the number #20009.

atrol:
> Please close the Issue with the number #20009.

Done
Thanks @QuestorX for the feedback.

PR #621, opened 2015-07-29
Master 1.2.x - Create version on the fly during bug status change

holodyn:
This is a simple feature that allows for a new version to be defined when a bug resolution is being applied.  A new `Create Version` input field is shown to permitted users that will trigger the creation of a new project version, and apply the version to the bug along with the status change.

### DIFF
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -274,6 +274,34 @@ if ( ( $f_new_status >= $t_resolved ) ) {
 		<select name="fixed_in_version">
 			<?php print_version_option_list( $t_bug->fixed_in_version, $t_bug->project_id, VERSION_ALL ) ?>
 		</select>
+    <?php if( access_has_project_level( config_get( 'manage_project_threshold' ), $t_bug->project_id, $t_current_user_id ) ){ ?>
+      <div id="create_fixed_in_version">
+        <label>
+          <?php echo lang_get( 'create_fixed_in_version' ); ?>
+          <input type="text" name="create_fixed_in_version" maxlength="10" />
+        </label>
+        <label>
+          <?php echo lang_get( 'released' ); ?>
+          <input type="checkbox" name="create_fixed_in_version_released" />
+        </label>
+        <script>
+          var createFixedInVersionToggle = function(){
+            document.querySelector('select[name=fixed_in_version]').onchange = function(){
+              document.getElementById('create_fixed_in_version').style.display = (this.value == '' ? 'block' : 'none');
+            }
+            document.getElementById('create_fixed_in_version').style.display
+              = (document.querySelector('select[name=fixed_in_version]').value == '' ? 'block' : 'none');
+          }
+          (window.addEventListener
+            ? window.addEventListener('load', createFixedInVersionToggle, false)
+            : (window.attachEvent
+              ? window.attachEvent('onload', createFixedInVersionToggle)
+              : createFixedInVersionToggle()
+              )
+            );
+        </script>
+      </div>
+    <?php } ?>
 	</td>
 </tr>
 <?php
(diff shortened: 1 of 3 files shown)


atrol:
I noticed your 4 PR's since yesterday but don't have time at the moment to comment in a adequate way.
Maybe another member of the core team can invest some time for it.
I just want to let you know that we don't accept new features in master-1.2.x branch. Just bug fixes are allowed.

FavioGalvis:
:-1: I don't support this change, while this can be usefull for your specific needs, many other projects are monitored by large groups of users, and the creation or removal of one project must not be that easy for any manager.

Some workflows inside many entreprises have specific teams for the development of new projects or versions (we are one) and this will hurt the security of mantis.

If you turn it into a toggle option, default to OFF, then maybe.

holodyn:
@atrol - thank you for the note - Since these are changes for our production environment I am offering them to the 1.2.x branch.  Any that you feel positive about I am happy to implement in the 1.3.x branch.

@FavioGalvis - I'm not sure I understand your resistance.  This feature is only enabled for users with access to control versions already.  This is simply removing a series of clicks required for the new version to be created.  This does NOT control projects, only a new project version, so no creation or removal or a project is present here.

I agree an option in the configuration would enhance this feature, and I'm happy to add that.

FavioGalvis:
@holodyn Then if you add the option in the configuration im happy too. Its not me, i don't mind this change, its the structure in my organization (wich should not be the only one).

syncguru:
Hi @holodyn, thanks for your contribution.

This is definitely a shortcut that would save a particular set of users few clicks. As I think about it, I find myself wondering about other similar shortcuts. For example, adding project category on the fly. 

Another issue I struggled with is that version definition has more fields than just name & obsolete flag; there is released flag and description ..not to mentioned addition in future versions.  

I think adding new versions, categories, custom fields, users .. etc. are unfrequent enough to demand shortcuts to be sprinkled around the product.

-1

holodyn:
@syncguru I appreciate your points.  With many of our products a new build or patch version is prepared almost daily.  Updating versions is a real chore in the current system.  

 - Press Manage 
 - Press Manage Project
 - ** Scroll through several dozen versions for a spot on the page
 - Enter version number only
 - Press Add and Edit Version
 - Select the Release Status
 - Press Save
 - Press OK on Confirmation Screen
 - Find and Open the bug I started marking as complete when I realized we need to add a version

I realize for large teams this type of cycle may not feel out of the ordinary, but for a smaller team these steps can mean the difference between marking an actual version number, or just using a generic version 3.x that everything ends up lumped into.

To me this all seems extremely cumbersome in this day and age of streamlined applications.  There are several other areas of the system where similar usability features could be implemented that would imho greatly improve the usability of Mantis.

Regarding the comment about the other fields - honestly this approach was a bit taken from the current use case, which is to enter a version only and then select Add or Add and Edit.  I agree this could be better - ideally a modal that loads and allows complete version record creation with a refresh of the selection list on the preceding page on save.

atrol:
> a new build or patch version is prepared almost daily

I use the field "Product Build" for it. I don't like that much to have to select from hundreds of entries in the version list.

holodyn:
@atrol Thanks for the feedback.  I assume "Product Build" is a custom field?  In our particular case the goal / was to have patches paired with their issues on the changelog and even roadmap, and avoid using the `2.1.x` wildcard approach.  Maybe we need to just stick with the `x` identifier.

atrol:
>  I assume "Product Build" is a custom field?

No, it's a standard field.
You have to activate it by setting
```$g_enable_product_build = ON;```

But you won't get a separate section in changelog/roadmap for it.

holodyn:
Thank you for the clarification.

holodyn:
This doesn't appear to agree with the vision of the team.

PR #623, opened 2015-08-02
Set anonymous user's cookie when accessing a page directly

dregad:
This ensures consistent system behavior when pages are accessed directly
without prior "formal" login (i.e. via login_anon.php)

Fixes [#19964](https://www.mantisbt.org/bugs/view.php?id=19964)

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -652,6 +652,11 @@ function auth_get_current_user_cookie( $p_login_anonymous = true ) {
 
 						$g_cache_anonymous_user_cookie_string = $t_cookie;
 						current_user_set( $t_row['id'] );
+
+						# Set anonymous user's cookies to ensure consistent system
+						# behavior when pages are accessed directly without prior
+						# "formal" login (i.e. via login_anon.php)
+						auth_set_cookies( $t_row['id'], false );
 					}
 				}
 			} else {


syncguru:
+1

dregad:
I would appreciate if you guys could actually perform some testing on this, to ensure I did not break anything related to authentication or introduce any security issues.

vboctor:
In this area, the most problematic scenarios are cases where users are signing up, processing a forgot password or confirmation link, etc.  Did you test these scenarios?

dregad:
Thanks for the advice.

No I didn't - focused mainly on behavior in "normal" usage scenarios (as anonymous with/without login and as a registered user, switching back and forth between these), with and without clearing cookies).

I'll check these additional cases later.

badfiles:
The problem actually is not in setting cookies or not, I think the solution is to make Mantis think anonymous cookies are set even if they are not.

dregad:
@badfiles as far as I can tell, setting the cookies fixes the issue... (the existence of the cookies is essentially the only difference between a direct page browsing vs the same after anonymous login).

> I think the solution is to make Mantis think anonymous cookies are set even if they are not.

If you have a better fix than the one I proposed, then by all means submit it for review.

badfiles:
I will test your PR, thank you.

badfiles:
Delete all cookies.
Refresh page. No report link.
Refresh page again. See the report link. Without the patch you would see no report link again.

This should be done with report_bug_threshold = anonymous_level = viewer.

dregad:
@badfiles thanks for testing, but I'm afraid your report is not really clear... Is this patch fixing your problem or not ? Could the fact that you did not see the report issue link initially be due to caching ?

badfiles:
Caching is off when I'm testing :)
I expected such behavior. There are two scenarios with no cookies on client side:
1. user requests location which generates redirect.
2. user requests location which does not.

1. cookies are set with first request, so user gets the page system wants (default) with cookies already set.
2. cookies are set during request, system gives a page as if there are no cookies (there really are none).

The patch partly solves the issue, If a user goes directly to the default page, he has to refresh to see the page as an authenticated anonymous should see,
if a user is redirected to the default page, he becomes authenticated anonymous, because in fact there were two requests and on the second the cookies were already set.

l266509:

PR #624, opened 2015-08-02
Improve performance of diskfile_is_name_unique()

dregad:
Fixes [#19900](https://www.mantisbt.org/bugs/view.php?id=19900)


vboctor:
:+1:

PR #625, opened 2015-08-02
Fix hardcoded email titles

dregad:
Regression introduced by 03c5a5d18cfef07dec45a1025072981f24a5f07c.

Fixes [#19984](https://www.mantisbt.org/bugs/view.php?id=19984)

### DIFF
--- a/bug_update.php
+++ b/bug_update.php
@@ -417,10 +417,10 @@ helper_call_custom_function( 'issue_update_notify', array( $f_bug_id ) );
 
 # Send a notification of changes via email.
 if( $t_resolve_issue ) {
-	email_generic( $f_bug_id, 'resolved', 'The following issue has been RESOLVED.' );
+	email_generic( $f_bug_id, 'resolved', 'email_notification_title_for_status_bug_resolved' );
 	email_relationship_child_resolved( $f_bug_id );
 } else if( $t_close_issue ) {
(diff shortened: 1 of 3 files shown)


vboctor:
I thought we were planning to revert the original change, rather than fixing just these calls, is this still the plan?

atrol:
I think it's better this way.

We have this piece of code in bug_api.php
email_generic( $c_bug_id, $t_status, 'email_notification_title_for_status_bug_' . $t_status );
So we should deal the same way with it at other places instead of having an additional function for each status.

atrol:
@dregad, you missed this one
```
email_generic( $t_sponsorship->bug_id, 'sponsor', 'A sponsorship of the following issue was withdrawn.' );
```

dregad:
@vboctor as mentioned in 03c5a5d18cfef07dec45a1025072981f24a5f07c I forgot about that discussion. That being said, I am not opposed to simply reverting, although like Roland I believe it makes sense to reduce the number of API calls.

Since as far as I can remember you never really explained what the *direction of where you would like notifications to go* is, I can't say whether it would be preferable to simply revert Paul's commit, or just merge that PR.

vboctor:
I've submitted a pull request with reverting the original change - see https://github.com/mantisbt/mantisbt/pull/626

The PR has an explanation of the reasons.  Happy to discuss further.

dregad:
Closing this in favor of #626.

PR #626, opened 2015-08-05
Revert email_* calls to not use generic API

vboctor:
It is undesirable to use this approach for the following:
- Having each caller having to know the right language string + call type.
- Having a properly named APIs provides better intelli-sense experience.
- It is easier to add event spent logic or extra parameters compared to switch statements in generic API.
- More inline with the approach of having different email formats based on the change event, rather than a single format which dumps the whole issue, rather than focus on the actual change.

Fixes #19984

This reverts commit 03c5a5d18cfef07dec45a1025072981f24a5f07c.

Conflicts:
- bug_update.php
- core/bugnote_api.php
- library/disposable
- library/ezc/Base
- library/ezc/Graph
- library/securimage

### DIFF
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -830,7 +830,7 @@ function mc_issue_add( $p_username, $p_password, stdClass $p_issue ) {
 		mci_tag_set_for_issue( $t_issue_id, $p_issue['tags'], $t_user_id );
 	}
 
-	email_new_bug( $t_issue_id );
+	email_bug_added( $t_issue_id );
 
 	if( $t_bug_data->status != config_get( 'bug_submit_status' ) ) {
 		history_log_event( $t_issue_id, 'status', config_get( 'bug_submit_status' ) );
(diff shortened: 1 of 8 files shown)


vboctor:
Also see the PR with the prototype for re-designing notifications that I shared earlier at:
https://github.com/mantisbt/mantisbt/pull/101

dregad:
Having read your arguments, I closed #625 in favor of the revert.

+1

atrol:
Would be fine if remaining email_generic calls are also replaced where possible (e.g. in function mc_issue_add, ...)

Adding
```
/**
 * send notices when a bug is updated
 * @param int $p_bug_id
 * @return null
 */
function email_updated( $p_bug_id ) {
	email_generic( $p_bug_id, 'updated', 'email_notification_title_for_action_bug_updated' );
}
```
would allow to remove some more email_generic calls.

vboctor:
@atrol I've pushed some more changes.  Now email_generic() is only called from within email_api.php.

atrol:
You changed my proposed email_updated to email_bug_updated.
I considered also this name but didn't find a clear rule.
E.g. should email_new_bug be changed to email_bug_submitted?
should email_reopen be changed to email_bug_reopened?

Maybe too much for this PR and not the right time for this kind of cleanup.
Feel free to merge the current version.

dregad:
I think that, while we're at it, improving consistency in the API names is worth doing.

PR #629, opened 2015-08-09
Review group status change

badfiles:
Fixes #16620


bug_actiongroup.php
```diff
@@ -206,19 +208,27 @@
...
+							# No need to call email_generic(), bugnote_add() does it
+						}
+						helper_call_custom_function( 'issue_update_notify', array( $t_bug_id ) );
 					} else {
-						email_bug_updated( $t_bug_id );
+					 	if( !is_blank( $f_bug_notetext ) ) {
+							bugnote_add( $t_bug_id, $f_bug_notetext, null, $f_bug_noteprivate );
+							helper_call_custom_function( 'issue_update_notify', array( $t_bug_id ) );
+						}
+						$t_failed_ids[$t_bug_id] = lang_get( 'bug_actiongroup_same_status' );
```
vboctor: Why do we consider the case where an issue is set to the same status as a failure?  The goal is to have a group in state X and if some are already there, then that is already a success.  I would remove this + the associated string.

bug_actiongroup.php
```diff
@@ -206,19 +208,27 @@
...
+						bug_set_field( $t_bug_id, 'status', $f_status );
+						if( is_blank( $f_bug_notetext ) ) {
+							$t_new_status_label = MantisEnum::getLabel( config_get( 'status_enum_string' ), $f_status );
+							$t_new_status_label = str_replace( ' ', '_', $t_new_status_label );
+							email_bug_status_changed( $t_bug_id, $t_new_status_label );
+						} else {
+							bugnote_add( $t_bug_id, $f_bug_notetext, null, $f_bug_noteprivate );
+							# No need to call email_generic(), bugnote_add() does it
+						}
+						helper_call_custom_function( 'issue_update_notify', array( $t_bug_id ) );
```
vboctor: Should we also call event_signal( 'EVENT_UPDATE_BUG', ...)?

bug_actiongroup.php
```diff
@@ -206,19 +208,27 @@
...
-						bugnote_add( $t_bug_id, $f_bug_notetext, null, $f_bug_noteprivate );
-						# No need to call email_generic(), bugnote_add() does it
+					if( $t_status !== $f_status ) {
+						bug_set_field( $t_bug_id, 'status', $f_status );
+						if( is_blank( $f_bug_notetext ) ) {
+							$t_new_status_label = MantisEnum::getLabel( config_get( 'status_enum_string' ), $f_status );
+							$t_new_status_label = str_replace( ' ', '_', $t_new_status_label );
+							email_bug_status_changed( $t_bug_id, $t_new_status_label );
+						} else {
+							bugnote_add( $t_bug_id, $f_bug_notetext, null, $f_bug_noteprivate );
```
vboctor: If a user provides a note + change status, I would rather get the email for "issue has been resolved" than a "note has been added'.  Specially that the flow we are in right now is the "Change Status" flow.  So I would trigger the status change email anyway and not trigger the note added email notification.

bug_actiongroup.php
```diff
@@ -206,19 +208,27 @@
 		case 'UP_STATUS':
 			$f_status = gpc_get_int( 'status' );
 			if( access_has_bug_level( access_get_status_threshold( $f_status, $t_bug->project_id ), $t_bug_id ) ) {
-				if( true == bug_check_workflow( $t_status, $f_status ) ) {
+				if( true == bug_check_workflow( $t_status, $f_status ) ||
+					access_compare_level( current_user_get_access_level(), config_get_access( 'status_enum_workflow' ) ) ) {
```
vboctor: Is this access check different than the one in the if statement above?

bug_actiongroup.php
```diff
@@ -206,19 +208,27 @@
...
+						} else {
+							bugnote_add( $t_bug_id, $f_bug_notetext, null, $f_bug_noteprivate );
+							# No need to call email_generic(), bugnote_add() does it
+						}
+						helper_call_custom_function( 'issue_update_notify', array( $t_bug_id ) );
 					} else {
-						email_bug_updated( $t_bug_id );
+					 	if( !is_blank( $f_bug_notetext ) ) {
+							bugnote_add( $t_bug_id, $f_bug_notetext, null, $f_bug_noteprivate );
+							helper_call_custom_function( 'issue_update_notify', array( $t_bug_id ) );
```
vboctor: Same here relating to event signal.

vboctor:
Thanks @badfiles - the change looks good.  I've added some comments to address.  I suggest updating the commit name to reflect the issue being fixed (e.g. "Fix email notification for group status change" + referencing the fixed issue "Fixes #16620" in the commit message body.  You may also want to squash the commits.

bug_actiongroup.php
badfiles: This check allows a user who can alter workflow set status not allowed in workflow settings.

bug_actiongroup.php
badfiles: I think one of the goals of group status change might be email notification.
A group-status-changer should be informed which selected issues have not been updated and thus notifications have not been sent.

bug_actiongroup.php
badfiles: I followed previous logic. The proposed is much better.

bug_actiongroup.php
badfiles: I'm not really sure I have enough knowledge to add this call. Is this call supposed to include previous and new bugdata?

event_signal( 'EVENT_BUG_ACTION',... called right after switch...case.

badfiles:
Done

dregad:
I don't think it's right to send a notification for something that *didn't happen* (i.e. the new 'You tried to set status the issue was already in' message).

In my opinion, if the status didn't change, then there should be no notification unless a note was added, in which case the e-mail should be about a new note.

badfiles:
As far as I can remember the logic is:
If status has not been changed and a comment added --> send only comment notification and show 'same status' result
if status has been changed --> send status change notification (include comment if any)
if status has not been changed and no comment --> show 'same status result'

There is no 'same status' email notification.

dregad:
Thanks for the feedback. Maybe I was confused - I'll re-test this.

bug_actiongroup.php
vboctor: Wouldn't that confuse the person who is setting the workflow or not allow them to test what other users would experience?  I would rather not have a different there.

bug_actiongroup.php
```diff
@@ -202,20 +204,26 @@
...
+						email_bug_status_changed( $t_bug_id, $t_new_status_label );
 					} else {
-						email_bug_updated( $t_bug_id );
+						$t_failed_ids[$t_bug_id] = lang_get( 'bug_actiongroup_same_status' );
+					}
+					if( $t_status_changed || !is_blank( $f_bug_notetext ) ) {
+						helper_call_custom_function( 'issue_update_notify', array( $t_bug_id ) );
 					}
-
-					helper_call_custom_function( 'issue_update_notify', array( $t_bug_id ) );
```
vboctor: Why did you remove the issue_update_notify?  We used to trigger this before and trigger this in other events in the same group action file (e.g. update category).

badfiles: I did not remove it, instead I call it when there actually is an update.

vboctor: You are right, I missed that.

lang/strings_english.txt
```diff
@@ -1784,4 +1785,4 @@ $s_dropzone_cancel_upload_confirmation = 'Are you sure you want to cancel this u
 $s_dropzone_remove_file = 'Remove file';
 $s_dropzone_remove_file_confirmation = '';
 $s_dropzone_max_files_exceeded = 'You cannot upload any more files.';
-$s_dropzone_not_supported = 'Dropzone.js does not support older browsers!';
\ No newline at end of file
+$s_dropzone_not_supported = 'Dropzone.js does not support older browsers!';
```
k1931547: identical line

l266509:

badfiles:
Rebased to latest master

badfiles:
Any reason why this one does not get merged?

atrol:
> Any reason why this one does not get merged?

I am missing feedback / change for https://github.com/mantisbt/mantisbt/pull/629#discussion_r91238911

bug_actiongroup.php
badfiles: Other user workflow experience is being tested on an individual issue page.

I believe the group status change feature is more likely to be used by an experienced developer, let alone a person having privileges to change workflow should not get confused here.

atrol: > Wouldn't that confuse the person who is setting the workflow or not allow them to test what other users would experience?

Right, but @vboctor are you aware that there is no checking when using SOAP/REST API?
This can be even more confusing.

> I would rather not have a different there.

👍 at least there will be no confusion if just the UI is used for changes.

In future, all high level checks could be removed if the the workflow check takes place in function validate in a new command IssueStatusChangeCommand.

PR #630, opened 2015-08-12
Fix Copied bug attachments have wrong ownership

dregad:
https://www.mantisbt.org/bugs/view.php?id=20018

### DIFF
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -1068,7 +1068,7 @@ function file_copy_attachments( $p_source_bug_id, $p_dest_bug_id ) {
 		db_query( $t_query, array(
 			$p_dest_bug_id, $t_bug_file['title'], $t_bug_file['description'],
 			$t_new_diskfile_name, $t_new_file_name, $t_file_path,
-			$t_bug_file['filesize'], $t_bug_file['file_type'], $t_bug_file['date_added'],
+			$t_bug_file['filesize'], $t_bug_file['file_type'], db_now(),
 			auth_get_current_user_id(), $t_bug_file['content']
 		) );
 	}
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -55,7 +55,7 @@ $g_cache_file_count = array();
 /**
  * Processes the post files from a form by adding them to the specified
  * issue.
- * 
+ *
  * @param int $p_bug_id    The bug id.
  * @param array $p_files   The array of files, if null, then do nothing.
  */
@@ -1016,10 +1016,9 @@ function file_move_bug_attachments( $p_bug_id, $p_project_id_to ) {
 }
 
 /**
- *
  * Copies all attachments from the source bug to the destination bug
  *
- * <p>Does not perform history logging and does not perform access checks.</p>
+ * Does not perform history logging and does not perform access checks.
  *
  * @param integer $p_source_bug_id Source Bug.
  * @param integer $p_dest_bug_id   Destination Bug.
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -1055,19 +1055,22 @@ function file_copy_attachments( $p_source_bug_id, $p_dest_bug_id ) {
 			}
 		}
 
-		$t_query = 'INSERT INTO {bug_file} 
-							( bug_id, title, description, diskfile, filename, folder, filesize, file_type, date_added, content )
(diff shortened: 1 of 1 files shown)


grangeway:
Regarding the date_added -> I'd be cautious there. I can imagine users wanting both options, but would probably be inclined to think that the leaving it with the original date might be the safest option:

If a user reported a bug, and added an attachment with a log file and then some additional bugs/files were added - and at some point in time, we decided to copy the bug, I'd want the dates to remain the original dates (I think) - this would then mean that the files remain in date order, as opposed to have the *same* date i.e. now(). A bug with 10 attachments in this context could get confusing, and in this context keeping the original attachment dates probably makes more sense

On the flip side, however, it doesn't make sense to have attachments predating the bug. And in the context of a bug with a single attachment being cloned, having both the original date of the attachment, and the current date (with the current date making more sense) is probably appropriate.

Seperately to that, it might be safer to set the ownership of the file to that of the original owner, i.e. $t_bug_file['user_id'] as opposed to current authenticated user.

atrol:
> it might be safer to set the ownership of the file to that of the original owner

I asked for it the way it's implemented, see https://www.mantisbt.org/bugs/view.php?id=20018#c51228

If you want it another way, you have to think about the following scenarios and maybe some more:
- Clone an issue where attachments have been added by deleted/deactived user(s)
- Use "Copy" on "View Issues" page and select a target project for the copy where the original user(s) that added the attachment(s) have no access.

atrol:
> this would then mean that the files remain in date order

We could order the display by id instead of date_added

grangeway:
:) as I say, I think one can argue both ways around. Depending how you view the word "copy/clone" - if I print out a mantis issue and photocopy it, dates etc remain unchanged. My only reason for commenting on the attachment date was for the simple reason that if someone uploads a file - whilst it's been attached to a new bug, it's still the original file.

If I clone an issue in a tracker that 5 attachments, then added an attachment, once you wipe out the dates, it's not immediately clear whether there's 0, 1 or 5 new attachments as the dates of all would be new.

My comment regarding t_bug_file vs authenticated user was due to us saying that we don't do any access checks when copying. IF I have access to a bug, but not to view the attachments on the bug, but have access to copy the bug, then with this change, after copying the bug i'd have access to the attachments due to owning the attachments. That is why i commented - whereas in your example scenario's, I'd have cloned an issue, where the attachments were still listed as being by a deleted user (not an issue assuming I can see attachment) and if the original user didn't have access... well.. that's the same as if ownership is changed and they no longer have access.

In addtion: 
The wider question here is what you expect "cloning" or "copying" an issue to do. If I print out a mantis issue, photocopy it and give it to someone else the data and the times etc stay the same. I think for the bug itself, and for the amount of times i'd expect "copy" to get used - i think we've probably got reasonable behaviour. 

The copy bug notes side, currently keeps the original submitted date and the original author. 

The copy attachment side at the moment keeps the original date and nukes the author.

With this change, the copy attachment side would change the date and replace the author. This is then the opposite behaviour to the copy bugnote's functionality.

atrol:
I agree that attachments and bug notes should be treated the same way.
BTW, we should also display the user id for attachments the same way we do for notes.  

> The copy bug notes side, currently keeps the original submitted date

true when copying
not true when cloning

> we don't do any access checks when copying

true for attachments and bug notes when copying
true for attachments when cloning
not true for bug notes when cloning

Of course, my idea to set user_id = auth_get_current_user_id() should not be implemented at the moment as it would introduce a security issue.

Seems there is some more cleanup needed.

dregad:
Thanks Paul for your valuable insights. At lunch time yesterday this seemed like a simple and straightforward issue, but it would appear I opened a can of worms ;-)

I propose to close (reject) this PR as it currently stands; instead I'll submit a more focused fix that limits itself to addressing the issue 20018 as originally reported, i.e. preventing *bug_file_table.user_id* from being set to 0. In this context, I believe that setting it to the source bug's attachment owner (without access check) is the right way to go.

Thoughts ?

atrol:
+1 for first focusing on the obvious bug (user_id = 0)

vboctor:
:+1: for fixing the user id being 0 issue.  For the other aspects, we can open another tracking issue that points to this discussion.

dregad:
Closing this as per discussion.

PR #632, opened 2015-08-13
Fix Copied bug attachments have wrong ownership (v2)

dregad:
This is a more targeted fix for https://www.mantisbt.org/bugs/view.php?id=20018 following up on discussion in #630.

The fix just copies whatever user_id is in the source issue as it is. If that is already set to 0, then the cloned issue's attachment(s) will have the same, which I think is the desired behavior.


vboctor:
:+1:

atrol:
+1

PR #633, opened 2015-08-15
Load jquery from CDN

vboctor:
This is to improve performance for the following reasons:

- Browser loads more in parallel due to loading from different servers.
- CDN libraries likely to be already cached as it is referenced by other websites / web apps.
- CDN will deliver lower latencies with the possible exception of intranet.

Fixes #19932

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -3410,12 +3410,20 @@ $g_css_include_file = 'default.css';
  */
 $g_css_rtl_include_file = 'rtl.css';
 
-
 /**
  * meta tags
  * @global string $g_meta_include_file
  */
 $g_meta_include_file = '';
+
+/**
+ * A flag that indicates whether to use CDN (content delivery networks) for loading
+ * javascript libraries and their associated CSS.  This improves performance for
+ * loading MantisBT pages.  This can be disabled if it is desired that MantisBT
+ * doesn't reach out outside corporate network.
+ * @global integer $g_cdn_enabled
+ */
+$g_cdn_enabled = OFF;
 
 ################
 # Redirections #
--- a/core/constant_inc.php
+++ b/core/constant_inc.php
@@ -590,3 +590,7 @@ define( 'LOG_SOAP', 64 );                           # LOG_WEBSERVICE
 define( 'FTP', 1 );                                 # DISK
 define( 'ERROR_FTP_CONNECT_ERROR', 16 );            # N/A
 
+# JQuery and JQuery UI
+define ( 'JQUERY_VERSION', '1.11.3' );
+define ( 'JQUERY_UI_VERSION', '1.11.4' );
+
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -365,7 +365,13 @@ function require_css( $p_stylesheet_path ) {
 function html_css() {
 	global $g_stylesheets_included;
 	html_css_link( config_get( 'css_include_file' ) );
-	html_css_link( 'jquery-ui-1.11.4.min.css' );
+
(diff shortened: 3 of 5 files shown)


atrol:
This is a quite hardcoded approach in terms of the CDN (https://ajax.googleapis.com) and in terms of the libraries that are supported (maybe we will use more libraries in future, e.g. AngularJS)

What about having an option $g_cdn instead of $g_cdn_enabled that could be set to OFF or to the CDN url, e.g. https://ajax.googleapis.com?

If the CDN is limited to jQuery the option should be $g_jQuery_cdn

It might also be worth to have two more options for the used versions to reduce redundant numbers.
```php
$g_jQuery_ui_version = '1.11.4';
$g_jQuery_version = '1.11.3';
```

badfiles:
I think it's best to remove the protocol from the link and let browser decide which one to use like
https://ajax.googleapis.com/...  --> //ajax.googleapis.com/...

vboctor:
@atrol I don't see how this approach is hard-coded:
- The config flag applies to all javascript libraries and their associated css - not jquery only.  As we add more libraries, we would have the same wrapping check that will determine whether to load them from local code or from the CDN.
- The config flag enables us to use any CDN to retrieve libraries.  It even enables using more than one CDN based on availability of the libraries.  As we add CDNs, we will need to white list them, similar to what we do with other services like gravatar.
- I'm not sure whether all libraries that we would want to use would exist on all CDNs, and would exist under the same path, hence, having just a CDN URL in the config and concatenating paths to it, may not work.

As for the jquery version config options, I don't see a benefit from having them as a config option, but if you think it is worth it, we can add them as constants.  I believe these are part of the code and the users shouldn't be changing them via configuration, otherwise, it may break some functionality or jQuery plugins.

vboctor:
@atrol I've added the constants.
@badfiles I've used protocol-less urls.

atrol:
> @atrol I've added the constants.

Thanks, that's better than my idea with the configs.

atrol:
> I'm not sure whether all libraries that we would want to use would exist on all CDNs, and would exist under the same path, hence, having just a CDN URL in the config and concatenating paths to it, may not work.

That's why is asked for an option like $g_jQuery_cdn = '//ajax.googleapis.com/ajax/libs';

vboctor:
> That's why is asked for an option like $g_jQuery_cdn = '//ajax.googleapis.com/ajax/libs';

@atrol I'm not sure I see the value for adding such a config compared to just using a popular CDN for users.  Even if we provide the configuration option, then we will eventually have N of these options, one per library.  I would rather hold off on that until we really have good reason and feedback for adding such complexity.

syncguru:
+1  - default should be ON

In my opinion, this is getting way over thought.  CDNs for public use are very few and limited in scope. So, it should be per library and without all that customizations. I have hard time seeing the value of the cd_enabled defaulted to OFF, are we catering defaults for the majority of users or the few!?

vboctor:
I'm of the opinion that it should be ON by default too, I think for majority of users it will be faster or equivalent.  For intranet scenarios, the users will cache the scripts and won't hit any latency going forward.  But their customers would benefit from having the CDN.  I meant to add a note about this and forgot.  The default being OFF was based on the discussion in the bug which we didn't really finish.

Now if for some reason an admin has a reason why not to use CDN, he will be motivated to find the config option and change it.  But out of the box, we should optimize for the common case.

atrol:
I still prefer to have the default set to OFF
I updated my pros/cons
https://www.mantisbt.org/bugs/view.php?id=19932#c51067
I hope this is enough to convince you.

Keep also in mind that there are countries that block IP's.

PR #88, opened 2013-06-11
additions to api soap, user add, sub projects (#0010349)

wepl:
new functions to add user and assign to sub projects

PR #636, opened 2015-08-25
Fix email validation when user email has a space

vboctor:
Fixes #20053

### DIFF
--- a/manage_user_create.php
+++ b/manage_user_create.php
@@ -87,6 +87,7 @@ if( $f_password != $f_password_verify ) {
 	trigger_error( ERROR_USER_CREATE_PASSWORD_MISMATCH, ERROR );
 }
 
+$f_email = trim( $f_email );
 email_ensure_not_disposable( $f_email );
 
 if( ( ON == config_get( 'send_reset_password' ) ) && ( ON == config_get( 'enable_email_notification' ) ) ) {


dregad:
+1 but if you don't mind rewording the commit message, I initially read it as accepting spaces within the address and went 'what?' until I saw in the code that it was about trimming

atrol:
-1
IMO not the right place to trim as low level functions shouldn't manipulate input data.

We have already $p_email = trim( $p_email ) in function user_signup, so at least the most common use case doesn't need any additional trim.

vboctor:
Having the trimming in the low level API makes sure it is applied independent from the user creation code path: e.g. user creation from signup, administration, csv import, soap api, etc.  We could remove the trimming in email_signup() since it is now redundant.  This also makes the trimming and validation are called from the same API.

If we are not going to trim in user_create(), then why would it make sense to trim in any other API (e.g. user_signup).  At this point, we might as well duplicate this work in N scenarios that require trimming of user input.

I don't really see a disadvantage of trimming, since it is not really a manipulation of the actual data.  Logically there are never email addresses with spaces.

If you still disagree, also provide more clarity on exact locations you would like to see the trimming logic.

atrol:
I agree that this is a quite convenient way to fix this special single issue.
But you will hardly be able to convince me that this is a clean solution ;-)

I think this is the right order
1) Input data
2) Change data if needed for user convenience (e.g. trim)
3) Check access rights
4) Validate data
5) Store data

> We could remove the trimming in email_signup() since it is now redundant

After that we have
1) Input data
3) Check access rights
4) Validate data
2) Change data if needed for user convenience (e.g. trim)
5) Store data

This means that you validate the data and change it afterwards.
This will introduce some kind of backdoor.
I didn't try, but the following steps might work:
- Enter e-mail adress "xxx@disposable.com "
- email_ensure_not_disposable will not trigger a problem 
- "xxx@disposable.com " gets trimmed to "xxx@disposable.com"
- "xxx@disposable.com" gets stored

I don't think that having it this way is the right solution for it
1) Input data
2) Change data if needed for user convenience (e.g. trim)
3) Check access rights
4) Validate data
2) Change data if needed for user convenience (e.g. trim)
5) Store data

The main problem is, that we allow to call 5) without being sure that 3) and 4)
have been called before and without being sure that 2) is not called after 3) or 4)

Concerning the mentioned SOAP: We don't have an API to create users at the moment.
But there is a PR for it, have a look at my comment
https://github.com/mantisbt/mantisbt/pull/88#issuecomment-111431160
and the following one where I just added
"Forgot to mention that there is also no validation of $p_user['email'] in terms of email_ensure_not_disposable()."

All in all, I think that your small change makes things even worse than they are at the moment.
We should at least not break this rule: Don't change data after validation.

The first step for a better solution might be to have 3) + 4) + 5) in a single function.
No problem for me, if 2) is also contained at the begin of this function.

vboctor:
@atrol I've moved the trimming to the user creation action page.
@dregad I reworded the commit message

Merged.

PR #637, opened 2015-09-02
Generate valid CSS/class name for custom status with spaces

dregad:
Prior to this, for a custom status 'in progress', Mantis would generate
the following invalid code:

- HTML (class name): class="small-caption in progress-color"
- Dynamic CSS (status_config.php): .in progress-color { ... }

We now rely on a new API function html_get_css_identifier() to generate
a valid CSS class name by replacing invalid characters with '-'.

Fixes [#20068](https://www.mantisbt.org/bugs/view.php?id=20068)


atrol:
See comment https://www.mantisbt.org/bugs/view.php?id=20068#c51360

vboctor:
See comment in the bug.

atrol:
Comments at https://www.mantisbt.org/bugs/view.php?id=20068#c51425

atrol:
+1

PR #638, opened 2015-09-05
Fix custom field access denied on bug update page

vboctor:
The previous fix caused triggering of access_denied due to the addition of the BUG_UPDATE_TYPE_NORMAL.

Issue #20002

### DIFF
--- a/bug_update.php
+++ b/bug_update.php
@@ -276,12 +276,7 @@ $t_custom_fields_to_set = array();
 foreach ( $t_related_custom_field_ids as $t_cf_id ) {
 	$t_cf_def = custom_field_get_definition( $t_cf_id );
 
-	# if this is not a full update action and custom field is not on the form, then don't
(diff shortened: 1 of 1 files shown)


bug_update.php
```diff
@@ -293,6 +288,12 @@
 		}
 	}
 
+	# If the custom field is present or not required (checked above) and the it is not present,
```
badfiles: please fix 'the it'

syncguru:
:+1:

dregad:
+1

FYI, in case you didn't know, using `Fixes #xxx` instead of `Issue #xxx`in the commit message will make the source integration plugin automatically resolve the issue in addition to attaching the changeset to it.

vboctor:
@dregad Thanks for the reviews.  Note that the issue was already resolved and this is just a follow up fix.  So didn't need to use Fixes -- just used Issue.

PR #640, opened 2015-09-07
Fix AJAX for 'Match type' and 'Highlight changed'

dregad:
Add missing 'dynamic-filter-expander' class to the HTML anchor tags to enable the AJAX.

https://www.mantisbt.org/bugs/view.php?id=20080

Also fixes a minor inconsistency in language strings:
- the identifier was missing the standard `_label` suffix), and 
- the trailing `:` was hardcoded in the page instead of being part of the language string

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -3389,7 +3389,13 @@ function filter_draw_selection_area2( $p_page_number, $p_for_screen = true, $p_e
 
 		<!-- Match Type -->
 		<tr class="row-1">
-			<td class="small-caption category2"><a href="<?php echo $t_filters_url . FILTER_PROPERTY_MATCH_TYPE;?>" id="match_type_filter"><?php echo lang_get( 'filter_match_type' )?>:</a></td>
+			<td class="small-caption category2">
+				<a id="match_type_filter"
+					href="<?php echo $t_filters_url . FILTER_PROPERTY_MATCH_TYPE;?>"
+					<?php echo $t_dynamic_filter_expander_class; ?>>
+					<?php echo lang_get( 'filter_match_type' )?>:
+				</a>
+			</td>
 			<td class="small-caption" id="match_type_filter_target">
 			<?php
 				switch( $t_filter[FILTER_PROPERTY_MATCH_TYPE] ) {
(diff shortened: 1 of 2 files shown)


core/filter_api.php
```diff
@@ -3389,7 +3389,13 @@ function filter_draw_selection_area2( $p_page_number, $p_for_screen = true, $p_e
 
 		<!-- Match Type -->
 		<tr class="row-1">
-			<td class="small-caption category2"><a href="<?php echo $t_filters_url . FILTER_PROPERTY_MATCH_TYPE;?>" id="match_type_filter"><?php echo lang_get( 'filter_match_type' )?>:</a></td>
+			<td class="small-caption category2">
+				<a id="match_type_filter"
+					href="<?php echo $t_filters_url . FILTER_PROPERTY_MATCH_TYPE;?>"
+					<?php echo $t_dynamic_filter_expander_class; ?>>
+					<?php echo lang_get( 'filter_match_type_label' )?>
```
vboctor: filter_match_type lang string seems to be no longer being used.  Should we remove it?  Or do we expect to use it elsewhere?

lang/strings_english.txt
```diff
@@ -1187,7 +1187,7 @@ $s_issue_id = 'Issue #';
 $s_recently_visited = 'Recently Visited';
 $s_priority_abbreviation = 'P';
 $s_note_user_id_label = 'Note By:';
-$s_filter_match_type = 'Match Type';
+$s_filter_match_type_label = 'Match Type:';
```
vboctor: Is it worth breaking the translation completely until all languages are translated again?  Why don't we use the same language string?  Another option is to rename it in all languages.

dregad: I'd rather have consistency on this kind of things (ie label strings being identified as such), as it reduces the maintenance overhead in the long run. 

We could easily rename the strings for all languages but I'd like to check with @siebrand first to avoid any issues with translatewiki updates. As an alternative we could do this in 2 steps, first introduce new string, wait for translations to be updated then remove the old one.

vboctor: Let's go with the two steps approach.

vboctor:
:+1: with two steps approach for language string.

PR #641, opened 2015-09-07
Fix invalid query error when updating an issue

vboctor:
If a custom field is on the form but has value of null, it caused invalid query error.

Fixes #20082


syncguru:
+1

dregad:
I didn't do any test, and I'm not sure of the implications of this, but I'll go along with your judgment.

PR #642, opened 2015-09-07
Fire EVENT_LAYOUT_BODY_END directly before </body>

syncguru:
Fixes #20084: EVENT_LAYOUT_BODY_END should fire after loading JS libraries

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -757,8 +757,6 @@ function html_footer() {
 function html_body_end() {
 	global $g_scripts_included;
 
-	event_signal( 'EVENT_LAYOUT_BODY_END' );
-
 	echo "\t" . '<script type="text/javascript" src="' . helper_mantis_url( 'javascript_config.php' ) . '"></script>' . "\n";
 	echo "\t" . '<script type="text/javascript" src="' . helper_mantis_url( 'javascript_translations.php' ) . '"></script>' . "\n";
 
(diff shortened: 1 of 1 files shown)


vboctor:
:+1: for the change, but it should be a single commit that just moved the line down.

dregad:
When submitting PRs, I would prefer if you rebased your branch on top of latest master, instead of merging - allows for easier reviews and simple, clean merges.

In any case, :-1: to this for now, at least until the discussion in https://www.mantisbt.org/bugs/view.php?id=20081 is complete.

vboctor:
I've added my response in the bug.

syncguru:
@dregad, @vboctor - rebased the PR

I also commented on the bug with why JS at the bottom of the page is good for MantisBT.

core/html_api.php
```diff
@@ -777,6 +775,8 @@ function html_body_end() {
 
 	echo '</div>', "\n";
 
+	event_signal( 'EVENT_LAYOUT_BODY_END' );
```
dregad: As [suggested by @vboctor](https://www.mantisbt.org/bugs/view.php?id=20084#c51400), add comment explaining the significance of the position

dregad:
Comment was added in c3f52b74328ef9e50df745da3c6ae91dda226017

PR #643, opened 2015-09-07
Replace 'filter_match_type' by 'filter_match_type_label'

dregad:
This is step 2 (see f932c3aa67177500a7a6c53ad0a8a402ab859f87), as per the discussion in PR #640.

This should be merged once the translations of the new *filter_match_type_label* string have been updated.


vboctor:
:+1: after translations are done.

PR #645, opened 2015-09-10
fix #0020076: auto_set_status_to_assigned

cproensa:
only set "bug_assigned_status" if action is an assignment, not a status change
should fix #0020076

### DIFF
--- a/bug_update.php
+++ b/bug_update.php
@@ -365,6 +365,7 @@ if( $t_bug_note->note &&
 # Handle automatic assignment of issues.
 if( $t_existing_bug->handler_id === NO_USER &&
 	 $t_updated_bug->handler_id !== NO_USER &&
+	 $t_updated_bug->status == $t_existing_bug->status &&
 	 $t_updated_bug->status < config_get( 'bug_assigned_status' ) &&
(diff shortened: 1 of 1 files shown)


bug_update.php
```diff
@@ -364,6 +364,7 @@
 
 # Handle automatic assignment of issues.
 if( $t_existing_bug->handler_id === NO_USER &&
+         $t_updated_bug->status == $t_existing_bug->status &&
```
vboctor: Adjust indentation to match neighboring lines.  Move this check to be the first or the third check, since the two checks for handler makes sense to be next to each other.

vboctor:
I suggest revising the commit message.

First line should describe the fix and be shorter than or equal to 50 characters.
The details of the commit should include Fixes #20076 -- ideally on its own line after the detailed description of the fix.

Once the commit is revised + the comment about indentation and moving the check :+1:

cproensa:
Hi Victor, i am new to git and still learning the best practices
this branch is rewritten with the commit as you suggested.
Thanks!

vboctor:
:+1:

dregad:
Fixed in bd603b3841d5ee038594b5e707af7d6557642874

PR #646, opened 2015-09-10
fixing  manage_config_workflow_set.php

cproensa:
some corrections to "manage_config_workflow_set.php"
problem: when deleting configuration options if detected as same as parent level settings. but some of that logic was wrong, not correctly comparing against proper parent values, causing valid configurations to disappear (and reset to default)
related: #0019970


cproensa:
I have consolidated commits and rewritten commit text for cleanliness.

dregad:
I reviewed and tested this, found and fixed a bug. Then moved on with testing whether [#16771](https://www.mantisbt.org/bugs/view.php?id=16771) was actually fixed, found out it was not (or at least not completely), then fixed that plus a few other issues. 

TL/DR - this PR is superseded by #670.

PR #647, opened 2015-09-11
display custom field values overflows container

cproensa:
Displaying custom field values, adds ZWSP characters after each '|' to allow line wrapping

fixing for bug report [#0020103](https://www.mantisbt.org/bugs/view.php?id=20103)

### DIFF
--- a/manage_custom_field_page.php
+++ b/manage_custom_field_page.php
@@ -76,7 +76,7 @@ print_manage_menu( 'manage_custom_field_page.php' );
 				</td>
 				<td><?php echo count( custom_field_get_project_ids( $t_field_id ) ) ?></td>
 				<td><?php echo get_enum_element( 'custom_field_type', $t_desc['type'] ) ?></td>
-				<td><?php echo string_display( $t_desc['possible_values'] ) ?></td>
+                                <td><?php echo str_replace( '|', '|&#x200b;', string_display( $t_desc['possible_values'] ) ) ?></td>
 				<td><?php echo string_display( $t_desc['default_value'] ) ?></td>
 			</tr><?php
 		} # Create Form END ?>
--- a/manage_custom_field_page.php
+++ b/manage_custom_field_page.php
@@ -76,7 +76,11 @@ print_manage_menu( 'manage_custom_field_page.php' );
 				</td>
 				<td><?php echo count( custom_field_get_project_ids( $t_field_id ) ) ?></td>
 				<td><?php echo get_enum_element( 'custom_field_type', $t_desc['type'] ) ?></td>
-                                <td><?php echo str_replace( '|', '|&#x200b;', string_display( $t_desc['possible_values'] ) ) ?></td>
+				<?php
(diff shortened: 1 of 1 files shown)


dregad:
+1

atrol:
No time to have a deeper look at the moment, but isn't there a better solution for it?
e.g. something like http://caniuse.com/#search=word-wrap

dregad:
CSS would indeed be cleaner, but the word-wrap / overflow-wrap property you're proposing is only supported in fairly recent browsers. @cproensa's proposal is a simple and effective workaround.

atrol:
> in fairly recent browsers

Do you mean FF 38 with it?

We should add some comment lines like
```php
# workaround to enforce line break
# @todo replace by CSS after we don't support any longer browsers without CSS3 support
```

cproensa:
you are right on being a somewhat ugly hack.
however the way the values are stored with pipes is already ugly byitself

a drawback of this PR: copying text from the browser into a text editor will carry the non-printing-character. Example, into vim: "... AZ002|<200b>TAZ003|<200b>TAZ004|<200b>TAZ ..."

Two solutions here: line-wrap the contents in some way, or stretch the container around the inner table (but still pouting out of the screen horizontally, like it happens on 1.2.x)

Other option to avoid messing with css:
display the values in a readonly textfield, like the one on the editing page.

cproensa:
if the "copying text from browser, it carries non-printing chars" is an issue, maybe a simple approach would be
- display spaces alongside pipes: "VAL1 | VAL2 | VAL3 | ..."
instead of "VAL1|VAL2|VAL3|..."
- display as comma separated: "VAL1, VAL2, VAL3, ..."

The key is that spaces allow for line wrap, and do not confuse the user if text is copied, and would only be used for displaying this table

vboctor:
Why don't we just print it as 'option1 | option2'?  If the user inputs the same then we trim it so that it is saved as 'option1|option2'?

Another option is to just use css, and worst case it won't look great on older browsers in the case where there are too many options.

cproensa:
I'd think its appropiate to print  'option1 | option2', joined with atrol suggesion for comments:
    # workaround to enforce line break
    # @todo replace by CSS after we don't support any longer browsers without CSS3 support

Are you sure to trim the user input too? This would mean chasing all forms processing custom fields input to ensure the text is trimmed (only for enumerations, or all types?). Not allowing outer spaces in values, even if it may be a source of problems for users, may break current functionality.

I didn't intend this to be more than a cosmetic mod. But if you are willing to go all the way down, please consider my comment in #0017828: get rid of the '|' by storing these values as arrays.

cproensa:
Changed separator to display spaces as vboctor suggestion
Added "workaround" comments as indicated by atrol

Still willing to get rid of hardcoded separators in db, may open a separate feature request if supported.

dregad:
Fixed in bd49dc00cf221c6c393400aef7fc62b40bcfc2d7

PR #648, opened 2015-09-11
Fix 'Data Type Mismatch' errors in filters

dregad:
See
- https://www.mantisbt.org/bugs/view.php?id=20087
- https://www.mantisbt.org/bugs/view.php?id=20105


vboctor:
:+1:

PR #649, opened 2015-09-17
Force storing bugnote statistics

cproensa:
If bug is already in cache, we still want to store bugnote statistics
becasue the stored bug may not have them stored yet
Fixes #0020121

see buginfo for explanation


cproensa:
close, this would be part of another PR, when the change is required

PR #650, opened 2015-09-17
0006939 count only visible notes

cproensa:
This is another take on #0006939: Number of Private notes visible for reporters
It was also discussed in PR: https://github.com/mantisbt/mantisbt/pull/43
This PR accounts for each project, (since view bug page may list bugs from different projects, each with different access levels to private notes)
Instead of using count, max in sql query, there's need to traverse actual bugnote list, but it still done in only one query, and trying to get minimum api_calls in the loop.

* This modification has revealed bug #0020121 which should be fixed to improve performance on bug cache.

* This modification consolidates sql query which was duplicated (differently) in two apis, creating a new function to be called from both places.

* I have expanded statistics with new values:
last_modified_bugnote
last_submitted_bugnote
which can prove useful to future features (eg: #0020122)

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -729,6 +729,10 @@ function bug_cache_database_result( array $p_bug_database_result, array $p_stats
 	global $g_cache_bug;
 
 	if( !is_array( $p_bug_database_result ) || isset( $g_cache_bug[(int)$p_bug_database_result['id']] ) ) {
+		if( null !== $p_stats ) {
+			#force store the bugnote statistics
+			return bug_add_to_cache( $p_bug_database_result, $p_stats );
+		}
 		return $g_cache_bug[(int)$p_bug_database_result['id']];
 	}
 
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -1484,12 +1484,22 @@ function bug_get_bugnote_stats_array( array $p_bugs_id, $p_user_id = null ) {
 	$t_array_chunks = array_chunk( $t_unique_array, 1000 );
 
 	foreach( $t_array_chunks as $t_id_array ) {
-		$t_where_string = ' WHERE n.bug_id in (' . implode( ', ', $t_id_array ) . ')';
+		# Build sql IN list
+		$t_count = count( $t_id_array );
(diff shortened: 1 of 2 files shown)


cproensa:
May someone help me know why the checks have failed?

dregad:
I didn't check your code, but according to the Travis log
```
[Thu Sep 17 18:42:47 2015] [mantisconnect.php] Error Type: SYSTEM NOTICE,
Error Description: Undefined variable: t_stats
Stack Trace:
filter_api.php L1521 mc_error_handler()
filter_api.php L2106 bug_get_bugnote_stats_multi()
filter_api.php L2096 filter_cache_result()
mc_project_api.php L81 filter_get_bug_rows()
mantisconnect.php L84 mc_project_get_issues_for_user()
mantisconnect.php L84 handle()
mantisconnect.php L0 {main}()
```
this is most likely because the test script is hitting a code path where *$t_stats* variable has not been initialized.

cproensa:
Thanks Dreagd, you nailed it. variable not correctly initialized
Another error was using some code not compatible with php 5.3

dregad:
You're welcome.

cproensa:
cherry-picked commit from #649
as it is needed to review performance in this branch

core/bug_api.php
```diff
@@ -1461,6 +1463,88 @@ function bug_get_newest_bugnote_timestamp( $p_bug_id ) {
...
+	$t_unique_array = array_unique( $p_bugs_id );
+	if ( null === $p_user_id ) {
+		$t_user_id = auth_get_current_user_id();
+	}
+	else {
+		$t_user_id = $p_user_id;
+	}
+
+	$t_stats = array();
+	$t_array_chunks = array_chunk( $t_unique_array, 1000 );
```
cproensa: size '1000', should be defined as a constant, or global?

cproensa:
After solving the sql IN issue, i am not adding more big changes, as it matches current state of code functionality.
For future actions, i will propose some clean up, but keep separated from this changes

cproensa:
close, to refactor when a proper standard for long SQL clauses is implemented

PR #651, opened 2015-09-18
Actions for edit/clone configuration options

cproensa:
~~In adm_config_report, adds new button and rename current for editing options~~
- ~~action "Create Configuration Option" is the previous behaviour (create new
  option and keep old)~~
- ~~action "Move Configuration Option" creates new option and deletes old one
  (or keep if doesn't change key values)~~

In adm_config_report, show buttons for different editing actions:
- Action "Clone" creates new option and keeps old, overwritting if one
  already exists with those key values.
  exists
- Action "Edit" updates current configuration option, or moves the
  option if key values are changed, overwritting if one already exists

Fixes #0020058

### DIFF
--- a/adm_config_report.php
+++ b/adm_config_report.php
@@ -217,6 +217,7 @@ $t_edit_project_id      = gpc_get_int( 'project_id', $t_filter_project_value == 
 $t_edit_option          = gpc_get_string( 'config_option', $t_filter_config_value == META_FILTER_NONE ? '' : $t_filter_config_value );
 $t_edit_type            = gpc_get_string( 'type', CONFIG_TYPE_DEFAULT );
 $t_edit_value           = gpc_get_string( 'value', '' );
+$t_edit_action          = gpc_get_string( 'action', 'action_create' );
 
 # Apply filters
 
@@ -412,6 +413,21 @@ while( $t_row = db_fetch_array( $t_result ) ) {
 					'config_option' => $v_config_id,
 					'type'          => $v_type,
 					'value'         => $v_value,
+					'action'        => 'action_edit',
+				),
+				OFF );
+
+			# Clone button
+			print_button(
+				'#config_set_form',
+				lang_get( 'create_child_bug_button' ),
+				array(
+					'user_id'       => $v_user_id,
+					'project_id'    => $v_project_id,
+					'config_option' => $v_config_id,
+					'type'          => $v_type,
+					'value'         => $v_value,
+					'action'        => 'action_clone',
 				),
 				OFF );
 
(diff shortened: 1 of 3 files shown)


vboctor:
I would expect keeping the single action approach "Set Configuration Option".  This should do the right thing, whether it is to update in place or move based on the change of user/project id/option name.

The scenario that won't be covered by this approach is the ability to "copy" a config option from one user or project to another.  In such case, next to the "Edit" option, we can add a "Clone" or "Copy" button that opens the form, but in clone mode.  The clone mode is basically setup for creating a new entry unless the entry already exists, it would overwrite it.

Does that make sense?  I think having "Set" and "Move" options would be confusing.

adm_config_report.php
```diff
@@ -412,6 +413,21 @@ function check_config_value( $p_config ) {
...
 					'type'          => $v_type,
 					'value'         => $v_value,
+					'action'        => 'action_edit',
+				),
+				OFF );
+
+			# Clone button
+			print_button(
+				'#config_set_form',
+				lang_get( 'create_child_bug_button' ),
```
cproensa: TODO: crate proper lang strings for action buttons?

cproensa:
Changed as suggested
Actually looks cleaner

still TODO: lang strings

edit_config_option_2

dregad:
I like this implementation better than the previous one with 2 action buttons actually.

> still TODO: lang strings

Not sure what you mean by that... translations ? If so, this happens automatically through TranslateWiki.

cproensa:
Added Dregard's suggestions

The thing with translations... is that i really dont know how to do it. I just added needed strings in "strings_english.txt". Please tell me the correct procedure.

Also, the "clone" label is reused from lang_get( 'create_child_bug_button' )
And the (current) "edit" label is also reused from lang_get( 'edit_link' )
Should they have their own string?

vboctor:
Re-using existing strings is OK.  For new strings, just add them to strings_english.txt.  The translation team will handle the rest.  Till the strings are translated, the other languages will fallback to the English strings.

My expectation is the following:
- User clicks Edit, this ends up updating the issue in place or moving it as necessary when user click 'Set Configuration Option'.
- User click Clone, this ends up duplicating the config option to the new user/project/config option set on the form. If one already exist, it will overwrite it.  I'm ok with the button of the form being called 'Set Configuration Option' in such case as well.

vboctor:
Use Create/Clone/Edit is fine too.

cproensa:
> My expectation is the following: [...]

Thats how it's implemented at this point.

updated with rebased and cleaned up commits

vboctor:
@cproensa I've tested this and it works well.  Is there any work remaining on this before we get it in?

cproensa:
>  Is there any work remaining on this before we get it in?

its ready, and recently rebased.
Thanks

vboctor:
Merged via 13bda674ee2131f65edfd1eccc36b114b3b5ea69

PR #652, opened 2015-09-21
0020107 config report filter options

cproensa:
In configuration report (adm_config_report.php)
when selecting a specific configuration option, some options are not showing the filtered list and resets to the "any" state.
Fixes #0020107

after modifications, found that some values like "database version" were editable, which is also fixed here.


dregad:
Fixed in 15990874a610339b62afc361e1280ae49c31ce44, 4ac423751c7a813af1c4b1b6b52d536f1a81ba97

PR #653, opened 2015-09-25
improve sql performance print_project_option_list

cproensa:
cache subprojects before processing to save sql queries
fixes #0020139


dregad:
Fixed in 770df68eb22f18a8a4a2171c8c3bfea25704584f

PR #654, opened 2015-09-25
use selective bug cache for attachment count

cproensa:
use selective bug cache for attachment count instead of cache data for all bugs in db
fetch data only for bugs included in filter result set
fixes #0020138

The problem identified in bug #0020138, is that caching the attachment count query data for all bugs in database. This is unnecessary, and can penalize in db time, processing time and memory usage.

Modification uses the filter bug list and caches all of its values, only for that set. Usually all bugs accessed are those from the filter result, so subsequent calls to file_bug_attachment_count will hit the cache successfully. In other case, a single query will be issued.

Typical scenario will perform one query in view_all_bug_page, and several queries in my_view_page (one for each filter box). Hopefully the queries and its processing are more lightweight that original approach.

### DIFF
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -178,7 +178,7 @@ function file_bug_attachment_count( $p_bug_id ) {
 
 	# First check if we have a cache miss, if so try to load
 	if( !isset( $g_cache_file_count[$p_bug_id] ) ) {
-		file_bug_attachment_count_cache( array( $p_bug_id ) );
+		file_bug_attachment_count_cache( array( (int)$p_bug_id ) );
 	}
 
 	# If bug is valid, it should be in cache now.
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -98,6 +98,30 @@ function file_get_display_name( $p_filename ) {
 }
 
 /**
+ * Fills the cache with the attachement count from a list of bugs
+ * using one sql query
+ * @param array $p_bug_ids Array of bug ids
+ */
+function file_bug_attachment_count_cache( array $p_bug_ids ) {
+	global $g_cache_file_count;
+
+	if( null === $p_bug_ids || !is_array( $p_bug_ids ) ) {
+		return;
+	}
+
+	$t_sql_list= '(' . implode( ', ', $p_bug_ids ) . ')';
+	$t_query = 'SELECT B.id AS bug_id, COUNT(F.bug_id) AS attachments'
+			. ' FROM {bug} B LEFT OUTER JOIN {bug_file} F'
+			. ' ON (B.id = F.bug_id)'
+			. ' WHERE B.id IN ' . $t_sql_list . ' GROUP BY B.id';
+
+	$t_result = db_query( $t_query );
+	while( $t_row = db_fetch_array( $t_result ) ) {
+		$g_cache_file_count[$t_row['bug_id']] = $t_row['attachments'];
+	}
+}
+
+/**
  * Check the number of attachments a bug has (if any)
  * @param integer $p_bug_id A bug identifier.
  * @return integer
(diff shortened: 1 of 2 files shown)


cproensa:
With this modification I have compared the loading of "view_all_bug_page.php"
I go from about 2.5 segs and 17MB, to 0.8 segs 4MB

This is a profile of calls for the current view_all. Most of the work here is just fetching the attachment count:
profile_view_all

core/file_api.php
```diff
@@ -98,45 +98,49 @@ function file_get_display_name( $p_filename ) {
...
- * Check the number of attachments a bug has (if any)
- * @param integer $p_bug_id A bug identifier.
- * @return integer
+ * Fills the cache with the attachement count from a list of bugs
+ * using one sql query
+ * @param array $p_bug_ids Array of bug ids
  */
-function file_bug_attachment_count( $p_bug_id ) {
+function file_bug_attachment_count_cache( array $p_bug_ids ){
 	global $g_cache_file_count;
```
dregad: Missing space before {

core/file_api.php
```diff
@@ -98,45 +98,49 @@ function file_get_display_name( $p_filename ) {
...
-	#   the cache. If the cache isn't empty and we didn't have
-	#   a hit, then there are not attachments for this bug.
-	if( count( $g_cache_file_count ) > 0 ) {
-		return 0;
-	}
+	$t_sql_list= '(' . implode( ', ', $p_bug_ids ) . ')';
+	$t_query = 'SELECT B.id AS bug_id, COUNT(F.bug_id) AS attachments'
+			. ' FROM {bug} B LEFT OUTER JOIN {bug_file} F'
+			. ' ON (B.id = F.bug_id)'
+			. ' WHERE B.id IN ' . $t_sql_list . ' GROUP BY B.id';
```
dregad: I don't like the approach of using IN operator here. There is a limit in how many items can fit in (especially with Oracle DB), which will lead to errors when executing the code on large data sets.

Please consider an alternative approach.

core/file_api.php
```diff
@@ -98,45 +98,49 @@ function file_get_display_name( $p_filename ) {
...
+function file_bug_attachment_count( $p_bug_id ) {
+	global $g_cache_file_count;
+
+	# First check if we have a cache miss, if so try to load
+	if( !isset( $g_cache_file_count[$p_bug_id] ) ) {
+		file_bug_attachment_count_cache( array( $p_bug_id ) );
 	}
 
-	return $t_file_count;
+	# If bug is valid, it should be n cache now.
```
dregad: *in* cache ?

core/file_api.php
cproensa: > There is a limit in how many items can fit in (especially with Oracle DB), which will lead to errors when executing the code on large data sets

Right.
This is already an issue with bugnote count (see #696), with the current functions.
And probably more places: retrieving projects, cache bug rows, etc...

In standard usage (with current code base): bug listings, are limited to pagination size, so this is not practical problem (but still it exists and waiting to bite us is some corner cases)
A real use case where a massive enough bug list may trigger the fail situation would be CSV/Excel exports, which currently has disabled the inclusion of these two fields (again, #696)

So even if this would not result in fail, with current use cases, I agree to fix first the cache functions to address the long IN failure
So in order to get this done, i'd like some feedback on what proposed methods would be convenient to use, to provide a uniform approach on other similar situations. Brainstorming, some possibilities:
- loop through the array partitioned in pieces of 500 item size, performing ~N/500 queries. (or any other convenient chunk size)
- use a temporary table to store values and join.
- ¿any other?

dregad: Instinctively I'd go for temp tables myself.

IMO partitioning the data would be resource-intensive and complex to develop / maintain.

cproensa: > Instinctively I'd go for temp tables myself.

defining an api funcion, use a persistent table? (create is not an enforced permission in mantis db user), allow concurrency, etc 

> IMO partitioning the data would be resource-intensive and complex to develop / maintain.

Unless the method for insertion itself has to partition the data too. When Mysql fails its probably due to message size, not a limit for IN elements. Would it fail too if the insert is also a single "bulk" insert?

This would be the best approach, but you need to open a RFC for developers to discuss.

In the mean time, a mid-term solution, seems to be only partitioning the query for each case?

vboctor: I may be missing your points here, why can't we do this query for 50 bug ids at a time using the exact same query we support now?

- Get 50 ids
- Run the query for the 50 ids.
- Read results in cache
- Repeat until done

vboctor:
@cproensa nice view of the callee map.  What tool did you use for that?

core/file_api.php
cproensa: > I may be missing your points here, why can't we do this query for 50 bug ids at a time using the exact same query we support now?

Yes, that is the immediate solution. But having seen dregad oftenly complaining about the INs, i wanted to bring up the general issue, in case it may be someday worked on.

cproensa:
> @cproensa nice view of the callee map. What tool did you use for that?

i used kcachegrind to open a Xdebug profiler dump.

core/file_api.php
```diff
@@ -98,45 +98,96 @@ function file_get_display_name( $p_filename ) {
...
+ * TODO cproensa
+ * As a provisional fix for potentially breaking db limits on query size,
+ * the query is partitioning the supplied list.
+ * The sitaution referenced here: #0020423. If any methodology is
+ * proposed, this function will have to be reviewed
  */
-function file_bug_attachment_count( $p_bug_id ) {
+function file_bug_attachment_count_cache( array $p_bug_ids ) {
 	global $g_cache_file_count;
+	define( 'SQL_IN_MAX_COUNT', 500 );
```
cproensa: Should this be a global constant?

cproensa:
Regarding user permissions for attachment visibility:
This function provides the count number to screen (view, columns, etc)
but also to BugData->attachment_count

Actual "visible" attachment count is not calculated properly at the moment (eg https://www.mantisbt.org/bugs/view.php?id=20411). If checks for user access and configuration are implemented, should BugData contain the "real" count, or the "visible" count?

There may be a deeper problem here: some data is cached, most of what is presented in the view page, but it may be dependant of the user accessing it.
In the same session you may need to get that data for different context, like here: how many attachments exists vs how many are visible to current user. This is a problem when the same functions are used and data is already cached.
And another level of wrong: sometimes another user is impersonated to call the apis, (eg email_api:email_build_visible_bug_data), are we hitting the same cached data?

cproensa:
I plan to keep this branch as it's now, since this keeps current functionality (even though it may not be correct).
Leaving for future actions:
- Fix the attachment visibility count (#20411)
- Propose improvements about those issues of visibility, user context, cache, etc

cproensa:
closing, until underlying problems with big sql can be addressed with a standard (see #764)

PR #655, opened 2015-09-26
accept objects for event EVENT_FILTER_COLUMNS

cproensa:
As explained in [#0020140](https://www.mantisbt.org/bugs/view.php?id=20140)
Modify EVENT_FILTER_COLUMNS event processing, to allow plugin to pass objects, not only classes
Maintains compatibility with previous functionality

### DIFF
--- a/core/columns_api.php
+++ b/core/columns_api.php
@@ -179,12 +179,16 @@ function columns_get_plugin_columns() {
 		foreach( $t_all_plugin_columns as $t_plugin => $t_plugin_columns ) {
 			foreach( $t_plugin_columns as $t_callback => $t_plugin_column_array ) {
 				if( is_array( $t_plugin_column_array ) ) {
-					foreach( $t_plugin_column_array as $t_column_class ) {
-						if( class_exists( $t_column_class ) && is_subclass_of( $t_column_class, 'MantisColumn' ) ) {
-							$t_column_object = new $t_column_class();
-							$t_column_name = utf8_strtolower( $t_plugin . '_' . $t_column_object->column );
-							$s_column_array[$t_column_name] = $t_column_object;
+					foreach( $t_plugin_column_array as $t_column_item ) {
+						if( is_object( $t_column_item ) && $t_column_item instanceof MantisColumn ) {
+							$t_column_object = $t_column_item;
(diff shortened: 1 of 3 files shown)


cproensa:
I have updated this PR, rebased to master-1.3.x

core/columns_api.php
```diff
@@ -179,9 +179,15 @@ function columns_get_plugin_columns() {
...
-						if( class_exists( $t_column_class ) && is_subclass_of( $t_column_class, 'MantisColumn' ) ) {
-							$t_column_object = new $t_column_class();
+					foreach( $t_plugin_column_array as $t_column_item ) {
+						$t_column_object = null;
+						if( is_object($t_column_item) && $t_column_item instanceof MantisColumn ) {
+							$t_column_object = $t_column_item;
+						}
+						else if( class_exists( $t_column_item ) && is_subclass_of( $t_column_item, 'MantisColumn' ) ) {
+							$t_column_object = new $t_column_item();
+						}
```
dregad: I would remove `$t_column_object = null;` (line 183) and add 
```
else {
    continue;
}
```
here, then we can get rid of the if() below.

core/filter_api.php
```diff
@@ -106,11 +106,17 @@ function filter_get_plugin_filters() {
...
+						$t_filter_object = null;
+						if( is_object($t_filter_item) && $t_filter_item instanceof MantisFilter ) {
+							$t_filter_object = $t_filter_item;
+						}
+						else if( class_exists( $t_filter_item ) && is_subclass_of( $t_filter_item, 'MantisFilter' ) ) {
+							$t_filter_object = new $t_filter_item();
+						}
+						if( $t_filter_object ) {
+							$t_filter_name = utf8_strtolower( $t_plugin . '_' . $t_filter_object->field );
+							$s_field_array[$t_filter_name] = $t_filter_object;
```
dregad: Same comment as above

cproensa:
updated with @dregad suggestion

I have tested OK with my plugins.
If you already have any plugin that uses a column or filter, just change:
return array( 'ClassName' ), to
return array( new className() )
in the event hook.

cproensa:
Updated coding standards
Rebased to current master-1.3.x

PR #656, opened 2015-09-27
Bugnotes action buttons improvements

cproensa:
Pre-generate form security tokens for bugnote buttons to avoid performance issues in view bug page
Fixes #0018016
Modify button generation to send action parameters as POST
Fixes #0020141

### DIFF
--- a/bugnote_view_inc.php
+++ b/bugnote_view_inc.php
@@ -219,20 +219,36 @@ $t_num_notes = count( $t_bugnotes );
 
 				# show edit button if the user is allowed to edit this bugnote
 				if( $t_can_edit_bugnote ) {
-					print_button( 'bugnote_edit_page.php?bugnote_id='.$t_bugnote->id, lang_get( 'bugnote_edit_link' ), null, OFF );
+					print_button(
+						'bugnote_edit_page.php',
+						lang_get( 'bugnote_edit_link' ),
+						array( 'bugnote_id' => $t_bugnote->id ),
+						OFF );
 				}
 
 				# show delete button if the user is allowed to delete this bugnote
 				if( $t_can_delete_bugnote ) {
-					print_button( 'bugnote_delete.php?bugnote_id='.$t_bugnote->id, lang_get( 'delete_link' ), null, $t_security_token_delete );
+					print_button(
+						'bugnote_delete.php',
+						lang_get( 'delete_link' ),
+						array( 'bugnote_id' => $t_bugnote->id ),
+						$t_security_token_delete );
 				}
 
 				# show make public or make private button if the user is allowed to change the view state of this bugnote
 				if( $t_can_change_view_state ) {
 					if( VS_PRIVATE == $t_bugnote->view_state ) {
-						print_button( 'bugnote_set_view_state.php?private=0&bugnote_id=' . $t_bugnote->id, lang_get( 'make_public' ), null, $t_security_token_state );
+						print_button(
+							'bugnote_set_view_state.php',
+							lang_get( 'make_public' ),
+							array( 'private' => '0', 'bugnote_id' => $t_bugnote->id ),
+							$t_security_token_state );
 					} else {
-						print_button( 'bugnote_set_view_state.php?private=1&bugnote_id=' . $t_bugnote->id, lang_get( 'make_private' ), null, $t_security_token_state );
+						print_button(
+							'bugnote_set_view_state.php',
+							lang_get( 'make_private' ),
+							array( 'private' => '1', 'bugnote_id' => $t_bugnote->id ),
+							$t_security_token_state );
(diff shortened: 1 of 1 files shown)


bugnote_view_inc.php
```diff
@@ -216,20 +220,36 @@
 
 				# show edit button if the user is allowed to edit this bugnote
 				if( $t_can_edit_bugnote ) {
-					print_button( 'bugnote_edit_page.php?bugnote_id='.$t_bugnote->id, lang_get( 'bugnote_edit_link' ) );
+					print_button(
+						'bugnote_edit_page.php',
+						lang_get( 'bugnote_edit_link' ),
+						array( 'bugnote_id' => $t_bugnote->id ),
+						OFF );
```
vboctor: Why is this OFF?  If that is the correct behavior, then add a comment to explain.

bugnote_view_inc.php
```diff
@@ -122,6 +122,10 @@
 	$t_can_delete_all_bugnotes = access_has_bug_level( config_get( 'delete_bugnote_threshold' ), $f_bug_id );
 	$t_can_change_view_state_all_bugnotes = $t_can_edit_all_bugnotes && access_has_bug_level( config_get( 'change_view_status_threshold' ), $f_bug_id );
 
+	# Pre-generate form security tokens to avoid performance issues
+	$t_security_token_state = form_security_token( 'bugnote_set_view_state' );
```
vboctor: That is much better than before in terms of perf.  Now we will generate 2 instead of 2n tokens.  You could optimize further by having two functions that get the token.  If the function is called before, it returns the same token.  If the function wasn't called before it creates the token.  Hence, by lazy creation, we could end up generating 0, 1, or maximum 2 tokens.  I'm OK either way, it can be an easy improvement, but it could also be over-engineering :)

vboctor:
:+1: after addressing my comments.  Thanks @cproensa .

bugnote_view_inc.php
cproensa: It is OFF because "bugnote_edit_page.php" does not check for token
maybe we should check for it there too.

In any way, i agree on commenting about the OFF in case it stays

bugnote_view_inc.php
cproensa: You mean: if no buttons are going to be shown (eg, user cant delete), no token is needed
That can be done using IF $t_can_delete... THEN generate-token-for-delete-page.
Nitpicking, but simple enough. I agree on doing it.

cproensa:
added commit for vboctor suggestion: create tokens only if needed

waiting for feedback on: ¿should bugnote_edit_page.php include a validation for token? (currently it doesn't)

vboctor:
I suggest adding token validation in bugnote_edit_page.php.

cproensa:
Added token validation.
Let me know if i should squash all commits into one.

cproensa:
> I suggest adding token validation in bugnote_edit_page.php.

I am thinking that token for edit button is not actually needed. The "edit" button does not make any modification by itself, it only goes into the edit form.

tgulacsi:
+1 for the first commits! It solved my slowness problem for view_bug_page.php for issues with lots of notes.

dregad:
>> I suggest adding token validation in bugnote_edit_page.php.

>I am thinking that token for edit button is not actually needed. The "edit" button does not make any modification by itself, it only goes into the edit form.

I don't think the token is needed either in this case. Suggest to ditch the last commit.

The build is failing. See below sample error from [the TravisCI log](https://travis-ci.org/mantisbt/mantisbt/jobs/97993999)
```
There were 6 errors:

1) IssueAddTest::testCreateIssueWithTimeTrackingNote
SoapFault: Error Type: SYSTEM NOTICE,

Error Description: Undefined variable: t_user_id

/home/travis/build/mantisbt/mantisbt/tests/soap/IssueAddTest.php:295
/home/travis/build/mantisbt/mantisbt/tests/soap/IssueAddTest.php:295
[...]
```
I'm not sure the system notice was introduced by this PR, but I didn't check. Suggest you refresh the PR by rebasing your branch on top of latest master, and investigate further if it still fails.

cproensa:
> I don't think the token is needed either in this case. Suggest to ditch the last commit.

Ok, commit is nuked.
That can be reviewed at a later time if needed.

> The build is failing

It failed becasue at the time i uploaded it, master branch had some error that made the build fail.
Now, the error is different, and apparently unrelated to this PR:

    The command "phpenv global 5.5" failed and exited with 1 during .

cproensa:
> Now, the error is different

and now it works...  ¯\_(ツ)_/¯

dregad:
Merged in 2941f258c3bf5231dc33600927b666ffa33778c6

PR #657, opened 2015-10-14
Fix broken issue update page due to custom field

vboctor:
In some corner cases, the customer hits invalid query which is trying to set
the custom field string to null. This is caused in the following case:

- Custom field is visible on update page.
- Custom field is not required.
- Custom field is of type list.
- Custom field doesn't have any value selected.

In this case, the new custom field value evaluated to default value provided by
bug_update.php page which is null. The fix is to change it to ''.

Fixes #20196


dregad:
+1

syncguru:
+1

PR #659, opened 2015-10-15
Add config to disable caching of LDAP passwords

skoef:
our LDAP passwords are SSHA hashed for a reason. I don't like the idea of having a MD5 hashed version of my password in mantis' database so I introduced `use_ldap_password` to toggle whether mantis should keep a local copy of the LDAP password after logging in.

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -1946,6 +1946,13 @@ $g_ldap_uid_field = 'uid';
 $g_ldap_realname_field = 'cn';
 
 /**
+ * Keep a local copy of the LDAP password (ON) so we can
+ * authenticate even when LDAP is unavailable.
+ * @global integer $g_ldap_cache_passwords
+ */
+$g_ldap_cache_passwords = ON;
+
+/**
  * Use the realname specified in LDAP (ON) rather than the one stored in the
  * database (OFF).
  * @global integer $g_use_ldap_realname
@@ -4260,7 +4267,7 @@ $g_global_settings = array(
 	'class_path','library_path', 'language_path', 'absolute_path_default_upload_folder',
 	'ldap_simulation_file_path', 'plugin_path', 'bottom_include_page', 'top_include_page',
 	'default_home_page', 'logout_redirect_page', 'manual_url', 'logo_url', 'wiki_engine_url',
-	'cdn_enabled'
+	'cdn_enabled', 'ldap_cache_passwords'
 );
 
 # Temporary variables should not remain defined in global scope
--- a/core/ldap_api.php
+++ b/core/ldap_api.php
@@ -379,7 +379,11 @@ function ldap_authenticate_by_username( $p_username, $p_password ) {
 
 		if( false !== $t_user_id ) {
 
-			$t_fields_to_update = array('password' => md5( $p_password ));
+			$t_fields_to_update = array();
+
+			if ( ON == config_get( 'ldap_cache_passwords' ) ) {
+				$t_fields_to_update['password'] = md5( $p_password );
+			}
 
 			if( ON == config_get( 'use_ldap_realname' ) ) {
 				$t_fields_to_update['realname'] = ldap_realname( $t_user_id );
--- a/docbook/Admin_Guide/en-US/config/auth.xml
+++ b/docbook/Admin_Guide/en-US/config/auth.xml
@@ -213,6 +213,20 @@ ldaps://ldap.example.com:3269/
 			</varlistentry>
 
 			<varlistentry>
+				<term>$g_ldap_cache_passwords</term>
+
+				<listitem>
+					<para>Keep a local copy of the LDAP password (ON)
+					so we can authenticate even when LDAP is unavailable.
+					Defaults to
+					<emphasis>ON</emphasis>.</para>
+
+					<para>Note that MantisBT will update the database with
+					the data retrieved from LDAP when ON.</para>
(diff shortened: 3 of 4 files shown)


vboctor:
Thanks @skoef for your contribution.  I wonder if we are better of adding SHA support to MantisBT than providing such option.  That will improve security across the board and not just for LDAP.

skoef:
We sure can add that as well, but then I still see the need/desire of not caring about a local backup of your password as a whole, hashed in whatever way.

atrol:
No time for a deeper look at the moment.
This will fix https://www.mantisbt.org/bugs/view.php?id=12957

config_defaults_inc.php
```diff
@@ -1946,6 +1946,13 @@
 $g_ldap_realname_field = 'cn';
 
 /**
+ * Keep a local copy of the MD5 hashed LDAP password (ON) so we can
+ * authenticate even when LDAP is unavailable.
+ * @global integer $g_use_ldap_password
+ */
+$g_use_ldap_password = ON;
```
vboctor: How about $g_ldap_cache_passwords as the config option name?

config_defaults_inc.php
```diff
@@ -1946,6 +1946,13 @@
 $g_ldap_realname_field = 'cn';
 
 /**
+ * Keep a local copy of the MD5 hashed LDAP password (ON) so we can
```
vboctor: Just mention that the password is cached in the database without referencing the hashing type.

core/ldap_api.php
```diff
@@ -379,7 +379,11 @@ function ldap_authenticate_by_username( $p_username, $p_password ) {
 
 		if( false !== $t_user_id ) {
 
-			$t_fields_to_update = array('password' => md5( $p_password ));
+			$t_fields_to_update = array( );
```
vboctor: Remove the space ( ) -> ()

docbook/Admin_Guide/en-US/config/auth.xml
```diff
@@ -213,6 +213,20 @@ ldaps://ldap.example.com:3269/
 			</varlistentry>
 
 			<varlistentry>
+				<term>$g_use_ldap_password</term>
+
+				<listitem>
+					<para>Keep a local copy of the MD5 hashed LDAP password (ON)
```
vboctor: Update description as per comment above about referencing MD5.

vboctor:
Update the commit message to have "Fixes #12957" as part of the body (i.e. not the first line of the commit message).

atrol:
The option should be added to $g_global_settings in config_defaults_inc.php to avoid that it can be set in database.

I am not complete sure if the option should be added to function config_is_private to avoid it can be exposed by the web service.

vboctor:
Looking into getting this into master.  A couple of comments:

- Looking at the code, it seems that the only way we would use such passwords is in the case where the administrator would change the auth type from LDAP to MD5.  We don't actually fallback to it when LDAP server is available.  I believe the documentation for the config in config_defaults_inc/manual provides a different impression.

- Is it worth keeping this functionality at all?  Should we just drop it?  Should we at least have it OFF by default?

Thoughts?

dregad:
I added this reference in the documentation when I rewrote that section a few years back. I basically reflected [what the ldap api does](https://github.com/mantisbt/mantisbt/blob/master/core/ldap_api.php#L382), without actually checking whether it is was actually usable.

I have personnally used this "fallback" as you describe (i.e. changing $g_login_method to MD5) when copying a Prod DB in a test environment where I don't want to connect to the live LDAP.

In my opinion, for obvious security reasons, it would be best to completely remove the LDAP password caching (in ldap api, as well as [auth api (auth_auto_create_user)](https://github.com/mantisbt/mantisbt/blob/master/core/authentication_api.php#L212) which, as a side note, is not covered by this PR.

I have not analyzed if there is any impact is of not storing anything in the password field.

skoef:
@dregad I can completely remove the functionality, if that's what you're saying

dregad:
This is what I'm saying indeed (in response to @vboctor's [request for comments](https://github.com/mantisbt/mantisbt/pull/659#issuecomment-157920203)). I suggest you wait for his feedback before doing anything.

vboctor:
Let's remove this functionality.

atrol:
One more PR for it  https://github.com/mantisbt/mantisbt/pull/713

seveas:
I've been running without password cache and it has no negative sideeffects. I'd love for it to be killed, or at least configurable to be killed :)

vboctor:
@seveas I think we want to remove it without configuration.  Let's make sure we have a PR that has such removal so that we can get it checked in.

seveas:
@vboctor done. https://github.com/mantisbt/mantisbt/pull/714

dregad:
Since I think we all agree that caching LDAP password should be disabled entirely and not be optional, I'm closing this in favor of @seveas' PR #714

tcs-ulli:
LDAP passwords are cached till today in unsecure md5 sums that are decryptable within seconds with web services. WTF?!

tfnab:
This **security issue** was submitted to the project at least seven-and-a-half years ago (see https://mantisbt.org/bugs/view.php?id=12957) and I seriously don't understand why it can't be fixed. It's not like there aren't any suitable patches, pull requests (the latest being #718 ) and the like. What is required to get the project managers' attention?

PR #660, opened 2015-10-21
avoid html output for errors in ajax call

cproensa:
fixes [#0013048](https://www.mantisbt.org/bugs/view.php?id=13048)
Filter inputs are populated by ajax call to return_dynamic_filters.php
If there is an error in remote script, it should fail the request
instead of outputting an error page
A failed request is handled by the ajax call.

### DIFF
--- a/return_dynamic_filters.php
+++ b/return_dynamic_filters.php
@@ -154,7 +154,9 @@ if( function_exists( $t_function_name ) ) {
 
 	if( !$t_found ) {
 		# error - no function to populate the target (e.g., print_filter_foo)
-		# return err "400 Bad Request"
+		# return err "400 Bad Request", and log event
+		error_parameters( $f_filter_target );
+		log_event( LOG_AJAX, error_string(ERROR_FILTER_NOT_FOUND) );
 		header( ' ', true, 400);
 		exit;
 	}
--- a/return_dynamic_filters.php
+++ b/return_dynamic_filters.php
@@ -49,7 +49,11 @@ require_api( 'filter_constants_inc.php' );
 require_api( 'gpc_api.php' );
 require_api( 'helper_api.php' );
 
-auth_ensure_user_authenticated();
+#if user is not authenticated, return err "401 Unauthorized"
+if( !auth_is_user_authenticated() ){
+	header( ' ', true, 401 );
+	exit;
+}
 
 compress_enable();
 
@@ -150,7 +154,8 @@ if( function_exists( $t_function_name ) ) {
 
 	if( !$t_found ) {
 		# error - no function to populate the target (e.g., print_filter_foo)
-		error_parameters( $f_filter_target );
-		trigger_error( ERROR_FILTER_NOT_FOUND, ERROR );
+		# return err "400 Bad Request"
+		header( ' ', true, 400);
+		exit;
 	}
 }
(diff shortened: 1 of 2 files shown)


cproensa:
filter_error

dregad:
This looks good but I think we should add internal logging of the error, to allow troubleshooting by the admin if necessary, something like `log_event( LOG_AJAX, 'blah' );` where *blah* is some meaningful message, e.g. what was displayed by the original `trigger_error()` call.

cproensa:
Hi Damien, that makes sense.
added the event logging

return_dynamic_filters.php
```diff
@@ -150,7 +154,10 @@ function return_dynamic_filters_prepend_headers() {
 
 	if( !$t_found ) {
 		# error - no function to populate the target (e.g., print_filter_foo)
+		# return err "400 Bad Request", and log event
 		error_parameters( $f_filter_target );
-		trigger_error( ERROR_FILTER_NOT_FOUND, ERROR );
+		log_event( LOG_AJAX, error_string(ERROR_FILTER_NOT_FOUND) );
```
dregad: coding guidelines: space before/after parentheses

dregad:
Considering that `auth_ensure_user_authenticated()` also calls trigger_error(), I'm wondering if we should not also add a log_event() before returning with a 401.

cproensa:
@dregad:
> Considering that auth_ensure_user_authenticated() also calls trigger_error(), I'm wondering if we > > should not also add a log_event() before returning with a 401

Are you sure? i cant reproduce any log output for that action

cproensa:
I am wondering if a more proper fix to this, and other similar error screens, should be more on the line:
Make the error routines aware that we are in a script that is not intended for UI screen, so instead of printing formatted html, the standard error codes and messages are returned by setting headers, all from the error handlers

vboctor:
In my opinion, the right long term fix is:

- Error pages should return the appropriate status code whether they are targeted for scripts or normal web pages.  For example: if we browse for an issue that doesn't exist, we should show the page we currently show + status code 404.
- Business logic should throw an exception that we catch on the main script, and show error pages based on such exception or return some ajax/soap friendly error otherwise.

dregad:
Not sure what the status for this is. Can I merge it, or are there further changes required following @vboctor's latest comment ?

cproensa:
I feel this is not the cleanest solution

> Business logic should throw an exception that we catch on the main script, and show error pages based on such exception or return some ajax/soap friendly error otherwise.

This makes sense as the-right-thing-to-do
however, i think exceptions are not used as is in current code base, and this would require some major thinking

Probably an easier route is to add another entry to the error output configurations that sets output= headers
A script that works without gui would set error output to headers, and let the main error handler do its work

cproensa:
Regarding exceptions, there are other use cases in which exception handling is handy.
While looking at group actions, if an error happens inside one on the bug actions, the whole batch is affected. Ideally a try/catch should isolate errors and keep working with the batch.
A legit error can be a validation from a plugin event. AFAIK, the only way to stop an action from a plugin is to thrown an error (eg: bug_update events).

dregad:
I'm closing this PR assuming the new approach you proposed in #733 fully covers it. Feel free to reopen if not.

PR #664, opened 2015-10-23
CSS: Hide collapsed sections for js disabled presentation

cproensa:
as stated in #0019508
Disabling JavaScript in 1.3.x leads to various side effects, e.g. duplicated sections where collapse functionality is used.

This css modification solves the duplication of sections when javascript is disabled.

### DIFF
--- a/core/collapse_api.php
+++ b/core/collapse_api.php
@@ -110,7 +110,7 @@ function collapse_closed( $p_name, $p_section = '' ) {
 	ob_start();
 
 	$t_div_id = $t_block . '_closed';
-	echo "\n<div id=\"", $t_div_id, '"', ( $t_display ? ' class="collapse-open"' : ' class="collapse-closed"' ), '>';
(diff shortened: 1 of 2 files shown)


vboctor:
See the bug for feedback.

dregad:
The expected behavior when JS is disabled, is (or should be) to show all sections expanded.

With this patch, user is no longer able to access the functionality of a collapsed section at all, e.g. Filters (which is collapsed by default) or any section that was collapsed by the user before javascript got disabled.

I'm afraid a proper fix will be a bit more complex than a simple CSS tweak...

-1

cproensa:
Damien, this second commit may solve the issue with the saved collapse-state for section.
seems like the classes collapse_closed, collapse_open, sometimes reflect the contrary of its names... sigh

dregad:
Fixed in 9674df6cb7e46d7f5c37cb46a51b8b5f227131d6

PR #665, opened 2015-10-23
print_enum_string_option_list accept array of current values

cproensa:
This function was defined to accept a integer value, to check
against when building the option list, so it gets defaulted.
Some places (filter fileds) were calling with this value as
an array of one int, or several ints (for multiselection
option list).
Changed parameter to accept array, since the check_selected
function already accepts an array of values to check.
Adding a check for the case $p_val is a single value to cast
it to an int, becasue check_select should be called in
strict mode for enum keys anyways, and some callers may still
pass a number as string.
Then, in the case of "array of int", the caller must ensure
values are proper int type.

fixes #0019978 (or more)

### DIFF
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -915,13 +915,20 @@ function print_build_option_list( $p_build = '' ) {
 
 /**
  * select the proper enumeration values based on the input parameter
+ * Current value may be an integer, or an array of integers.
  * @param string  $p_enum_name Name of enumeration (eg: status).
- * @param integer $p_val       The current value.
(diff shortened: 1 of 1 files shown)


dregad:
I have tested this and confirm this resolves [#19978](https://www.mantisbt.org/bugs/view.php?id=19978), but did not check whether this introduces any regressions, although that seems unlikely.

+1

vboctor:
:+1: thanks.

dregad:
Fixed in b8dcb52a

PR #666, opened 2015-10-25
Remove reference to mantisbt-announce

vboctor:
Remove reference to mantisbt-announce and mention that announcements will be sent to
users registered on our official bugtracker.

Will backport to 1.2.x after this PR is closed.

Fixes #19378

### DIFF
--- a/docbook/Admin_Guide/en-US/About.xml
+++ b/docbook/Admin_Guide/en-US/About.xml
@@ -211,14 +211,11 @@
         <itemizedlist>
             <listitem>
                 <para>
-                    The <ulink url="mailto:mantisbt-announce@lists.sourceforge.net">
-                        mantisbt-announce mailing list</ulink>
-                    is a very low traffic
-                    list that is used for major announcements,  typically
-                    about new releases.  All MantisBT users are
-                    encouraged to subscribe to this mailing list.  The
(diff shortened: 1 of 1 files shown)


docbook/Admin_Guide/en-US/About.xml
```diff
@@ -211,14 +211,13 @@
...
-                    about new releases.  All MantisBT users are
-                    encouraged to subscribe to this mailing list.  The
-                    average traffic should be no more than one to two posts
-                    per month.
+                    We send release announcements and important updates to users registered
+                    on our <ulink url="http://www.mantisbt.org/bugs">official bugtracker</ulink>.
+                    To get onto our mailing list, users will have to signup there and verify
+                    their email address. This same account can also be used to report, monitor,
+                    and comment on issues relating to MantisBT.
+                    All MantisBT users are encouraged to subscribe to this mailing list.  The
```
dregad: ...are encouraged to *register on issue tracker*. ...

docbook/Admin_Guide/en-US/About.xml
```diff
@@ -211,14 +211,13 @@
...
+                    average traffic should be no more than one to two posts per month.
```
dregad: Probably a frequency indication by year would be more appropriate 😉

PR #667, opened 2015-10-29
Revert 738272f and reimplement hide graph link

cproensa:
commit 738272f to hide Mantisgraph link introduces
regression: failing at login when anonymous login is
disabled.
Here that change is reverted and the access level check
is now placed in the callback function, instead of the
hook initialization

#0020217


atrol:
+1

dregad:
Fixed in 2c0b2a59770588a1d6e7bf8bd40962563263919b

PR #668, opened 2015-10-31
Filters and print user lists

cproensa:
This is a rewrite of project_get_all_user_rows(). This function is the core functionality of higher level functions that creates user lists based on project assignation, access levels, etc. Mainly for printing user option list for selectable input fields (print_api, filter_api, ...)

The new functionality of this code allows querying on multiple projects, with a per-project access-level check. All the users are retrieved and sorted with one single database query.
Also, the printing of user list does not use temporary space and bypass cache in order to save memory usage.
This helps solving a bug of poor performance detected when user count is very large.

Also, as a direct consequence of allowing multi-project queries, some filter related bugs may be solved now. Example: include the actual filtered projects in the user list search fields, for reporter, monitoring user, etc. 

Regarding "print_user_option_list" improvements, here are some test that i have made to measure times and memory usage, to get an idea:

Test case, 20k users
(It was run in debug mode, and discarding the first run that usually takes longer)
Measuring memory usage with "memory_get_peak_usage(true)", and time with "microtime(true)" on begin and end of print_user_option_list()

For one single project (public)    
  * on branch master:
    * standard: 96468992 (96 MB) (5.34 s)
    * disabling user rows cache: 33554432 (33 MB) (2.9 s)
    
  * on modified branch:
    * sorted by username, realname: 5505024 (5 MB) (2.27 s)
    * sorted by lastname: 99876864 (99 MB) (4.99 s)
    * idem, but disabling cache: 30932992 (30 MB) (2.89 s)

For ALL_PROJECTS best case:
  * master: 115605504 (115 MB)   10.20 s
  * mod: 5505024 (5 MB) 2.11 s

### DIFF
--- a/api/soap/mc_project_api.php
+++ b/api/soap/mc_project_api.php
@@ -1056,6 +1056,7 @@ function mc_project_get_issue_headers( $p_username, $p_password, $p_project_id, 
 
 /**
  * Get appropriate users assigned to a project by access level.
+ * @TODO cproensa: Do we need to call user_cache_array_rows() here?
  *
  * @param string  $p_username   The name of the user trying to access the versions.
  * @param string  $p_password   The password of the user.
--- a/core/access_api.php
+++ b/core/access_api.php
@@ -643,3 +643,28 @@ function access_get_status_threshold( $p_status, $p_project_id = ALL_PROJECTS ) 
 		}
 	}
 }
+
+/**
+ * Converts an access level threshold, which can be an integer or an array
+ * of access levels, to an array of access levels.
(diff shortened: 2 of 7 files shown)


cproensa:
I have reviewed the call to "user_cache_array_rows()" that was made inside "project_get_all_user_rows()"
For most of the callers it's not needed, especially on the cases of building very large user lists.

Instead it has been moved to the caller when necessary (for some email functions).

cproensa:
close, to refactor when standard solution for SQL long clauses is implemented

PR #669, opened 2015-11-01
CSS change margin unit px into rem

cproensa:
Added the margin measure in 'rem' units,
keeping 'px' as fallback.
Seems like Safari on retina display renders
better the 0.1rem (as 1px equivalent)

Fixes #0017816

### DIFF
--- a/css/default.css
+++ b/css/default.css
@@ -587,6 +587,7 @@ div.field-container {
 	width: 100%;
 	float: left;
 	margin: 0em 0em 1px 0em;
+	margin: 0em 0em 0.1rem 0em;
 	padding: 0em;
 }
 .field-container>.display-label,


css/default.css
```diff
@@ -587,6 +587,7 @@ div.field-container {
 	width: 100%;
 	float: left;
 	margin: 0em 0em 1px 0em;
+        margin: 0em 0em 0.1rem 0em;
```
dregad: Use tabs, not spaces

dregad:
I can't actually test this since I don't have a platform where the issue can be replicated.

Is there any point of keeping 2 margin definitions ? Wouldn't the 2nd always override the first ?

cproensa:
> Use tabs, not spaces

done! note to self: remember that netbeans does not format css...

> Is there any point of keeping 2 margin definitions ? Wouldn't the 2nd always override the first ?

Yes, that's the point. Seems like 'rem' is not supported by some older browsers. In that case, it fallbacks to 'px'

dregad:
Thanks for the explanation.

For the record, according to http://caniuse.com/#search=rem, this fallback is only needed for IE <= 8, Safari <= 4, iOS Safari <= 3.2, Opera <= 11.5, Firefox <= 3.5

dregad:
I just did a regression test on my PC here (tested both FF38 and Chrome46), and this actually messes up the presentation because the 0.1rem actually expands to 1.6px, causing the separator's height to  be sometimes 1px and sometimes 2px, as shown in the screenshot below.

Snapshots were taken from login page, with code from current master (1px) on the left shown right next to your proposal (0.1rem) on the right; the red box is a 300% zoom to make the difference more visible.

screenshot_20151102-095621_clipboard

So, considering that Safari has about 5% market share, and only a subset of this is on retina display, I don't think this should be merged.

:-1:

cproensa:
@dregad 
Actually, the 1pixel/2pixel is going to happen no matter what
With the current 'px' measures, i usually see a variaton of 1 to 2 pixels wide separations (i see it in chromium and firefox, linux, and a standard display).
The underlying problem is that css 'px' is not an actual screen pixel, when different system settings, dpi, custom page zoom, etc... all comes into play.

I am observing "empirically":
- In my system (safari, retina display), with 'px' units the rounding is: sometimes 1 pixel, sometimes 0 pixel
- safari, with 'rem' units, sometimes its 1 pixel, sometimes 2 pixel.

So probably, what you say is a regression, its actually the expected behaviour?

Another fact:
In my usual linux chromium system (with 'px' units) Decreasing zoom level about 3 or 4 steps (like pressing 4 times ctrl-), is the point that the separator disappear (rounding to 0 pixels).
In safari, with the 'rem' units, takes about the same amount of zooming out to get it to disappear.
  
Anyway, maybe its just a safari bug?

cproensa:
so is this actually the same as doing "1.6px"? 
the 'rem' thing may be coincidential (idk why i thought the 1rem=10px...)

dregad:
I do know it's not an actual pixel (note my use of *px* unit and not the word *pixel*) and that the "real" size will vary depending on the user's display, but if you take a "normal" situation with 100% zoom, using 1px gives the same apparent margin size (1 *pixel* on my screen here) which is not the case with *rem*.

> So probably, what you say is a regression, its actually the expected behaviour?

IMO, the expected behavior is to have *consistent* spacing between the divs, whether it is 1 pixel, 1.6 pixel or whatever, does not matter.

> 1rem=10px

em and rem are *relative* to the *font size*, of the parent container or the html root respectively (whereas px is an absolute specification)

> so is this actually the same as doing "1.6px"? 

No, that depends on your font size...

cproensa:
> IMO, the expected behavior is to have consistent spacing between the divs, whether it is 1 pixel, 1.6 pixel or whatever, does not matter.

I guess that's the key point and not a trivial one.
As i said, with current css i already see a mix of 1/2 pixels separators. Which, at least, is better than 0/1 pixel... tonight i will take some screenshots.
Maybe its because in linux there are more facilities for scaling at various levels, and "classic" Windows is more strict in that aspect. Like most of these issues, with newer version( Windows 8/10, Edge, etc) all these issues may start to arise too.

cproensa:
I'm fearing that the only failproof method to get a **consistent** separator may be to use some trick with images (pixel gifs,...) from css

dregad:
Which may not be worth the effort, especially considering that the UI will be changed radically in 2.x

cproensa:
> Which may not be worth the effort

Exactly what i was thinking...

cproensa:
Another WTF: In Chrome at 110% zoom, 1px margins disappear...

so, if anything, its possible to choose 2px, to get less chances to get it rounded to 0, at the cost of thicker separators
...or set 1.1px, which most of the time it gets a consistent 1 pixel (or whaterver it shows) for all separators, at the usual display configuration.

Surprisingly, the same Safari that messes 1px margin, does a perfect work on 1px table borders. Even when zoomed out to max, it keeps a nice an clean 1 pixel border...

i will close this PR, since it doesn't add anything useful.

PR #671, opened 2015-11-02
CSS add outline to field-container div

cproensa:
Added the property 'outline' to the field-container
div class.
Seems like it has a more consistent rendering that
margins or borders defined at 1px
Keep the previous definition of margin 1px as fallback
for old browsers

see issue [#0017816](https://www.mantisbt.org/bugs/view.php?id=0017816)

### DIFF
--- a/css/default.css
+++ b/css/default.css
@@ -591,6 +591,7 @@ div.field-container {
 	width: 100%;
 	float: left;
 	margin: 0em 0em 1px 0em;
+	outline: 1px solid #fff;
 	padding: 0em;
 }
 .field-container>.display-label,


vboctor:
@syncguru can you please take a look at this css tweak?

syncguru:
Looks good. Never used outline before though but seems like a clever fix to this issue.

cproensa:
With "outline" there are other issues:
When zooming out, the outline overlaps the container border.
Example in chrome at 75% zoom level:
seleccion_081

This happens because the "outline" border is drawn outside of the element outer dimension
This is still not perfect, but I would say this use case is less frequent than those which caused the original problem with margins
Maybe tweaking the container paddings would address this one?

dregad:
Fixed in c5e99534fe65619a7ee223f857d5c4ee2d64af91

PR #673, opened 2015-11-04
Check workflow for reopening bug

cproensa:
Add a check to access_can_reopen_bug() for
testing if the reopen status is reachable
by current workflow configuration.
In case the status is not reachable, returns
false.

Currently, the reopen operation would fail
anyway at the moment of changing status.
With this early check, a fail can be detected
before the action is made.

A direct benefit from this is that the
"reopen" button is not shown to the user if
the reopen action is not possible

Fix #0020256


dregad:
Fixed in 9eb915c34fe0b300d551b9c5598b7cc337710ac3

PR #675, opened 2015-11-05
Cache user data for view_all page

cproensa:
Cache the array of user ids that appear in the bug set to show.
This saves later individual sql queries from user_cache_row()

https://www.mantisbt.org/bugs/view.php?id=20258

### DIFF
--- a/view_all_bug_page.php
+++ b/view_all_bug_page.php
@@ -74,17 +74,19 @@ if( $t_rows === false ) {
 }
 
 $t_bugslist = array();
-$t_users_handlers = array();
-$t_project_ids  = array();
+$t_unique_user_ids = array();
+$t_unique_project_ids = array();
 $t_row_count = count( $t_rows );
 for( $i=0; $i < $t_row_count; $i++ ) {
 	array_push( $t_bugslist, $t_rows[$i]->id );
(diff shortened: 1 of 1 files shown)


vboctor:
:+1:

atrol:
We cache already all handlers in view_all_bug_page.php
```php
user_cache_array_rows( $t_unique_users_handlers );
```

Wouldn't this be the right place to add also the reporters?

cproensa:
> Wouldn't this be the right place to add also the reporters?

Initially i wanted to cache only reporters, but added handlers, being in the loop anyways.
So, yes, it would be better to integrate the reporter users in that call from view_all_bug_page.php,

but also move the category cache to that place to group all cache related code in the same place.
However, that code for categories probably needs a rework? is it caching effectively categories for subprojects, or only current project (with inherited)? ... this may go to another bug report, will try to investigate.

cproensa:
As discussed, i have rewritten the modification to go into "view_all_bug_page.php"
Also, changed how the arrays are built so "array_unique" is not needed.

dregad:
Looks good to me.

syncguru:
+1

dregad:
Merged in 7f997e8712ebc70be670f02b209f96b45e50dace

PR #677, opened 2015-11-11
Code cleanup

atrol:


core/string_api.php
```diff
@@ -721,75 +718,53 @@ function string_get_bug_view_url_with_fqdn( $p_bug_id ) {
...
 /**
- * return the name and GET parameters of a bug REPORT page for the given bug
- * account for the user preference and site override
- * @param integer $p_user_id A valid user identifier.
+ * return the name of a bug REPORT page
  * @return string
  */
-function string_get_bug_report_url( $p_user_id = null ) {
-	return string_get_bug_report_page( $p_user_id );
+function string_get_bug_report_url() {
```
vboctor: Why do we still need string_get_bug_report_url() and string_get_bug_report_page() rather than just one of them?

vboctor:
:+1:

dregad:
+1

core/string_api.php
atrol: Removed string_get_bug_report_page

PR #679, opened 2015-11-17
FEATURE mc_issue_set_handler()

fmancardi:
I've developed this (may be quick & dirty) using mc_issue_update() as starting point. My need was to have possibility to update/set _just_ issue handler (using userID), a thing that seems not simple using mc_issue_update().
Please consider to add this to stable code
Thanks a lot for your wonderful application.

### DIFF
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -1597,3 +1597,59 @@ function mci_issue_data_as_header_array( $p_issue_data ) {
 
 		return $t_issue;
 }
+
+/**
+ * Set Issue Handler
+ *
+ * @param string $p_username The name of the user trying to add the issue.
+ * @param string $p_password The password of the user.
+ * @param integer $p_issue_id 
+ * @param integer $p_handler_id 
+ * @return boolean
+ */
+function mc_issue_set_handler( $p_username, $p_password, $p_issue_id, $p_handler_id ) {
+	global $g_project_override;
+
+	$t_user_id = mci_check_login( $p_username, $p_password );
+	if( $t_user_id === false ) {
+		return mci_soap_fault_login_failed();
+	}
+
+	if( !bug_exists( $p_issue_id ) ) {
+		return SoapObjectsFactory::newSoapFault('Client', "Issue '$p_issue_id' does not exist.");
+	}
+
+	if( bug_is_readonly( $p_issue_id ) ) {
+		return SoapObjectsFactory::newSoapFault('Client', "Issue '$p_issue_id' is readonly");
+	}
+
+	$t_project_id = bug_get_field( $p_issue_id, 'project_id' );
+
+	if( !mci_has_readwrite_access( $t_user_id, $t_project_id ) ) {
+		return mci_soap_fault_access_denied( $t_user_id );
+	}
+
+	$g_project_override = $t_project_id; // ensure that helper_get_current_project() calls resolve to this project id
+
+    if( is_null($p_handler_id) && $p_handler_id <= 0) {
+		return SoapObjectsFactory::newSoapFault( 'Client', "Invalid Handler" );
+    }
+	$t_handler_id = intval($p_handler_id);
+    
+	if( !access_has_bug_level( config_get( 'update_bug_threshold' ), $p_issue_id, $t_user_id ) ) {
+		return mci_soap_fault_access_denied( $t_user_id,  "Not enough rights to update issues" );
+	}
+
+	$t_bug_data = bug_get( $p_issue_id, true );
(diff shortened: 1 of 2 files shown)


api/soap/mc_issue_api.php
```diff
@@ -1597,3 +1597,59 @@ function mci_issue_data_as_header_array( $p_issue_data ) {
...
+
+	$t_project_id = bug_get_field( $p_issue_id, 'project_id' );
+
+	if( !mci_has_readwrite_access( $t_user_id, $t_project_id ) ) {
+		return mci_soap_fault_access_denied( $t_user_id );
+	}
+
+	$g_project_override = $t_project_id; // ensure that helper_get_current_project() calls resolve to this project id
+
+    if( is_null($p_handler_id) && $p_handler_id <= 0) {
```
vboctor: Did you mean to have an || (OR) here?   Adjust spacing around parenthesis to match rest of code and indentation.

vboctor:
Thanks @fmancardi for your contribution.

I suggest we consider the following:
- Rename field to mc_issue_assign().
- Send handler as AccountData similar to mc_issue_add() enable clients to provide username, id, or email.
- Should we enable an optional IssueNoteData to provide optional addition of a note as part of assignment.

It also shouldn't be too hard to achieve the same via an mc_issue_get() and mc_issue_update(), where the mc_issue_update() if the last update field is provided in the incoming request, yet it doesn't match the one in the data (i.e. someone else update the bug between the get and the update).  However, I can understand the need for a more specific API that requires a single round trip.

@rombert can also provide feedback on this PR.

fmancardi:
First thanks for your fat answer
1) Yes, I was wrong when using AND, instead was OR; excuse me
2) I've thought about using AccountData as input, OK I will change do all changes in method interface and name that you have suggested.

3) Before developing this I've given a look to the mc_issue_update() because I suppose it can do the trick, but I've discovered that this method requires lot of mandatory fields when you only want to do the assignment, then IMHO seems an overkill. I think we can still use you approach i.e. rewriting this method using mc_issue_get() and mc_issue_update() in this way:
using p_issue_id => use mc_issue_get() to get issue data, then on the object you got, set ONLY handler and notes (is any exists), then call mc_issue_update().
Do you prefer this approach?

4) I can change the method name, no problems.
    Only issue is that I do not how to proceed: i.e. I need to open a new merge request ? or this can be change?
Can I ask via mail to rmobert?

5) Do you want the opposite method mc_issue_unassign() we written? this is neede if we want to have same feature present on GUI

Best regards
Francisco

dregad:
> 4) I can change the method name, no problems.
Only issue is that I do not how to proceed: i.e. I need to open a new merge request ? or this can be change?

Just add new commits to your local branch and push them to your github fork, they will appear here automatically. In the future, I suggest you create a local feature branch instead of updating "master-1.2.x".

rombert:
> Before developing this I've given a look to the mc_issue_update() because I suppose it can do the trick, but I've discovered that this method requires lot of mandatory fields when you only want to do the assignment, then IMHO seems an overkill. 

If I read the code for `mc_issue_update` correctly ( https://github.com/mantisbt/mantisbt/blob/8e35696dd1d097696ca8532a96eef245964ce195/api/soap/mc_issue_api.php#L931-L944 ) then the following fields are required:

* project_id
* reporter_id
* handler_id
* category_id
* summary
* description

Actually, I think that no field should be required for updates, so IMO a better approach would be to change the `mc_issue_update` method to not require any fields to be set in the call, so it's easier to change just one field.

I understand why it's simpler to add a `mc_issue_set_handler` method, but then again the same argument can be made for (almost) any field that can be changed. I'd rather see `mc_issue_update` being improved.

vboctor:
There are a couple of things to take into consideration with the update method:
- There needs to be a distinction between "not supplied" and "not set".  For example, in an update request, handler_id can be not supplied or set to something that unassigns the issue.
- We should fail update requests that are based on an older version of the issue if the date_modified field is supplied, but doesn't match the database.

fmancardi:
IHMO there are other things to think about
If the target of API is to do things with same constraints that actions
done by a user on GUI then maybe some issues attributes related to audit
like date submitted, datemodified may be need to be ignored by API.
if API allows setting properties that can not be normally managed on GUI
this means that using API is like accessing DB using a SQL client.
It's not a black or white choice (IMHO) because to integrate system YOU
want to be able to set the reporter and normally when using GUI you can not
do this.
Hope i was clear enough.

Regards
Il 20/Nov/2015 04:56, "Victor Boctor" <notifications@github.com> ha scritto:

dregad:
What is the status of this PR ?

l266509:

PR #680, opened 2015-11-19
Fix query order in history get range result

cproensa:
Fixes the query order to be consistent.
When the query is requested as DESC ordered,
some events, created simultaneously, were
showed in reverse order than their creation,
because id field was still being sorted by
default (ASC) order.

However: probably, order by 'date_modified' is redundant if 'id' sequence is assumed to be chronological already?

### DIFF
--- a/core/history_api.php
+++ b/core/history_api.php
@@ -213,7 +213,7 @@ function history_get_range_result( $p_bug_id = null, $p_start_time = null, $p_en
 		$t_query .= ' WHERE ' . implode( ' AND ', $t_where );
 	}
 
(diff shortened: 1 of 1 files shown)


vboctor:
:+1: -- given that we have an index on date_modified, keeping it is not a big deal there.  Just in case there is imported data or some race condition.

vboctor:
Please also add "Fixes #20298" in the commit description.

cproensa:
> Please also add "Fixes #20298" in the commit description.
Thanks, I forgot...

commit is amended with the description

vboctor:
Merged via 2d3da823f2c3b602a97cab824456e0dd1783a1b2 - thanks @cproensa

PR #681, opened 2015-11-22
Support tokens in addition to passwords for API usage

vboctor:
- Schema change to add token field to user table with empty default.
- Account page populates the token, if empty, before displaying it.
- Ability to revoke a token and get a new one from account page.
- API authentications accepts tokens or passwords in the password field.  Header support will be added later.

This is phase 1 which provides the basic functionality + the language strings + the schema change.  There will be follow up work to support tokens to be passed in headers in the API.

screenshot 2015-11-22 09 00 59

Issue [#17766](https://www.mantisbt.org/bugs/view.php?id=17766)

### DIFF
--- a/account_page.php
+++ b/account_page.php
@@ -86,11 +86,18 @@ current_user_ensure_unprotected();
 
 html_page_top( lang_get( 'account_link' ) );
 
+$t_user_id = auth_get_current_user_id();
+
 # extracts the user information for the currently logged in user
 # and prefixes it with u_
-$t_row = user_get_row( auth_get_current_user_id() );
+$t_row = user_get_row( $t_user_id );
 
 extract( $t_row, EXTR_PREFIX_ALL, 'u' );
+
+if ( is_blank( $u_token ) ) {
+	$u_token = auth_generate_secret_token();
+	user_set_field( $t_user_id, 'token', $u_token );
+}
 
 $t_ldap = ( LDAP == config_get( 'login_method' ) );
 
@@ -229,6 +236,15 @@ if( $t_force_pw_reset ) {
 				<span class="input"><span class="field-value"><?php echo get_enum_element( 'access_levels', current_user_get_access_level() ); ?></span></span>
 				<span class="label-style"></span>
 			</div>
+			<div class="field-container">
+				<span class="display-label">
+					<span>
+						<?php echo lang_get( 'user_token' ) . '<br />' . lang_get( 'user_token_description' ); ?>
+					</span>
+				</span>
+				<span class="input"><span class="field-value"><?php echo $u_token; ?></span></span>
+				<span class="label-style"></span>
+			</div>
 			<?php
 			$t_projects = user_get_assigned_projects( auth_get_current_user_id() );
 			if( count( $t_projects ) > 0 ) {
@@ -257,18 +273,25 @@ if( $t_force_pw_reset ) {
 		</fieldset>
 	</form>
 </div>
+
+<div class="form-container">
+	<fieldset>
+	<span class="submit-button">
+<form method="post" action="account_revoke_token.php">
+		<?php echo form_security_field( 'account_revoke_token' ) ?>
+		<input type="submit" class="button" value="<?php echo lang_get( 'revoke_token_button' ) ?>" />
+</form>
+
 <?php # check if users can't delete their own accounts
 if( ON == config_get( 'allow_account_delete' ) ) { ?>
-
-<!-- Delete Button -->
-<div class="form-container">
+	<!-- Delete Button -->
 	<form method="post" action="account_delete.php">
-		<fieldset>
 			<?php echo form_security_field( 'account_delete' ) ?>
-			<span class="submit-button"><input type="submit" class="button" value="<?php echo lang_get( 'delete_account_button' ) ?>" /></span>
-		</fieldset>
+			<input type="submit" class="button" value="<?php echo lang_get( 'delete_account_button' ) ?>" />
 	</form>
-</div>
 <?php
 }
+
+echo '</span></fieldset></div>';
+
 html_page_bottom();
--- a/account_revoke_token.php
+++ b/account_revoke_token.php
@@ -1,0 +1,47 @@
+<?php
+# MantisBT - A PHP based bugtracking system
+
+# MantisBT is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# MantisBT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * This page stores the reported bug
+ *
+ * @package MantisBT
+ * @copyright Copyright 2000 - 2002  Kenzaburo Ito - kenito@300baud.org
+ * @copyright Copyright 2002  MantisBT Team - mantisbt-dev@lists.sourceforge.net
+ * @link http://www.mantisbt.org
+ *
+ * @uses authentication_api.php
+ * @uses core.php
+ * @uses print_api.php
+ * @uses user_api.php
+ */
+
+require_once( 'core.php' );
+require_api( 'authentication_api.php' );
+require_api( 'print_api.php' );
+require_api( 'user_api.php' );
+
+auth_ensure_user_authenticated();
+auth_reauthenticate();
+
+$t_user_id = auth_get_current_user_id();
+
+user_ensure_unprotected( $t_user_id );
+
+$t_token = auth_generate_secret_token();
+user_set_field( $t_user_id, 'token', $t_token );
+
+print_successful_redirect( 'account_page.php' );
+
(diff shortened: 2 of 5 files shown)


account_page.php
```diff
@@ -86,12 +86,19 @@
...
 # extracts the user information for the currently logged in user
 # and prefixes it with u_
-$t_row = user_get_row( auth_get_current_user_id() );
+$t_row = user_get_row( $t_user_id );
 
 extract( $t_row, EXTR_PREFIX_ALL, 'u' );
 
+if ( is_blank( $u_token ) ) {
+	$u_token = helper_generate_uuid();
+	user_set_field( $t_user_id, 'token', $u_token );
```
atrol: Token should not be stored unencrypted in database.

core/helper_api.php
```diff
@@ -692,3 +692,30 @@ function helper_duration_to_minutes( $p_hhmm ) {
 function shutdown_functions_register() {
 	register_shutdown_function( 'email_shutdown_function' );
 }
+
+/**
+ * Generate v4 UUID
+ * @return string The v4 UUID
+ */
+function helper_generate_uuid() {
+	return sprintf('%04x%04x-%04x-%04x-%04x-%04x%04x%04x',
```
atrol: openssl_random_pseudo_bytes might be a better/shorter solution for it (I am no specialist for this).
http://php.net/manual/en/function.openssl-random-pseudo-bytes.php
http://stackoverflow.com/questions/18830839/generating-cryptographically-secure-tokens

account_revoke_token.php
```diff
@@ -0,0 +1,48 @@
...
+ * @uses print_api.php
+ * @uses user_api.php
+ */
+
+require_once( 'core.php' );
+require_api( 'authentication_api.php' );
+require_api( 'helper_api.php' );
+require_api( 'print_api.php' );
+require_api( 'user_api.php' );
+
```
atrol: Prevent from being called directly by something like
if( !defined( 'ACCOUNT_PREFS_REVOKE_TOKEN_ALLOW' ) ) {
	return;
}
define ACCOUNT_PREFS_REVOKE_TOKEN_ALLOW in account_page.php

account_revoke_token.php
```diff
@@ -0,0 +1,48 @@
...
+auth_ensure_user_authenticated();
+
+$t_user_id = auth_get_current_user_id();
+
+user_ensure_unprotected( $t_user_id );
+
```
atrol: auth_reauthenticate() to prevent unauthorized users from clicking the "Revoke Token" button when having access to a unlocked screen where the account page is opened.

account_page.php
vboctor: How is this different from the cookies that are stored unencrypted?  The password is hashed because it can be re-used elsewhere and that we never need to display it.  That is not the case with token and cookie.

account_revoke_token.php
vboctor: This is an action page and is not an include file, hence, we don't need such check.  Actually such check would break the code.

account_revoke_token.php
vboctor: I thought about it when I was writing the code, but I thought that the form auth token is enough.  But you are right if the page is rendered. we don't want the token revoked after the re-authenticate window, so I've added it.

core/helper_api.php
vboctor: I've changed the code to re-use the same code that we use to generate the cookie_string.

account_page.php
atrol: If you have access to the database you can act as another user by using the unencrypted token.
This is not possible with encrypted password.

account_page.php
```diff
@@ -257,18 +269,26 @@
...
 </div>
-<?php # check if users can't delete their own accounts
-if( ON == config_get( 'allow_account_delete' ) ) { ?>
 
-<!-- Delete Button -->
 <div class="form-container">
+	<fieldset>
+	<span class="submit-button">
+<!-- Revoke Token: not we know that user account is not protected -->
+<form method="post" action="account_revoke_token.php">
```
atrol: not we know?
Do you mean: now we know?

account_page.php
vboctor: You can do the same with the cookie string, which is already in the database in un-encrypted format.

atrol: No time to have a deeper look at the moment, but isn't there a major difference that the cookie string has a limited lifetime (session)?

dregad:
I don't have much time to analyze this at the moment.

A couple of remarks on the UI:
a306e1b4-90f7-11e5-8027-47248a31836f

> Account page populates the token, if empty

Wouldn't it be better to not generate a token automatically, and instead let the user request one if/when it is needed ?

>  There will be follow up work to support tokens to be passed in headers in the API

In other words, this PR does not actually introduce any useful functionality. I would prefer it to include API changes as well, so that when merged, a single branch covers all the related changes.

Is this something you're targeting at 1.3 ?

vboctor:
> In other words, this PR does not actually introduce any useful functionality

This PR enables the following functionality:
- Life cycle management for the token (generate, get, revoke).
- Logging in with the token instead of password via the API.  The token is accepted in the API "password" parameter.

The work that will be added is also supporting the token in the header.

> token label adjustment

Will do.

> Moving button

The reason we don't have the button at the location you highlighted are:
- There is another "Delete" button that is already implemented in the same location as "Revoke Token" but it is hidden by default based on configuration.  So it is not a newly introduced well of buttons.
- In the case where we have a form (in a div/table) + a button, we don't typically add other buttons next to it to avoid breaking html order of start/end tags (from memory).

> Wouldn't it be better to not generate a token automatically, and instead let the user request one if/when it is needed ?

Not sure why this is better.

> Is this something you're targeting at 1.3 ?

Yes, I'm targeting this for 1.3.  This includes the schema change, we can always build more features on the functionality later.  Just signing into the API with a token instead of the password is the main win.

cproensa:
Hello Victor
Why not using a dedicated table for these tokens? Seems like extending the current user table is limiting prematurely for future enhancements.

One benefit from a dedicated table is a 1..n relation, even if not implemented at the moment, its cleaner to have it now that to move the data out of the user table in the future.
As a potential feature, I'd think that having multiple tokens may be useful. This way, one could differentiate the access from different external agents, and revoke them individually, for example. 

Also, it can be useful to have some additional fields related to each token, like a date of last access, creation date, etc, which are directly associated to 'token' but not to 'user'

atrol:
> Why not using a dedicated table for these tokens?

Has been considered in description of https://www.mantisbt.org/bugs/view.php?id=17766
Would be Ok for 2.x, to many changes for 1.3.

cproensa:
>> Has been considered in description of https://www.mantisbt.org/bugs/view.php?id=17766

>Would be Ok for 2.x, to many changes for 1.3.

True, I know. But my point is: the functionality is delayed for later release. The data model can be implemented for this 1.3 target.

If the features are planned for v2, this data model will be valid only for the 1.3 cycle, and is destined to be deprecated.

account_page.php
vboctor: The cookie may have a limited lifetime in the browser (1 session or 1 year in case of "remember me"), but the value of the cookie is valid and doesn't change.  So if user gets it from the database, then they can login as the user without issue.  We only rotate the value of the cookie when the user changes their password, hence, causing them to be logged out of all browsers.  Hence, there is no difference in terms of access I can get between the token vs. cookie except that cookie gives me interactive web access and token gives me API access.

account_page.php
vboctor: Yes.  But the whole form is only viable because we have ensure that user in unprotected, hence, I removed the comment.

vboctor:
@cproensa The goal of this PR is to get enable us to accept a long auto-generated token instead of a password for our APIs in v1.3.  There is a lot more we can do with tokens and the schema will depend on our approach.  For example, do we support scopes, description of each token, etc.  We can always go and write a schema upgrade that would move the existing tokens to a new table.

@dregad I've applied the form change and answered your questions.  We could build a new page for this and only show / generate token when the user navigates to it.  If you really think we need this, let me know and I'll try to fit it in.

@atrol I also addressed your feedback.

@atrol @dregad I would like to get this in, please let me know if there is anything that is blocking.

account_page.php
atrol: Seems there is an existing security issue (cookie) and you want to add another one (token).

vboctor: @atrol What approach do you have in mind?

If we only disclose the token to user once, then we can use a SHA implementation.  In such case we will need to support multiple tokens, etc.

If we want to show the token to the user multiple times, then we need to encrypt / decrypt and if we do so, then the data is accessible if someone has access to config + db rather than just db in the current implementation.

If you have access to the db, then game over for MantisBT data.  However, we try to protect the password since users tend to re-use them elsewhere and we really don't need to have them beyond in the hash form.

dregad:
>> In other words, this PR does not actually introduce any useful functionality. I would prefer it to include API changes as well, so that when merged, a single branch covers all the related changes.

I apologize if that came across as rude or something. What I actually meant to say is that it does not introduce something that is of immediate benefit to the users, since it's just the foundation, and it's not actually used anywhere.

>> Wouldn't it be better to not generate a token automatically, and instead let the user request one if/when it is needed ?

> Not sure why this is better.

The general idea idea was, following atrol's securitarian argument, to avoid opening a door when it is not actually needed by the user to use, i.e. if you want to use the SOAP API then you need to generate your token first.

> In the case where we have a form (in a div/table) + a button, we don't typically add other buttons next to it to avoid breaking html order of start/end tags (from memory).

I think you're right. I rest my case.

With regards to 1:N relationship to tokens, I'm with @cproensa's here... The suggestion to implement the "proper" data model now makes sense, even if we only use it as a 1:1 relationship initially. The overhead is not that big, and it would be fairly simple to code now. I think it's quite likely we may be needing multiple tokens in the future, so that will save us the trouble of having to implementing an upgrade function later.

+1 for dedicated auth tokens table

vboctor:
See https://github.com/mantisbt/mantisbt/pull/685 instead of this pull request.

PR #682, opened 2015-11-27
Fixed var scope and added missing semi columns.

JeremyVriens:


dregad:
Fixed in d01292246b303473ab8a5bd515f2b9d93929d833. Thanks !

PR #683, opened 2015-11-28
Fix email notifications for re-assign / unassign

vboctor:
- When an issue is unassigned, trigger an "unassigned" notification and
include the user that was unassigned.
- When an issue is re-assigned, trigger an "owner" change notification,
rather than issue update.

Fixes #19731
Fixes #20321


dregad:
Commit message lines seem a bit long, could you wrap at 72 chars ? Also,

> When an issue as unassigned, trigger an "unassigned" notification and include the user that was unassigned

When an issue *is* unassigned [...] ?

+1

PR #684, opened 2015-11-28
Add 'email on updated' to notifications config

vboctor:
Fixes #20322

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -1555,3 +1555,27 @@ function email_shutdown_function() {
 		email_send_all();
 	}
 }
+
+/**
+ * Get the list of supported email actions.
+ *
+ * @return array List of actions
+ */
+function email_get_actions() {
+	$t_actions = array( 'updated', 'owner', 'reopened', 'deleted', 'bugnote', 'relation' );
+
+	if( config_get( 'enable_sponsorship' ) == ON ) {
+		$t_actions[] = 'sponsor';
+	}
+
+	$t_statuses = MantisEnum::getAssocArrayIndexedByValues( config_get( 'status_enum_string' ) );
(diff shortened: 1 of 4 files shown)


core/email_api.php
```diff
@@ -1555,3 +1555,29 @@ function email_shutdown_function() {
...
+ * @return array List of actions
+ */
+function email_get_actions() {
+	$t_actions = array( 'updated', 'owner', 'reopened', 'deleted', 'bugnote' );
+
+	if( config_get( 'enable_sponsorship' ) == ON ) {
+		$t_actions[] = 'sponsor';
+	}
+
+	$t_actions[] = 'relation';
```
dregad: Any particular reason why this is not in the initial array ?

core/email_api.php
```diff
@@ -1555,3 +1555,29 @@ function email_shutdown_function() {
...
+
+	$t_statuses = MantisEnum::getAssocArrayIndexedByValues( config_get( 'status_enum_string' ) );
+	ksort( $t_statuses );
+	reset( $t_statuses );
+
+	foreach( $t_statuses as $t_status => $t_label ) {
```
dregad: No need for `$t_status =>` if you're not using it

dregad:
+1 with minor comments

Good call moving redundant code to dedicated email_get_actions() API.

core/email_api.php
vboctor: Nope, this was the case for the original code, I'll combine it with the initial array initialization.

core/email_api.php
vboctor: Yep, I removed it.  It was used in one case of the old code and not in the other.  I generally don't mind including it for readability, but will remove it.

PR #685, opened 2015-11-29
Use API tokens instead of user passwords for API auth

vboctor:
Here is an alternate implementation for API tokens.  It includes the following:

- A dedicated table for api tokens containing id, name, hash, date_created, date_used.
- Using of sha1() hashing for verifying tokens.
- Tokens are 32 characters long.
- Tokens are valid without or with spaces allowing us to format them for display, but for now we don't.
- Users by default have no tokens.
- Users can specify the purpose/name of token on creation and then get to see the token once.
- Users can revoke specific tokens.
- Users have an API tokens page where they can see all their tokens (name, creation time, last used time).
- API token management is done via its own 'API Tokens' page, not 'My Account' page.  Similar to preferences page for example.


atrol:
Thanks for the new PR.
Looks quite good at first sight.
Will hopefully find some time the next days for a deeper look and tests.

Some PHPDoc (parameter types) missing in api_token_api.php

dregad:
Thanks for the alternative approach. I'll review this PR some time next week.

vboctor:
Thanks @atrol and @dregad.  The PR is now complete and ready for checkin.  I would appreciate your feedback so that I can address before releasing 1.3.0rc1 this weekend.

For picking the token from the headers, I'm thinking of doing that when we introduce the REST API since there won't be a username/password fields in the methods there.  For now, leverage these fields to pass the username/password.

core/api_token_api.php
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Calculate the hash for the token.
+ * @param string $p_token The plain token.
+ * @return string The one way hash for the token
+ * @access public
+ */
+function api_token_hash( $p_token ) {
+	# We ignore spaces in tokens.  We inject spaces for users for readability only.
+	return sha1( str_replace( ' ', '', $p_token ) );
```
atrol: str_replace is no longer needed, the injection of spaces has been removed.

lang/strings_english.txt
```diff
@@ -362,6 +362,18 @@ $s_update_user_button = 'Update User';
 $s_verify_warning = 'Your account information has been verified. The account confirmation message you have received is now invalid.';
 $s_verify_change_password = 'You must set a password here to allow you to log in again.';
 
+# API tokens page
+$s_api_tokens_link = 'API Tokens';
+$s_api_token_create_form_title = 'Create API Token';
+$s_api_token_create_button = 'Create API Token';
+$s_api_token_name = 'Token Name';
+$s_api_token_disclose_message = 'Token to use for accessing API.  Note that this token will only be displayed once and spaces are ignored.';
```
atrol: Mentioning spaces no longer needed.
Maybe show in red/bold that the token is displayed only one time.

core/api_token_api.php
```diff
@@ -0,0 +1,162 @@
...
+	db_query( $t_query, array( $p_api_token_id, $p_user_id ) );
+}
+
+/**
+ * Gets the max length for the API token name.
+ * @return integer Maximum length for API token name.
+ * @access public
+ */
+function api_token_name_max_length() {
+	return 128;
```
atrol: We implemented another approach for similar fields (e.h. user name).
Used a constant for it, define( 'DB_FIELD_SIZE_USERNAME', 255 );

core/api_token_api.php
```diff
@@ -0,0 +1,162 @@
...
+ * @access public
+ */
+function api_token_create( $p_token_name, $p_user_id ) {
+	if( is_blank( $p_token_name ) ) {
+		error_parameters( lang_get( 'token_name' ) );
+		trigger_error( ERROR_EMPTY_FIELD, ERROR );
+	}
+
+	$t_token_name = trim( $p_token_name );
+
```
atrol: Check also length of token_name (api_token_name_max_length or maybe a constant, see other comment) to avoid database error message.
The existing maxlength in form is not enough as it can be manipulated.

core/api_token_api.php
```diff
@@ -0,0 +1,162 @@
...
+ * Create an API token
+ *
+ * @param string $p_token_name The name (description) identifying what the token is going to be used for.
+ * @param integer $p_user_id The user id.
+ * @return string The plain token.
+ * @access public
+ */
+function api_token_create( $p_token_name, $p_user_id ) {
+	if( is_blank( $p_token_name ) ) {
+		error_parameters( lang_get( 'token_name' ) );
```
atrol: Should be 
```PHP
lang_get( 'api_token_name' )
```

core/api_token_api.php
atrol: Ensure that token names are unique by
a) creating unique index in database +
b) using a function to check before, something like api_token_is_name_unique similar to custom_field_is_name_unique

core/api_token_api.php
rombert: If we want to make sure that an attacker does not obtain the plain-text value of the token, we should not relay on SHA-1. Maybe use the `hash` function with a `sha256` argument?

vboctor: @atrol str_replace + comment about spaces removed.
@rombert changed to SHA256 instead of SHA1.

lang/strings_english.txt
vboctor: Removed reference to spacing and added formatting.

core/api_token_api.php
vboctor: Removed the max length functional and replaced it by a constant.

core/api_token_api.php
vboctor: Added validation for max length.  Though, we don't typically do that for other APIs.

I'm not sure about the benefit for enforcing uniqueness for token names.  For example, I may create two tokens to use for the same purpose.  Hence, same name makes sense.  What is really going to differentiate them is the creation timestamp and last used timestamp.  So I'm not sure there is value of having the user try to use "token for iPhone app 2" just to make it unique.

core/api_token_api.php
vboctor: Fixed.

vboctor:
@atrol @rombert Thanks for your review.  I've applied all the suggested changes.  The only exception is enforcing uniqueness for token names.  See my reasoning in dedicated comment.

core/api_token_api.php
```diff
@@ -0,0 +1,156 @@
...
+ * @access public
+ */
+function api_token_create( $p_token_name, $p_user_id ) {
+	if( is_blank( $p_token_name ) ) {
+		error_parameters( lang_get( 'api_token_name' ) );
+		trigger_error( ERROR_EMPTY_FIELD, ERROR );
+	}
+
+	$t_token_name = trim( $p_token_name );
+	if( strlen( $t_token_name ) > DB_FIELD_SIZE_API_TOKEN_NAME ) {
```
atrol: use utf8_strlen

core/api_token_api.php
atrol: > I may create two tokens to use for the same purpose

Why should I create two tokens for one purpose?
Can you provide a use case for this?

Wouldn't it be easier to revoke a certain token if it can be distinguished from similar tokens by name?

core/api_token_api.php
vboctor: Fixed.

atrol:
So we are near +1 from my side.
Remaining:
- Open discussion about unique token names
- Request to salt the token
- Missing documentation

Hope that @dregad will also find some time to review .

core/api_token_api.php
vboctor: I'll enforce uniqueness.  My original position was that I didn't really care if user ends up creating two tokens for their phone.  One for a phone and another when they replace their phone.  Sure they can make it unique by include phone model or new phone, but I didn't see the reason to make them do that.  I generally just want to create new tokens and every now and then I delete tokens that haven't been used for a long time.

vboctor:
@atrol here is the summary.
- Documentation - I think I've addressed it.  Please let me know if I missed something.
- Salt - Last I checked it was possible to change the crypto salt without impacting the cases where it was used.  If we use it in tokens, then we won't be able to change the crypto salt without having to invalidate all user tokens, which would be a bad thing.  Also salt is adding random string to avoid making it easy to use a big dictionary of hashes.  In our cases, we have randomly generated 32-character tokens that are hashed via SHA256.  I think we are over doing it here.  What we should really fix is to use a salt for password + SHA256, compared to no salt + MD5.
- Unique names - I've fixed something in the schema to get user_id out of the primary index, which was a mistake.  Then added an index for user id + token name that is unique.  I've also added validation to make sure token is unique for attempting to add it to the database.

So unless you are not convinced about the salt argument, I think I've addressed all your feedback.

atrol:
> Documentation - I think I've addressed it. Please let me know if I missed something.

I should have written **user** documentation
https://www.mantisbt.org/docs/master/en-US/Admin_Guide/html-single/#admin.pages.account

>  If we use it in tokens, then we won't be able to change the crypto salt without having to invalidate all user tokens

What could be the reason to change the crypto salt?
Maybe if you have been hacked.
Wouldn't it be a good thing in this case if this invalidates all tokens?

> What we should really fix is to use a salt for password + SHA256
Of course, we should.

> Unique names 
Thanks for changing
Code looks good at first sight.
At the moment I don't have time to test it.

vboctor:
@atrol I've updated the documentation.  As for the salting, I'm not convinced about the need for salting at all for the API tokens.  And if we want to consider this, then we would consider it when we implement the same for user passwords where salting is really useful.  I think revoke-all should be a very explicit action and not a side effect of changing a config option.  When we implement salting for passwords, I think it would be a good idea to have a salt per user rather than a global salt value.

api_token_create.php
```diff
@@ -0,0 +1,66 @@
...
+
+<div align="center">
+<br /><br />
+<?php echo $t_disclose_message ?>
+<br />
+<span class="important-msg"><?php echo $t_display_once_message ?></span>
+<br /><br />
+<span class="api-token"><?php echo string_display_line( $t_token ) ?></span>
+<br /><br />
+<?php echo print_bracket_link( 'api_tokens_page.php', lang_get( 'api_tokens_link' ) ) ?>
```
atrol: Remove the echo
print_bracket_link does not return any value and does the echoing itself.

atrol:
Tested also unique names.
Found one more minor issue (print_bracket_link)
+1

api_token_create.php
vboctor: Removed echo statement.

vboctor:
Thanks @atrol for the review.  @dregad any comments from your side?

dregad:
I apologize, but I've been completely under the water this week and despite my intentions, I have not been able to perform proper testing so far. I'll try and do it this week-end; if I don't manage, I'll put my trust in the quality of your work and atrol's review.

admin/schema.php
```diff
@@ -750,5 +750,19 @@ function installer_db_now() {
...
+
+$g_upgrade[200] = array(
+		'CreateTableSQL',array( db_get_table( 'api_token' ), '
+			  id           I UNSIGNED NOTNULL PRIMARY AUTOINCREMENT,
+			  user_id      I DEFAULT \'0\',
+			  name         C(128) NOTNULL,
+			  hash         C(128) NOTNULL,
+			  date_created I UNSIGNED NOTNULL DEFAULT \'0\',
+			  date_used    I UNSIGNED NOTNULL DEFAULT \'0\''
+		) );
```
dregad: While working on [#20357](https://www.mantisbt.org/bugs/view.php?id=20357), I realized that upgrade step 200 is not setting the standard table options that all other tables have.

I opened https://www.mantisbt.org/bugs/view.php?id=20443 to track this.

grangeway: also, look more closely at the table

dregad: @grangeway instead of playing riddles, if you noticed an error say it clearly.

atrol: I asked grangeway via e-mail yesterday and got

> We normally set DEFAULT value of " \'\' " for NOTNULL strings I believe
> - used to be case oracle or mssql I think needed it 
> - it might be that adodb handles internally now

dregad: > We normally set DEFAULT value of " \'\' " for NOTNULL strings I believe

These are not strings, they are integers (dates); other dates in the schema are defaulted to `'1'`, not `''`. I'm not really sure what Oracle or MSSQL have to do with this though. In any case, for consistency, we should probably change the default to 1.

We should also change the *user_id* column to `I UNSIGNED` too, to match user.id.

atrol: name and hash are strings

dregad: Duh.

Ok, so this means every column from the initial definition (except id) will have to be altered :unamused:

dregad: Follow-up in https://www.mantisbt.org/bugs/view.php?id=20472 / PR #700

PR #686, opened 2015-12-05
Provide config option to disable subprojects

vboctor:
This will keep the feature enabled by default.  However, it provides the ability to disable this feature so it won't be used by managers and administrators.

Fixes #20348

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -4069,6 +4069,13 @@ $g_due_date_view_threshold = NOBODY;
 ################
 
 /**
+ * Whether sub-projects feature should be enabled.  Before turning this flag OFF,
+ * make sure all sub-projects are moved to top level projects, otherwise
+ * they won't be accessible.
+ */
+$g_subprojects_enabled = ON;
+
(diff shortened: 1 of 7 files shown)


config_defaults_inc.php
```diff
@@ -4069,6 +4069,13 @@
 ################
 
 /**
+ * Whether sub-projects feature should be enabled.  Before turning this flag OFF,
+ * make sure all sub-projects are moved to top level projects, otherwise
```
dregad: Move the `Before turning...` bit to the next line for [better phpDoc formatting](http://www.phpdoc.org/docs/latest/guides/docblocks.html#summary) (summary vs description)

dregad:
It would be worth adding the description in the PR's comment to the commit message.

I'm fine with this change on principle, but don't have time to actually test it atm. Did you verify that the project selector works properly ?

vboctor:
Yes, it works fine.  However, I elected not to mess up with the queries, but require that the administrator move sub-projects to top level before disabling the feature.

PR #687, opened 2015-12-10
Do not print inline errors in non-HTML output

dregad:
There are several PHP scripts in MantisBT which are used to dynamically
generate non-HTML content, e.g. javascript_*.php, css/common_config.php,
etc. Printing HTML code in these files generates invalid ouput.

This commit allows these special scripts to prevent inline errors
messages from being printed by defining a constant prior to including
core.php (or the Error API), as follows:

define( 'DISABLE_INLINE_ERROR_REPORTING', true );

Fixes [#20372](https://www.mantisbt.org/bugs/view.php?id=20372)

### DIFF
--- a/browser_search_plugin.php
+++ b/browser_search_plugin.php
@@ -27,6 +27,9 @@
  * @uses gpc_api.php
  */
 
+# Prevent output of HTML in the content if errors occur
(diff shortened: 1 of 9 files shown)


dregad:
Without feedback for 7 days, I assume nobody objects to this. Will merge shortly.

PR #688, opened 2015-12-11
Use global default_notify if flag not in DB config

vboctor:
The web UI currently doesn't support the explicit notify flag,
hence, when default_notify_flag and notify_flag are saved based
on administrator using web UI to override configs, such settings
will not include the 'explicit' flag causing it to be considered
false.  The fix is to fallback to the global config when overridden
configs in db for notify_flags and default_notify_flags don't have
the requested flag.

Fixes #20371

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -215,22 +215,34 @@ function email_ensure_not_disposable( $p_email ) {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


dregad:
+1

PR #689, opened 2015-12-12
Disallow deleting or disabling last admin

vboctor:
- When checking for remaining admins, exclude disabled ones.
- Don't worry about changes to already disabled users.
- Complain when disabling last administrator, not just reducing their
  access level.

Fixes #20381

### DIFF
--- a/manage_user_delete.php
+++ b/manage_user_delete.php
@@ -62,10 +62,10 @@ $t_user = user_get_row( $f_user_id );
 # current user.
 access_ensure_global_level( $t_user['access_level'] );
 
-# check that we are not deleting the last administrator account
+# Check that we are not deleting the last administrator account
 $t_admin_threshold = config_get_global( 'admin_site_threshold' );
 if( user_is_administrator( $f_user_id ) &&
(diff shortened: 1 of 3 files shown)


manage_user_delete.php
```diff
@@ -62,10 +62,11 @@
 # current user.
 access_ensure_global_level( $t_user['access_level'] );
 
-# check that we are not deleting the last administrator account
+# Check that we are not deleting the last administrator account
 $t_admin_threshold = config_get_global( 'admin_site_threshold' );
-if( user_is_administrator( $f_user_id ) &&
-	 user_count_level( $t_admin_threshold ) <= 1 ) {
+if( (int)$t_user['enabled'] === 1 &&
```
atrol: When reading this I thought:
Isn't it questionable in terms of security that user_is_administrator and current_user_is_administrator return true for disabled administrators.

vboctor: Good point @atrol.  I've opened a dedicated bug for this and fixed it as part of this pull request.

atrol:
+1

PR #690, opened 2015-12-12
Support notes field as a column field

vboctor:
- Added support for notes field in View Issues, Print Issues, CSV,
  and Excel.
- Add vertical align top to the style for column values which is
  useful when the view includes a multi-line column.
- Apply CSS styles for column values to Print Issues page.

Fixes #20383

### DIFF
--- a/core/bugnote_api.php
+++ b/core/bugnote_api.php
@@ -433,6 +433,51 @@ function bugnote_get_all_visible_bugnotes( $p_bug_id, $p_user_bugnote_order, $p_
 }
 
 /**
+ * Build a string that captures all the notes visible to the logged in user along with their
+ * metadata.  The string will contain information about each note including reporter, timestamp,
+ * time tracking, view state.  This will result in multi-line string with "\n" as the line
+ * separator.
+ *
+ * @param integer $p_bug_id             A bug identifier.
+ * @param integer $p_user_bugnote_order Sort order.
+ * @param integer $p_user_bugnote_limit Number of bugnotes to display to user.
+ * @param integer $p_user_id            An user identifier.
+ * @return string The string containing all visible notes.
+ * @access public
+ */
+function bugnote_get_all_visible_as_string( $p_bug_id, $p_user_bugnote_order, $p_user_bugnote_limit, $p_user_id = null ) {
+	$t_notes = bugnote_get_all_visible_bugnotes( $p_bug_id, $p_user_bugnote_order, $p_user_bugnote_limit, $p_user_id );
+	$t_date_format = config_get( 'normal_date_format' );
+	$t_show_time_tracking = access_has_bug_level( config_get( 'time_tracking_view_threshold' ), $p_bug_id );
+
+	$t_output = '';
+
+	foreach( $t_notes as $t_note ) {
+		$t_note_string = '@' . user_get_name( $t_note->reporter_id );
+		if ( $t_note->view_state != VS_PUBLIC ) {
+			$t_note_string .= ' (' . lang_get( 'private' ) . ')';
+		}
+
+		$t_note_string .= ' ' . date( $t_date_format, $t_note->date_submitted );
+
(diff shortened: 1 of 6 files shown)


core/bugnote_api.php
```diff
@@ -433,6 +433,51 @@ function bugnote_get_all_visible_bugnotes( $p_bug_id, $p_user_bugnote_order, $p_
...
+ * @param integer $p_bug_id             A bug identifier.
+ * @param integer $p_user_bugnote_order Sort order.
+ * @param integer $p_user_bugnote_limit Number of bugnotes to display to user.
+ * @param integer $p_user_id            An user identifier.
+ * @return string The string containing all visible notes.
+ * @access public
+ */
+function bugnote_get_all_visible_as_string( $p_bug_id, $p_user_bugnote_order, $p_user_bugnote_limit, $p_user_id = null ) {
+	$t_notes = bugnote_get_all_visible_bugnotes( $p_bug_id, $p_user_bugnote_order, $p_user_bugnote_limit, $p_user_id );
+	$t_date_format = config_Get( 'normal_date_format' );
```
atrol: lower case config_get

vboctor: Fixed.  Not sure how this happened :)

core/columns_api.php
```diff
@@ -900,6 +902,21 @@ function print_column_title_description( $p_sort, $p_dir, $p_columns_target = CO
...
+ * Print table header for notes column.
+ *
+ * @param string  $p_sort           Sort.
+ * @param string  $p_dir            Direction.
+ * @param integer $p_columns_target See COLUMNS_TARGET_* in constant_inc.php.
+ * @return void
+ * @access public
+ */
+function print_column_title_notes( $p_sort, $p_dir, $p_columns_target = COLUMNS_TARGET_VIEW_PAGE ) {
+	echo '<th class="column-description">';
```
atrol: column-notes?

core/csv_api.php
```diff
@@ -344,6 +344,17 @@ function csv_format_description( BugData $p_bug ) {
 }
 
 /**
+ * return the description
```
atrol: return the notes

core/excel_api.php
```diff
@@ -463,6 +463,16 @@ function excel_format_description( BugData $p_bug ) {
 }
 
 /**
+ * Gets the formatted description field.
```
atrol: notes fields

core/excel_api.php
```diff
@@ -463,6 +463,16 @@ function excel_format_description( BugData $p_bug ) {
...
+ * @param BugData $p_bug A bug object.
+ * @return string The formatted description (multi-line).
```
atrol: formatted notes

vboctor:
@atrol Thanks for the review.  I pushed an updated version with your feedback addressed.

css/default.css
```diff
@@ -85,12 +85,13 @@ td.my-buglist-description	{ width: 100%; vertical-align: top; }
 /**
  * view_all_bug_page.php
  */
-#buglist td, #buglist th	{ text-align: center; }
+#buglist td, #buglist th	{ text-align: center; vertical-align: top; }
```
dregad: I don't think the `th` should be top-aligned.

dregad:
Depending on the number of notes associated to issues, including this column may have a non-negligible impact on performance.

What happens if bugnote text already contains "\n" ? In other words, is it an issue that the column contents can't be broken down further after e.g. download to XL.

css/default.css
vboctor: Fixed.

vboctor:
@dregad The notes column is not really practical in the common usage scenario, i.e. View Issues page.  It is mostly requested for csv/excel export where performance isn't really an issue.

As for the new lines in notes, it is not really an issue.  We have a line separator between notes and first line of each notes refers to author, timestamp, private vs. public, etc.  It is mainly for human consumption re-parsing is do-able, but I don't see it as the main scenario.

PR #691, opened 2015-12-13
Support exporting billing report to CSV and Excel

vboctor:
- Added billing_api.php
- Implemented 'Export to Excel' using billing api
- Implemented 'Export to CSV' using billing api

Will open a code cleanup bug to move more code to use billing_api.

Fixes #7932

### DIFF
--- a/core/billing_api.php
+++ b/core/billing_api.php
@@ -1,0 +1,145 @@
+<?php
+# MantisBT - A PHP based bugtracking system
+
+# MantisBT is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# MantisBT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Billing API
+ *
+ * @package CoreAPI
+ * @subpackage BillingAPI
+ * @copyright Copyright 2000 - 2002  Kenzaburo Ito - kenito@300baud.org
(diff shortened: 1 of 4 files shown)


dregad:
2 high-level comments, without looking at the code...

- I don't think we should be spending effort on a deprecated feature [[1]](http://thread.gmane.org/gmane.comp.bug-tracking.mantis.devel/3348/focus=3355),[[2]](https://www.mantisbt.org/blog/?p=416)
- Now that we're in release candidate mode, it may be best to postpone adding a new API to introduce new features until after 1.3.0 goes live.

That being said, the decision to deprecate time tracking was not mine, and I don't feel strongly about maintaining it even though I do agree that its design is dated and functionality limited. There are a few plugins out there which do a much better job at it.

vboctor:
I think the export functionality is useful specially if we are going to deprecate the feature.  I haven't done a comparison with existing plugins, so not sure about how they compare.

As for locking down 1.3.0, I don't see this change as disruptive, as you can see, other than adding the link to the billing page, the changes are mostly new APIs to provide the export functionality.  I think we will keep allowing non-disruptive changes to master until we fork master-1.3.x.

core/billing_api.php
```diff
@@ -0,0 +1,144 @@
...
+
+	access_ensure_project_level( config_get( 'time_tracking_reporting_threshold' ), $p_project_id, $p_user_id );
+}
+
+/**
+ * Returns an array of bugnote stats
+ * @param integer $p_project_id A project identifier.
+ * @param string  $p_from       Starting date (yyyy-mm-dd) inclusive, if blank, then ignored.
+ * @param string  $p_to         Ending date (yyyy-mm-dd) inclusive, if blank, then ignored.
+ * @param integer $p_cost       Cost.
```
atrol: $p_cost_per_hour

vboctor: Fixed.

vboctor:
@atrol I've addressed your comments.
@dregad let me know your feedback.

dregad:
In terms of implementation, considering that CSV vs Excel export are nearly identical in terms of structure, I would personally have gone for an object-based approach (e.g. having a *BillingExport* class and 2 child classes *BillingExportExcel* and *BillingExportCsv*) to reduce code duplication.

That being said, I don't mind if you merge this.

PR #692, opened 2015-12-15
Fix required on resolve/close for custom fields

vboctor:
If a custom field is marked as required on resolved or closed, it wasn't
enforced.  It was only enforced for the regular issue update page.

Fixes #20386


dregad:
+1

PR #693, opened 2015-12-20
Add entry for new log level text

cproensa:
After changes introduced by:
95f947f "improve email logging" #0020403
there are system notices for undefined offset
due to a missing entry in $g_log_levels array.


dregad:
Fixed in e54d31bf468a5b6bda28c5df343b8420773e629e

Thanks !

PR #694, opened 2015-12-20
Fix variable out of scope in email_api (regression #0020403)

cproensa:
After changes introduced by:
95f947f "improve email logging" #0020403


dregad:
Fixed in 6cbcb2541eb442cf3f13588513060554f28a3f7b

Thank you !

PR #696, opened 2015-12-21
Implement bugnotes_count, attachment_count in CSV, Excel export

cproensa:
As per issue
#0019284

Columns bugnotes_count, attachment_count are being silently dropped from the columns set for CSV and Excel export, even when the user is able to select them.

This is still an incomplete fix, because it opens another show-stopper:
The bugnote cache function chokes when the list of bug ids passed is long enough. This happens easily as the csv/excel export are not limited by pagination.
The query used has an IN clause, which grows out of the runtime limits.
(@dregad This brings up the issue of long IN sql clauses, which has been talked before.)

seleccion_084

For further addressing this case, (and until a more unified approach is taken for the IN issue), the query could be partitioned. In that case, i suggest to review first Pr #650, which is already refactoring the affected code (for changes to be done on the query, i'd rather do it on top of those)

### DIFF
--- a/core/csv_api.php
+++ b/core/csv_api.php
@@ -438,3 +438,35 @@ function csv_format_due_date( BugData $p_bug ) {
 function csv_format_sponsorship_total( BugData $p_bug ) {
 	return csv_escape_string( $p_bug->sponsorship_total );
 }
+
+/**
+ * return the attachment count for an issue
+ * @param BugData $p_bug A BugData object.
+ * @return string
+ * @access public
+ */
+function csv_format_attachment_count( BugData $p_bug ) {
+	# Check for attachments
+	$t_attachment_count = 0;
+	if( file_can_view_bug_attachments( $p_bug->id, null ) ) {
+		$t_attachment_count = file_bug_attachment_count( $p_bug->id );
+	}
+	return csv_escape_string( $t_attachment_count );
+}
+
+/**
+ * return the bug note count for an issue
+ * @param BugData $p_bug A BugData object.
+ * @return string
+ * @access public
+ */
+function csv_format_bugnotes_count( BugData $p_bug ) {
+	# grab the bugnote count
+	$t_bugnote_stats = bug_get_bugnote_stats( $p_bug->id );
+	if( null !== $t_bugnote_stats ) {
+		$t_bugnote_count = $t_bugnote_stats['count'];
+	} else {
+		$t_bugnote_count = 0;
+	}
+	return csv_escape_string( $t_bugnote_count );
+}
--- a/core/excel_api.php
+++ b/core/excel_api.php
@@ -252,15 +252,6 @@ function excel_format_reporter_id( BugData $p_bug ) {
 }
 
 /**
- * Gets the formatted number of bug notes.
- * @param BugData $p_bug A bug object.
- * @return string The number of bug notes.
- */
-function excel_format_bugnotes_count( BugData $p_bug ) {
-	return excel_prepare_number( $p_bug->bugnotes_count );
-}
-
-/**
  * Gets the formatted handler id.
  * @param BugData $p_bug A bug object.
  * @return string The handler user name or empty string.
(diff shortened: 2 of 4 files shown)


core/csv_api.php
```diff
@@ -438,3 +438,35 @@ function csv_format_due_date( BugData $p_bug ) {
...
+/**
+ * return the attachment count for an issue
+ * @param BugData $p_bug A BugData object.
+ * @return string
+ * @access public
+ */
+function csv_format_attachment_count( BugData $p_bug ) {
+	# Check for attachments
+	$t_attachment_count = 0;
+	if( file_can_view_bug_attachments( $p_bug->id, null ) ) {
```
vboctor: I wonder if we should move these access checks inside the API that returns the count, rather than having to check before it is called.  I can also imagine a scenario where a user can view their own attachments but not other people's (though this feature is not implemented yet), in such case having the API take the access check call into consideration makes sense.

core/excel_api.php
```diff
@@ -558,6 +549,39 @@ function excel_format_sponsorship_total( BugData $p_bug ) {
...
+ * @return string
+ * @access public
+ */
+function excel_format_attachment_count( BugData $p_bug ) {
+	# Check for attachments
+	$t_attachment_count = 0;
+	if( file_can_view_bug_attachments( $p_bug->id, null ) ) {
+		$t_attachment_count = file_bug_attachment_count( $p_bug->id );
+	}
+	return excel_prepare_string( $t_attachment_count );
```
vboctor: Should use excel_prepare_number() instead of excel_prepare_string().

core/excel_api.php
```diff
@@ -558,6 +549,39 @@ function excel_format_sponsorship_total( BugData $p_bug ) {
...
+ */
+function excel_format_bugnotes_count( BugData $p_bug ) {
+	# grab the bugnote count
+	$t_bugnote_stats = bug_get_bugnote_stats( $p_bug->id );
+	if( null !== $t_bugnote_stats ) {
+		$t_bugnote_count = $t_bugnote_stats['count'];
+	} else {
+		$t_bugnote_count = 0;
+	}
+	return excel_prepare_string( $t_bugnote_count );
```
vboctor: Should use excel_prepare_number() instead of excel_prepare_string().

core/excel_api.php
```diff
@@ -558,6 +549,39 @@ function excel_format_sponsorship_total( BugData $p_bug ) {
...
+}
+
+
```
vboctor: Remove extra blank line.

core/helper_api.php
```diff
@@ -528,15 +528,9 @@ function helper_get_columns_to_view( $p_columns_target = COLUMNS_TARGET_VIEW_PAG
 	if( $p_columns_target == COLUMNS_TARGET_CSV_PAGE || $p_columns_target == COLUMNS_TARGET_EXCEL_PAGE ) {
 		$t_keys_to_remove[] = 'selection';
 		$t_keys_to_remove[] = 'edit';
-		$t_keys_to_remove[] = 'bugnotes_count';
-		$t_keys_to_remove[] = 'attachment_count';
 		$t_keys_to_remove[] = 'overdue';
```
vboctor: How about adding the 'overdue' column as well.  For example, true/false or 1/0.

cproensa:
Regarding issues with 0 notes, or comments, should the field result in blank, or explicitly show the number 0?

core/csv_api.php
cproensa: file_bug_attachment_count() should remain as unfiltered count, it is used by core to move bugs (see file_api:file_move_bug_attachments)
However, a new function in bug/user/access/etc api may be convenient.

cproensa: @vboctor 
> I can also imagine a scenario where a user can view their own attachments but not other people's (though this feature is not implemented yet),

Actually, it's implemented, but is not working quite right for bug listing: #0020411

(note to self: review callers for "file_can_view_bug_attachments")
Will try a fix here including your suggestion: access logic is better placed in the queried api.
However, this may collide with my other PR #654. I'd rather settle that one before.

vboctor:
> Regarding issues with 0 notes, or comments, should the field result in blank, or explicitly show the number 0?

@cproensa the field should always be a number representing the accessible notes for the issue.  It shouldn't behave differently when the issue really has no notes vs. has notes that are inaccessible to me.

core/csv_api.php
vboctor: My main point is to not having the calling scripts/apis worry about access checks.  Whatever access checks that we already support should live in the APIs of the resource being called.  For examples, bugnote_api apis should apply all access checks rather than csv_api/excel_api.  I was just giving examples of access checks, and not advocating implementing them as part of this PR.

vboctor:
btw, here is a related issue - doesn't need to be part of this pull request, but I thought I would raise it since we have been improving the supported fields in exported list:
https://www.mantisbt.org/bugs/view.php?id=9096

cproensa:
I am keeping this branch as done, apart from corrections, etc
This one is blocked by #650, which solves the long sql problem that arises from allowing the bugnote count field.

Suggestions for more fields: overdue, tags, ... will be treated as future improvements after this one is settled.
Also the fixing of attachment access checks, and moving it into api functions, is kept for future actions

cproensa:
close, until related problem with long SQL queries is resolved

PR #697, opened 2015-12-21
Print caller trace in file log

cproensa:
Include info for call trace when the log
destination is "file". This information is
already printed on "page" destination

Fix #20408


vboctor:
:+1:

vboctor:
Merged via 952a7aeabbdc97d6ebb8a83d0a82c64688613cde.

vboctor:
Thanks @cproensa.

PR #698, opened 2015-12-25
0020424 csv memory

cproensa:
There is a problem with the exports to CSV and Excel functions when retrieving a large list of issues to export.
It is getting the list through filter_api:filter_get_bug_rows(), which pre-process the list loading it into memory, and doing cache of some bug data. Because of this, the memory usage grows linearly and php memory limit may be reached.
Excel export has a work around that uses pagination to get the data in batches, however there is still some redundancy because each time the function filter_get_bug_rows is called, a new query is being composed and executed.

The changes here are:
filter_get_bug_rows is a function which does all in one:
- prepare pagination
- compose query conditions
- execute query and load results in memory, add to cache related data

First modification is refactoring filter_get_bug_rows by separating the outer layers of logic (the core query composer is untouched and still that ugly block of code with hundreds of lines...).
However, now the result set can be accessed before its loaded and cached into memory.
To me it makes sense, to separate the query building and execution, from pagination which is very screen oriented.

Next modifications are the export procedures, which now reads directly from the result set. 
By reading rows in batches, memory used to store data and cache is controlled, and cleared between each batch.

fixes #0020424

### DIFF
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -749,8 +749,8 @@ function mc_issue_add( $p_username, $p_password, stdClass $p_issue ) {
 	$t_bug_data->resolution = $t_resolution_id;
 	$t_bug_data->projection = $t_projection_id;
 	$t_bug_data->category_id = $t_category_id;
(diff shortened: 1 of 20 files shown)


cproensa:
opened PR #832 to properly target 1.3.x
after rebasing the branch, this PR is mostly broken, since it was created against master

PR #699, opened 2016-01-02
Add utf8mb4 support and various fixes/improvements to schema.php

dregad:
This includes the following changes

1. utf8mb4 support (only applies if MySQL >= 5.5.3), fixes [#20431](https://www.mantisbt.org/bugs/view.php?id=20431)
  - Set default charset to utf8mb4 
  - Add admin check to check for utf8 tables and recommend use of utf8mb4
  - reduce user.username column size to 191 chars (fixes [#20465](https://www.mantisbt.org/bugs/view.php?id=20465))

2. Code cleanup and consistency in schema.php
  - consistent use of `"` for create/alter tables/column strings
  - move functions to install API, removed dead code
  - define table options globally
  - use DB's default charset instead of specifying it for each table
  - improve comments

3. Other fixes 
  - do not specify engine when creating tables (fixes [#20357](https://www.mantisbt.org/bugs/view.php?id=20357))
  - use standard table options for new *api_tokens* table (fixes [#20443](https://www.mantisbt.org/bugs/view.php?id=20443))
  - reduce


dregad:
Hmmm looks like I broke pgsql builds somehow :disappointed:

dregad:
From Travis log:
```
Schema step 51: InsertData ( mantis_user_table ) BAD

INSERT INTO mantis_user_table(
		username, realname, email, password,
		date_created, last_visit, enabled, protected, access_level,
		login_count, lost_password_request_count, failed_login_count,
		cookie_string
	)
	VALUES (
		'administrator', '', 'root@localhost', '63a9f0ea7bb98050796b649e85481845',
		'2016-01-02 00:25:17', '2016-01-02 00:25:17', '1', '0', 90,
		3, 0, 0, '
		abb54c48423f22df0aa0954d3f49b2ff19d7239593bdac8e5c0a585a88e74f6f'
	)

ERROR:  value too long for type character varying(64)
```

admin/schema.php
```diff
@@ -16,44 +16,40 @@
...
+ * Array key is the ADOdb datadict driver's name
  */
-function installer_db_now() {
-	global $g_db;
-
-	# Timezone must not be set to UTC prior to calling BindTimestamp(), as
-	# ADOdb assumes a local timestamp and does the UTC conversion itself.
-	return $g_db->BindTimeStamp( time() );
-}
+$t_table_options = array(
```
atrol: I think that $t_table_options for MySQL should include the default charset.
This is to cover the case where Mantis is installed in an existing database where the default charset is not utf8 or utf8mb4.

atrol:
Looks good in general, didn't test.

admin/schema.php
dregad: Good point. I didn't consider that, as we generally create a new DB as part of the installer. I'll try to see if we can somehow set this depending on whether we created the database ourselves or if using an existing one.

admin/schema.php
```diff
@@ -16,44 +16,40 @@
...
+ * included in the corresponding release
  */
 
-if( !function_exists( 'db_null_date' ) ) {
-	/**
-	 * Legacy null date function for installer backwards compatibility
-	 * @return integer
-	 */
-	function db_null_date() {
-		return 0;
```
vboctor: Note that null date is 0 not 1 here, which makes sense.

core/constant_inc.php
```diff
@@ -577,7 +577,7 @@
 define( 'EMAIL_SHUTDOWN_FORCE', 2 );
 
 # Lengths - NOTE: these may represent hard-coded values in db schema and should not be changed.
-define( 'DB_FIELD_SIZE_USERNAME', 255 );
+define( 'DB_FIELD_SIZE_USERNAME', 191 );
 define( 'DB_FIELD_SIZE_REALNAME', 255 );
```
vboctor: Why is realname being 255 is not an issue?

core/constant_inc.php
```diff
@@ -577,7 +577,7 @@
 define( 'EMAIL_SHUTDOWN_FORCE', 2 );
 
 # Lengths - NOTE: these may represent hard-coded values in db schema and should not be changed.
-define( 'DB_FIELD_SIZE_USERNAME', 255 );
+define( 'DB_FIELD_SIZE_USERNAME', 191 );
```
vboctor: I would add a comment explaining why we came up with this magic number and if there is a point in the future where we can revisit.

vboctor:
Looks like the pull request description is truncated at the end.

vboctor:
Good cleanup, thanks @dregad :+1:.  Though it is a tricky part to change in such a big way :) -- but I like that in the future, it will be easier to get things right.  We should capture the schema guidelines for future tables, indices, default values, etc in a wiki page as I mentioned earlier (not sure if you have responded to a comment about this that I posted somewhere).

core/constant_inc.php
atrol: There is no index on realname.
The length of the index is limited.
https://www.mantisbt.org/bugs/view.php?id=20431#c52209

core/constant_inc.php
dregad: > add a comment explaining why we came up with this magic number

Done 3411e0a2d2bb545fa67f0a23019480db3f75623e.

dregad:
> Looks like the pull request description is truncated at the end.

Not sure what happened. Fixed now.

> We should capture the schema guidelines for future tables, indices, default values, etc in a wiki page as I mentioned earlier (not sure if you have responded to a comment about this that I posted somewhere).

That's would indeed be a useful thing to do, and no I did not respond to that, sorry. Here is a stub https://www.mantisbt.org/wiki/doku.php/mantisbt:db_schema

dregad:
There's one thing that bothers me with this, and I'm not really sure how to properly handle it.

Creating the schema (schema.php/install.php) and checking it (admin checks) is one thing, and that PR pretty much covers it.

But there is also the `SET NAMES UTF8` statement in database API. To be consistent, this should be set to *utf8mb4* to match the database setting, but only if the DB actually *is* utf8mb4; if not, charset should be *utf8*.

We could detect the charset dynamically at run-time (e.g. when initializing database API) and store it in a global var, or we could define it in config_inc.php...

Thoughts and advice welcome.

admin/schema.php
atrol: Just got aware that simple including default charset in $t_table_options could end in mixed tables (some utf8 and some utf8mb4)
e.g. this would happen if you update from 1.2 and your database version is >= 5.5.3

atrol:
Consider also that database setting can be utf8mb4 but the table is utf8.

Maybe we should postpone utf8mb4 support until our minimum requirement is MySQL >= 5.5.35.

dregad:
I decided to change this PR's scope, as described in the revised opening text.

The purpose is to avoid delaying bugfixes for the master branch, as well as the code layout improvements which are effectively distinct issues. These have been merged into master a few minutes ago, and the original branch rebased on top of that.

dregad:
In reply to @atrol's latest comment

> Consider also that database setting can be utf8mb4 but the table is utf8.

I did actually, and this is one of the reasons I can't think of a proper solution to actually *detect* which charset is in use.

> Maybe we should postpone utf8mb4 support until our minimum requirement is MySQL >= 5.5.35.

That's one way to handle it. (and that's 5.5.3 by the way - I assume you just made a typo).

Another could be, as I mentioned previously, to store the charset in a config variable which gives the admin control over the whole thing.

I also thought of a 3rd option: create/update the DB back-end to utf8mb4 if we can, but continue to use MySQL's 3-byte utf8 charset (i.e. `SET NAMES UTF8`) for the time being. I think this would offer the same functionality as we have today without side-effects, since utf8mb4 is a superset of utf8, while reducing future effort for admins to upgrade their DB's if and when we actually decide to make utf8mb4 mandatory.

admin/check/check_api.php
```diff
@@ -269,8 +269,12 @@ function check_print_test_warn_row( $p_description, $p_pass, $p_info = null ) {
 /**
  * Verifies that the given collation is UTF-8
  * @param string $p_collation
+ * @param boolean $p_mb4 Check for utf8mb4 instead of utf8
```
k1931547: fix indentation

l266509:

okainov:
Any update on that?

PR #700, opened 2016-01-02
Fixing api_token table structure

dregad:
After careful review following @grangeway's comment in PR 685 [1]:

- user_id field should be *unsigned* int
- default value for varchar columns should be ''
- default value for date fields should be 1 not 0

[1] https://github.com/mantisbt/mantisbt/pull/685#discussion_r48686670

Fixes [#20472](https://www.mantisbt.org/bugs/view.php?id=20472)

### DIFF
--- a/admin/schema.php
+++ b/admin/schema.php
@@ -876,6 +876,14 @@ $g_upgrade[207] = array( 'AlterColumnSQL', array( db_get_table( 'user' ), "
 $g_upgrade[208] = array( 'AlterColumnSQL', array( db_get_table( 'user' ), "
 	email					C(191)	$t_notnull DEFAULT \" '' \"" ) );
 
+$g_upgrade[209] = array( 'AlterColumnSQL', array( db_get_table( 'api_token' ), "
+	user_id					I		UNSIGNED NOTNULL DEFAULT '0',
+	name					C(128)	NOTNULL DEFAULT \" '' \",
+	hash					C(128)	NOTNULL DEFAULT \" '' \",
+	date_created			I		UNSIGNED NOTNULL DEFAULT '1',
+	date_used				I		UNSIGNED NOTNULL DEFAULT '1'"
+	) );
+
 # Release marker: 1.3.0
 
 
--- a/admin/schema.php
+++ b/admin/schema.php
@@ -878,8 +878,6 @@ $g_upgrade[208] = array( 'AlterColumnSQL', array( db_get_table( 'user' ), "
 
 $g_upgrade[209] = array( 'AlterColumnSQL', array( db_get_table( 'api_token' ), "
 	user_id					I		UNSIGNED NOTNULL DEFAULT '0',
-	name					C(128)	NOTNULL DEFAULT \" '' \",
-	hash					C(128)	NOTNULL DEFAULT \" '' \",
 	date_created			I		UNSIGNED NOTNULL DEFAULT '1',
 	date_used				I		UNSIGNED NOTNULL DEFAULT '1'"
 	) );
--- a/api_tokens_page.php
+++ b/api_tokens_page.php
@@ -87,7 +87,7 @@ if ( count( $t_tokens ) > 0 ) {
 
 			$u_date_created  = date( $t_date_format, $u_date_created );
 
-			if ( (int)$u_date_used === 0 ) {
+			if( api_token_is_used( $t_token ) ) {
 				$u_date_used = lang_get( 'api_token_never_used' );
 			} else {
 				$u_date_used = date( $t_date_format, $u_date_used );
--- a/core/api_token_api.php
+++ b/core/api_token_api.php
@@ -153,6 +153,16 @@ function api_token_get_all( $p_user_id ) {
 }
 
 /**
+ * Determines whether the specified token has ever been used.
+ * @param array $p_token token to check
+ * @return bool True if used
+ * @access public
+ */
+function api_token_is_used( array $p_token ) {
+	return (int)$p_token['date_used'] <= 1;
+}
+
+/**
  * Updates the last used timestamp for the api token.
  *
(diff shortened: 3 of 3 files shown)


admin/schema.php
```diff
@@ -770,4 +770,12 @@ function installer_db_now() {
 
 # Release marker: 1.3.0-rc.1
 
+$g_upgrade[202] = array( 'AlterColumnSQL', array( db_get_table( 'api_token' ), "
+	user_id					I		UNSIGNED NOTNULL DEFAULT '0',
+	name					C(128)	NOTNULL DEFAULT \" '' \",
+	hash					C(128)	NOTNULL DEFAULT \" '' \",
+	date_created			I		UNSIGNED NOTNULL DEFAULT '1',
```
vboctor: Do we have a reason for defaulting dates to 1 instead of 0 other than consistency?  This seems like a strange default value.  You will need two things to make this change correct:
- API Token has a check for field not set for the case where tokens aren't used yet.  Make sure this check considers <= 1 to be never used.
- You may want to have upgrade steps to normalize already existing instances (if we care enough to do so). i.e. replace values where date == 0 with date = 1.

admin/schema.php
```diff
@@ -770,4 +770,12 @@ function installer_db_now() {
 
 # Release marker: 1.3.0-rc.1
 
+$g_upgrade[202] = array( 'AlterColumnSQL', array( db_get_table( 'api_token' ), "
+	user_id					I		UNSIGNED NOTNULL DEFAULT '0',
+	name					C(128)	NOTNULL DEFAULT \" '' \",
```
vboctor: I'm not sure why you are adding a default.  It is by design that there is no default since this field was part of the initial definition of the table and we don't want any entries to be added without setting this field.  Hence, we shouldn't have a default value.

admin/schema.php
dregad: I believe the 1 is historical, goes back to before the dates fields were converted to int (unix timestamps). I do agree it's kind of strange, but IMO maintaining consistency is reason enough.

I'll verify the API for the unused tokens check.

I'm not sure normalization is really needed if the above-mentioned check goes for <= 1 but I'll verify that too.

syncguru:
I am not sure about using default for this. I'd rather make the column nullable instead of defaulting to some value that does not have specific semantic meaning. So Null means it was not set.

For use_id column - it should not be nullable but there should not be default value either.

vboctor:
@dregad How about simplifying this change to just making user id unassigned and leave the rest as is.  I'm not convinced that we should do the other changes.  If so, please submit an update PR or let me know and I can do it.

dregad:
For the record, I tried without success to find out why datetime fields are defaulted to 1.

Prior to 1.2.0rc1, date fields were created with ADOdb *T* metatype, resulting in them having a *datetime* type in the underlying RDBMS. In 1.2.0rc1, starting with schema step 87, these fields were replaced by *unsigned integers* (storing unix timestamps). Having date fields defaulted to 1 pre-dates this change though. 

Historically (at least as far back as 0.15.0, as far as I can tell), zero dates in the tables were set to `1970-01-01 00:00:01` by UPDATE statements in the upgrade script. Starting with 0.16.0 (b827897bf22a8ab357cf670c61462c8e85ca2bd8) the date fields are defaulted to `1970-01-01 00:00:01` in the table definition. 

I could not find any documentation for this. I am guessing it could be due to 0 (0000-00-00 00:00:00) being an invalid date, but I have no way to find out.

Anyway, as mentioned previously we should try to be consistent. In answer to @vboctor's points in https://github.com/mantisbt/mantisbt/pull/700#discussion_r48688793:

> - API Token has a check for field not set for the case where tokens aren't used yet. Make sure this check considers <= 1 to be never used.

- **date_created** is always set to `db_now()` when the token is created, so there should be no issue with that field
- **date_used**: api_token API itself does not perform any check against this field; it is only used in api_tokens_page.php

> - You may want to have upgrade steps to normalize already existing instances (if we care enough to do so). i.e. replace values where date == 0 with date = 1.

I don't think that's necessary. Updating the check on date_used to `<= 1` instead of `=== 0` will cover any legacy data.

And to be clean, I'll move the logic from api_tokens_page.php to a new API function *api_token_is_used()* 

>> `+	name					C(128)	NOTNULL DEFAULT \" '' \",`

> I'm not sure why you are adding a default. It is by design that there is no default since this field was part of the initial definition of the table and we don't want any entries to be added without setting this field. Hence, we shouldn't have a default value.

I understand why you'd like the table structure match the logic. However, this is not how things are done everywhere else in the MantisBT schema. For example, take the *project* table: we have the exact same situation with the project name, yet the field is also defined as `name	C(128)	NOTNULL DEFAULT \" '' \"`. Data integrity is handled in the code, not at DB level.

So again, for consistency, we should follow the same rules for the api_token table.

dregad:
I rebased the branch and updated the PR, it is now ready for merge. I will do it tomorrow.

vboctor:
@dregad Looks good.  But please revert the default of the name and hash.  There are things that were not done right historically, that doesn't mean we should keep doing them going forward.  Consistency here has no value since it doesn't even impact the way the code is written against the database.

Otherwise looks good.

dregad:
For the record that I think your approach is wrong - if what was done in the past is incorrect, then we should have a global approach to fix things everywhere instead of doing it incrementally like that.

Along the same lines, we could also argue that setting a default value for *date_used* is unnecessary and we could have the field defined as `date_used I UNSIGNED NULL`; likewise for *date_created*, NOT NULL is somewhat useless if we set a default value, and furthermore the default should really be the current date/time and not 0 or 1. I could go on...

Anyway, all that to say that I still disagree, but don't feel strongly enough about it to continue arguing ;) I'll revert per your demand.

PR #701, opened 2016-01-05
Fix gpc_get_xxx_array for associative arrays

cproensa:
Functions:
gpc_get_int_array(),
gpc_get_bool_array()
dont work properly for html fields named with
associative indexes (numeric out or order, or
strings), becasue it assumes indexes as zero
based sequential numbers.
The code is here changed to keep the original
keys as defined in the field names.

Fixes #20476


dregad:
Fixed in b53f7caace3513d544149a18beb8f1e422e9a96e

PR #702, opened 2016-01-05
Plugin filter fields

cproensa:
While developing plugin that uses the filter fileds event, found these two issues

https://www.mantisbt.org/bugs/view.php?id=20480
Plugin filter fields throw warnings while rendering filter section

https://www.mantisbt.org/bugs/view.php?id=20481
Plugin filter fields dont use dynamic filter expansion


dregad:
Merged in d6ea7fb4085224747068aa3048137bd33ed98072

PR #703, opened 2016-01-06
Fix gpc_get regression by b53f7ca

cproensa:
This fixes a regression introduced by commit
b53f7ca
When loop is changed into 'foreach', null values
produce error if not checked.

This happens when default value passed as parameter is explicitly null.
Sorry i didnt catch it in original commit


dregad:
> Sorry i didnt catch it in original commit

Don't worry, it happens.

Fixed in 2ed3c644e50c8e3959f06c0a8f2b49d874438c79.

PR #704, opened 2016-01-07
0020483 pgsql manage user page

cproensa:
Fix pgsql query error manage_user_page

Due to pgsql having positional parameters in
parametrized queries, the page is throwing  an
error in some combination of conditions.
The error is caused by the reutilization of a
prepared WHERE condition in a string from
previous query. In the new query, a dp_param()
call is issued, which has the parameter reseted
to 1 (returning "$1").
Then, the resulting query string contains two
"$1" parameters for different values.

With this commit, the query build code is cleaned
to avoid said situation.

Fixes #20483

### DIFF
--- a/manage_user_page.php
+++ b/manage_user_page.php
@@ -203,22 +203,17 @@ $t_total_user_count = 0;
 # Get the user data in $c_sort order
 $t_result = '';
 
-if( 1 == $c_show_disabled ) {
-	$t_show_disabled_cond = '';
-} else {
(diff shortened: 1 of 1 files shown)


vboctor:
@cproensa looks good, haven't tested.  @dregad is the one that typically has pgsql and other dbs + the db expert, so I would wait for his feedback on this.

dregad:
Fixed in 9d8d010d92d13b3c19365a6ead6e5f282b4faddc, thanks @cproensa for your contribution.

PR #705, opened 2016-01-08
Consistently push query params in all APIs

dregad:
Prior to building a database query, our APIs must ensure that they leave
the parameters numbering in the same state as it was prior to the
function call.

While failure to do so does not have any impact with MySQL, it
causes SQL errors with RDBMS which rely on strict parameter ordering
(PostgreSQL and Oracle) when a query is executed while another is being
built, because the parameter count is reset (see #20479, #20483 for
recent examples).

All MantisBT APIs have been reviewed, and db_push_param() calls added
where needed to ensure that the caller always finds the environment in
the same state as it was prior the API call.

A new db_pop_param() API was added to cover rare cases where a query is
called multiple times, or built but never executed.

Fixes [#20479](https://www.mantisbt.org/bugs/view.php?id=20479)

### DIFF
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -387,15 +387,8 @@ function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset
 							'}' => $s_suffix,
 							) );
 
-	# We assume that '%' tokens won't mix with db_param() generated tokens
-	# Param push is needed in case this query is happening inside an outer
-	# query which is using db_param (eg as part of an api call)
+	# Pushing params to safeguard the ADOdb parameter count (required for pgsql)
 	$g_db_param->push();
-	$p_query = preg_replace_callback(
-							array( "/%s/", "/%d/", "/%b/", "/%l/"),
-							function() use (&$p_pop_param) { $p_pop_param = false; return db_param(); },
-							$p_query
-							);
 
 	if( db_is_oracle() ) {
 		$p_query = db_oracle_adapt_query_syntax( $p_query, $p_arr_parms );
@@ -407,9 +400,7 @@ function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset
 		$t_result = $g_db->Execute( $p_query, $p_arr_parms );
 	}
 
-	# undo the previous param push.
-	# This also restores parameter count in ADOdb, in case this is pgsql,
-	# so that an api sql call dont break token numbering
+	# Restore ADOdb parameter count
 	$g_db_param->pop();
 
 	$t_elapsed = number_format( microtime( true ) - $t_start, 4 );
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -448,7 +448,8 @@ function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset
 		array_push( $g_queries_array, array( '', $t_elapsed ) );
 	}
 
-	if( $p_pop_param ) {
+	# Restore param stack: only pop if asked to AND the query has params
+	if( $p_pop_param && !empty( $p_arr_parms ) ) {
 		$g_db_param->pop();
 	}
 
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -324,7 +324,9 @@ function db_query_bound() {
 /**
  * execute query, requires connection to be opened
  * An error will be triggered if there is a problem executing the query.
- * This will pop the database parameter stack {@see MantisDbParam} after a successful execution
+ * This will pop the database parameter stack {@see MantisDbParam} after a 
+ * successful execution, unless specified otherwise
+ * 
  * @global array of previous executed queries for profiling
  * @global adodb database connection object
  * @global boolean indicating whether queries array is populated
@@ -332,9 +334,10 @@ function db_query_bound() {
  * @param array   $p_arr_parms Array of parameters matching $p_query.
  * @param integer $p_limit     Number of results to return.
  * @param integer $p_offset    Offset query results for paging.
+ * @param boolean $p_pop_param Set to false to leave the parameters on the stack
  * @return IteratorAggregate|boolean adodb result set or false if the query failed.
  */
-function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset = -1 ) {
+function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset = -1, $p_pop_param = true ) {
 	global $g_queries_array, $g_db, $g_db_log_queries, $g_db_param;
 
 	$t_db_type = config_get_global( 'db_type' );
@@ -450,7 +453,9 @@ function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset
 		trigger_error( ERROR_DB_QUERY_FAILED, ERROR );
 		return false;
 	} else {
-		$g_db_param->pop();
+		if( $p_pop_param ) {
+			$g_db_param->pop();
+		}
 		return $t_result;
 	}
 }
@@ -473,6 +478,19 @@ function db_param() {
 function db_param_push() {
 	global $g_db_param;
 	$g_db_param->push();
+}
+
+/**
+ * Pops the previous parameter count from the stack
+ * It is generally not necessary to call this, because the param count is popped
+ * automatically whenever a query is executed via db_query(). There are some
+ * corner cases when doing it manually makes sense, e.g. when a query is built
+ * but not executed.
+ * @return void
+ */
+function db_param_pop() {
+	global $g_db_param;
+	$g_db_param->pop();
 }
 
 /**
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -385,11 +385,12 @@ function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset
 	$p_query = strtr($p_query, array(
 							'{' => $s_prefix,
 							'}' => $s_suffix,
-							'%s' => db_param(),
-							'%d' => db_param(),
-							'%b' => db_param(),
-							'%l' => db_param(),
 							) );
+	$p_query = preg_replace_callback(
+							array( "/%s/", "/%d/", "/%b/", "/%l/"),
+							"db_param",
+							$p_query
+							);
 
 	if( db_is_oracle() ) {
 		$p_query = db_oracle_adapt_query_syntax( $p_query, $p_arr_parms );
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -386,9 +386,14 @@ function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset
 							'{' => $s_prefix,
 							'}' => $s_suffix,
 							) );
+
+	# We assume that '%' tokens won't mix with db_param() generated tokens
+	# Param push is needed in case this query is happening inside an outer
+	# query which is using db_param (eg as part of an api call)
+	$g_db_param->push();
 	$p_query = preg_replace_callback(
 							array( "/%s/", "/%d/", "/%b/", "/%l/"),
-							"db_param",
+							function() use (&$p_pop_param) { $p_pop_param = false; return db_param(); },
 							$p_query
 							);
 
@@ -401,6 +406,11 @@ function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset
 	} else {
 		$t_result = $g_db->Execute( $p_query, $p_arr_parms );
 	}
+
+	# undo the previous param push.
+	# This also restores parameter count in ADOdb, in case this is pgsql,
+	# so that an api sql call dont break token numbering
+	$g_db_param->pop();
 
 	$t_elapsed = number_format( microtime( true ) - $t_start, 4 );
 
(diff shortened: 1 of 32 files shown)


cproensa:
ouch
> does have any impact not impact

To be precise 20483 would not have been solved by this:
#20483 actually is not caused by this issue (core not pushing params), however its related to the resetting of param counter after execution.

core/database_api.php
```diff
@@ -332,9 +334,10 @@ function db_query_bound() {
  * @param array   $p_arr_parms Array of parameters matching $p_query.
  * @param integer $p_limit     Number of results to return.
  * @param integer $p_offset    Offset query results for paging.
+ * @param boolean $p_pop_param Set to false to leave the parameters on the stack
  * @return IteratorAggregate|boolean adodb result set or false if the query failed.
  */
-function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset = -1 ) {
+function db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset = -1, $p_pop_param = true ) {
```
cproensa: .

cproensa:
> A new db_pop_param() API was added to cover rare cases where a query is
called multiple times, **or built but never executed.**

Wouldtn that use case, be more useful a db_reset_param(), that reset param count?

For the raw pop function, its supported by ADO library, so it may be good idea to expose it.

vboctor:
- How will that work with plugins that don't push?
- Do we really need one more thing to worry about and potentially forget every time we need to run a query?  Aren't there other options?

cproensa:
> Do we really need one more thing to worry about and potentially forget every time we need to run a query? Aren't there other options?

These days i've wondered the same question. To which i think:
- Premise: its mandatory to reset parameter count at some point, if we want to avoid "nested query collision". This is mandated by the db engine way of doing things! As Damien says, this is an actual problem, but only with pgsql, which is the minority,
- Option 1: this method of explicitly push parameters, in case a caller up the stack is in the middle of a sql build.
- Option 2: Impose a start_param() style call mandatory to whoever that start to build a parametrized sql.
So whats a bigger annoyance? 
With (1) only need to account for this defenshive push when developing a function that may be called from a unkown scenario. Thats what API does.

Plugins dont need to push(\*), base pages of mantis ui, don't either, becasue there the code is known to not be part of a inner cycle.
(\*) However, if plugins has some kind of api, they should push for whoever is using them.

dregad:
> ouch
>> does have any impact not impact

Oops. Will edit commit message

> To be precise 20483 would not have been solved by this

You're right. I'll remove the reference.

> Wouldtn that use case, be more useful a db_reset_param(), that reset param count?

I don't think so, because in this case the param count has been pushed, so the stack needs to be reset (see for example in tag api or custom_field api).

> How will that work with plugins that don't push?

As far as I can tell, plugins won't be affected, just like our base page and update scripts are not. If a plugin's internal API needs to build queries in parallel, it can call db_param_push().

> Do we really need one more thing to worry about and potentially forget every time we need to run a query? 

I'm afraid we do. But keep in mind it's not everytime we do a query, it's only for those functions that could potentially be called from somewhere else (i.e. API calls).

> Aren't there other options?

I couldn't think of another way to fix this, but I'm open to suggestions if have any.

cproensa:
@dregad  sorry i preceded to your response, it's just i was around writting too

dregad:
NP :wink:

cproensa:
>>   Wouldtn that use case, be more useful a db_reset_param(), that reset param count?

> I don't think so, because in this case the param count has been pushed, so the stack needs to be reset (see for example in tag api or custom_field api).

Right
I misunderstood as kind of "decrease param count by one"

core/database_api.php
dregad: Not sure what you mean by `.` - is this in relation with [your comment below](https://github.com/mantisbt/mantisbt/pull/705#issuecomment-170029549) ? Or is there a problem with the parameter ?

cproensa: not related to the comment, 
actually i cant seem to make up a case for using
$p_pop_param = false

it may be justified, but i cant make up the scenario.

On a side note, shouldn't we also pop() when the query fails? (for the sake of correctness)

vboctor:
Here is an idea.  What if we flip this the other way around?  If you are preparing a query, you must not make an API call?  This will require the code to look as follows:

collect your variables using API calls.
use variables to build the query (can't make API calls here).

In the case, where this is not followed, we will get a database error with pgsql and oracle and will fix it to follow the pattern I just described.

Would that solve the issue?

cproensa:
> collect your variables using API calls.
use variables to build the query (can't make API calls here).

for example: filter_api

and leaving aside the monstruosity that is "filter_get_bug_rows"...
one of the features of the filter query, is that it allows plugins to insert its own sql clauses to implement column ordering and filter fields. This is the foundation of MantisColumn and MantisFilter classes.
This implies doing api calls (a lot) to trigger events. Also, those plugin implementations will need to use api calls for sure (access, config_get, etc)

dregad:
Your suggestion sounds like a lot more effort to implement (and maintain). Look at how many functions I modified in this PR - while technically not all every single one actually *requires* to push params (I modified all of them to be on the safe side), a good number of them would require refactoring to comply with your approach.

Furthermore, as @cproensa pointed out, there are existing situations where it would simply not work, and I would not even wish in my worst nightmares to rewrite the filter API :fearful: 

My solution is not perfect, and it adds a tiny bit of overhead, but it's safe and simple. There's only one rule which can be summarized like this: **if developing an API function, call `db_param_push()` before any use of `db_param()`.**

core/database_api.php
cproensa: OK, not sure if its the same reasons you had in mind:
In the situation of a nested query call, the inner query always pops the params.
So, even if the inner query dont use db_param in any way, we should also call db_param_push to keep it safe.

This doesn't sound quite right.
In this PR you are using db_param_push() in every api function that uses db_param().
But what about other functions with queriess that dont use db_params()? those are a failure point too... unless you modify the db_query() call to include this parameter to false.

I suggest to check inside db_query if the query has params. If not, then do not pop().

cproensa: > On a side note, shouldn't we also pop() when the query fails? (for the sake of correctness)

Made some debugging, and i'm positive it should also pop() on query fail, although the problem is quite subtle here.

First, note that an sql error is not a script halt. We can explicitly supress errors in api calls, e.g "@config_get(...)", or manage error recovery, exception, etc.

So, when the query fails:
1) ADOdb internal param count is resetted to 1 (pgsql)
2) $g_db_param param count is not cleared, so it will keep counting on next db_param()

**This happens in disconnected code**, not just in the nested query situation, as $g_db_param param is persistent and global

How can this be a failure point? i'm thinking Oracle, which uses named parameters, which mantis api names after its param count.
Probably what wil happen, on next db_param calls:
mysql: ?, ?, ? .. (ok)
pgsql: $1, $2, $3 ... (the number comes from ADOdb)
oracle: :5, :6, :7  (fail!)

cproensa:
>Do we really need one more thing to worry about and potentially forget every time we need to run a query? Aren't there other options?

> I couldn't think of another way to fix this, but I'm open to suggestions if have any.

This is all about a concurrency problem. The db-params object is persistent and shared through all session.
Probably a better design would be use a dbParams object that is instantiated at each usage, instead of passing a plain array around
$params = new dbParams();
$sql = "..." . $params->add($value) . "..";
db_query( $sql, $params);

However this is a major refactor.

cproensa:
@dregad 
bug_api.php is not included

tgulacsi:
I think it would be waaaay easier and less code, to just use "?" or any other placeholder (db_param() could return it, but get rid of that global state) for the params in the query building phase, and replace them to the RDBMS-specific placeholder right before execution.

We already do this for Oracle (replace $n with :n), and that renumbering is easy enough.

dregad:
The more I think about it, the more I believe the way we handle DB params is broken by design...

Consequently, continuing to update this PR feels a bit like applying a bandage on a wooden leg (that's a French idiom which I translated litterally, but I'm sure you get the idea).

That being said, it *does* get the job done in terms of fixing the pgsql issue, and leave the code in a better state than it was before, so it's probably worth merging anyway but we seriously need to revisit the way we build and execute queries.

dregad:
@cproensa the commits I just pushed should address the points you mentioned [1](https://github.com/mantisbt/mantisbt/pull/705#issuecomment-170156222), [2](https://github.com/mantisbt/mantisbt/pull/705#discussion_r49217695), [3](https://github.com/mantisbt/mantisbt/pull/705#discussion_r49221420)

@tgulacsi it might indeed be easier, but I'm concerned about the potential for regressions doing this could introduce. 

If the problem is fixed like that, then let's leave it at that and focus on a proper, long-term solution (probably using named parameters for all queries).

cproensa:
> @cproensa the commits I just pushed should address the points you mentioned 1, 2, 3

Looks good. I will try to test those corner cases again, aside from not introducing regressions.

> If the original problem is fixed, then let's leave it at that and focus on a proper, long-term solution 

Agree.

> (probably using named parameters for all queries).

Which is related to other api inconsistency we talked about (not in this PR)
Your idea goes the way of not exposing the db specific syntax for parameters until query time. +1 on that if time comes to do it.

cproensa:
Sorry to call again, but still I'd like to understand the purpose of the $p_pop_param parameter in db_query()
I'm seeing something that can be wrong there, but first i'd like to know better the reasoning behind that.

cproensa:
Ok, bad news
Now we have this trouble code in db_query():

	$p_query = strtr($p_query, array(
							'{' => $s_prefix,
							'}' => $s_suffix,
							'%s' => db_param(),
							'%d' => db_param(),
							'%b' => db_param(),
							'%l' => db_param(),
							) );

Those db_param() are always evaluated, which has two side effects:
1) Substitutions of % parameters always use the same numbered token 
2) The param count is increased by 4 on every call to db_query

The consequence of (2), previously was masked by always popping the paramater stack. With latest fix, when the params are not popped (by means of 2f35d3d, or the $p_pop_param), next query using db_param() will receive a shifted numbered token.
Eg: expected first param, but receive $5.

atrol:
What about the approach that grangeway had in his PR's, e.g.
https://github.com/mantisbt/mantisbt/pull/490/files
```
-	$t_query = 'SELECT * FROM {user} WHERE username=' . db_param();
+	$t_query = 'SELECT * FROM {user} WHERE username=%s';
 	$t_result = db_query( $t_query, array( $p_username ) );
```

tgulacsi:
That has its own problem: "LIKE with 'aa%space'".
That's why I suggest "?" or any other unlikely string (#?# for example) as a placeholder.

cproensa:
> That has its own problem: "LIKE with 'aa%space'".
That's why I suggest "?" or any other unlikely string (#?# for example) as a placeholder.

That issue is valid for every token, as every string MAY appear as literal in a query.
current mysql token "?" can appear as a character in a string expression, and IIRC, ADOdb would fail on it. The compromise is to inject the string as parameter:
LIKE %
and use parameter = (string) "aa%space"

DB engine is lexically parsing the sql and can distinguish proper tokens from literal characters, but that would be overkill on our db layer, even for ADOdb, i think...

dregad:
@cproensa as I was reviewing the [PR you submitted in my private repository](https://github.com/dregad/mantisbt/pull/8) against the source branch of this PR, and I started wondering why this parameter replacement logic was there in db_query() in the first place.

It was added in ad4b7bf0f31f7983e7cdb12b717c856bf2993897 - as part of the new DB API that grangeway was planning to implement .

Maybe we should simply revert that commit and be done with it, since 

1. the syntax it deals with (i.e. `SELECT * FROM {table} WHERE column = %s`) does not appear to be used anywhere in the code, 
1. it does not add any immediate value,
1. it introduces potential problems of its own, as you pointed out

cproensa:
> Maybe we should simply revert that commit and be done with it, since 

I thought it was some code to support legacy syntax. Being not that case, you may be right with your suggestion.

However, The second point i got in that PR:
In whatever situation, that our parameters are popped or not, the query execution resets the ADOdb parameter count, which is what is actually used for pgsql tokens. 

In my PR a push/pop is used to restore ADOdb counter after the query, also fixing the % stuff. If the later is not needed, still it is for pgsql (at least the part that restores ADOdb counter)

dregad:
@cproensa I merged dregad/mantisbt#8 and then removed the "new query" as per our discussion.

I've done some local testing and it seems to work as expected, but if you can spare the time, I think some additional testing would be useful...

Thanks in advance !

cproensa:
OK @dregad 
I will check again with the test cases i have

cproensa:
> I will check again with the test cases i have

Checked your latest commits with pgsql cases. they're ok now.

pkolaric:
Are there still any issues that preven this PR from being merged to master?

dregad:
> Are there still any issues that preven this PR from being merged to master?

Other than me finding the time to resolve the merge conflicts, you mean ? No, I don't think so.

dregad:
I rebased the branch. From my perspective, this should now be ready to merge; let me know if there are any objections.

dregad:
Note: I documented the use of `db_push_param()` in the wiki
https://www.mantisbt.org/wiki/doku.php/mantisbt:executing_db_queries

PR #707, opened 2016-01-19
Update readme.md

spidgorny:
Fix admin/check/ link

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -18,7 +18,7 @@ Installation
 ------------
 
  * Extract the tarball into a location readable by your web server
- * Point your browser to http://path/to/mantisbt/admin/check.php to ensure that
+ * Point your browser to http://path/to/mantisbt/admin/check/ to ensure that
    your webserver is compatible with MantisBT and configured correctly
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -18,7 +18,7 @@ Installation
 ------------
 
  * Extract the tarball into a location readable by your web server
- * Point your browser to http://path/to/mantisbt/admin/check.php to ensure that
+ * Point your browser to http://path/to/mantisbt/admin/check/ to ensure that
```
atrol: Should be
```
* Point your browser to http://path/to/mantisbt/admin/check/index.php
```
as we should not depend on web server configuration.

vboctor:
thanks @spidgorny :+1:

dregad:
I'll merge this, thanks @spidgorny 

As a side note, I'm wondering if we would not be better off removing all installation instructions from the Readme file, and refer people to the Admin Guide instead, to avoid duplicating information...

dregad:
Fixed in f2a8172cfebaa5a3ad6f2116d1cb16c1e6dcbab5

PR #708, opened 2016-01-20
RFC 0008204 temporary filter set

cproensa:
This is an incomplete work in progress
There are some issues that need discussion as they require changes in some core functionality.

Line comments will hightlight key points

The point of this PR is fixing temporary filters that get lost after changing sort order for columns
https://www.mantisbt.org/bugs/view.php?id=8204

### DIFF
--- a/core/current_user_api.php
+++ b/core/current_user_api.php
@@ -247,6 +247,7 @@ function current_user_get_bug_filter( $p_project_id = null ) {
 				$t_filter = json_decode( $t_token, true );
 			}
 		} else {
+			// TODO @cproensa remove this, remainders of ancient #4609
 			$t_filter = json_decode( $f_filter_string, true );
 		}
 		$t_filter = filter_ensure_valid_filter( $t_filter );
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -1271,6 +1271,8 @@ function print_formatted_severity_string( BugData $p_bug ) {
  * @return void
  */
 function print_view_bug_sort_link( $p_string, $p_sort_field, $p_sort, $p_dir, $p_columns_target = COLUMNS_TARGET_VIEW_PAGE ) {
+	// TODO @cproensa url parameters are carried from custom_function_default_print_column_title()
+	// needing new parameters means also changing parameter definition on all print_column_title_xxxx functions!
 	if( $p_columns_target == COLUMNS_TARGET_PRINT_PAGE ) {
 		if( $p_sort_field == $p_sort ) {
 			# We toggle between ASC and DESC if the user clicks the same sort order
--- a/view_all_set.php
+++ b/view_all_set.php
@@ -60,6 +60,7 @@ auth_ensure_user_authenticated();
 $f_type					= gpc_get_int( 'type', -1 );
 $f_source_query_id		= gpc_get_int( 'source_query_id', -1 );
 $f_print				= gpc_get_bool( 'print' );
+$f_filter_token			= gpc_get( 'filter', null );
 $f_temp_filter			= gpc_get_bool( 'temporary' );
 
 # validate filter type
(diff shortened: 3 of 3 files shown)


cproensa:
General description of how to fix:
When using a temporary filter, pagination works, becasue is using the parameter 'filter' in the liks, example:
view_all_bug_page.php?filter=233070&page_number=4
this parameter is used by current_user_get_bug_filter() that returns the temporary filter (which was previously saved in tokens table).
To get columns sorting to work, a similar thing has to be done, the column title link should carry the same parameter to signal that is the temporary filter which has to be used (and modified)

The are some issues
1) In order to include parameter in the column sort link, custom function and all column_api individual functions has to be modified to be aware of the temporary nature of the filter.
2) The parameter 'filter' is an id for the saved filter in tokens table, but this id is never used, becasue tokens table is accesed by user and type, not id (current_user_get_bug_filter)
3) current_user_get_bug_filter() accepts a json encoded filter object as parameter. This can be removed, as it seems to be legacy behaviour from version <1.0. (It actually works in some reports like csv export, but breaks in view_all_bug_page)
4) saving temporary filters as an unique token per user, means that multiple views (eg: browser tabs) interact with each other. Ideally, a temporary filter should be tracked to be used only on current views. This would work by allowinf multiple token per user, and actually using the id token in 'filter' parameter

core/print_api.php
```diff
@@ -1271,6 +1271,8 @@ function print_formatted_severity_string( BugData $p_bug ) {
  * @return void
  */
 function print_view_bug_sort_link( $p_string, $p_sort_field, $p_sort, $p_dir, $p_columns_target = COLUMNS_TARGET_VIEW_PAGE ) {
+	// TODO @cproensa url parameters are carried from custom_function_default_print_column_title()
+	// needing new parameters means also changing parameter definition on all print_column_title_xxxx functions!
```
cproensa: .

core/current_user_api.php
```diff
@@ -247,6 +247,7 @@ function current_user_get_bug_filter( $p_project_id = null ) {
 				$t_filter = json_decode( $t_token, true );
 			}
 		} else {
+			// TODO @cproensa remove this, remainders of ancient #4609
```
cproensa: Here is some code that probably should be cleaned.
Do not use raw json filter strings as url parameters.
Refactor the tokens get/set to allow multiple filters per user, (and get-by-id)

view_all_set.php
```diff
@@ -359,6 +360,16 @@
...
+# TODO @cproensa
+# It is assumed that 'temporary' parameter means setting a new filter, so here type is set to 1
+# however, the temporary nature has to be carried through the pagination links, and the column sorting
+# Currently this is achieved by passing the filter token id (stored temporary filter)
+# BUT current_user_get_bug_filter() doesnt use that number at all, only if its present,
+# and it is a number, it gets the filter from token table, regardless of its id
+# Proposal is to change semantics:
+# * 'temporary' means it is stored/read from tokens table (no id, just current user)
+#   and carry this parameter through pagination and sorting
+# * caller, eg, my_view, should set type=1 on first call, to force creation of new temporary filter.
```
cproensa: For this comment i didnt account for actually using the token ids to allow multiple temporary filters per user, by actually using the passed id number

view_all_set.php
```diff
@@ -414,26 +425,32 @@
...
+				exit; # stop here
+			}
 		}
+	} else {
+		# no cookie found, set it
+		$f_type = 1;
 	}
-} else {
-	# no cookie found, set it
-	$f_type = 1;
```
cproensa: Type 1 is to create a new filter
'temporary' parameter is used so as to save in tokens instead of DB.

What is changed in this page basically is:
Do we have 'temporary' parameter? create new filter, and store it in tokens, not in DB
Do we have 'filter' parameter? retrieve filter from tokens, modify, store in tokens again.

cproensa:
opened #717 to replace this

PR #709, opened 2016-01-22
Fix passing parameters for EVENT_TYPE_OUTPUT events

cproensa:
Fix passing parameters for EVENT_TYPE_OUTPUT events

Some events of EVENT_TYPE_OUTPUT specify a parameter
to be passed to the callback function.
However the event dispatcher for that type was not
designed to receive a parameter in that way.

Modified the event function to allow for that parameter
to be passed on, thus matching documented specification
and current use in code base.

Affected events are:
EVENT_MANAGE_OVERVIEW_INFO
EVENT_VIEW_BUG_ATTACHMENT

Those events have not been working properly until now.

Fixes #20535

### DIFF
--- a/core/event_api.php
+++ b/core/event_api.php
@@ -230,26 +230,27 @@ function event_type_execute( $p_event, array $p_callbacks, $p_params = null ) {
  * If there are no callbacks, then nothing will be sent as output.
  * @param string $p_event     Event name.
  * @param array  $p_callbacks Array of callback function/plugin base name key/value pairs.
- * @param mixed  $p_params    Output separator (if single string) or indexed array of pre, mid, and post strings.
+ * @param mixed  $p_params    Parameters to event callback (array, or single object)
+ * @param mixed  $p_format    Output separator (if single string) or indexed array of pre, mid, and post strings.
  * @access public
  * @return void
  */
-function event_type_output( $p_event, array $p_callbacks, $p_params = null ) {
+function event_type_output( $p_event, array $p_callbacks, $p_params = null, $p_format = null ) {
 	$t_prefix = '';
 	$t_separator = '';
 	$t_postfix = '';
 
-	if( is_array( $p_params ) ) {
-		switch( count( $p_params ) ) {
+	if( is_array( $p_format ) ) {
+		switch( count( $p_format ) ) {
 			case 3:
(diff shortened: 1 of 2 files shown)


vboctor:
I'm confused by this change:

- You added a parameter to the event_type_output() but not to the single place where it is called.
- Why are we making the assumption that the parameter  needed to be added has to do with formatting and providing a separator?
- Since these events were always broken, should we send an object as the event parameter? The object class will dependent on the event type?

cproensa:
There are several things wrong
1) The dispatcher for EVENT_TYPE_OUTPUT was already using the formatting parameter which:
- was never used though mantis base
- is passed to the plugin callback (shouldn't, this only affects the dispatcher)

2) The two events EVENT_MANAGE_OVERVIEW_INFO,EVENT_VIEW_BUG_ATTACHMENT, are triggered by passing a parameter, which is not the formatting array.
- It is expected that parameter to reach the plugin callback
- the event dispatcher does not propagate caller parameters into the plugin callback

> You added a parameter to the event_type_output() but not to the single place where it is called.

added a propagating parameter array (which would reach the callback, like other events do), and moved the fromatting parameter to a second position, as it is optional, like, more optional and less used than the callback-parameters.
Event callers are not modified because they already (misused) the parameter in that order. Implementation now match that intention.

> Why are we making the assumption that the parameter needed to be added has to do with formatting and providing a separator?

That was already there.

> Since these events were always broken, should we send an object as the event parameter?

Not sure. EVENT_VIEW_BUG_ATTACHMENT already sends a BugData object. EVENT_MANAGE_OVERVIEW only sends a bool. 
Returning objects is not a need, since these only print to output.


Other variations can be: define the events parameter in different order, or even change the event type for those two events, so no change is made in type_output...

However, i can see some potential on using parameter on current output events like those for resources, html_body, etc... that can be discussed on another feature request.

dregad:
@vboctor are you satisfied by @cproensa's answer ? If yes, I'm OK with merging this.

tgulacsi:
Sorry to reply late, but yes, this PR works for me, so please merge it, if possible!

vboctor:
Merged via da98b63795216b5fc116730bf42a16df3f964955 and e541d060397a2da7d0c90829c89ccd147eced0cb.

PR #710, opened 2016-01-24
Add plugin events for user accounts lifecycle

vboctor:
- EVENT_MANAGE_USER_CREATE_FORM
- EVENT_MANAGE_USER_CREATE
- EVENT_MANAGE_USER_UPDATE_FORM
- EVENT_MANAGE_USER_UPDATE
- EVENT_MANAGE_USER_DELETE
- EVENT_MANAGE_USER_PAGE

Fixes #8779


vboctor:
This change was made consistent with the EVENT_MANAGE_PROJECT_* and EVENT_MANAGE_VERSION_*.  However, there are also existing events related to user accounts:

	# User account pages
	'EVENT_ACCOUNT_PREF_UPDATE_FORM' => EVENT_TYPE_EXECUTE,
	'EVENT_ACCOUNT_PREF_UPDATE' => EVENT_TYPE_EXECUTE,

I would be happy to change the prefix of the new events to EVENT_ACCOUNT_XXXX.  I'm also thinking of adding events for auth related events like login and logout), etc. So these can use EVENT_ACCOUNT_LOGIN, etc.  or EVENT_USER_LOGIN, or EVENT_AUTH_LOGIN, etc.

I personally like USER over ACCOUNT.  So I would use EVENT_MANAGE_USER_* for management, and EVENT_USER_* for stuff like login/logout/..etc.

So, let me know your thoughts about the naming convention.

There is also a question about whether we want to continue having events to enable plugins to pump html in a specific place in the page, vs. use jquery to insert such elements in the exact space that makes sense using the DOM.  For that, they can just hook to one of the end of body events for example.

cproensa:
Have you thought about a "processing" event, like, used before any data is saved, to be able modify, validate, or reject any change (with bug data there are these kind of events)

> There is also a question about whether we want to continue having events to enable plugins to pump html in a specific place in the page, vs. use jquery to insert such elements in the exact space that makes sense using the DOM. For that, they can just hook to one of the end of body events for example

I like how that sounds, but there is a need to review those page/scripts related events. Better to discuss it as a separate thread?

core/user_api.php
```diff
@@ -647,6 +649,8 @@ function user_delete_profiles( $p_user_id ) {
 function user_delete( $p_user_id ) {
 	$c_user_id = (int)$p_user_id;
 
+	event_signal( 'EVENT_MANAGE_USER_DELETE', array( $p_user_id ) );
```
cproensa: event is placed before user_ensure_unprotected()?
should be better placed after it?

vboctor: Fixed. Thanks.

vboctor:
@cproensa Let's see if there is demand for the processing events.  I haven't added them since they aren't there for other MANAGE events for projects and versions.

PR #711, opened 2016-01-25
Support avatar plugins

vboctor:
Add the necessary events to enable development of avatar providers
for scenarios like retrieving avatars from local file syste, LDAP,
social identities, etc.

As a proof of the events and a sample, this change also refactors
Gravatar integration into a core plugin.

Fixes #13788


vboctor:
Some thoughts to get your feedback on:

- There is a schema change to auto-add the Gravatar plugin, though we can skip that and have
the users install the plugin to enable plugins.  Unless we have show_avatar defaulted to ON,
installing the plugin makes sense and we can make this clearer in the show_avatar documentation.

- When avatars are disabled, we show the default avatar, rather than hiding the avatar completely.  However, if avatars are OFF, we won't be communicating with gravatar.  I'm planning to add support for url/blob where users or plugins can set the avatars.  This addresses the concern that was raised before where an admin may want to disable communicating with external services but yet get us towards modern times in terms of having first class support for avatars.

- Add CSP event, but it seems that we could use the RESOURCE event to have plugins emit their own CSP headers.  Let me know your thoughts, I'm thinking of removing this event if using RESOURCE events works.

dregad:
I actually started implementing the very same thing about a year ago [1], but didn't complete back then because I put the work aside as we agreed to focus on 1.3 release.

I'll want to review your work, compare it with what I did myself and see if there are things that we implemented differently that deserve to be merged.

[1] https://github.com/dregad/mantisbt/tree/avatar

vboctor:
@dregad I had a look at your implementation and did some tweaks based on it.  I've also removed CSP specific event as indicated earlier and switched to RESOURCES event.

vboctor:
@dregad would love to get your feedback on this.

dregad:
It's on my todo list, but unfortunately I have too many things to do and not enough time. This is a non-critical feature, so has lower overall priority. Please bear with me.

vboctor:
@dregad I'll check this in later today.  If you want to tweak and improve further, we can have follow up pull requests.

dregad:
Sorry this fell off the radar :-/

c-schmitz:
I hope this feature will be included RC2 as I am really looking foward to this (I considerGravatar as a privacy leak)

dregad:
@c-schmitz Yes it will be.

weberhofer:
Hi! Are there any news on this pr?

dregad:
> Hi! Are there any news on this pr?

@weberhofer It has been merged over a year ago...

c-schmitz:
That might be but I can't see in the latest stable version?

c-schmitz:
Ok, take that back. I was blind, sorry.

weberhofer:
Sorry, didn't see the "merged commit" line. Thanks!

PR #712, opened 2016-01-29
Support login by email address

vboctor:
In addition to login by username and password, this change adds
support for login by email address and password when the following
conditions are true:

- email address is not blank.
- email_login_enabled is ON.
- there is a single enabled email account w/ such email.

Fixes #20543

This change also includes enforcing of email address uniqueness
for all new user signups, creation by manager, update by manager,
and update by user.  A configuration option is available to turn this
OFF.

Note that there can still be duplicate emails for data that was already
in the system or was updated while this option is turned OFF.  But
for new installs, we should be clean.

Fixes #9093

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -547,7 +547,7 @@ $g_validate_email = ON;
  *
  * @global integer $g_email_login_enabled
  */
-$g_email_login_enabled = ON;
+$g_email_login_enabled = OFF;
 
 /**
  * Ensure that email addresses are unique.
--- a/docbook/Admin_Guide/en-US/config/email.xml
+++ b/docbook/Admin_Guide/en-US/config/email.xml
@@ -174,7 +174,7 @@
 					Enable support for logging in by email and password, in addition to
 					username and password.  This will only work as long as there is a single
 					user with the specified email address and the email address is not blank.
-					The default value is ON.
+					The default value is OFF.
 				</para>
 			</listitem>
 		</varlistentry>


cproensa:
Maybe this should be merged at the same time that there is implemented a verification system for newly created accounts?
In other case, if this option is enabled, a user could register with an existant email address, and would invalidate other user ability to login with email.

vboctor:
I was thinking about the same kind of issue but was thinking of a separate fix, but it makes sense to provide something in here as well.

In order to guard against a user adding another account with their email address or another user changing their address to someone else's email, I have created a configuration option to ensure uniqueness that is ON by default.  It applies to all new changes like signup, user creation, my account update, manage updating users, etc.

With both options - enable login via email and email addresses should be unique, I think we have a clean story going forward.  As for existing data, we don't allow login via email when there exist multiple accounts with the same email address, hence we are safe there.

config_defaults_inc.php
```diff
@@ -541,6 +541,20 @@
...
+ * user with the specified email address and the email address is not blank.
+ *
+ * @global integer $g_email_login_enabled
+ */
+$g_email_login_enabled = ON;
+
+/**
+ * Ensure that email addresses are unique.
+ */
+$g_email_ensure_unique = ON;
```
atrol: missing @global integer $g_email_ensure_unique

core/authentication_api.php
```diff
@@ -223,6 +223,32 @@ function auth_auto_create_user( $p_username, $p_password ) {
 }
 
 /**
+ * Given a login username provided by the user via the web UI or our API,
+ * get the user id.  The login username can be a username or an email address.
+ * The email address will work as long there is a single enabled account with
+ * such address and it is not blank.
+ *
+ * @param string $p_login_name The login name.
+ * @return user id or false.
```
atrol: missing int|boolean

core/user_api.php
```diff
@@ -254,6 +254,53 @@ function user_ensure_name_unique( $p_username ) {
...
+ */
+function user_is_email_unique( $p_email, $p_user_id = null ) {
+	if( is_blank( $p_email ) ) {
+		return true;
+	}
+
+	$p_email = trim( $p_email );
+
+	if ( $p_user_id === null ) {
+		$t_query = 'SELECT email FROM {user} WHERE email=' . db_param();
```
atrol: Create index for email on user table.

config_defaults_inc.php
```diff
@@ -541,6 +541,20 @@
...
 
 /**
+ * Enable support for logging in by email and password, in addition to
+ * username and password.  This will only work as long as there is a single
+ * user with the specified email address and the email address is not blank.
+ *
+ * @global integer $g_email_login_enabled
+ */
+$g_email_login_enabled = ON;
+
```
atrol: Would like to see default OFF as the new functionality introduces
- less security
- decreased performance, stresses server/database especially in SOAP scenario

vboctor: Users re-use same email addresses and logins across the internet.  Hence, I would argue it is not less secure.  As for the load, I've added a check that the provided email is a valid email address before looking up the DB for emails.

core/user_api.php
vboctor: Good catch.  I thought we had this already.  Adding now.

vboctor:
@atrol Pushed an update that addresses feedback.

login_page.php
```diff
@@ -241,7 +247,10 @@ function debug_setting_message ( $p_type, $p_setting, $p_value ) {
 			</ul>
 			<div class="field-container">
 				<label for="username"><span><?php echo lang_get( 'username' ) ?></span></label>
-				<span class="input"><input id="username" type="text" name="username" size="32" maxlength="<?php echo DB_FIELD_SIZE_USERNAME;?>" value="<?php echo string_attribute( $f_username ); ?>" class="<?php echo $t_username_field_autofocus ?>" /></span>
+				<span class="input">
+					<input id="username" type="text" name="username" size="32" maxlength="<?php
+						echo DB_FIELD_SIZE_USERNAME;?>" value="<?php echo string_attribute( $f_username ); ?>" class="<?php
+						echo $t_username_field_autofocus ?>" <?php echo $t_username_placeholder; ?> /></span>
```
atrol: We did hardly use placeholder until now (of course we should).
IE9 (the latest one on Windows Vista) does not support placeholder.
Not sure if we should also introduce $s_username_or_email and use it for the label.

config_defaults_inc.php
atrol: Setting to ON means having two user names matching a password.
This is decreased security.

There are users that use the same password but different user names for various accounts.
They shouldn't, but they do.
Typically their email address is known to others but not their user name.
Offering email address as user name is decreased security in this case.

I am not telling that we are opening a serious security issue with this, but keep in mind that hackers are using each single information they have.

> added a check that the provided email is a valid email address before looking up the DB for emails.

Not sure if you are aware that we still have the additional load with each SOAP call.

login_page.php
vboctor: Now I change the label rather than add a place holder.  Though not because of IE9 :) -- just checked evernote.com and they have it in the label.  I remember seeing it else where in the place holder.

login_page.php
```diff
@@ -66,6 +66,12 @@
 	$f_username = '';
 }
 
+if( config_get_global( 'email_login_enabled' ) ) {
+	$t_username_label = lang_get( 'username_or_email' );
+} else {
+	$t_username_label = lang_get( 'email' );
```
atrol: should be lang_get( 'username');

vboctor: Fixed.

PR #713, opened 2016-01-29
ldap authentication: add an option not to cache passwords

seveas:
Not everyone wants to cache passwords, even in hashed format. This new
option caters to those users.

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -201,15 +201,18 @@ function auth_auto_create_user( $p_username, $p_password ) {
 
 	if( $t_login_method == BASIC_AUTH ) {
 		$t_auto_create = true;
+		$t_auto_create_password = md5( $p_password );
 	} else if( $t_login_method == LDAP && ldap_authenticate_by_username( $p_username, $p_password ) ) {
 		$t_auto_create = true;
+		$t_cache_password = config_get( 'ldap_cache_passwords' );
+		$t_auto_create_password = $t_cache_password ? md5( $p_password ) : 'external-password';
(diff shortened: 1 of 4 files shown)


atrol:
No time to have a deeper look at the moment.
Just want to mention that 
- there is another PR for it https://github.com/mantisbt/mantisbt/pull/659
- there is an issue to track the feature request  https://www.mantisbt.org/bugs/view.php?id=12957

seveas:
That bug has been open for 5 years, damn :(

I'm also fine with complete removing the caching, as it's a big security issue.

core/authentication_api.php
```diff
@@ -201,15 +201,18 @@ function auth_auto_create_user( $p_username, $p_password ) {
 
 	if( $t_login_method == BASIC_AUTH ) {
 		$t_auto_create = true;
+		$t_auto_create_password = md5( $p_password );
 	} else if( $t_login_method == LDAP && ldap_authenticate_by_username( $p_username, $p_password ) ) {
 		$t_auto_create = true;
+		$t_cache_password = config_get( 'ldap_cache_passwords' );
+		$t_auto_create_password = $t_cache_password ? md5( $p_password ) : 'external-password';
```
vboctor: Let's use auto-generated password (instead of 'external-password') so that the password can't be guessed if auth method is changed for example.  I would use a model like random cookie generation we already do.

vboctor:
@seveas let's remove this without configuration.

seveas:
Closing in favor of #714, based on discussin  in #659

PR #714, opened 2016-01-29
Stop caching LDAP passwords

seveas:
When authenticating against LDAP, mantis does not need to know the
user's password, so it should not remember it.

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -209,7 +209,7 @@ function auth_auto_create_user( $p_username, $p_password ) {
 
 	if( $t_auto_create ) {
 		# attempt to create the user
-		$t_cookie_string = user_create( $p_username, md5( $p_password ) );
+		$t_cookie_string = user_create( $p_username, "!" + auth_generate_random_password() );
 		if( $t_cookie_string === false ) {
 			# it didn't work
 			return false;
--- a/core/ldap_api.php
+++ b/core/ldap_api.php
@@ -379,7 +379,11 @@ function ldap_authenticate_by_username( $p_username, $p_password ) {
 
 		if( false !== $t_user_id ) {
 
-			$t_fields_to_update = array('password' => md5( $p_password ));
+			$t_fields_to_update = array();
+			# Do we need to wipe a previously cached password?
+			if( substr( user_get_field( $t_user_id, 'password' ), 0, 1) != '!' ) {
+				user_set_field( 'password', '!' + auth_generate_random_password() );
+			}
 
 			if( ON == config_get( 'use_ldap_realname' ) ) {
 				$t_fields_to_update['realname'] = ldap_realname( $t_user_id );


core/authentication_api.php
```diff
@@ -209,7 +209,7 @@ function auth_auto_create_user( $p_username, $p_password ) {
 
 	if( $t_auto_create ) {
 		# attempt to create the user
-		$t_cookie_string = user_create( $p_username, md5( $p_password ) );
+		$t_cookie_string = user_create( $p_username, "" );
```
vboctor: Set the password to:
crypto_generate_uri_safe_nonce( 32 ) instead of blank string.

seveas: The password is never used, so why does it need to be set to any value? It being empty is a better indicator that it's unused.

vboctor:
I would also set the password on login to crypto_generate_uri_safe_nonce( 32 ) to wipe any previously cached passwords on login.

seveas:
I wouldn't mind setting it to empty upon login iff there is a previously cached value, but a random password doesn't add any value.

core/authentication_api.php
vboctor: If administrator want to switch away from LDAP you don't want to end up with a lot of users that have an open well known password.  You want such users to reset their password and continue to work.  Also it protects against an end point, e.g. a bug in API that doesn't honor LDAP and just goes directly to the MD5 hash.

One day we may also want to support a MantisBT instance that enables internal users to use LDAP and external to use normal auth.  That gives us good protection against any mistakes there as well.

Plus, it is cheap to set to a random password.

seveas: Then we should not set it to a random password but a string that clearly indicates 'disabled password', such as '!' followed by a random hash. That way we don't need to generate a random password (and update the user account) on every login...

vboctor: We already update the user often, e.g. last visit which is updated on every page, so I wouldn't over think it.  There is really two options to clear existing data:
- Generate a random password.
- Add a schema update step to reset all passwords in case of LDAP -- I think this is risky since if someone has an approach where they having mixed login, we will delete all their user passwords.

Now if you want ! + random hash, so that you can figure it out and avoid an extra update, then I'm OK with that.  Just make sure the final string fits.  Also you may want a function that generates this string and is used from the appropriate locations.

For example:
- ldap_generate_random_password()
- ldap_is_random_password()

seveas:
Force-pushed an updated commit based on @vboctor's feedback.

core/ldap_api.php
```diff
@@ -379,7 +379,11 @@ function ldap_authenticate_by_username( $p_username, $p_password ) {
 
 		if( false !== $t_user_id ) {
 
-			$t_fields_to_update = array('password' => md5( $p_password ));
+			$t_fields_to_update = array();
+			# Do we need to wipe a previously cached password?
+			if( user_get_field( $t_user_id, 'password' )[0] != '!' ) {
```
dregad: Function array dereferencing is not supported in PHP 5.3, which is our minimal supported version.

dregad:
I would suggest to perform the cleanup of cached passwords globally as a one-time upgrade step.

dregad:
Also, if you could please amend your commit message to add `Fixes #12957` so the commit gets properly referenced by source integration, that would be great.

seveas:
@dregad force-pushed a new commit that addresses both your comments.

dregad:
Thanks !

As a side note, while in review mode, I generally find it preferable to submit a series of corrective commits as they occur, (even if they are squashed in the end) rather than force-pushing a single amended commit, as it makes things easier on reviewers by maintaining the history.

dregad:
I'm closing this, since the proposed code is included in #718 together with additional fixes. Please continue the discussion there.

PR #716, opened 2016-01-31
Russian translations bulk update

dregad:
Contributed by Alexey Chumakov (@achumakov)

Fixes #20531


dregad:
As per @siebrand's request in https://www.mantisbt.org/bugs/view.php?id=20531

badfiles:
quotes should also be Russian, and a native speaker (other than author) review is required, I guess.

siebrand:
@badfiles changes can be reviewed at translatewiki.net.

siebrand:
It looks like every line in the file was replaced from the patch statistics. That's susicious, and most of the time points to a change in line breaks. Please fix that.

siebrand:
It also looks like the header copyright year was changed back from 2016 to 2014. But as the patch is so large, that github doesn't display a diff, I cannot assess additional formatting issues.

dregad:
@siebrand fixed line endings.

PR #718, opened 2016-02-05
Stop caching LDAP passwords (take 3)

dregad:
This builds upon @seveas pull request #714, with the following changes.

- original commit rebased on top of latest master
- corrected minor issues found during review
- added 2 new API calls to handle *no-cache passwords*, as per @vboctor's suggestion in https://github.com/mantisbt/mantisbt/pull/714#discussion_r51292184

Fixes [#12957](https://www.mantisbt.org/bugs/view.php?id=12957)

In https://github.com/mantisbt/mantisbt/pull/714#discussion_r51292184, @vboctor also wrote

> Add a schema update step to reset all passwords in case of LDAP -- I think this is risky since if someone has an approach where they having mixed login, we will delete all their user passwords.

The approach to wipe previously cached passwords incrementally as they are used still bothers me. I believe that doing this in a single pass would be much better because

- we increase security by removing all unsalted passwords stored in the DB once and for all
- the code is more efficient because we don't need to check for a 'no-cache' password all the time

I am not aware of a scenario where we can have mixed logins today. Do you have a specific case in mind ? If so, we could try to detect that in the upgrade step. As an alternative, maybe we could provide an admin check combined with some admin tool to wipe cached passwords.

Let me know your thoughts.

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -209,7 +209,7 @@ function auth_auto_create_user( $p_username, $p_password ) {
 
 	if( $t_auto_create ) {
 		# attempt to create the user
-		$t_cookie_string = user_create( $p_username, md5( $p_password ) );
+		$t_cookie_string = user_create( $p_username, "!" + auth_generate_random_password() );
 		if( $t_cookie_string === false ) {
 			# it didn't work
 			return false;
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -209,7 +209,7 @@ function auth_auto_create_user( $p_username, $p_password ) {
 
 	if( $t_auto_create ) {
 		# attempt to create the user
-		$t_cookie_string = user_create( $p_username, "!" + auth_generate_random_password() );
+		$t_cookie_string = user_create( $p_username, auth_generate_nocache_password() );
 		if( $t_cookie_string === false ) {
 			# it didn't work
 			return false;
@@ -538,6 +538,30 @@ function auth_process_plain_password( $p_password, $p_salt = null, $p_method = n
  */
 function auth_generate_random_password() {
 	return crypto_generate_uri_safe_nonce( 16 );
+}
+
+/**
+ * Generate a random string to use when we should not store the actual password.
+ * When relying on an external authentication system (e.g. LDAP), we should not
+ * store the user's password for security reasons. To avoid creating another
+ * security hole by populating the password field with a known value, we use a
+ * random string that can be identified as a 'nocache' password by its prefix.
+ * @return string Random password, prefixed by '!'
+ * @access public
+ */
+function auth_generate_nocache_password() {
+	return '!' . crypto_generate_strong_random_string( 31 );
+}
+
+/**
+ * Determine whether the given string is a 'no-cache' password.
+ * @see auth_generate_nocache_password()
+ * @param string
+ * @return boolean true if the string is a 'nocache' password, false otherwise
+ * @access public
+ */
+function auth_is_nocache_password( $p_password ) {
+	return $p_password != '' && $p_password[0] == '!';
 }
 
 /**
--- a/core/ldap_api.php
+++ b/core/ldap_api.php
@@ -379,7 +379,11 @@ function ldap_authenticate_by_username( $p_username, $p_password ) {
 
 		if( false !== $t_user_id ) {
 
-			$t_fields_to_update = array('password' => md5( $p_password ));
+			$t_fields_to_update = array();
+			# Do we need to wipe a previously cached password?
+			if( substr( user_get_field( $t_user_id, 'password' ), 0, 1) != '!' ) {
+				user_set_field( 'password', '!' + auth_generate_random_password() );
+			}
 
 			if( ON == config_get( 'use_ldap_realname' ) ) {
 				$t_fields_to_update['realname'] = ldap_realname( $t_user_id );
--- a/core/ldap_api.php
+++ b/core/ldap_api.php
@@ -373,16 +373,16 @@ function ldap_authenticate_by_username( $p_username, $p_password ) {
 
 	# If user authenticated successfully then update the local DB with information
 	# from LDAP.  This will allow us to use the local data after login without
-	# having to go back to LDAP.  This will also allow fallback to DB if LDAP is down.
+	# having to go back to LDAP.
 	if( $t_authenticated ) {
 		$t_user_id = user_get_id_by_name( $p_username );
 
 		if( false !== $t_user_id ) {
-
 			$t_fields_to_update = array();
-			# Do we need to wipe a previously cached password?
+
+			# Wipe previously cached password if necessary
 			if( substr( user_get_field( $t_user_id, 'password' ), 0, 1) != '!' ) {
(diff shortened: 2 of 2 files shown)


core/ldap_api.php
```diff
@@ -373,13 +373,17 @@ function ldap_authenticate_by_username( $p_username, $p_password ) {
...
+	# having to go back to LDAP.
 	if( $t_authenticated ) {
 		$t_user_id = user_get_id_by_name( $p_username );
 
 		if( false !== $t_user_id ) {
+			$t_fields_to_update = array();
 
-			$t_fields_to_update = array('password' => md5( $p_password ));
+			# Wipe previously cached password if necessary
+			if( auth_is_nocache_password( user_get_field( $t_user_id, 'password' ) ) ) {
```
vboctor: Are you missing a ! (NOT) fore auth_is_...?

core/authentication_api.php
```diff
@@ -541,6 +541,30 @@ function auth_generate_random_password() {
...
 /**
+ * Generate a random string to use when we should not store the actual password.
+ * When relying on an external authentication system (e.g. LDAP), we should not
+ * store the user's password for security reasons. To avoid creating another
+ * security hole by populating the password field with a known value, we use a
+ * random string that can be identified as a 'nocache' password by its prefix.
+ * @return string Random password, prefixed by '!'
+ * @access public
+ */
+function auth_generate_nocache_password() {
```
vboctor: Let's use auth_generate_no_login_password() -- since nocache is influenced by LDAP, but there are other scenarios where we don't want to enable logins, e.g. when a user is created and we want them to reset the password, bot accounts, etc.  Apply the same naming to the other method.

core/authentication_api.php
```diff
@@ -541,6 +541,30 @@ function auth_generate_random_password() {
...
+	return '!' . crypto_generate_strong_random_string( 31 );
```
vboctor: From memory even though you supply 31, it will still generate a 32 character since it goes to the next number that is divisible by 4.  I would generate the 32 and then truncate to 31 +  a comment to explain.

vboctor:
@dregad thanks for doing this.  I had it on my todo list to augment the change similar to what you have done, but I was waiting to close the other open PRs before doing more :) - thanks.

I worry about wiping the passwords field.  However, I'm OK with an admin check + script to do that with the appropriate warning.  However, this can be a separate change.

core/authentication_api.php
dregad: crypto_generate_strong_random_string() will return a string of exactly the specified length. You're probably confusing it with crypto_generate_uri_safe_nonce() which increases the size to a multiple of 4, to avoid base64 padding.

core/authentication_api.php
dregad: Ok, will change

core/ldap_api.php
dregad: Good catch, thanks

dregad:
Repeating discussion held on Gitter:

@vboctor Oct 25 08:01
> @atrol I think we can implement the LDAP issue in much simpler way. When creating a user generate a random password and not save login password to the database when user logs in. We also have to decide whether we want to apply the same model to basic auth (which is the case now - though not sure all code paths - have to review further). I don’t like wiping passwords from the database. If wonder if it is better for admin to wipe passwords via a script or sql query. That would also wipe passwords that have logged-in in the past but didn’t log in anymore. - cc @dregad
not sure I like the ! prefix.

@atrol Oct 26 20:07
>> I think we can implement the LDAP issue in much simpler way

>@vboctor What do you mean with it? Much simpler than what the PR does?

>> When creating a user generate a random passowrd and not save login password to the database when user logs in

> Isn't this what the PR implements?

@vboctor Nov 1 16:53
> @atrol Relating the LDAP password caching @dregad is back, so he can look at the PR. I think have some comments above and some on the PR. I would like to have it done in a way that achieves the following:
   - Don't overwrite existing passwords.
   - Don't cache new usages of passwords.
   - Don't worry about ! at the beginning of passwords, such generate complex passwords.
   - Administrators can use SQL to clear all passwords if they want.

dregad:
@vboctor 

> Don't overwrite existing passwords.
> Don't cache new usages of passwords.

I'm not sure I understand what you mean by that.

> not sure I like the ! prefix
> Don't worry about ! at the beginning of passwords, such generate complex passwords.

The `!` was not my idea, that's the original contributor's implementation. 

Can you please explain why you don't like it ?

I don't really care how we do it, but this approach has the advantage of making sure that the generated "password" will *never* match a real one, since md5 hash cannot produce a `!`. In addition, it allows identifying dummy passwords vs real ones, which could be useful in a mixed (multiple login methods) or  transitional (prior to cleaning up previously cached passwords) scenario.

> Administrators can use SQL to clear all passwords if they want.

I don't like the SQL approach, that's too error-prone and not user-friendly. I'd rather stick to my previously proposed approach of admin check + script to perform the cleanup.

vboctor:
@dregad 

> Don't overwrite existing passwords.
> Don't cache new usages of passwords.

I think the issue we need to fix is to stop caching LDAP passwords in the db.  Typically companies have a policy that force rotating passwords every 90 days or so.  Existing passwords cached will be obsolete or can be removed using an admin action (script or sql query).

> The ! was not my idea, that's the original contributor's implementation.

I don't like it because it introduces custom logic for generating and checking on these new kind of passwords.  It also not more secure that a secure long password that we can auto-generate as needed.  A proper fix would be to set to a secure password and disallow password reset functionality from Mantis in case of LDAP.  Haven't checked if latter is already done.

> I don't like the SQL approach, that's too error-prone and not user-friendly. I'd rather stick to my previously proposed approach of admin check + script to perform the cleanup.

That's fine.  I just don't think it should be a runtime thing.

l266509:

rfrovarp:
Can this be applied? I can't have my credentials cached as md5, that just can't happen. I can apply the patch locally, but then I have to apply each and every time we upgrade. It would be much better for this to be in the core.

tfnab:
This **security issue** was submitted to the project at least seven-and-a-half years ago (see https://mantisbt.org/bugs/view.php?id=12957) and I seriously don't understand why it can't be fixed. What is required to get the project managers' attention? Please accept this request!

yosh-se:
Is there a reason why hasn't this been resolved yet? The first mention of this issue is 12 years at this point. I don't want to maintain a local patch, is there any way that I can contribute to push this forward?

core/authentication_api.php
yosh-se: Shouldn't this be using crypto_generate_uri_safe_nonce() instead of crypto_generate_strong_random_string()? I had database issues with the output of the latter.

raspopov:
Also, in the current PHP version, `md5( null )` is deprecated and will generate a warning at **authentication_api.php**:441:
```
439:	if( $t_auto_create ) {
440:		# attempt to create the user
441:		$t_cookie_string = user_create( $p_username, md5( $p_password ) );
442:		if( $t_cookie_string === false ) {
```
For LDAP or other external authentication the $p_password is **null** here.

PR #719, opened 2016-02-08
Roadmap for entries without target version

dforstner:
Refers to https://www.mantisbt.org/bugs/view.php?id=17632


vboctor:
Looks like the discussion on the issue lead to not implement this issue.  Plus the PR is full of space changes, so it is not really consumable.

Any reason not to close this PR?

dforstner:
Well, I'm quite new to github and realized late, that my editor made heavy space changes. I tried to correct them which with limited success. If you want, I can drop this change and try it again without impacting the spacings.

The reason why I picked this issue is, that I try to harmonize our companies Mantis installation (version 1.0 with a couple of small changes) with the official project, in order to be able to update to version 1.2. My view is that Mantis has a quite bad usability when unclosed entries are hidden from the roadmap. Since there is also an entry for this in the official Mantis, I'm quite sure that many others share my view.

Regards,
Daniel

atrol:
@dforstner, did you read the whole discussion at https://www.mantisbt.org/bugs/view.php?id=17632?

Our user pauldouglas was not able to convince dregad or me that the feature request makes sense.
I was finally able to convince him (at least I think so) that it makes no sense to populate automatically the roadmap.

I understand the roadmap as a list of issues where someone promised to resolve the issue in a future version.
I can't imagine that you automatically want to promise to resolve any incoming issue without having a deeper look at it.

Keep also in mind that we will not accept PR's for version 1.2.x as we are near releasing 1.3.x.

dforstner:
@atrol, I read the whole discussion and I tried to do, what you suggested before you changed your mind in your last post.
I understand your point, but I feel uncomfortable with introducing dummy-versions for accepted entries only for not loosing issues out of focus, which are promissed to be solved once in the future. Maybe it would improve this change, if I introduce a status threshold for showing unassigned entries in the roadmap (e.g. min. accepted).

If this PR is accepted I would be glad to provide the same change for 1.3.x as well. Currently, I want to stick to 1.2.x because some plugins, I want to use are not available at 1.3.x and I didn't see any breaking changes in 1.3.x, which I want to use (well, besides of the timeline-feature). 

I hope, I can convince you with this explanation.

dregad:
@dforstner considering the discussion in the issue, I find it unlikely that your suggestion would get merged.

Also note that the probability of it happening is virtually zero if we can't review the code, which is not possible at the moment due to the whitespace mess. 

> If this PR is accepted I would be glad to provide the same change for 1.3.x as well. 

Enhancements *must* be targeted at master branch, 1.2.x only gets bug fixes.

dforstner:
OK. I'm going to close this PR. For now, I'm going to use my own fork of 1.2.x and later on I'll going to bother you again with clean PRs against the master branch.

PR #720, opened 2016-02-14
Move Bug update events into BugData

cproensa:
EVENT_UPDATE_BUG_DATA and EVENT_UPDATE_BUG
are now triggered in the BugData update function, as suggested in #0020577

A new parameter for a callback is introduced in
BugData::update(), to allow for bug related data
to be saved before EVENT_UPDATE_BUG

The callback thing is to guarantee that all bug related data has been created/saved at the point the final event is triggered.
This should be a transitional situation... to keep current code structure with minimal changes, when migrating other functions into the use of BugData for modifications
As a first step, i've adapted bug_update page, and refactored bug_move function.

Ideally, the functionality for checks, side effects, notifications, etc should be moved into a managed object system. Something like what BugData tries to achieve, but also accounting for custom fields, bugnotes, attachments, etc. to implement atomic modifications, and non-redundant "bussiness logic".

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -1841,10 +1841,13 @@ function bug_reopen( $p_bug_id, $p_bugnote_text = '', $p_time_tracking = '0:00',
 	# Add bugnote if supplied
 	# Moved bugnote_add before bug_set_field calls in case time_tracking_no_note is off.
 	# Error condition stopped execution but status had already been changed
+	# @todo cproensa: is the comment above still valid? bugnote will be added even if bug-update cant be performed
 	bugnote_add( $p_bug_id, $p_bugnote_text, $p_time_tracking, $p_bugnote_private, 0, '', null, false );
 
-	bug_set_field( $p_bug_id, 'status', config_get( 'bug_reopen_status' ) );
-	bug_set_field( $p_bug_id, 'resolution', config_get( 'bug_reopen_resolution' ) );
+	$t_bugdata = bug_get( $p_bug_id );
+	$t_bugdata->status = config_get( 'bug_reopen_status' );
+	$t_bugdata->resolution = config_get( 'bug_reopen_resolution' );
+	$t_bugdata->update( false, true );
 
 	email_bug_reopened( $p_bug_id );
 
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -1657,36 +1657,41 @@ function bug_assign( $p_bug_id, $p_user_id, $p_bugnote_text = '', $p_bugnote_pri
 		trigger_error( ERROR_USER_DOES_NOT_HAVE_REQ_ACCESS );
 	}
 
-	# extract current information into history variables
-	$h_status = bug_get_field( $p_bug_id, 'status' );
-	$h_handler_id = bug_get_field( $p_bug_id, 'handler_id' );
+	$t_bugdata = bug_get( $p_bug_id );
+	$t_original_handler_id = $t_bugdata->handler_id;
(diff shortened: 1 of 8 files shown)


cproensa:
I'm thinking that maybe different events should be used, the reasoning is this:
1) Events that are used on the update page from bug_update. These are where the current events are placed.
2) A different set of events that are called from the core BugData update.

The reason for (1) is, that these events are known to be called on the processing page for bug_update. This allows to reliably get information from the http request, like some arbitrary fields that where inserted in the update form by other form events.
This is a common work flow to add functionality from a plugin: use the form to add inputs, and read them on the update page.

On the other hand, events like (2), could be called from anywhere in the code. At this point we can't know if its a bug_update processing page, to read POST data, etc.

Currently, most of the events are designed to be closely tied to the web frontend pages, such as that each event happens at only one point of the web flow.
Having an event inside BugData core methods would imply creating a new "core" event.
Plugins could hook on the current bug events if they need to process form data, or hook to the core BugData event for more general actions and validations.

core/bug_api.php
```diff
@@ -1720,65 +1747,74 @@ function bug_resolve( $p_bug_id, $p_resolution, $p_fixed_in_version = '', $p_bug
...
+	# only set handler if specified explicitly or if bug was not assigned to a handler
+	if( null == $p_handler_id ) {
+		if( $t_bugdata->handler_id == 0 ) {
+			$t_bugdata->handler_id = auth_get_current_user_id();
+		}
+	} else {
+		$t_bugdata->handler_id = $p_handler_id;
+	}
+
+	if( !class_exists( 'Bug_assign_post_callback' ) ) {
```
atrol: No time to have a deeper look at the whole PR, but this should certainly be Bug_resolve_post_callback

cproensa: Thanks, fixed
However this classes are a proof of concept, they should certainly be placed in the classpath, and defined in a more general way.

cproensa:
As a heads up, and summary of this work:

If we agree that:
- EVENT_UPDATE_BUG_DATA and EVENT_UPDATE_BUG should be triggered inside the BugData update method
- All bug modifications should be done through the BugData object. (We'll still keep some direct table updates for some specific core code, that should be under control)
That would guarantee the consistency of the application for core, plugins, and soap, operations.

The problem with this approach is that other modifications beside the main bug table may exist (eg: adding notes). These operations has to be done:
- After EVENT_UPDATE_BUG_DATA, because this event can be used for validation to stop the update operation. 
- before EVENT_UPDATE_BUG, becasue this event should indicate finalization of the update, including all related tasks.

Since the BugData->update() is an atomic call, the method proposed is to include a callback system that stores the operations to be performed after "validation" and before "finalization".

A clear example of this is updating custom fields.
CF are not updated by the BugData operation, and the changes has to be performed by the caller.
If you update *before*, then if a plugin stops the update (due to some failed validation), the fields are already changed. If you update *after*, a plugin that listens to bug update event dont know the new CF values, since at that moment they are not still committed.

As for implementation, i am thinking this as a better implementation:
- Define some specific operation classes, for: adding a note, update a CF, add a relation, etc,
- These classes are callables.
- Create a callback queue in BugData, which would be executed at update time. These callbacks are not limited to the defined objects, since they are PHP standard callables.
The classes are needed due to most of the mantis code not being OOP, so it cant persist the data until the callback is executed (in another time and scope)

This solution is nice, because then a plugin can hook on EVENT_UPDATE_BUG_DATA and add its own update operation and keep consistency with the proposed update methodology.

cproensa:
updated in #834

PR #722, opened 2016-02-19
Don't generate separate delete token for each attachment

tgulacsi:
Fixes Issue #0020142.


tgulacsi:
With PR #656 it solves my bug_view slowness problems.

core/print_api.php
```diff
@@ -1819,7 +1820,7 @@ function print_bug_attachments_list( $p_bug_id ) {
  * @param array $p_attachment An attachment array from within the array returned by the file_get_visible_attachments() function.
  * @return void
```
dregad: PHPdoc missing for new param

core/print_api.php
```diff
@@ -1850,7 +1851,7 @@ function print_bug_attachment( array $p_attachment ) {
...
```
dregad: PHPdoc missing for new param

core/print_api.php
```diff
@@ -1877,8 +1878,9 @@ function print_bug_attachment_header( array $p_attachment ) {
 	}
 
 	if( $p_attachment['can_delete'] ) {
+		$t_delete_token = $p_bug_file_delete_token ? $p_bug_file_delete_token : form_security_param( 'bug_file_delete' );
```
cproensa: I think it's cleaner to pass only the token id, not the prebuilt string.
However, you'll need to add the option to form_security_param(), to accept the token id as parameter. Like it's already done in form_security_field()

cproensa:
+1, tested ok

dregad:
Fixed in 90c17a52ee0eb0b0ad5dfbc65260d2af382df086

PR #724, opened 2016-02-23
Highlight topmost priorities in roadmap.

leonardvimond:
Make issues ordering by priority (instead of last_updated)
+
Put in bold issues that have a significant priority, according to config value 'priority_significant_threshold'

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -988,6 +988,18 @@ function bug_is_resolved( $p_bug_id ) {
 }
 
 /**
+ * Check if a given bug has a significant priority
+ * @param int p_bug_id integer representing bug id
+ * @return bool true if bug is significant, false otherwise
+ * @access public
+ * @uses config_api.php
+ */
+function bug_has_significant_priority( $p_bug_id ) {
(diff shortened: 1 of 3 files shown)


vboctor:
Thanks @leonardvimond for your contribution.

I would rather stick with the current behavior which seems to have worked for most users until we have a way to provide users with control on how they would like to sort the view.  Similar to how users are able to control the ordering on the View Issues page.  Changing from date based to priority based, is just going to annoy the rest of the users.

The View Issues page provides users with the ability to control the sorting on a per user bases with a fair bit of flexibility.  If we want flexibility in roadmap/changelog, then we should possibly consider something similar.

leonardvimond:
yes, you're probably right to keep the current behavior for the ordering, instead the second commit "put in bold issues that have a significant priority" does not change anything for users and make anyway the first commit insignificant.

dregad:
> the second commit "put in bold issues that have a significant priority" 

@leonardvimond, considering that you can customize the display to satisfy your requirement by defining your own custom function, I don't see the need to change the core.

PR #725, opened 2016-02-24
Added link format option to automatically-generated www links

quayzar:
I've added a format option to the `string_insert_hrefs` function. 

If the existing flag `$g_html_make_links` is set to `ON` and the new flag `$g_html_make_links_new_window` is _also_ set to `ON`, the function will automatically convert www URLs into clickable links that open in a new window themselves, without the `[^]` link.


core/string_api.php
```diff
@@ -508,7 +508,12 @@ function string_insert_hrefs( $p_string ) {
 		$s_url_regex,
 		function ( $p_match ) {
 			$t_url_href = 'href="' . rtrim( $p_match[1], '.' ) . '"';
-			return "<a ${t_url_href}>${p_match[1]}</a> [<a ${t_url_href} target=\"_blank\">^</a>]";
+			$t_url_html = "<a ${t_url_href}>${p_match[1]}</a> [<a ${t_url_href} target=\"_blank\">^</a>]";
```
vboctor: This should move to the else, rather than being done and thrown away in some cases.

vboctor:
Thanks @quayzar for your contribution.

Is it really an issue to have the default behavior?  Is it about making it look better? e.g. show an icon instead of ^.  I would rather not have this configurable.

The change is also missing the following:
1. A bug in our bug tracker, that is referenced from the body of the commit with "Fixes #1234".
2. Documentation of the new configuration option in the manual.

I would rather not add the configuration option, but open for arguments why this is necessary.  If it is about the current format being confusing or can be made to look better, then let's fix that.

quayzar:
I've updated the condition to employ if/else, as you requested. 

Per your notes:
1. This change wasn't made in response to a reported bug, so there was no bug number to reference.
2. I didn't find an entry in the manual for `$g_html_make_links` so I've added one, as well as an entry for `$g_html_make_links_new_window`.

Regarding the need for this option, I made the change locally in response to several clients requesting that links open in a new window by default. They made this request even after I pointed out the `[^]` link so I don't believe a change in appearance would solve the problem.

I don't think this is a _necessary_ change, just a UI improvement based on user feedback. Feel free to merge or decline at your discretion. Thanks for the useful software!

atrol:
As a side note:
We removed the [^] links on "My View" page in 1.3.

I prefer to have just one type of links and to use Shift and Ctrl key to decide if I want to open in a new window or a new tab.
Maybe there are users that are not aware of this basic functionality of browsers.

dregad:
> This change wasn't made in response to a reported bug, so there was no bug number to reference.

If there is no issue, then ideally you should create one for reference.

> I prefer to have just one type of links

+1

> I would rather not add the configuration option

I don't feel strongly about this, although I tend to agree that we should try to avoid the YACO (Yet Another Config Option) syndrome - we have so many of these already ;) 

As an alternative, instead of having 2 settings with a boolean value, what about expanding the existing one and use numeric values, i.e. `$g_html_make_links` = 

- *OFF* (0) - no links
- *ON* or *LINK_SAME_WINDOW* (1) - links as per current behavior
- *LINK_NEW_WINDOW* (2) - links with target=_blank

quayzar:
> If there is no issue, then ideally you should create one for reference.

Thanks for the tip, @dregad. I'll do this in the future.

> we should try to avoid the YACO (Yet Another Config Option) syndrome

Agreed. I like your suggestion and will revise my edits accordingly and recommit.

config_defaults_inc.php
```diff
@@ -1784,11 +1784,15 @@
...
+ * This flag controls whether www URLs and email addresses
+ * are automatically converted to clickable links as well as
+ * where the www links open when clicked. Options:
+ * 0 = Do not convert www URLs or email addresses to links
+ * 1 = Convert to links that open in the current window (DEFAULT)
+ * 2 = Convert to links that open in a new window
  * @global integer $g_html_make_links
  */
-$g_html_make_links = ON;
+$g_html_make_links = 1;
```
dregad: Don't use "magic numbers". Please define constants as appropriate (in core/constants_inc.php).

core/string_api.php
```diff
@@ -503,12 +503,19 @@ function string_insert_hrefs( $p_string ) {
...
 	}
 
-	# Find any URL in a string and replace it by a clickable link
+	# Find any URL in a string and replace it with a clickable link
 	$p_string = preg_replace_callback(
 		$s_url_regex,
 		function ( $p_match ) {
 			$t_url_href = 'href="' . rtrim( $p_match[1], '.' ) . '"';
-			return "<a ${t_url_href}>${p_match[1]}</a> [<a ${t_url_href} target=\"_blank\">^</a>]";
+			if( config_get( 'html_make_links' ) == 2 ) {
```
dregad: Same comment as above

dregad:
Also, with regards to commit messages, kindly read https://www.mantisbt.org/wiki/doku.php/mantisbt:git_commit_messages

quayzar:
Add link target options to $g_html_make_links

- Replace "magic numbers" with constants defined in core/constants_inc.php
- Update manual and config_defaults_inc.php with revised constants

Fixes #[20694](https://www.mantisbt.org/bugs/view.php?id=20694).

Signed-off-by: Ian MacKenzie <ian@quayzar.com>

dregad:
Merged in aa91878cd8bdb28fd302598d8a26471a30c56a3b. Thanks for your contribution.

PR #726, opened 2016-02-27
Fix deletion of default category on MSSQL/Oracle

dregad:
$f_category_id is of type int, but the config table's config_id column
is defined as VARCHAR(64). This causes a type mismatch on some RDBMS:

- Oracle (ORA-00932: inconsistent data types: expected -, found CLOB)
- MSSQL

Problem resolved by casting the variable to string when passing it to
db_query().

Fixes [#16384](https://www.mantisbt.org/bugs/view.php?id=16834)

### DIFF
--- a/manage_proj_cat_delete.php
+++ b/manage_proj_cat_delete.php
@@ -67,9 +67,9 @@ access_ensure_project_level( config_get( 'manage_project_threshold' ), $t_projec
 
 # Protect the 'default category for moves' from deletion
 $t_default_category_id = config_get( 'default_category_for_moves', /* default */ null, ALL_USERS, ALL_PROJECTS );
-$t_query = 'SELECT count(config_id) FROM {config} WHERE config_id = ' . db_param() . ' AND value = ' . db_param();
-$t_default_cat_count = db_result( db_query( $t_query, array( 'default_category_for_moves', (string)$f_category_id ) ) );
-if( $t_default_cat_count > 0 || $f_category_id == $t_default_category_id ) {
+if(    $f_category_id == $t_default_category_id
+	|| config_is_defined( 'default_category_for_moves', $f_category_id )
+) {
 	trigger_error( ERROR_CATEGORY_CANNOT_DELETE_DEFAULT, ERROR );
 }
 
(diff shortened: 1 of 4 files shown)


manage_proj_cat_delete.php
```diff
@@ -68,7 +68,7 @@
 # Protect the 'default category for moves' from deletion
 $t_default_category_id = config_get( 'default_category_for_moves', /* default */ null, ALL_USERS, ALL_PROJECTS );
 $t_query = 'SELECT count(config_id) FROM {config} WHERE config_id = ' . db_param() . ' AND value = ' . db_param();
-$t_default_cat_count = db_result( db_query( $t_query, array( 'default_category_for_moves', $f_category_id ) ) );
+$t_default_cat_count = db_result( db_query( $t_query, array( 'default_category_for_moves', (string)$f_category_id ) ) );
```
vboctor: Shouldn't this be a call to config api?

dregad: We need to find if there is *any* occurence of the given category in the config table (i.e. in all projects and/or in specific ones and for any user); I don't think the current API offers this functionality. Are you saying we should add a new function for that ?

vboctor: Ah I missed the type of config check.  At least, it feels that this logic should moved to category API.  That can in turn leverage a new API in config api.

dregad:
Commit 2309587d3d3d5d5b2c16eb04ac43c34aceac0b48 did not fix the issue, [as reported](https://www.mantisbt.org/bugs/view.php?id=16834#c52606).

@vboctor 
- I added a new config API function as you suggested; this relies on the config cache instead of querying the DB. Let me know what you think.
- Not really sure what you had in mind for category API

The new commits should hopefully address the problem.

vboctor:
Looks good. :+1: 

For the category api, I was thinking adding category_can_delete( $p_category_id ) or something along these lines.  This would do the necessary checks.

Anyway, feel free to checkin either way.

dregad:
> adding category_can_delete( $p_category_id ) 

Ah OK, that makes sense. Will do.

PR #727, opened 2016-03-02
Allow adding comma-separated monitor list

tgulacsi:
This helps adding the same herd of monitors to similar bugs: just copy
the list, comma-separated into the "username" field.

### DIFF
--- a/bug_monitor_add.php
+++ b/bug_monitor_add.php
@@ -55,26 +55,35 @@ form_security_validate( 'bug_monitor_add' );
 
 $f_bug_id = gpc_get_int( 'bug_id' );
 $t_bug = bug_get( $f_bug_id, true );
-$f_username = gpc_get_string( 'username', '' );
+$f_usernames = trim( gpc_get_string( 'username', '' ) );
 
 $t_logged_in_user_id = auth_get_current_user_id();
 
-if( is_blank( $f_username ) ) {
-	$t_user_id = $t_logged_in_user_id;
-} else {
-	$t_user_id = user_get_id_by_name( $f_username );
-	if( $t_user_id === false ) {
-		$t_user_id = user_get_id_by_realname( $f_username );
-
-		if( $t_user_id === false ) {
-			error_parameters( $f_username );
-			trigger_error( ERROR_USER_BY_NAME_NOT_FOUND, E_USER_ERROR );
-		}
+if( is_blank( $f_usernames ) ) {
+	if( user_is_anonymous( $t_logged_in_user_id ) ) {
+		trigger_error( ERROR_PROTECTED_ACCOUNT, E_USER_ERROR );
 	}
-}
+	$t_user_ids = array( $t_logged_in_user_id );
+} else {
+	$t_usernames = preg_split( '/[,|]/', $f_usernames, -1, PREG_SPLIT_NO_EMPTY );
+	$t_usernames = array_unique( $t_usernames );
+	$t_user_ids = array();
+	foreach( $t_usernames as $t_username ) {
+		$t_username = trim( $t_username );
+		$t_user_id = user_get_id_by_name( $t_username );
+		if( $t_user_id === false ) {
+			$t_user_id = user_get_id_by_realname( $t_username );
 
-if( user_is_anonymous( $t_user_id ) ) {
-	trigger_error( ERROR_PROTECTED_ACCOUNT, E_USER_ERROR );
+			if( $t_user_id === false ) {
+				error_parameters( $t_username );
+				trigger_error( ERROR_USER_BY_NAME_NOT_FOUND, E_USER_ERROR );
+			}
+		}
+		if( user_is_anonymous( $t_user_id ) ) {
+			trigger_error( ERROR_PROTECTED_ACCOUNT, E_USER_ERROR );
+		}
+		$t_user_ids[$t_user_id] = $t_user_id;
+	}
 }
 
 bug_ensure_exists( $f_bug_id );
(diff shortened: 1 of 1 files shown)


bug_monitor_add.php
```diff
@@ -55,26 +55,32 @@
...
-	$t_user_id = $t_logged_in_user_id;
+if( is_blank( $f_usernames ) ) {
+	$t_user_ids = array( $t_logged_in_user_id );
 } else {
-	$t_user_id = user_get_id_by_name( $f_username );
-	if( $t_user_id === false ) {
-		$t_user_id = user_get_id_by_realname( $f_username );
-
+	$t_user_ids = array();
+	foreach( explode( ',', $f_usernames ) as $f_username ) {
```
vboctor: We should support splitting based on ',' and '|'.

vboctor: Don't use $f_username, use $t_username instead.  $f_ are used for form fields.

bug_monitor_add.php
```diff
@@ -55,26 +55,32 @@
...
+		$t_user_id = user_get_id_by_name( $f_username );
```
vboctor: Trip username before passing in to this method or trim() within the method.

bug_monitor_add.php
```diff
@@ -55,26 +55,32 @@
...
+			}
 		}
+		$t_user_ids[] = $t_user_id;
 	}
 }
 
-if( user_is_anonymous( $t_user_id ) ) {
-	trigger_error( ERROR_PROTECTED_ACCOUNT, E_USER_ERROR );
+foreach( $t_user_ids as $t_user_id ) {
+	if( user_is_anonymous( $t_user_id ) ) {
```
vboctor: Should this check be moved in the same loop as the access check below?

bug_monitor_add.php
```diff
@@ -85,13 +91,15 @@
...
-}
+foreach( $t_user_ids as $t_user_id ) {
+	if( $t_logged_in_user_id == $t_user_id ) {
+		access_ensure_bug_level( config_get( 'monitor_bug_threshold' ), $f_bug_id );
+	} else {
+		access_ensure_bug_level( config_get( 'monitor_add_others_bug_threshold' ), $f_bug_id );
+	}
 
-bug_monitor( $f_bug_id, $t_user_id );
+	bug_monitor( $f_bug_id, $t_user_id );
```
vboctor: Maybe have a separate loop to add all monitors.  This way if this loop fails, we end up with no changes, rather than half the changes.

bug_monitor_add.php
```diff
@@ -85,13 +91,15 @@
...
+}
 
 form_security_purge( 'bug_monitor_add' );
```
vboctor: Give the amount of possible errors above, I wonder if this call should be earlier on this page.  Not part of this change, but should be easy to change.

bug_monitor_add.php
```diff
@@ -55,26 +55,34 @@
...
-		$t_user_id = user_get_id_by_realname( $f_username );
-
-		if( $t_user_id === false ) {
-			error_parameters( $f_username );
-			trigger_error( ERROR_USER_BY_NAME_NOT_FOUND, E_USER_ERROR );
+	$t_usernames = array( $f_usernames );
+	foreach( array( "|", "," ) as $t_sep ) {
+		if( strpos( $f_usernames, $t_sep ) !== false ) {
+			$t_usernames = explode( $t_sep, $f_usernames );
+			break;
```
dregad: Wouldn't it be simpler to do 
``` php
$t_usernames = preg_split( '/[,|]/', $f_usernames, -1, PREG_SPLIT_NO_EMPTY )
```
And maybe add an `array_unique()` call too, just in case.

bug_monitor_add.php
```diff
@@ -85,13 +93,23 @@
...
+
+	if( $t_logged_in_user_id == $t_user_id ) {
+		access_ensure_bug_level( config_get( 'monitor_bug_threshold' ), $f_bug_id );
+	} else {
+		access_ensure_bug_level( config_get( 'monitor_add_others_bug_threshold' ), $f_bug_id );
+	}
 }
 
-bug_monitor( $f_bug_id, $t_user_id );
+// then add only if all can be added.
```
dregad: We use `#` for inline comments

bug_monitor_add.php
```diff
@@ -85,13 +93,23 @@
 	$g_project_override = $t_bug->project_id;
 }
 
-if( $t_logged_in_user_id == $t_user_id ) {
-	access_ensure_bug_level( config_get( 'monitor_bug_threshold' ), $f_bug_id );
-} else {
-	access_ensure_bug_level( config_get( 'monitor_add_others_bug_threshold' ), $f_bug_id );
+// Check all monitors first,
```
dregad: We use `#` for inline comments

bug_monitor_add.php
```diff
@@ -55,26 +55,34 @@
...
+			$t_user_id = user_get_id_by_realname( $t_username );
 
-if( user_is_anonymous( $t_user_id ) ) {
-	trigger_error( ERROR_PROTECTED_ACCOUNT, E_USER_ERROR );
+			if( $t_user_id === false ) {
+				error_parameters( $t_username );
+				trigger_error( ERROR_USER_BY_NAME_NOT_FOUND, E_USER_ERROR );
+			}
+		}
+		$t_user_ids[] = $t_user_id;
```
dregad: `$t_user_ids[$t_user_id] = $t_user_id;` prevents duplicates

bug_monitor_add.php
```diff
@@ -85,13 +93,23 @@
 	$g_project_override = $t_bug->project_id;
 }
 
-if( $t_logged_in_user_id == $t_user_id ) {
-	access_ensure_bug_level( config_get( 'monitor_bug_threshold' ), $f_bug_id );
-} else {
-	access_ensure_bug_level( config_get( 'monitor_add_others_bug_threshold' ), $f_bug_id );
+// Check all monitors first,
+foreach( $t_user_ids as $t_user_id ) {
```
dregad: Is there any benefit of doing these 2 steps in an extra loop ?

dregad:
Thanks for your contribution.
Can you please reference an issue in our tracker (create one if needed.

bug_monitor_add.php
tgulacsi: If I'd know PHP, then yes :) So thanks for the simpler solution!

bug_monitor_add.php
tgulacsi: Nope, just conserved the logic of the original.

dregad:
Merged in d7953e10ff9c0b48320448a0309e15071786003c. Thank you !

dregad:
@tgulacsi as a side note, it would appear the e-mail address you used to commit https://github.com/mantisbt/mantisbt/commit/b694c1434ac366ed67744ca47c837518065f6005 is not registered with your GitHub account, you may want to fix that. 

2016-03-07 10_07_07-allow adding or _ separated monitor list mantisbt_mantisbt b694c14

PR #728, opened 2016-03-02
Print NOTHING if $g_show_avatar === OFF

tgulacsi:


vboctor:
Would you want to hide even if users were able to upload their own image, i.e. not just gravatar support?  We are planning to add another avatar plugin to provide such functionality.

dregad:
I suggest to have the discussion in the tracker 
https://www.mantisbt.org/bugs/view.php?id=20641

vboctor:
I've submitted a PR for several follow up fixes for avatar plugin support
https://github.com/mantisbt/mantisbt/pull/734

PR #729, opened 2016-03-03
Fix defaulting of issue page time tracking range

vboctor:
The start date should be defaulted to issue submission date.
The end date should be defaulted to today.

They were defaulted to Jan 1st of current year.

Fixes #20655


dregad:
+1

atrol:
Not sure if this is the better approach as there can't be anything better in filter API without a major rewrite.
```php
filter_init( $t_filter )
```

vboctor:
@atrol I've updated the change to use filter_init().  Thanks @dregad and @atrol for fast review.

PR #730, opened 2016-03-04
Fix unicode encoding for csv exports

vboctor:
Fixes #20659
Fixes #20660

### DIFF
--- a/core/csv_api.php
+++ b/core/csv_api.php
@@ -45,17 +45,23 @@ require_api( 'project_api.php' );
 require_api( 'user_api.php' );
 
 /**
- * Emits headers for csv file download pages.
+ * Emits the headers and byte order marker.  This must be called at the beginning
+ * of scripts that export CSV file before any other content is called.
  *
- * @param  string $p_filename The filename
+ * @param  string  $p_filename The csv filename and extension.
+ * @param  boolean $p_utf8     Use UTF8 or ASCII.
  * @return void
  */
-function csv_emit_headers( $p_filename ) {
+function csv_start( $p_filename ) {
+	$t_filename = urlencode( file_clean_name( $p_filename ) );
+
 	header( 'Pragma: public' );
 	header( 'Content-Encoding: UTF-8' );
-	header( 'Content-Type: text/csv; name=' . urlencode( file_clean_name( $p_filename ) ) . ';charset=UTF-8' );
+	header( 'Content-Type: text/csv; name=' . $t_filename . ';charset=UTF-8' );
 	header( 'Content-Transfer-Encoding: BASE64;' );
-	header( 'Content-Disposition: attachment; filename="' . urlencode( file_clean_name( $p_filename ) ) . '"' );
+	header( 'Content-Disposition: attachment; filename="' . $t_filename . '"' );
+
+	echo UTF8_BOM;
 }
 
 /**
--- a/core/csv_api.php
+++ b/core/csv_api.php
@@ -59,14 +59,6 @@ function csv_emit_headers( $p_filename ) {
 }
 
 /**
- * Generates the byte order marker (BOM) to be generated at the top of the csv file.
- * @return string The byte order marker.
- */
-function csv_get_byteordermarker() {
-	return "\xEF\xBB\xBF";
-}
-
-/**
  * get the csv file new line, can be moved to config in the future
  * @return string containing new line character
  * @access public
--- a/core/csv_api.php
+++ b/core/csv_api.php
@@ -28,6 +28,7 @@
  * @uses category_api.php
  * @uses config_api.php
  * @uses constant_inc.php
+ * @uses file_api.php
  * @uses helper_api.php
  * @uses project_api.php
  * @uses user_api.php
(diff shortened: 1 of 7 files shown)


core/csv_api.php
```diff
@@ -38,11 +39,34 @@
...
+	header( 'Content-Type: text/csv; name=' . urlencode( file_clean_name( $p_filename ) ) . ';charset=UTF-8' );
+	header( 'Content-Transfer-Encoding: BASE64;' );
+	header( 'Content-Disposition: attachment; filename="' . urlencode( file_clean_name( $p_filename ) ) . '"' );
+}
+
+/**
+ * Generates the byte order marker (BOM) to be generated at the top of the csv file.
+ * @return string The byte order marker.
+ */
+function csv_get_byteordermarker() {
```
atrol: Is there a special reason to have a function instead of a constant?

dregad:
It's a good idea to move the sending of headers to an API, although I would recommend put that in *HTTP API* instead, together with other `http_xxxx_headers()` functions.

core/csv_api.php
```diff
@@ -38,11 +39,34 @@
...
+ *
+ * @param  string $p_filename The filename
+ * @return void
+ */
+function csv_emit_headers( $p_filename ) {
+	header( 'Pragma: public' );
+	header( 'Content-Encoding: UTF-8' );
+	header( 'Content-Type: text/csv; name=' . urlencode( file_clean_name( $p_filename ) ) . ';charset=UTF-8' );
+	header( 'Content-Transfer-Encoding: BASE64;' );
+	header( 'Content-Disposition: attachment; filename="' . urlencode( file_clean_name( $p_filename ) ) . '"' );
```
dregad:
``` php
$t_filename = urlencode( file_clean_name( $p_filename ) );
...
header( 'Content-Type: text/csv; name=' . $t_filename . ';charset=UTF-8' );
...
header( 'Content-Disposition: attachment; filename="' . $t_filename . '"' );
```

core/csv_api.php
dregad: +1 for constant

csv_export.php
```diff
@@ -68,16 +68,8 @@
 
 $t_filename = csv_get_default_filename();
 
-# Send headers to browser to activate mime loading
```
dregad: Why wasn't this comment copied to the new API function ?

csv_export.php
```diff
@@ -68,16 +68,8 @@
...
-
-# Make sure that IE can download the attachments under https.
-header( 'Pragma: public' );
-
-header( 'Content-Type: text/csv; name=' . urlencode( file_clean_name( $t_filename ) ) );
-header( 'Content-Transfer-Encoding: BASE64;' );
-
-# Added Quotes (") around file name.
-header( 'Content-Disposition: attachment; filename="' . urlencode( file_clean_name( $t_filename ) ) . '"' );
+csv_emit_headers( $t_filename );
```
dregad: Now we echo the BOM twice if $g_csv_add_bom = ON (see a few lines below)

core/csv_api.php
dregad: Actually the function would make sense, if we were to take *$g_csv_add_bom* config into consideration. Is it intentional that you are not checking it ?

csv_export.php
vboctor: Didn't notice that.  I would like to remove this configuration option, add it to obsolete configs and always have Excel/CSV files have this byte order marker.  I can't see why we wouldn't always have it.

atrol: I am quite sure there are apps that are not able to deal with a BOM in CVS.
There is no clear specification for CSV https://tools.ietf.org/html/rfc4180
A check box in UI to add BOM could be a solution for it.

vboctor: Did you ever use an app that provides such check box?  I doubt most users understand what this is about.  How about we make the config ON by default?

vboctor: Or maybe we can have "Export to CSV" and "Export to CSV (UTF-8)".  I checked checked Excel save-as functionality and they have CSV and CSV (UTF-16).

If we do that, then we remove the configuration option.

atrol: > I doubt most users understand what this is about

Sure, but they will have the option to try if there is a problem.

> Or maybe we can have "Export to CSV" and "Export to CSV (UTF-8)"

Sounds good as the end user can decide.

csv_export.php
vboctor: Does this really add value?  I'm not sure this is useful.  I actually don't understand the point it is making.

core/csv_api.php
vboctor: Moved to a constant and removed the configuration option.

core/csv_api.php
vboctor: Done.

vboctor:
@dregad for the headers, I would rather keep it in csv_api.  Now I have a csv_start( $p_filename ) method which emits the headers + the bom.  I would rather most of the code to be in the specific API.

csv_export.php
vboctor: I did all the work suggested with the exception of having an option to export ASCII.  Should we wait and see the feedback before we add this complexity and extra strings?

csv_export.php
dregad: > Does this really add value?

I'm not sure, I was just curious as to why it got removed.

I just went through history, this comment is there since the CSV export feature was added by Ken in 2002... 49659dd8e1d824966a3c8b1c5050677fe3824300, and TBH I don't really see the point either.

csv_export.php
dregad: I think that nowadays UTF8 encoding is pretty much ubiquitous, and the risk of the receiving end not being able to deal with it is quite low. 

+1 for not adding extra complexity until someone screams for it.

dregad:
It just seemed cleaner to me to have the headers function in http API, and have` csv_start()` call it, but I don't feel strongly about it. It wouldn't be the only case where we issue headers outside of http API anyway (e.g. file download)

dregad:
:+1:

PR #731, opened 2016-03-04
CSS clean up

cproensa:
Remove redundant selectors


vboctor:
@syncguru can you please review this change.

dregad:
This change is harmless, because the removed selector `table#manage-overview-table > tr:nth-child` is covered by the generic `div.table-container > table > tr:nth-child`.

dregad:
Fixed in f8fd3f7fd11bf4a6ab98e97e73d2d22864209559

PR #732, opened 2016-03-06
Dont use serialization for Session variables

cproensa:
There is no need to serialize data inserted in
$_SESSION, as php does it at the end of script

This is a performance hit when a session key
is used multiple times (eg: 'form_security_tokens')

Fixes #20142, #18016

### DIFF
--- a/core/session_api.php
+++ b/core/session_api.php
@@ -135,7 +135,7 @@ class MantisPHPSession extends MantisSession {
 	 */
 	function get( $p_name, $p_default = null ) {
 		if( isset( $_SESSION[$this->key][$p_name] ) ) {
-			return unserialize( $_SESSION[$this->key][$p_name] );
+			return $_SESSION[$this->key][$p_name];
 		}
 
 		if( func_num_args() > 1 ) {
@@ -153,7 +153,7 @@ class MantisPHPSession extends MantisSession {
 	 * @return void
 	 */
 	function set( $p_name, $p_value ) {
-		$_SESSION[$this->key][$p_name] = serialize( $p_value );
+		$_SESSION[$this->key][$p_name] = $p_value;
 	}
 
 	/**
--- a/core/session_api.php
+++ b/core/session_api.php
@@ -100,7 +100,7 @@ class MantisPHPSession extends MantisSession {
 	function __construct( $p_session_id = null ) {
 		global $g_cookie_secure_flag_enabled;
 
-		$this->key = hash( 'whirlpool', 'session_key' . config_get_global( 'crypto_master_salt' ), false );
+		$this->key = hash( 'whirlpool', 'session_key_v_2' . config_get_global( 'crypto_master_salt' ), false );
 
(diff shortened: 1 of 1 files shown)


vboctor:
What is the lifetime of these sessions?  Should we worry about the updated code getting sessions that has data that was serialized by the old code after upgrade?

cproensa:
after changing the code, i get errors only in the active browser session.
closing the page, and open again it works fine, even if the auto-login is enabled

vboctor:
Should we version the session or do some trick to avoid such errors?  I expect that heavy users of MantisBT will typically have a browser tab or two open most of the time.  Thoughts?

cproensa:
I think to force logout of all users should be enough?
once they reauthenticate everything seems to work fine.
Also, i think there is not important data stored in session, as its stored in db tokens table

dregad:
Not sure what the status for this PR is - can it be merged as it is, or is there additional code needed to handle the force logout ?

cproensa:
@dregad 
I may try to implement a upgrade step to handle mass log out. Lets see if i do it right, as its the first time.

grangeway:
that would break people's rss links etc as that users cookie as key, remembered passwords etc - seems a bit of a heavy handed way to "expire" a session. You'd probably be better adding a new session variable (version) into and then letting the existing session_validate routine check for it, if a sesion exists, and the variable doesn't then you know it's old version. After all, a) that might be useful in future, b) sessions probably have a timeline of <24 hours in most cases - so only a subset of users would hit the issue.

cproensa:
Thanks @grangeway 
I was hoping feedback on this.
Your suggestion is the plan B i had thought, and I agree that requires adding compelxity (manage versions)

cproensa:
btw, do we have a proper way to expire a user session, meaning: force an individual user to log in again?
That seems a potentially useful feature

grangeway:
"btw, do we have a proper way to expire a user session, meaning: force an individual user to log in again?" -> you can reset a password... and given only use case I can think of you would want to manually reset a session would be likely require a password change...

In terms of session_clean at user login - given how much use of sessions are actually made - i'd be inclined to think not. However without looking at the original code - i'd do it on the logon_page.php and not login.php. I think you'll find that both initiate session, and that in the case of login.php that will set secure_session (i.e. the ip stuff). If i remember the api correctly, you'd effectively be doing php's session_start(), session_destroy() and then writing stuff back to the session.  I might be wrong but I think there's a chance it'll do one of either a) moan that session_start needs to be called, b) not save the session data or c) restore the original  data - as opposed to what you expect.

cproensa:
OK. I discarded the upgrade step.
The main point here is to avoid the use of existing data in session store, it there is any.
To avoid any further modifications in session handling, i think that changing the "session key" used to index the $_SESSION array, is enough. This provides an implicit versioning for session format, and ignores gracefully old data if there is any.

dregad:
Merged in 9e532d6b264d4c4c119acc44c0b5112569d03600

PR #733, opened 2016-03-07
Http header error method

cproensa:
Replacing PR #660 

Fixes #0013048

### DIFF
--- a/core/error_api.php
+++ b/core/error_api.php
@@ -216,6 +216,13 @@ function error_handler( $p_type, $p_error, $p_file, $p_line, array $p_context ) 
 					ob_clean();
 				}
 
+				# If HTML error output was disabled, set an error header and stop
+				if( defined( 'DISABLE_INLINE_ERROR_REPORTING' ) ) {
+					# @TODO Have a mapping for mantis error codes to appropiate HTTP error codes
+					header( ' ', true, 400 );
+					exit(1);
+				}
+
 				# don't send the page header information if it has already been sent
 				if( $g_error_send_page_header ) {
 					if( $t_html_api ) {
--- a/js/common.js
+++ b/js/common.js
@@ -85,6 +85,9 @@ $(document).ready( function() {
 			context: $('#' + targetID),
 			success: function(html) {
 				$(this).html(html);
+			},
+			error: function(obj,status,error) {
+				$(this).html('<span class="error-msg">' + status + ': ' + error + '</span>');
 			}
 		});
 	});
--- a/return_dynamic_filters.php
+++ b/return_dynamic_filters.php
@@ -36,6 +36,9 @@
  * @uses helper_api.php
  */
 
+# Prevent output of HTML in the content if errors occur
+define( 'DISABLE_INLINE_ERROR_REPORTING', true );
+
 require_once( 'core.php' );
 require_api( 'authentication_api.php' );
 require_api( 'compress_api.php' );
@@ -49,7 +52,9 @@ require_api( 'filter_constants_inc.php' );
 require_api( 'gpc_api.php' );
 require_api( 'helper_api.php' );
 
-auth_ensure_user_authenticated();
+if( !auth_is_user_authenticated() ) {
+	trigger_error( ERROR_ACCESS_DENIED, ERROR );
+}
 
 compress_enable();
 


config_defaults_inc.php
```diff
@@ -4140,6 +4140,8 @@
  * - DISPLAY_ERROR_INLINE  Display a one line error and continue execution.
  * - DISPLAY_ERROR_NONE    Suppress the error (no display). This is the default
  *                         behavior for unspecified errors constants.
+ * - DISPLAY_ERROR_HEADERS Stop and return an HTTP error header. No output
```
vboctor: Should we have a define or an API that when called at the beginning of a script page, it directs how the error is handled, without changing the valid values for this?  Having this option here means that users can use this globally.

core/error_api.php
```diff
@@ -284,6 +284,12 @@ function error_handler( $p_type, $p_error, $p_file, $p_line, array $p_context )
 				$g_error_handled = true;
 				break;
 
+			case DISPLAY_ERROR_HEADERS:
+				# @TODO Have a mapping for mantis error codes to appropiate HTTP error codes
```
vboctor: type: appropriate

core/error_api.php
```diff
@@ -284,6 +284,12 @@ function error_handler( $p_type, $p_error, $p_file, $p_line, array $p_context )
...
+				header( ' ', true, 500 );
+				exit(1);
+				break;
```
vboctor: Why is there a break here?

core/error_api.php
```diff
@@ -548,3 +554,12 @@ function error_alternate_class() {
 		return 'class="row-2"';
 	}
 }
+
+/**
+ * Shorthand function to swith error output method to use HTTP headers
```
vboctor: typo: switch

core/error_api.php
```diff
@@ -284,6 +284,12 @@ function error_handler( $p_type, $p_error, $p_file, $p_line, array $p_context )
 				$g_error_handled = true;
 				break;
 
+			case DISPLAY_ERROR_HEADERS:
+				# @TODO Have a mapping for mantis error codes to appropiate HTTP error codes
+				header( ' ', true, 500 );
```
vboctor: Shouldn't we return error_string( $p_error )?  Also for code, I would return the code for bad request (400) except in the case where the error is access denied, in which case I would return a forbidden (403) error code.

config_defaults_inc.php
vboctor: See install.php for example where it sets up expected behaviors for error handling, timeout, and maintenance mode.

    error_reporting( E_ALL );
    @set_time_limit( 0 );

    # Load the MantisDB core in maintenance mode. This mode will assume that
    # config/config_inc.php hasn't been specified. Thus the database will not be opened
    # and plugins will not be loaded.
    define( 'MANTIS_MAINTENANCE_MODE', true );

cproensa:
> Should we have a define or an API that when called at the beginning of a script page, it directs how the error is handled, without changing the valid values for this? Having this option here means that users can use this globally.

That makes sense
i'd say a "supress output" mode. Where some behaviour is overrriden:
- Error handler: 
  - ignore error display methods that write HTML output
  - if error is type "halt", stop execution with an HTTP error header
- Log api:
  - ignore log display method for "screen"

switching this "supress error" mode may be through a define, or a api function (if it needs to change other options, termorary, or permanent)

Additionally, i'd add functionality to send error to the logging api, wether:
- having a log level for "errors"
- having an error display method "system log"

vboctor:
The model makes sense.  I personally like the method approach over define.

Don't understand what you mean by "Additionally, i'd add functionality to send error to the logging api, wether: ...".

cproensa:
> Don't understand what you mean by "Additionally, i'd add functionality to send error to the logging api,

Afaik, errors are not written into the log system, unless explicitly doing an "event_log" call
I think it is useful to have the errors logged too. So on the error handler, a event_log call could be used to log the detailed error, (besides the current display method used)
For controlling this kind of log messages, a new log level for "ERRORS", can be added.

And back to having a flag for "no html output" scripts, any error can be then registered by enabling the ERROR log level, and log destination to file.

cproensa:
Actually, @dregad implemented already a define DISABLE_INLINE_ERROR_REPORTING
ab91f9eb84d7947bc7f318fa4adfa7c7ed4a1d1a

cproensa:
Rewrited the PR using the already existant define: DISABLE_INLINE_ERROR_REPORTING

I'd move this constant to be a global variable, for this reason:
as explained by Damien, the usage is "by defining a constant prior to including core.php"
i find some cases where i want to disable error messages on ly through a specific part of code (im thinking of bug action group, for example). Seems like a define can't be used that way.

Also:
> Shouldn't we return error_string( $p_error )? Also for code, I would return the code for bad request (400) except in the case where the error is access denied, in which case I would return a forbidden (403)

I'm not sure if you can return any string as part of the header. I understand that the text associated to the error must be the standard one, that why i simply used the error code, with a dummy string. PHP fills that text by itself

dregad:
I'm OK with the proposed approach to use a global variable rather than a define

dregad:
@cproensa let me know if you want me to merge this as-is, and deal with the conversion from define to global as a separate issue/PR.

cproensa:
I'm OK with this as is.
Thanks @dregad

dregad:
Just realized there are merge conflicts. Could you kindly rebase on top of master to get rid of them ? Thanks.

cproensa:
Ok, it's rebased now.
(I didn't get any conflicts though)

dregad:
> (I didn't get any conflicts though)

Weird, Github complained that the merge could not be performed automatically... (I didn't actually try to merge)

PR #734, opened 2016-03-10
Avatar plugin support related fixes

vboctor:
- Don't show avatars when they are disabled.
- Avatar threshold should check user for which to display avatar and not logged in user.
- If avatar is off and a user doesn't match threshold, then show default avatar.  This way we don't end up with a mix of no image and image.
- Remove gravatar specific settings from core config.
- Gravatar plugin to use its own default_config instead of show avatar.

Fixes #20641
Fixes #20642

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -1067,19 +1067,8 @@ $g_differentiate_duplicates = OFF;
 $g_sort_by_last_name = OFF;
 
 /**
- * Show user avatar
- *
- * The config can be either set to OFF (avatars disabled) or set to a string
- * defining the default avatar to be used when none is associated with the
- * user's email. Valid values:
- * - OFF (default)
- * - ON (equivalent to 'identicon')
- * - One of Gravatar's defaults (mm, identicon, monsterid, wavatar, retro)
- *   @link http://en.gravatar.com/site/implement/images/
- * - An URL to the default image to be used (for example,
- *   "http:/path/to/unknown.jpg" or "%path%images/no_avatar.png")
- *
- * @global integer|string $g_show_avatar
+ * Show user avatars
+ * @global integer $g_show_avatar
  * @see $g_show_avatar_threshold
  */
 $g_show_avatar = OFF;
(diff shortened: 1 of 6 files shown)


dregad:
Tested OK:

- $g_show_avatar = OFF; (regardless of whether Gravatar plugin is installed)  
  **OK** no avatars displayed
- $g_show_avatar = ON; Gravatar plugin *not* installed
  **OK** default avatar displayed
- $g_show_avatar = ON; Gravatar plugin *installed*
  **OK** gravatar displayed
- $g_show_avatar = DEVELOPER; Gravatar plugin *installed*; login with *reporter* user
  **OK** no avatars displayed 

In the last test case, we could debate whether the avatars should be completely hidden (as it is now) or the default one shown. Since the config is *not* OFF, it can be implied that the Administrator actually *wants* to use avatars, so the question is: does being below the threshold mean "do not use avatars at all" or "restrict viewing them to authorized users". I'm fine either way, the latter probably gives more consistency to the UI.

dregad:
Oh, and +1

cproensa:
> so the question is: does being below the threshold mean "do not use avatars at all" or "restrict viewing them to authorized users".

or:  only show avatar of users above that threshold. (like, only show the avatar of developers)

The other way: "restrict viewing them to authorized users", Is it any good (eg, me as a developer), to see the avatars of reporters, without they knowing i can see them?

dregad:
> or: only show avatar of users above that threshold. (like, only show the avatar of developers)

Re-reading @vboctor's [documentation update submitted as part of this PR](https://github.com/mantisbt/mantisbt/pull/734/files#diff-2b5232793167963cfb322f0a8ec9622bR205):

*The threshold of users for which MantisBT should show the avatar (default DEVELOPER).
Note that the threshold is related to the user for whom the avatar is being shown,
rather than the user who is currently logged in.*

I now realize that I was actually mixing 2 settings: $g_show_avatar and $g_show_avatar_threshold. The latter behaves exactly as documented, which also matches what @cproensa mentioned. Not sure what I had been smoking when I ran my tests :wink: 

My last test case's result is as follows:

- $g_show_avatar = DEVELOPER is actually equivalent to having it set to *ON*.

New test case:

- $g_show_avatar_threshold = DEVELOPER; Gravatar plugin installed; login with any user  
  **OK** avatars are displayed for DEVELOPER and above, regardless of logged-in user's access level; users below DEVELOPER are shown with the default avatar.

dregad:
Merged.

PR #735, opened 2016-03-14
New password activation

cproensa:
Manage the password reset hash as a token.
Refactor verify.php to be a not-logged-in page, like login_page.php, so the only action the user can do is change the password, and not navigate into the site.
If the user does not change the password and quits tue page, the activation token is still valid until the change is effectively done (or token timeout)

Fix #0020686, #0006009

### DIFF
--- a/account_page.php
+++ b/account_page.php
@@ -171,13 +171,13 @@ if( $t_force_pw_reset ) {
 			} else {
 			?>
 			<div class="field-container">
-				<label for="password" <?php echo $t_force_pw_reset_html ?>><span><?php echo lang_get( 'current_password' ) ?></span></label>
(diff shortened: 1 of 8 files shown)


dregad:
Tested the proposed code OK.
+1

cproensa:
@dregad
I have some doubts if it works fine with LDAP authentication, where the user activation does not use a user provided password. I cannot test with a LDAP now, so i'd appreciate feedback on this

dregad:
I don't have a live LDAP setup either, sorry.

cproensa:
I have tested LDAP with the $g_ldap_simulation_file_path option
Seems OK, as i made the adjustments to not show the password form if the user is not able to change, and show the proper warning. However, the reset password option is not available in the first place, when using LDAP mode.

dregad:
Merged in d7b8d33e79250e67888e2d6e82ee1e5f2bfd091f, thank you @cproensa

PR #739, opened 2016-03-17
Support default value for due date.

lukelee1124:
1. Add default due date value for newly submitted issues.
2. Access check for CSV/Excel export when displaying due date, and handling null due date in CSV/Excel.
3. Update manual for due date configs.

Fixes [#20710](https://www.mantisbt.org/bugs/view.php?id=20710)
Fixes [#20711](https://www.mantisbt.org/bugs/view.php?id=20711)
Fixes [#20712](https://www.mantisbt.org/bugs/view.php?id=20712)

### DIFF
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -172,7 +172,7 @@ if( $f_master_bug_id > 0 ) {
 	$f_steps_to_reproduce	= gpc_get_string( 'steps_to_reproduce', config_get( 'default_bug_steps_to_reproduce' ) );
 	$f_additional_info		= gpc_get_string( 'additional_info', config_get( 'default_bug_additional_info' ) );
 	$f_view_state			= gpc_get_int( 'view_state', (int)config_get( 'default_bug_view_status' ) );
-	$f_due_date				= gpc_get_string( 'due_date', strtotime( config_get( 'due_date_default' ) ) );
+	$f_due_date				= gpc_get_string( 'due_date', date_strtotime( config_get( 'due_date_default' ) ) );
 
 	if( $f_due_date == '' ) {
 		$f_due_date = date_get_null();
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -172,7 +172,7 @@ if( $f_master_bug_id > 0 ) {
 	$f_steps_to_reproduce	= gpc_get_string( 'steps_to_reproduce', config_get( 'default_bug_steps_to_reproduce' ) );
 	$f_additional_info		= gpc_get_string( 'additional_info', config_get( 'default_bug_additional_info' ) );
 	$f_view_state			= gpc_get_int( 'view_state', (int)config_get( 'default_bug_view_status' ) );
-	$f_due_date				= gpc_get_string( 'due_date', '' );
+	$f_due_date				= gpc_get_string( 'due_date', strtotime( config_get( 'due_date_default' ) ) );
 
 	if( $f_due_date == '' ) {
 		$f_due_date = date_get_null();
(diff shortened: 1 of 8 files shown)


bug_report_page.php
```diff
@@ -172,7 +172,7 @@
 	$f_steps_to_reproduce	= gpc_get_string( 'steps_to_reproduce', config_get( 'default_bug_steps_to_reproduce' ) );
 	$f_additional_info		= gpc_get_string( 'additional_info', config_get( 'default_bug_additional_info' ) );
 	$f_view_state			= gpc_get_int( 'view_state', (int)config_get( 'default_bug_view_status' ) );
-	$f_due_date				= gpc_get_string( 'due_date', '' );
+	$f_due_date				= gpc_get_string( 'due_date', strtotime( config_get( 'due_date_default' ) ) );
```
vboctor: strtotime() won't handle the case where due_date_default is set to empty string properly.  Please add date_strtotime() in core/date_api.php which does the following:
- If input is empty string, return date_get_null()
- otherwise return strtotime() of the input value.

Then you can call it here.

docbook/Admin_Guide/en-US/config/duedate.xml
```diff
@@ -0,0 +1,34 @@
...
+				</para>
+			</listitem>
+		</varlistentry>
+		<varlistentry>
+			<term>$g_due_date_default</term>
+			<listitem>
+				<para>Default due date value for newly submitted issues:
+ 					Empty string for no due date set.
+ 					Related date that is accepted by <ulink url="http://php.net/manual/en/function.strtotime.php">strtotime()</ulink>, 
+ 					e.g. 'today' or '+2 days'.
```
vboctor: Add the following sentence at the end:
Default is ''.

vboctor:
@lukelee1124 generally looks good.

The bug_report.php page needs to be update to default the due_date field properly when the field is not visible on the form and hence the value is not passed in.  Currently it is always defaulted to a null date instead of the new default value.

See https://github.com/mantisbt/mantisbt/blob/master/bug_report.php#L130

vboctor:
Please add the following lines to the commit and pull request description:
Fixes #20710
Fixes #20711
Fixes #20712

config_defaults_inc.php
```diff
@@ -4066,6 +4066,14 @@
  */
 $g_due_date_view_threshold = NOBODY;
 
+/**
+ * Default due date value for newly submitted issues:
+ * Empty string for no due date set.
+ * Related date that is accepted by strtotime (http://php.net/manual/en/function.strtotime.php), e.g. 'today' or '+2 days'.
+ * @global $g_due_date_default
```
atrol:
```
@global string $g_due_date_default
```

config_defaults_inc.php
```diff
@@ -4397,6 +4405,7 @@
 	'download_attachments_threshold',
 	'due_date_update_threshold',
 	'due_date_view_threshold',
+	'due_date_default',
```
atrol: The list is sorted, please add before 'due_date_update_threshold'

core/date_api.php
```diff
@@ -68,6 +70,19 @@ function date_get_null() {
 }
 
 /**
+ * get timestamp when date is accepted by strtotime (http://php.net/manual/en/function.strtotime.php)
+ * @param $p_date
```
vboctor: string $p_date and add documentation for the parameter.

core/date_api.php
```diff
@@ -68,6 +70,19 @@ function date_get_null() {
...
+ * @return int a timestamp on success, empty when $p_date is blank. 
+ * @access public
+ */
+function date_strtotime( $p_date ) {
+	if ( is_blank( $p_date ) ) {
```
vboctor: "if (" -> "if("

dregad:
Thanks for your contribution. I made minor changes to your commits, added a couple corrections to take into consideration @vboctor's latest comments and merged the changes in 84481739cc443be5d29b0c652cd1debc74273bd9

PR #740, opened 2016-03-17
Gravatar CSP logic overwrites any existing CSP headers.

bkraul:
The current csp logic in Gravatar replaces any existing content security policy. This includes the default policy set in core/http_api.php, and also any CSP declared in $g_custom_headers (permissible per the Admin Guide).

PR retrieves existing headers, tries to find any existing CSP header, and add the necessary gravatar whitelist without affecting the existing policy.

### DIFF
--- a/plugins/Gravatar/Gravatar.php
+++ b/plugins/Gravatar/Gravatar.php
@@ -118,9 +118,35 @@ class GravatarPlugin extends MantisPlugin {
 	function csp_headers() {
 		# Policy for images: Allow gravatar URL
 		if( config_get_global( 'show_avatar' ) !== OFF ) {
-			# Set CSP header
-			header( "Content-Security-Policy: img-src 'self' " .
-				self::getAvatarUrl() );
+			# Retrieve headers and attempt to find a CSP header entry.
+			$t_headers = headers_list();
+			foreach( $t_headers as $t_key => $t_item ) {
+				$t_pos = stripos( $t_item, 'Content-Security-Policy:' );
+				if ( $t_pos === 0 ) {
+					$t_csp = $t_item;
+					break;
+				}
+			}
+			
+			if ( $t_csp ) {
+				# Found a CSP header.
+				if ( stristr( $t_csp, 'img-src' ) ) {
+					# If we have an img-src policy already declared, inject.
+					$t_csp = preg_replace( '/(img\-src)(.*?(?=;)|(.*))/is', '$1$2 ' . 
+						self::getAvatarUrl(), $t_csp );
+				} else {
+					# Otherwise append.
+					$t_csp = $t_csp . "; img-src 'self' " .
+						self::getAvatarUrl();
+				}
+			} else {
+				# No current custom CSP header found, replace.
+				$t_csp = "Content-Security-Policy: img-src 'self' " .
+					self::getAvatarUrl();
+			}
+
+			# Set CSP header.
+			header( $t_csp );
 		}
 	}
 


vboctor:
Thanks @bkraul for the fix.

I just wonder the code should be simplified by leveraging an event or an API to the core.  In my intermediate commits for the gravatar pull request, I had an event that enabled plugins to easily add CSP segments to the CSP header.  I wonder if we should go back to this.

See https://github.com/mantisbt/mantisbt/pull/711/commits/f28e25ba535052cae9ff0d1cbfd3d75753895fab#diff-c058fb329046e311791b26f57ef4743bR155

bkraul:
Thanks @vboctor. After looking at the event placement, it would greatly benefit, because it would eliminate the need to scan all headers to find the CSP header. However, unless the elements in the CSP header are made into the likes of a modifiable array, we would still to run some of the code to make sure the existing CSP will not be overwritten by a new one.

A rough example of this could be:
```php
# global configurable array (not all keys needed).
$g_csp_policy = array(
	"style-src" => "'self' css.example.com",
	"img-src" => "'self' img.example.com",
	"connect-src" => "'self'",
	"font-src" => "font.example.com",
	"object-src" => "'self'",
	"media-src" => "media.example.com",
	"frame-src" => "'self'",
	"sandbox" => "allow-forms allow-scripts",
	"report-uri" => "/some-report-uri",
	"child-src" => "'self'",
	"form-action" => "'self'",
	"frame-ancestors" => "'none'",
	"plugin-type" => "application/pdf",
);

# -- Gravatar plugin ---
# Add Gravatar URL to img-src (to be performed during EVENT_LAYOUT_CSP_RECORDS hook)
if ( array_key_exists( 'img-src', $g_csp_policy ) ) {
	# img-src exists, inject.
	$g_csp_policy[ 'img-src' ] = preg_replace( '/(img\-src)(.*?(?=;)|(.*))/is', '$1$2 ' . "http://www.gravatar.com", $t_csp );	
} else {
	# no img-src exists. add.
	$g_csp_policy[ 'img-src' ] = "'self' http://www.gravatar.com";
}

# Build CSP from gobal array (applied during core CSP definition http_api.php)
# listen to EVENT_LAYOUT_CSP_RECORDS here.
$t_csp = 'Content-Security-Policy: ' . urldecode( http_build_query( $g_csp_policy, '', '; ' ) );

# Output CSP.
echo $t_csp;
)
```
I figure it would take more to set up this logic, which would be more re-usable, but I figure both approaches should work.

dregad:
I like the approach of having a global array for CSP settings, it provides flexibility, and lets user selectively override specific elements or even disable CSP entirely within config_inc.php. Not sure if the array elements should be strings as in your sample code above, or arrays.

The event hook (type chain would be best I think) should accept as parameter an array with the same structure as the global config, allowing individual CSP elements to be appended.

I'm not sure about the use of http_build_query(), it would produce a string like `img-src='self' example.com`, it should be `img-src 'self' example.com` (no `=` sign) - a *foreach* loop is probably more appropriate.

bkraul:
@dregad. Thanks for pointing out the overlooked flaw in build_query. I was trying to come up with a rough draft that would use loops as little as possible, as a proof of concept. I agree that (I suppose you mean nested?) arrays would be better in the long run.

I figure by this point (per my own experiences with CSP issues with my BBCode+ plugin) CSP is important enough to be developed into a comprehensive, configurable structure.

bkraul:
I have amended the snippet to correct the issue pointed out by @dregad. Hopefully some of that is ultimately usable.

dregad:
>  (I suppose you mean nested?) arrays

TBH, I have not thought this all the way through.

We could either have a structure like the one you proposed above, i.e. an array of strings with CSP elements as keys (adding new values to an element by appending to the string) or, instead of  strings, define the elements as arrays (I guess that's what you mean by *nested arrays*)

``` php
$g_csp_policy = array(
    "style-src" => array( 'self', 'css.example.com' ),
    "img-src" => array( 'self', 'img.example.com' ),
    # etc.
);

# Append new element
$g_csp_policy['img-src'][] = 'http://www.gravatar.com';

# Build CSP
$t_csp = 'Content-Security-Policy: ';
foreach ( $g_csp_policy as $t_key => $t_elements) {
    $t_csp .= $t_key . " " . implode( ' ', $t_elements ) . ';';
}
```

bkraul:
Guys I am sorry to ask through here but I believe the messaging feature of GitHub went away at some point. I am trying to find what is the proper way of keeping my fork synced with upstream in such away that any of my open pull requests wont be adversely affected.

Everything I am seeing says I am supposed to do a ```git merge upstream/master```, but that generates a separate commit message (usually "Merge remote-tracking branch 'upstream/master'"). I have also seen this affect the CI process in my open pull requests.

Am I supposed to ```git rebase upstream/master``` instead, to keep a cleaner commit history?

dregad:
@bkraul no worries. You might want to use our [mailing list](http://mantisbt.org/mailinglists.php) or [Gitter](https://gitter.im/mantisbt/mantisbt) for such questions in the future.

Here's the basics (assuming *upstream* is our repo's remote, and *origin* is your fork's):
- if you never commit to it, keeping your own *master* branch in sync with ours is as simple as executing `git pull upstream` (it will do a fast-forward merge)
- always use a dedicated branch for your patches or features, never the *master* branch
- your feature branches should always be based on our latest master
- to update your pull request, just add commits to the feature branch
- if Github tells you there is a merge conflict, our preferred way to resolve them is by *rebasing* your feature branch on top of master
```
git fetch upstream
# rebase branch, resolve conflicts 
git rebase upstream/master feature
# you need to force push to update the pull request
git push origin feature -f
```

Hope this helps. Ping me on Gitter if you need further guidance.

bkraul:
Guys I'm going to begin the attempt to implement what we discussed. Hopefully the result will be an acceptable PR. As such, I will be closing this PR, resetting my fork to upstream in order to follow @dregad's git practice recommendations. Thanks!

PR #741, opened 2016-03-19
Improved checkin script to allow for more elaborate regular expressions

gene-pavlovsky:
Added new config setting: source_control_issue_regexp
Checkin script now matches this regexp within matches of
source_control_regexp and source_control_fixed_regexp.
This allows those regexps to match a more elaborate pattern,
e.g. a sentence mentioning more than one issue id.
The source_control_issue_regexp is then used to extract
issue ids from those matches.

PS: Should not break existing setups.

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2679,6 +2679,29 @@
 	 * @global string $g_source_control_fixed_regexp
 	 */
 	$g_source_control_fixed_regexp = "%source_control_regexp%";
+	
+	/**
+	 * Regular expression used to detect issue ids within $g_source_control_regexp and
+	 * $g_source_control_fixed_regexp matches. This two-step matching makes it possible
+	 * to match more elaborate patterns.
+	 *
+	 * Example:
+	 *   $g_source_control_regexp = '/\b (?: feature | bug | issue)s? (?: \s+ | : \s*) (\#\d+ (?: (?: \s* , \s* | \s+ and \s+ | \s+) \#\d+)*) \b/xi';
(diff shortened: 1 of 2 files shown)


dregad:
Hello Gene,

Many thanks for your contribution.

Unfortunately, the built-in source control integration is deprecated in 1.2, and has been entirely [removed in upcoming version 1.3](https://www.mantisbt.org/bugs/view.php?id=11732).

Despite its value, we will therefore not merge your pull request in 1.2.x branch, which only accepts bug fixes.

I strongly suggest you switch to using the [Source Integration plugin](https://github.com/mantisbt-plugins/source-integration) instead of relying on the obsolete built-in feature.

gene-pavlovsky:
Hi,
I've noticed it's removal in the master branch, and suspected you might not accept the PR.
I would really love to use the Source Integration plugin, which I've already reviewed, but none of the available plug-ins are easy for me to install.
I'm developing a project all by myself, and have (on my laptop running Windows 7) a local git repository, an nginx webserver (with php). Recently I've installed mantis for issue (feature) tracking. Since nginx doesn't have CGI support, and FastCGI wrapper is not available on Windows, I can't seem to install either cgit or gitweb. Well I actually managed to make gitweb sorta working (found a perl script that runs other perl scripts as FastCGI, it didn't work out of the box and had to be hacked), but it was pretty slow and unreliable.
I was not inclined to install Apache with CGI as I have a strong dislike for it, therefore I decided to write my own git post-commit hook and use the checkin.php script. This will have to do for now but I'm not sure what to do in the future.
Are there any plans to make a Source Integration plugin which supports git using the command-line git to access a local repository?
Thanks.

PR #742, opened 2016-03-20
Tag with description are attached incorrectly

vboctor:
The tag dropdown shows the "tag - description", selecting it adds
"tag - description" as a new tag rather than using "tag".

Fixes #20724

### DIFF
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -405,9 +405,6 @@ function print_tag_option_list( $p_bug_id = 0 ) {
 	echo '<option value="0">', string_html_specialchars( lang_get( 'tag_existing' ) ), '</option>';
 	foreach ( $t_rows as $t_row ) {
 		$t_string = $t_row['name'];
-		if( !empty( $t_row['description'] ) ) {
-			$t_string .= ' - ' . utf8_substr( $t_row['description'], 0, 20 );
-		}
 		echo '<option value="', $t_row['id'], '" title="', string_attribute( $t_row['name'] ), '">', string_attribute( $t_string ), '</option>';
 	}
 }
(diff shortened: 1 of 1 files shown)


atrol:
I would like to see another approach for it, see https://www.mantisbt.org/bugs/view.php?id=20724#c52810

vboctor:
@atrol I removed the description from the dropdown.  The description was already being showed when hovering over the tags, so I didn't need to make changes there.

atrol:
> The description was already being showed when hovering over the tags, so I didn't need to make changes there.

@vboctor seems I was not clear when describing the issue:
We have a tag with name "documentation-needed" at https://www.mantisbt.org/bugs
Description is: "Used to tag issues which need to be documented, whether they are features which need to be explained or bugs which have workaround."

When when hovering over the tags **in the dropdown list** I get just the name in the bubble text. Thus I see two times the names instead of seeing the name in the list and the description  as bubble text (certainly a good help when adding tags).

vboctor:
@atrol I fixed the tooltip.

core/print_api.php
```diff
@@ -405,10 +405,7 @@ function print_tag_option_list( $p_bug_id = 0 ) {
 	echo '<option value="0">', string_html_specialchars( lang_get( 'tag_existing' ) ), '</option>';
 	foreach ( $t_rows as $t_row ) {
 		$t_string = $t_row['name'];
-		if( !empty( $t_row['description'] ) ) {
-			$t_string .= ' - ' . utf8_substr( $t_row['description'], 0, 20 );
-		}
-		echo '<option value="', $t_row['id'], '" title="', string_attribute( $t_row['name'] ), '">', string_attribute( $t_string ), '</option>';
+		echo '<option value="', $t_row['id'], '" title="', string_attribute( $t_row['description'] ), '">', string_attribute( $t_string ), '</option>';
```
atrol: Assignment of $t_string in line 407 is no longer needed.
You could use now $t_row['name'] in line 408.

PR #743, opened 2016-03-21
Fixes#20646:Rename javascript folder to js

lukelee1124:
Fix all references to javascript to use the js folder

Issue : https://www.mantisbt.org/bugs/view.php?id=20646


dregad:
Just noticed Github added a way to "+1" a PR...

dregad:
Just a note regarding the commit message: the description is redundant, so I would suggest making it a one-liner, as as follows:

```
Fix #20646: Rename 'javascript' folder to 'js'
```

vboctor:
Merged after changing the commit message.  Thanks @lukelee1124

PR #744, opened 2016-03-21
Fix #0020275

tgulacsi:
form_api.form_security_token returns constant 1 if p_security_token is
not null.

### DIFF
--- a/core/form_api.php
+++ b/core/form_api.php
@@ -127,7 +127,7 @@ function form_security_param( $p_form_name, $p_security_token = null ) {
 		return '';
 	}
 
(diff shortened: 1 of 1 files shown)


dregad:
You're right, not sure what I was thinking.

dregad:
Fixed in 419756478a7b154e5f95b6a71259f01a16b47ff5

dregad:
Unfortunate typo in issue reference...

tgulacsi:
Sorry :)
(I had to look three times at the  "changed the title ..." line to see the difference...(

dregad:
I said unfortunate, because it triggered a bug causing the *My View* page to error out at http://mantisbt.org/bugs :fearful:

PR #746, opened 2016-03-24
Status legend doesn't show final workflow states

dregad:
The patch for issue [#11553](https://www.mantisbt.org/bugs/view.php?id=11553) did not properly fix the problem, as the code only checked for the workflow allowing to leave a state.

We now also cover the case of entering the state.

Fixes [#20746](https://www.mantisbt.org/bugs/view.php?id=20746)

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -1390,10 +1390,21 @@ function html_status_legend( $p_display_position, $p_restrict_by_filter = false 
 	$t_workflow = config_get( 'status_enum_workflow' );
 	if( !empty( $t_workflow ) ) {
 		foreach( $t_status_array as $t_status => $t_name ) {
+			# Check if the workflow allows leaving the state
 			if( !isset( $t_workflow[$t_status] ) ) {
+				# Check if the workflow allows entering the state
+				$t_can_enter = false;
+				foreach( $t_workflow as $t_values ) {
+					if( MantisEnum::hasValue( $t_values, $t_status ) ) {
+						$t_can_enter = true;
+						break;
+					}
+				}
 
 				# drop elements that are not in the workflow
-				unset( $t_status_array[$t_status] );
+				if( !$t_can_enter ) {
+					unset( $t_status_array[$t_status] );
+				}
 			}
 		}
 	}


core/html_api.php
```diff
@@ -1390,10 +1390,22 @@ function html_status_legend( $p_display_position, $p_restrict_by_filter = false
...
+					if( MantisEnum::hasValue( $t_values, $t_status ) ) {
+						$t_can_enter = true;
+						break;
+					}
+				}
+
 				# drop elements that are not in the workflow
-				unset( $t_status_array[$t_status] );
+				if( !$t_can_enter ) {
+					unset( $t_status_array[$t_status] );
```
vboctor: It is generally not recommended to change an array within a foreach on the array.  In some languages (e.g. C#) this is banned.  Not sure about PHP, but you may consider doing this outside the loop.

core/html_api.php
```diff
@@ -1390,10 +1390,22 @@ function html_status_legend( $p_display_position, $p_restrict_by_filter = false
 	$t_workflow = config_get( 'status_enum_workflow' );
 	if( !empty( $t_workflow ) ) {
 		foreach( $t_status_array as $t_status => $t_name ) {
+			# Check if the workflow allows leaving the state
 			if( !isset( $t_workflow[$t_status] ) ) {
```
vboctor: This seems like an extra blank line.

dregad: You're right, but it was already there before. Anyway I'll remove it.

core/html_api.php
dregad: Not saying this is right, but I've always seen this done that way in PHP. As far as I know, this piece of code has been working just fine like this since its introduction 2008 (d953981e59743ddf6b6ccbc9c3838353ee25b648). 

Moving it out of the loop would require setting up a temp array storing statuses to unset, and iterating over it after the main loop. This adds complexity to the code, without immediate benefit (if it ain't broke...)

Are you aware of a case where this logic causes issues ?

dregad: Out of curiosity, I setup a [quick test ](http://ideone.com/re3qeN) which demonstrates that unsetting an array element within the foreach has no impact on the loop variables themselves, although a PHP notice is triggered (as expected) when the accessing the array element that was unset. When iterating by reference, the removed value is not even processed.

vboctor: No strong opinion here.  Feel free to checkin.

PR #749, opened 2016-04-01
Allow administrators to impersonate users

vboctor:
There are multiple scenarios where user impersonation is a useful feature. For example:

1. When an administrator is troubleshooting an error reported by one of the users.
2. When an administrator wants to verify what issues/projects are and are not visible to specific user.
3. When a user creates an (service) account and wants to create an API key for such account.

Any user with the access of managing users, should be able to do the above. Hence,
instead of specifically checking for "administrator" access level, use the config
option 'manage_user_threshold'.

Fixes #20772

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2764,6 +2764,12 @@ $g_project_user_threshold = MANAGER;
 $g_manage_user_threshold = ADMINISTRATOR;
 
 /**
+ * Threshold needed to impersonate a user or NOBODY to disable the feature.
+ * @global integer $g_impersonate_user_threshold
+ */
+$g_impersonate_user_threshold = ADMINISTRATOR;
+
+/**
  * Delete bug threshold
  * @global integer $g_delete_bug_threshold
  */
@@ -4429,6 +4435,7 @@ $g_public_config_names = array(
 	'html_make_links',
 	'html_valid_tags_single_line',
 	'html_valid_tags',
+	'impersonate_user_threshold',
 	'inline_file_exts',
 	'limit_reporters',
 	'logo_image',
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -325,9 +325,7 @@ function auth_impersonate( $p_user_id ) {
  * @return bool true: can impersonate, false: can't.
  */
 function auth_can_impersonate( $p_user_id ) {
-	# Make sure the logged in user can impersonate other users.  If user can
-	# manage users, then they would be able to impersonate them.
-	if( !access_has_global_level( config_get( 'manage_user_threshold' ) ) ) {
+	if( !access_has_global_level( config_get( 'impersonate_user_threshold' ) ) ) {
 		return false;
 	}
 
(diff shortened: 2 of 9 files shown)


atrol:
This would mean that a user having 'manage_user_threshold' is able to fake activity of other users.
I am not sure if the convenience functionality
> When a user creates an (service) account and wants to create an API key for such account. 

justifies this.

I tend to check for "administrator" access level.

dregad:
I have not reviewed the code, but I like the idea on principle.
@atrol see [my note in the issue](https://www.mantisbt.org/bugs/view.php?id=20772).

dregad:
I gave this a quick spin, and have the following comments / suggestions :

- I feel that the entry point for impersonation is hidden too deep, it would be nice to make it more accessible, e.g.
  - a link from the header (e.g. an icon next to the administrator's user name) allowing to select a user to impersonate from a list, something like the mock-up below
selection_002
  - from *view_user_page.php* 
  - directly from *manage_user_page.php* without having to open *manage_user_edit_page.php* (e.g. same icon as before, next to each entry in the list)

- Impersonating a user should not bump their *last_visit* timestamp
- I think the system should clearly show the admin that he is in "impersonating mode" (e.g. show `Impersonating: username (Realname - reporter)` in the header instead of `Logged in as:`

- When logging out, system should go back to the admin's account without making him to login again
- Maybe inform the user that someone used their account (e.g. e-mail notification)

vboctor:
@dregad 

**Feature Discovery**

- The impersonation option is now also available via the View User page.
- The impersonation option is no more hidden than 'Reset Password' or 'Delete User', which is good enough to start with.  So I don't feel a need to put it in too many places.
- Manage Users page has no user actions, so if we add the other actions, we can add this one.
- Adding an "Impersonate" option as a top level action is an overkill with new UI that I don't think is necessary at this point.

**Logout from impersonated user to administrator:**

- We already ask the user to re-login if they haven't logged-in in the last N minutes when they are doing administration tasks.  Hence, saving this extra login is not worth the complexity.  Though if someone feels that is needed, feel free to add.

** Impersonating rather than Logged in as **

Isn't this obvious from the fact that the user name is "xyz" rather than "vboctor" for example?  I don't see the need to add such complexity.

** last visit **

If we end up having a "impersonation" cookie, then we can add this and impersonation message in the header.  Should we add this as we get feedback?

** Email **

Given that the impersonating user has all the necessary access anyway, I don't see the need for this.  Our assumption is that such users are well trusted, otherwise, they can do a lot of damage anyway.  As an administrator, they can even modify the database.

**Access Level**

@atrol We can require administrator access level.  However, I think it is useful to leverage the manage users access level because: - typically this is the administrator anyway, if you are trusted to create/delete users, then you may need to also test such users to make sure they have the right access.

I've checked in some of the changes requested.  Happy to revise the access level if needed.  But would rather not do unnecessary work to start with.  The current implementation really does the trick.  Let's see how much use and feedback we get on it.

vboctor:
I'm planning to add documentation.  Didn't find documentation for manage users thresholds config option, so will figure out where to add them both.

vboctor:
Added 'impersonate_user_threshold' config option + documentation.

vboctor:
I'm targeting to check this in by Saturday.  Let me know if there is any issues with latest change.

PR #750, opened 2016-04-05
Font error when export issue

thansau239:
Dear all,
I have used export function for report.But it font error (i'm using vietnamese language) when i export . Please see in attach file
[https://www.mantisbt.org/forums/download/file.php?id=876](url)
sorry maybe my english not good but please help me


dregad:
This is not the right place to report problems. Please submit an issue in our tracker http://mantisbt.org/bugs 

Moreover, your pull request is asking us to merge 1.2.x branch into master, which is never going to happen, so I'm closing this.

PR #752, opened 2016-04-07
Fixes #0020784

elwlwlwk:
Sign up item appears when allow_signup is set OFF.
login_method and enable_email_notification are not checked because they are already checked top of the lost_pwd_page.

### DIFF
--- a/lost_pwd_page.php
+++ b/lost_pwd_page.php
@@ -65,7 +65,9 @@ html_page_top2a();
 			</legend>
 			<ul id="login-links">
 				<li><a href="login_page.php"><?php echo lang_get( 'login_link' ); ?></a></li>
+				<?php if( ON == config_get_global( 'allow_signup' ) ) { ?>
 				<li><a href="signup_page.php"><?php echo lang_get( 'signup_link' ); ?></a></li>
+				<?php } ?>
 			</ul>
 <?php
 	echo form_security_field( 'lost_pwd' );


lost_pwd_page.php
```diff
@@ -64,7 +64,9 @@
 				<span><?php echo lang_get( 'lost_password_title' ); ?></span>
 			</legend>
 			<ul id="login-links">
+				<?php if(  ( ON == config_get_global( 'allow_signup' ) )) { ?>
```
atrol: This is not what you want. You are deactivating login.

elwlwlwk:
Sorry for my mistake. I was confused because tested and modified code in different machine. I forcepushed.

lost_pwd_page.php
```diff
@@ -65,7 +65,9 @@
 			</legend>
 			<ul id="login-links">
 				<li><a href="login_page.php"><?php echo lang_get( 'login_link' ); ?></a></li>
+				<?php if(  ( ON == config_get_global( 'allow_signup' ) )) { ?>
```
vboctor: The spacing around the brackets seems to be wrong two spaces on one side and no spaces on the other.  There also seem to be an extra set of ().

vboctor:
Thanks @elwlwlwk for your contribution.

Please open a bug in our bug tracker and references it from the commit messages as:
Fixes #1234

vboctor:
Merged via 2e0e1438d4d9c6c4dbbacc99674428d3c7f27543 - thanks.

PR #755, opened 2016-04-11
Replace images with [FontAwesome](http://fontawesome.io)

Kirill:
Please review my work about replace different images (they don't have unified views) with font awesome


cproensa:
I dont know if this has place in 1.3 or should be targeted at v2, anyway, i've had a look and these are my impressions:
- I'd not delete the old images, since they may be still in use by other users enhancements and plugins
- Seems like the icons are all monochrome. I noticed the RSS icon to be specially "ugly".
- File type icons, they look too small to be readable. Here it adds the fact of monochome design. Old icons, even in small size are easily identifiable for file types.
- I liked the "check" instead of 'X' on some tables.

vboctor:
@Kirill I agree that this seems to be something that should be part of the modern UI.  Can you please submit the pull request against http://github.com/mantishub/mantisbt 

Plugins are being revised to work with modern UI anyway, so deleting images that are not used is OK, since plugin need to be updated and tested with the next version anyway.

Thanks!

Kirill:
@vboctor It's look nice and my patch will integrated with modern-ui more simple. When it's branch come to master?

vboctor:
@Kirill Glad you like it.  I was expecting it to be easier to integrate there since it already uses font awesome.  The current plan is to get it to master after 1.3 is out, in the meantime we are keeping it up to date with master.

vboctor:
Closing this in favor of resubmitting based on modern UI.

PR #757, opened 2016-04-15
Use JSON format in Configuration Report

cproensa:
Use JSON format for displaying and parsing complex type configuration
values.
Due to some errors in the current parser (eg, #20787, #20812, #20813), moving to JSON may be a solution to forget all the troubles that come with parsing regexps.
The core problem is that some complex expressions, as stored and displayed by this page, cannot be parsed back correctly.

This PR is a stub, that if found valid, can be further developed and refined.

### DIFF
--- a/adm_config_report.php
+++ b/adm_config_report.php
@@ -124,7 +124,7 @@ function print_config_value_as_string( $p_type, $p_value, $p_for_display = true 
 	if( $t_corrupted ) {
 		$t_output = $p_for_display ? lang_get( 'configuration_corrupted' ) : '';
 	} else {
-		$t_output = var_export( $t_value, true );
+		$t_output = json_encode( $t_value, JSON_PRETTY_PRINT | JSON_NUMERIC_CHECK );
 	}
 
 	if( $p_for_display ) {
--- a/adm_config_set.php
+++ b/adm_config_set.php
@@ -114,7 +114,7 @@ switch( $t_type ) {
 		break;
 	case CONFIG_TYPE_COMPLEX:
 	default:
-		$t_value = process_complex_value( $f_value );
+		$t_value = process_complex_json_value( $f_value );
 		break;
 }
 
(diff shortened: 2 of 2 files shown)


vboctor:
I would generally prefer dealing with json for configuration over php code.  I was thinking for a while of doing this for following reasons:

- Easier to pipe through our APIs -- at the moment we fail to retrieve arrays via SOAP API.
- Easier to consume by client apps once retrieved through our APIs.
- Easier to generate and consume - the example of the web configuration page.
- Easier for other tools to read / generate configs without having to include Mantis code or even being a PHP app.  For example, imagine a tool that loads config_inc.json, update/add a config, and then re-saving it.  This can even be used by plugins for their own configs if such access is available.

There is probably overlap in the above, but you get the point.

Using json to view/update would be a good step, but my biggest issue is that the user will have no reference to start from.  Currently, I can copy something from config_defaults_inc.php, tweak it, and add it to the UI.  With json, I have to figure out myself how to map to json, and then tweak.

Options to consider:

- config_defaults_inc would use json_decode( '...json...' ) instead of native arrays.
- config_defaults_inc would include json equivalent in the documentation for the config options that users would use 'complex' types for.
- possibly consider support config_inc.json as a way to override config instead of config_inc.php or in addition to it.  That would be part of considering json as a first class config option (likely longer term).

cproensa:
The proposal to use json for config files has its point. But its a different scope than this PR
Note the point here: the current editor/parser is not capable of loading configuration values that are present in the system: edit a value, and save it unmodified, and the value is corrupted. 
Probably most of the arrays in config_inc work fine, but some plugin configurations may fail. For example, email reporting stores its config as db config values, a multilevel array for mailboxes, if you edit that, they are gone.

> Using json to view/update would be a good step, but my biggest issue is that the user will have no reference to start from. Currently, I can copy something from config_defaults_inc.php, tweak it, and add it to the UI. With json, I have to figure out myself how to map to json, and then tweak.

I understand that. A better approach is to show global configurations (loaded from file) alongside database saved values. As the presentation in this page should be already Json, copying, editing, tweaking, or dupliclate for a project/user scope, is resovled.

My concern here is the effort needed to mantain a buggy parser, and user friendliness is not that important, since this page should be used only by administrators. Also, the main use of this page is to visualize, delete or tweak configurations, so once having the global configs available too, it's not that much harder that its now.

Also, it can be good to have the option for visualize as php syntax (current var_Export), for those situations when you want to copy that config to file.

dregad:
:-1: 

I agree that user friendliness is not the most important aspect of this admin page, but Mantis is a PHP application so it makes sense that our config relies on PHP syntax. 

I don't like the idea of using JSON in config_inc at all. And by extension, for consistency reasons, the admin page should use the same syntax as the config file. The API is a different story, and I have no issues with using JSON there.

With regards to processing the user's input for the admin page, I believe we could achieve this fairly easily by relying on PHP's `token_get_all()` parser instead of using regex as we do now. I wrote something similar once, I'll dig it up and adapt it to the task at hand, then post a proof-of-concept.

vboctor:
@dregad code as a config is not a good idea.  So I disagree with you.  Configurations are typically in json, yaml, xml, etc.  I don't often see java config, C# config, python config, etc as a common programming pattern.  We can still provide support for code based config overrides, but that shouldn't be the norm.

I'm OK with continuing to use PHP to minimize change for now, but I wouldn't choose PHP in the fullness of time.

cproensa:
> I don't like the idea of using JSON in config_inc at all. And by extension, for consistency reasons, the admin page should use the same syntax as the config file. The API is a different story, and I have no issues with using JSON there.

The short term issue is to fix the configuration parser. Proposal of json (only for this page) is aimed to rely on a trusted parser, instead of a custom made, and the maintenance overhead. 
If you think it can be done by parsing PHP syntax, without overkill effort, then that'd be a more suitable solution at present.

Regarding config_inc format.
I see the benefits of having a managed configuration format. 
Current scheme is loading the configurations as globals... This puts out of control from the application, for example, if some value was defined from config_defaults_inc, or config_inc.
Another benefit is to have configs structured logically (like firefox about:config)
I'd go further, and remove the file based config_inc, and move into DB. Keep the most basic in file for db connections, etc, but all bussines globals stored in db. Maybe import a file from admin tools, for loading first defaults or mass update...

But. That belongs to another thread. Victor: open a dedicated issue for that, and we can share the ideas there.

cproensa:
@dregad 
From what i see, token_get_all is a plain tokenizer. That's a start, but the other part needed is the grammar analyzer. So either write one, o look for a 3rd-p library (eg: http://pear.php.net/package/PHP_Parser).
Either way, i still dont like to put much work here, for a not-so-critical requeriment (dealing with php expression syntax)

cproensa:
A practical proposal can be:
- Add a switch to selection filter to choose the displayed syntax: php or json.
- Edition form for complex types uses json syntax (put a label/text indicating so)
- Add a link to documentation, showing a php code snippet for converting php syntax to json

dregad:
> the other part needed is the grammar analyzer. 

Correct. As I was saying earlier, I have some similar code which I wrote a while back for other purposes, and that can probably be adapted quite easily.

> i still dont like to put much work here

Agreed. The parser would be quite simple though, since we only need to deal with scalar and array types so I don't think it would be a big effort to implement.

 >  Configurations are typically in json, yaml, xml, etc. 

As an example, DokuWiki also stores config as PHP code. Other PHP apps (e.g. Wordpress, phpBB) also use a PHP script for their basic configuration.

> I'm OK with continuing to use PHP to minimize change for now, but I wouldn't choose PHP in the fullness of time.

I'm not opposed to having a discussion on revising the way we store and manage configs in the future, but as mentioned by Carlos it is outside the scope of fixing the admin config page parser and belongs in a separate thread.

cproensa:
closed, replaced by #761

PR #759, opened 2016-04-18
Collapse cookie issues

dregad:
This PR addresses 2 related issues which I faced while testing [Cloud9 IDE](https://c9.io/).

- [#20822](https://www.mantisbt.org/bugs/view.php?id=20822): Collapsing/Expanding sections triggers error 2300 
- [#20824](https://www.mantisbt.org/bugs/view.php?id=20824): collapse_cache_token() always update token ID # 5 instead of the user's token

### DIFF
--- a/core/collapse_api.php
+++ b/core/collapse_api.php
@@ -217,11 +217,15 @@ function collapse_cache_token() {
 			}
 		}
 
+		if( !$t_update ) {
+			$t_token = token_get( TOKEN_COLLAPSE );
+			$t_update = $t_token !== null;
+		}
 		if( $t_update ) {
-			$t_token = json_encode( $g_collapse_cache_token );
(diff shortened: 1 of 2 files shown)


vboctor:
Is there any backward compatibility needed here?  Were cookies with commas ever successfully set in some of the cases?  Otherwise, if it works go for it :+1:

dregad:
As far as I can tell based on my tests, if a user has an "old style" cookie with commas (e.g. `|relationship,0`), collapse_cache_token() will simply ignore it as it is [not a valid pair](https://github.com/mantisbt/mantisbt/pull/759/files#diff-01b084543ec8a766edadb648e77ba2e9R214) anymore since we *explode* with `:` as delimiter.

This means the only possible regression (in the unlikely event that the cookie's contents has not yet been stored in the tokens table), is for a user to see a previously collapsed section displayed again or vice-versa.

PR #760, opened 2016-04-21
Support @ mentions

vboctor:
This provides the basic support for @ mentions.  We can grow this feature
further over time.

Fixes #20837


vboctor:
This feature is pretty much done with the exception of some further testing specially around email notifications being generated as expected.  Thought I would get feedback in parallel.

config_defaults_inc.php
```diff
@@ -4729,3 +4729,9 @@
  */
 $g_webservice_version_when_not_found = '';
 
+/**
+ * Enables or disables @ mentions feature.
+ *
+ * @global integer $g_mentions_enabled
+ */
+$g_mentions_enabled = ON;
```
dregad: - this config does not belong in webservice section, suggest to move it to *MantisBT Misc Settings*, maybe next to the related *bug_link_tag*
- maybe call this `g_user_mentions_enabled` to be more accurate

dregad:
Do we really want to hardcode use of '@' ? I would suggest to provide end-user flexibility to pick the tag they want, like we do for `#` and `~`.

vboctor:
@dregad Thanks for the feedback:

- I've moved the config option along with the reminder config options
- I don't think I want the 'tag' to be configurable.  The @ mention is an industry standard.  We can revisit in the future if we determine this is really needed.
- I've implemented and tested the handling of mentions in email notifications in general.
- I have intentionally left the terminology for the config option to be open, since I would to evolve this to mentioning of other entities that are defined in Mantis.  For example, ability to mention a version or a project.  For example, "This is fixed in @1.3.x" saved as "This is fixed in @{V15}.  "The fix for this issue should be in @mantisbt-web" saved as @{P2}, etc.  Different entity type may have different rendering and hyperlinking characteristics.

vboctor:
@dregad one thing to consider with @ mention that is worse discussing is the handling of placeholders.  There are some options to consider:

1. Don't use placeholders - this will break if the referenced entity gets renamed.
2. Don't use placeholders but update all issues when an entity is renamed - this will protect against breaking, but makes renaming an entity more complex / expensive.
3. Use placeholders in db access layer - when calling issue_get(), you get the issue without placeholders.  In such case, the UI will have re-parse for the entities rather than the easier placeholders.
4. Use placeholders in UI/API - the placeholders are replaced in the UI layer and SOAP API layers.  Clients built on the API don't have to handle placeholders.  Better for backward compatibility.  If a client wants to provide special handling for @ mention they should re-use our regular expressions + similar logic to extract valid @ mentions.
5. Use placeholders except in UI - the UI layer, whether it is web or a client that is built on the API understands placeholders and does necessary substitution.  We can have issue_get return the placeholders and client do extra calls or we return an extra map of mentions as part of call to issue_get.

So far the code checked in implements option 4.

Thoughts?

dregad:
> I don't think I want the 'tag' to be configurable. The @ mention is an industry standard. We can revisit in the future if we determine this is really needed.

I disagree. `#` for designating issues is also a de-facto "standard", but still we allow admins to customize it. I don't see why we should treat `@` differently. Most people wouldn't bother changing the default, but it doesn't cost us much to allow it instead of hardcoding.

Re: placeholders, option 4 sounds like the best approach at first glance, but you obviously gave this a lot more thought than I did, and your plans for the feature are going much further than I originally considered.

vboctor:
The implementation is now complete.  I've ended up using the following convention:

- Replace mentions with placeholders before saving (close to INSERT/UPDATE sql statements).
- Replace placeholders with mentions after loading (close to SELECT sql statements).
- Decorate mentions in the UI.

This is option 3 from the list above, where the API handles the conversion to/from placeholders making it as transparent as possible to rest of code base.  Initially I had a mix of 3 + 4.

The code is now complete, tested manually and added a bunch of test cases for parsing/soap.

FSmith80:
Please **don't ** start modifying text submitted by users :(

We already make use of a public mantis plugin implementing mentions that replaces @foo on the fly as it were.

In our use case, we store debug data containing @, {} etc in bugs - we've found that occasionally this public plugin gets in the way, but you can always get the raw data from the database if need-be. In the case of this, it looks like you are starting to modify the raw data that comes in from users directly in the database - which means there's no chance to go back - this is something that mantis has never done. 

In terms of the example that you guys discussed: ""This is fixed in @1.3.x" saved as "This is fixed in @{V15}". I'd argue that NOT modifying the raw data is more important. If we create a planned version e.g. 1.3.x in your example, then delete that version from the database, you've then 'lost' the information of 1.3.x

cproensa:
sorry for getting late to this.
I had a mixed feeling... and previous comment hits the nail: user submitted text should not be modified.

I understand that changing configurations, deleting users, etc can lead to inconsistent functionality.but i dont think that is a problem (just as writing user names in plain text isn't either)
The functionality that "mentions" should cover is a short term one:
when the text is created, you should send notification, etc, but not after for example, 1 year later... At this point having an obsolete @olduser tag is almost irrelevant.

If a persistent tracking of mentioned user is wanted, it should be done outside the raw text. Use a dedicated table to track users mentioned in which bug (or should it be for each text element?).
Such persistent scheme could be used for filtering (and my-view style filter). How would you get the issues where a user is mentioned, without a proper relational table?
That is how monitoring is implemented.

So my opinion is: in any way that this is going to be used, inserting the metadata in the raw text is not a good idea. I suggest keeping the @xxxx detection on the fly like other current formatting logic, and for a more advanced features, either link to monitoring, or implement a proper mention tracking scheme.

Nevertheless, from a user point of view, i like this feature!

atrol:
> user submitted text should not be modified.

+1

vboctor:
I've removed the logic that alters the text.  I've also added a configuration option for mention tag.

At one point, I would like to also add the mentions table as per my comments in the issue.  I think it can power interesting scenarios.  But that is not going to be part of this PR.

This should address all feedback so far.

config_defaults_inc.php
atrol: Should be included in $g_public_config_names

config_defaults_inc.php
```diff
@@ -1607,6 +1607,19 @@
...
+
+/**
+ * The tag to use for mentions.
+ * @var string $g_mentions_tag
+ */
+$g_mentions_tag = '@';
```
atrol: Should be included in $g_public_config_names

core/bug_api.php
```diff
@@ -544,6 +546,52 @@ function create() {
...
+			if( !is_blank( $this->additional_information ) ) {
+				$t_mention_text .= lang_get( 'email_additional_information' ) . "\n\n";
+				$t_mention_text .= $this->additional_information . "\n\n";
+			}
+
+			mention_process_user_mentions(
+				$this->id,
+				$t_filtered_mentioned_user_ids,
+				$t_mention_text,
+				$t_removed_mentions_user_ids );
```
atrol: There is no parameter for $t_removed_mentions_user_ids in function mention_process_user_mention
```php
/**
 * Process users that are mentioned on the specified bug.
 *
 * @param int $p_bug_id The bug id.
 * @param array $p_mentioned_user_ids An array of user ids
 * @param string $p_message The message containing the mentions.
 * @return void
 */
function mention_process_user_mentions( $p_bug_id, $p_mentioned_user_ids, $p_message = '' ) {
	email_user_mention( $p_bug_id, $p_mentioned_user_ids, $p_message );
}
```

core/bugnote_api.php
```diff
@@ -192,25 +194,27 @@ function bugnote_add( $p_bug_id, $p_bugnote_text, $p_time_tracking = '0:00', $p_
 
 	antispam_check();
 
+	$t_bugnote_text = $p_bugnote_text;
+
```
atrol: Assignment not needed if $p_bugnote_text is used (again) in following two is_blank() checks.

core/bugnote_api.php
```diff
@@ -257,6 +261,21 @@ function bugnote_add( $p_bug_id, $p_bugnote_text, $p_time_tracking = '0:00', $p_
...
+		$t_bugnote_id,
+		$t_mentioned_user_ids );
+
+	$t_removed_mentions_user_ids = array_diff( $t_mentioned_user_ids, $t_filtered_mentioned_user_ids );
+
+	mention_process_user_mentions(
+		$p_bug_id,
+		$t_filtered_mentioned_user_ids,
+		$p_bugnote_text,
+		$t_removed_mentions_user_ids );
```
atrol: Same as in bug-api.php, there is no parameter for $t_removed_mentions_user_ids in function mention_process_user_mentions.

core/email_api.php
```diff
@@ -1301,6 +1301,77 @@ function email_bug_reminder( $p_recipients, $p_bug_id, $p_message ) {
...
+	$t_sender = user_get_name( $t_sender_id );
+
+	$t_subject = email_build_subject( $p_bug_id );
+	$t_subject = sprintf( lang_get( 'mentioned_in' ), $t_subject );
+	$t_date = date( config_get( 'normal_date_format' ) );
+	$t_user_id = auth_get_current_user_id();
+	$t_users_processed = array();
+
+	foreach( $p_removed_mention_user_ids as $t_remove_mention_user_id ) {
+		log_event( LOG_EMAIL_VERBOSE, 'skipped mention email for U' . $t_mention_user_id . ' (no access to issue or note).' );
```
atrol: Should be $t_remove_mention_user_id instead of $t_mention_user_id

core/mention_api.php
```diff
@@ -0,0 +1,185 @@
...
+	return $t_mentions;
+}
+
+/**
+ * Given a string find the @ mentioned users.  The return list is a valid
+ * list of valid mentioned users.  The list will be empty if the mentions
+ * feature is disabled.
+ *
+ * @param string $p_text The text to process.
+ * @return Array with valid usernames as keys and their ids as values.
```
atrol: array instead of Array

core/mention_api.php
```diff
@@ -0,0 +1,185 @@
...
+ * @param string $p_text The text to process.
+ * @param bool $p_html true for html, false otherwise.
+ * @return string The processed text.
+ */
+function mention_format_text( $p_text, $p_html = true ) {
+	if ( !mention_enabled() ) {
+		return $p_text;
+	}
+
+	$t_text = $p_text;
```
atrol: Not needed, should be removed.

core/mention_api.php
```diff
@@ -0,0 +1,185 @@
...
+	$t_mentioned_users = mention_get_users( $p_text );
+	if( empty( $t_mentioned_users ) ) {
+		return $p_text;
+	}
+
+	$t_mentions_tag = mentions_tag();
+	$t_formatted_mentions = array();
+
+	foreach( $t_mentioned_users as $t_username => $t_user_id  ) {
+		if( user_exists( $t_user_id ) ) {
```
atrol: Do we need this check?
Is there a way that mention_get_users() delivers non existent users?

tests/Mantis/MentionParsingTest.php
```diff
@@ -0,0 +1,113 @@
...
+		$this->checkMentions( 'some random string with @@vboctor sign.', array() );
+	}
+
+	public function testJustMention() {
+		$this->checkMentions( '@vboctor', array( 'vboctor' ) );
+	}
+
+	public function testWithDotInMiddle() {
+		$this->checkMentions( '@victor.boctor', array( 'victor.boctor' ) );
+	}
```
cproensa: Add test to email address format
Eg:
xxxx@vboctor
xxxx@victor.boctor
should not be matched as user mention

core/email_api.php
```diff
@@ -1301,6 +1301,77 @@ function email_bug_reminder( $p_recipients, $p_bug_id, $p_message ) {
...
+ * or an issue note.
+ *
+ * @param integer       $p_bug_id     Issue for which the reminder is sent.
+ * @param array         $p_mention_user_ids User id or list of user ids array.
+ * @param string        $p_message    Optional message to add to the e-mail.
+ * @param array         $p_removed_mention_user_ids  The users that were removed due to lack of access.
+ * @return array List of users ids to whom the reminder e-mail was actually sent
+ */
+function email_user_mention( $p_bug_id, $p_mention_user_ids, $p_message, $p_removed_mention_user_ids = array() ) {
+	if( OFF == config_get( 'enable_email_notification' ) ) {
```
cproensa: should be
config_get_global( 'enable_email_notification' )
?

core/email_api.php
```diff
@@ -1301,6 +1301,77 @@ function email_bug_reminder( $p_recipients, $p_bug_id, $p_message ) {
...
+		log_event( LOG_EMAIL_VERBOSE, 'email notifications disabled.' );
```
cproensa: Is this really helpful as a log message?

core/bug_api.php
```diff
@@ -544,6 +546,52 @@ function create() {
 		history_log_event_direct( $this->id, 'status', $t_original_status, $t_status );
 		history_log_event_direct( $this->id, 'handler_id', 0, $this->handler_id );
 
+		# Now that the issue is added process the @ mentions
+		$t_all_mentioned_user_ids = array();
```
cproensa: I'm not convinced this is the right place to process the mention tags
as BugData->create() is the core method for creating a bug, this may arise to triggering mentions in more situation than those intended.
Example of a debatible situation: Cloning an issue
1) Does it have to trigger mentions contained in the texts? and/or in the notes?
2) The user that is informed as mentioner is "auth_get_current_user_id", this means that if user A clones an issue where B did a comment mentioning C, a new mention will be triggerd as "user A mentions B" (i tested that it does so)

This login probably fits better in the processing of submitted text, right after user interaction, where current user is still  the "mentioner", and this way, mentions are triggered only at the time of submitting.

atrol: > Example of a debatible situation: Cloning an issue
Another situation: Plugins like XmlImport

core/bug_api.php
vboctor: Added the parameter and passed it through to the email method.

core/email_api.php
vboctor: No, since this config option can be overridden in the db.  It is also based on the code for reminders.

core/email_api.php
vboctor: One of the hardest area to troubleshoot for users is "why didn't person X receive this notification".  I think doesn't hurt and is useful.

core/mention_api.php
vboctor: Removed.  mention_get_users() will never return non-existent users, but in the model where we had placeholders, this was a possible case.

core/email_api.php
cproensa: ok, so offtopic here: can we have 'enable_email_notification' overriden at user/project level? That is wicked...

core/email_api.php
cproensa: ok, i was seeing the line without context. I guess the log says somewhere that this happened in the mentions method.

vboctor:
@cproensa and @atrol thanks for your comments.  I've addressed all of them with the exception of where the @ mentions are processed.  Will follow up on this one tonight.

core/email_api.php
vboctor: Unless we in the code we specify to only look at all users/projects, then it can be done.  That doesn't mean we really think it is a good idea or that it will work right.  I use this explicit approach when checking if mentions are enabled.  We often should but don't use this format, maybe we should have a method that makes it easy, e.g. config_get_global_or_db() that just takes the config name and makes a call like the one below.
config_get( 'enable_email_notification', null, ALL_USERS, ALL_PROJECTS )

core/bug_api.php
vboctor: I've just pushed the change to separate mention processing from issue/note creation API.  Please review.

cproensa:
> 	Make mention processing an explicit API action … 

Thanks victor, this is looking fine now

Still I have some doubt: should the user mentioning be guessed at the email method?. Being an api call, i'd play safe and pass this user id as a parameter, which could be then defaulted to current user if none is provided. 
Just a thought, since another alternative is to rely on current_user_set() if another user is needed as originator for mentions...

But, being like its now the BugData and bugnotes stuff, i don't care much. Those parts needs some work anyway.

vboctor:
> Still I have some doubt: should the user mentioning be guessed at the email method?. Being an api call, i'd play safe and pass this user id as a parameter, which could be then defaulted to current user if none is provided. 

@cproensa The current approach works based on the current scenarios.  We can always add the user when needed as different from logged in user.  In the meantime, we have already existing code in email api to filter out the "receive_own" email notifications (see email_collect_recipients()):

		# Possibly eliminate the current user
		if( ( auth_get_current_user_id() == $t_id ) && ( OFF == config_get( 'email_receive_own' ) ) ) {
			log_event( LOG_EMAIL_RECIPIENT, 'Issue = #%d, drop @U%d (own action)', $p_bug_id, $t_id );
			continue;
		}

PR #756, opened 2016-04-14
Configuration options arrays

cproensa:
Some fixes for parsing of config options
Fixes: #0020787
Fixes: #0020850
Fixes: #0020812
Fixes: #0020851

PR #761, opened 2016-04-22
Parse complex configs using PHP tokenizer

dregad:
This PR offers an alternative approach to #756 and #757.

It Implements 2 classes to parse 'complex' config options in admin Configuration Report page using PHP tokenizer instead of relying on regular expressions.

- Tokenizer calls token_get_all() to build a list of tokens from the
  given code
- Parser uses a Tokenizer object to convert the code (string) to an
  actual PHP array variable

TODO: Move classes to an API and write automated tests

Fixes [#20787](https://www.mantisbt.org/bugs/view.php?id=20787), [#20812](https://www.mantisbt.org/bugs/view.php?id=20812), [#20813](https://www.mantisbt.org/bugs/view.php?id=20813), [#20850](https://www.mantisbt.org/bugs/view.php?id=20850), [#20851](https://www.mantisbt.org/bugs/view.php?id=20851), [#16931](https://www.mantisbt.org/bugs/view.php?id=), [#16932](https://www.mantisbt.org/bugs/view.php?id=16932)

All the test cases given in the 5 issues listed above, as well as [@vboctor's notify_flags case](https://github.com/mantisbt/mantisbt/pull/756#issuecomment-210287456) were parsed successfully with this code.

**This is a work-in-progress and requires further testing to ensure it works properly**. 

Any suggestions to include as "tricky" test cases for PHPUnit tests would be appreciated.


cproensa:
Thanks @dregad 
if it helps, i have some stub pages for tests in #756, you can pull those and adapt for this

dregad:
Great, I'll make use of that. Thanks !

dregad:
@cproensa @vboctor I think this should pretty much be done now. 

I spent a bit more time than I expected to implement the unit tests, but hopefully they're fairly comprehensive, covering all the cases I could think of.

Let me know your feedback.

adm_config_set.php
```diff
@@ -102,177 +102,32 @@
...
-		$t_value = process_complex_value( $f_value );
-		break;
+# Parse the value
+if( $t_type == CONFIG_TYPE_STRING ) {
+	# Return strings as is
+	$t_value = $f_value;
+} else {
+	try {
+		$t_parser = new ConfigParser( $f_value );
+		$t_value = $t_parser->parse( ConfigParser::EXTRA_TOKENS_IGNORE );
```
vboctor: Does this code need to path in this parameter?  Or should it be the default for the parser?

tests/Mantis/ConfigParserTest.php
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	/**
+	 * Test no errors if we get extra tokens when the parser is set to ignore
+	 *
+	 * @throws Exception
+	 */
+	public function testExtraTokensIgnore() {
+		$t_parser = new ConfigParser( '1; 2' );
+		$t_result = $t_parser->parse( ConfigParser::EXTRA_TOKENS_ERROR );
```
vboctor: Shouldn't this be throwing?

tests/Mantis/ConfigParserTest.php
```diff
@@ -0,0 +1,432 @@
...
+	private function initScalarTestCases() {
+		# Integer
+		$this->addScalarCase( '1', PHPUnit_Type::TYPE_INT );
+		$this->addScalarCase( " 1\n", PHPUnit_Type::TYPE_INT );
+
+		# Float
+		$this->addScalarCase( '1.1', PHPUnit_Type::TYPE_FLOAT );
+
+		# String
+		$this->addScalarCase( '"1"', PHPUnit_Type::TYPE_STRING );
```
vboctor: Add test with single quote "'1'".

tests/Mantis/ConfigParserTest.php
```diff
@@ -0,0 +1,432 @@
...
+array( '"a""b"""', "'a''b'''" )
+EOT
+		);
+
+		/**
+		 * Associative arrays
+		 */
+
+		# associative
+		$this->addArrayCase( "array( 0 => 'a', 1 => 'b' )" );
```
vboctor: Test with double quotes inside the array.

tests/Mantis/ConfigParserTest.php
```diff
@@ -0,0 +1,432 @@
...
+		);
+
+		# both (e.g. set_status_threshold)
+		$this->addArrayCase( 'array( NEW_ => REPORTER )' );
+
+		/**
+		 * Multidimensional arrays
+		 */
+
+		# notify_flags sample
```
vboctor: Add test cases for assoc array of assoc arrays.  We use this for notify flags I believe where the first array covers the event and the second level covers who gets notified.

vboctor: You already have it, can you add two elements on the top level to make sure all is OK.

vboctor:
:+1: good work.  I wonder if this work can help us not end up with normal arrays prompted to user as:

array( 0 => 'element 1', 1 => 'element 2' ) but have array( 'element 1', 'element 2' ) which makes it easier to read and easier to edit (e.g. when removing an element from the middle of the report page fields.

Not necessarily part of the scope of this PR, but just suggesting it.

adm_config_set.php
dregad: As mentioned elsewhere, I reused this parser which I initially wrote for other purposes; the original code was more generic than this, and was designed to throw an exception.

As I was working through adapting it, I realized that in the specific usage of the admin config page, even though it is invalid for users to submit more than one value, it is more user-friendly to silently drop the extra stuff rather than choke on it. 

I decided to leave this as an option, in case we want to make use of it later (e.g. have an ajax to validate the value prior to actually submitting it). I'm fine with changing the default if you prefer.

tests/Mantis/ConfigParserTest.php
dregad: Ok, will add

tests/Mantis/ConfigParserTest.php
dregad: You mean [like this](https://github.com/mantisbt/mantisbt/pull/761/files#diff-19723a564ef6fb2d9e6afda09158a8d5R410) ?

adm_config_set.php
vboctor: Doesn't matter.

tests/Mantis/ConfigParserTest.php
dregad: You mean add 2 more events, from the full *notify_flags* array ?
```
  'owner' => 
  array (
    'reporter' => 1,
    'handler' => 1,
    'monitor' => 1,
    'bugnotes' => 1,
    'threshold_max' => 0,
  ),
  'reopened' => 
  array (
    'reporter' => 1,
    'handler' => 1,
    'monitor' => 1,
    'bugnotes' => 1,
    'threshold_max' => 0,
  ),
```

dregad:
Need to adjust the syntax check in Tokenizer's contructor for PHP7, as [eval() now throws an exception](https://travis-ci.org/mantisbt/mantisbt/jobs/125235617#L876).

>  I wonder if this work can help us not end up with normal arrays prompted to user as: `array( 0 => 'element 1', 1 => 'element 2' )` but have `array( 'element 1', 'element 2' )` which makes it easier to read and easier to edit (e.g. when removing an element from the middle of the report page fields.

We currently rely on *var_export()* to present the array to the user, which addings the numeric indexes on display:
```
php > var_export( array('a','b','c') );
array (
  0 => 'a',
  1 => 'b',
  2 => 'c',
)
```

I guess we could renumber the array (using *array_values()*), if we detect a numerically-indexed array, with the potential risk of destroying information if it turns out a gap in numbers was intentional (although I don't think we have this scenario in Mantis). 

Personally I'd rather play it safe; in the example you gave, the user preferences page is the canonical way of rearranging columns, and that will take care of renumbering the array even if contains a gap in the keys sequence.

tests/Mantis/ConfigParserTest.php
vboctor: yes.

cproensa:
Regarding implementation, so far all i can see is it works very fine.

> if we detect a numerically-indexed array, with the potential risk of destroying information if it turns out a gap in numbers was intentional 

That can be a possibility, but only if keys are 1) numeric 2) 0 based 3) sequential, to keep all data integrity.

> although I don't think we have this scenario in Mantis

must account for plugins...

Anyway the array, as php manages it, has keys. Var_export also prints those keys.
Only alternative is a custom array printer, that ignore keys when applicable.

dregad:
> Only alternative is a custom array printer, that ignore keys when applicable.

Indeed.

tests/Mantis/ConfigParserTest.php
dregad: You're right, this is a copy/paste error it should be *EXTRA_TOKENS_IGNORE*... And PHPUnit didn't catch it because there was an error in ConfigParser::parse(). I'll fix that.

tests/Mantis/ConfigParserTest.php
dregad: OK, it's done.

cproensa:
add references to #16931 #16932

dregad:
> add references to #16931 #16932

Done

Will rebase and merge shortly.

PR #762, opened 2016-04-22
Add email notifications for users assigned to categories

evilchewits:
Users who are assigned to categories can now receive email notifications.
It's configurable on [ E-mail Notifications ] page per-project.

A common use case:
 - A user was assigned to some category;
 - When new issue is submitted to this category and "Assigned To" field is empty, this issue is assigned to Category Owner (hence he receives a notification);
 - If then we re-assign issue to other man, the Category Owner will no longer receive notifications.

This patch makes it possible to receive notifications by Category Owners in this case as well.

See also: https://www.mantisbt.org/forums/viewtopic.php?f=3&t=22917

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -308,6 +308,20 @@ function email_collect_recipients( $p_bug_id, $p_notify_type, array $p_extra_use
 		}
 	}
 
+	# add Category Owner
+	if( ON == email_notify_flag( $p_notify_type, 'category' ) ) {
+		$t_category_id = bug_get_field( $p_bug_id, 'category_id' );
+
+		if( $t_category_id > 0 ) {
+			$t_category_assigned_to = category_get_field( $t_category_id, 'user_id' );
+
+			if( $t_category_assigned_to > 0 ) {
+				$t_recipients[$t_category_assigned_to] = true;
+				log_event( LOG_EMAIL_RECIPIENT, sprintf( 'Issue = #%d, add Category Owner = @U%d', $p_bug_id, $t_category_assigned_to ) );
+			}
+		}
+	}
+
 	# add users who contributed bugnotes
 	$t_bugnote_id = bugnote_get_latest_id( $p_bug_id );
 	$t_bugnote_date = bugnote_get_field( $t_bugnote_id, 'last_modified' );
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -445,6 +445,7 @@ $g_enable_email_notification	= ON;
  *       'handler': the handler of the bug
  *       'monitor': users who are monitoring a bug
  *      'bugnotes': users who have added a bugnote to the bug
+ *      'category': category owners
  *      'explicit': users who are explicitly specified by the code based on the
  *                  action (e.g. user added to monitor list).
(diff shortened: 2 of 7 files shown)


dregad:
Hello @evilchewits and thanks for the pull request.

I have not yet had the time to review in depth or test this, but I think it may be a useful feature. At first glance, I see 2 things missing from your submission

1. a reference to an issue in our bugtracker - please create it if none exist (we don't use the forums for that purpose), and reference it in your commit message (i.e. mention `Fixes #xxxx`)
2. updating the documentation (https://mantisbt.org/docs/master/en-US/Admin_Guide/html-desktop/#admin.config.email in this case)

core/email_api.php
```diff
@@ -308,6 +308,20 @@ function email_collect_recipients( $p_bug_id, $p_notify_type, array $p_extra_use
...
 	}
 
+	# add Category Owner
+	if( ON == email_notify_flag( $p_notify_type, 'category' ) ) {
+		$t_category_id = bug_get_field( $p_bug_id, 'category_id' );
+
+		if( $t_category_id > 0 ) {
+			$t_category_assigned_to = category_get_field( $t_category_id, 'user_id' );
+
+			if( $t_category_assigned_to > 0 ) {
```
vboctor: Should we check that category owner has access to the issue? e.g. what if the issue is private, but category owner only has access to public issues?  What if the category is global and this owner doesn't have access to the project containing the issue?

vboctor:
:+1: once access check is added in addition to @dregad 's comments.

evilchewits:
@dregad, thanks, I'll try to finalize.
@vboctor, there are additonal checks below in the code of email_collect_recipients(...) function:
```
...
		# exclude users who don't have at least viewer access to the bug,
		# or who can't see bugnotes if the last update included a bugnote
		if( !access_has_bug_level( config_get( 'view_bug_threshold', null, $t_id, $t_bug->project_id ), $p_bug_id, $t_id )
		 || $t_bug_date == $t_bugnote_date && !access_has_bugnote_level( config_get( 'view_bug_threshold', null, $t_id, $t_bug->project_id ), $t_bugnote_id, $t_id )
		) {
			log_event( LOG_EMAIL_RECIPIENT, 'Issue = #%d, drop @U%d (access level)', $p_bug_id, $t_id );
			continue;
		}
...
```
I think that's enough. No?

vboctor:
@evilchewits you are right about the checks.

evilchewits:
I've updated the commit and added an issue:
https://www.mantisbt.org/bugs/view.php?id=20876

By the way, I could not add bugnotes to my own issue, as well as add the relationship (#5275), etc.. Is that okay?

evilchewits:
What's more I consider enabling this option by default a good idea. I.e to make:

> $g_default_notify_flags = array(
> 	...
> 	'category'      => **ON**,
> 	...
> );

What do you think?

dregad:
> I could not add bugnotes to my own issue, 

That should work.

> as well as add the relationship (#5275), etc.. Is that okay?

As Reporter, it's normal that you can't add relationships.

dregad:
Fixed in 040cde6230ff165aaacfb73413748abbf350eb6a.

PR #763, opened 2016-04-22
Only cache visible plugin columns

cproensa:
On issues view/prin/export, a call is made to all MantisColumn plugin objects to
cache the list of issues. This call is now done only for those columns
that are part of the user column selection

Fixes: #0020854


dregad:
+1 (not tested though)

dregad:
Fixed in da3d2be7e05df21614720abe4f3080601242a7fd

PR #764, opened 2016-04-24
Use temporary data table for sql IN clauses

cproensa:
This PR implements the creation of tables to store temporary data, and a system to easily manage the operation of that data.

The class DBTempData allowed to create a temporary dataset, which is uploaded to the temp table, identified by unique id, which can then be used to be joined on any custom sql query. This dataset is deleted at session finalization.
Currently, only supporting int values, but the class can be expanded & inherited to support char fields/tables, if the need comes.

Additionally, a new db-api function: db_sql_in(), manages the creation of value lists for using with IN clauses.
Supports single value list, and tuples list. When the dataset is big enough, and the data types allows it, will make use of the DBTempData class to manage the data through a temporary table. This functionality should be transparent to the caller, as the resulting sql would be a compatible select subquery.

The aim of this implementation is to solve with a standard method the problems occurring with queries with very long list of parameters, that exceeds the limits of some DB engines

Edit to clarification:
This is dealing with "temporary data", with a limited time-to-live, that is stored in a standard table.
Ideally, these tables should be "in memory" tables, but that has different specifics for each db engine. That could be addressed on further iterations, at the moment standard tables are being used as a generic solution / fallback.


dregad:
Thanks for this Carlos ! I remember talking about this limitation for IN clauses several years ago already, I'm glad that you decided to tackle the problem.

I have not yet had time to go into details yet, but I already have a few comments
- How did you decide on the 500 limit ? Shouldn't this be DB-specific ? Oracle limits number of entries in IN clause to 1000 items, MySQL is [limited by packet size](http://dev.mysql.com/doc/refman/5.5/en/comparison-operators.html#function_in), not sure about other DBs.
- You should move the Class to a dedicated file in ./core/classes/ClassName.class.php` the autoloader will take care of including it

More later, probably...

cproensa:
The feelings are mixed: whether this is a solution, or the real problem has to be solved at the core of design (why do we have such situation in the first place). Probably both approaches can be combined.

However, the main code here is to manage the temporary rows. One of its applications is for the IN issue, but can be used in other useful ways. Eg: a big set of ids to be reused at several queries, 

> How did you decide on the 500 limit

Not a firm reason. Probably its better set at 1000. From looking at failing situations, usually the 1000 limit is reached far sooner that the usual packet size limits, for the same query. And 1000 is a big enough number to start playing it safe.

> You should move the Class to a dedicated file in ./core/classes

you're right..

cproensa:
I've looked the library code and its not possible to do many things with ADOdb create syntax.
Not possible to "create temporary table", or "unlogged" for pgsl, etc, becasue table options are just appended at the end of create statement.
Only possibility is for "engine=memory" for the mysql specific case.

cproensa:
The original issue has been solve by managing the size of items when doing operation on issues groups.
Closed because at the moment there is not an explicit need for temporary data tables

PR #768, opened 2016-04-29
Add tags for email notifications

lukelee1124:
Add tags field for email notifications when set and visible to user

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -1464,6 +1464,11 @@ function email_format_bug_message( array $p_visible_bug_data ) {
 	$t_message .= email_format_attribute( $p_visible_bug_data, 'email_project' );
 	$t_message .= email_format_attribute( $p_visible_bug_data, 'email_bug' );
 	$t_message .= email_format_attribute( $p_visible_bug_data, 'email_category' );
+
+	if( isset( $p_visible_bug_data['email_tag'] ) ) {
+		$t_message .= email_format_attribute( $p_visible_bug_data, 'email_tag' );
+	}
+
 	$t_message .= email_format_attribute( $p_visible_bug_data, 'email_reproducibility' );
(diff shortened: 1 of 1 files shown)


core/email_api.php
```diff
@@ -1635,6 +1640,12 @@ function email_build_visible_bug_data( $p_user_id, $p_bug_id, $p_message_id ) {
 	$t_category_name = category_full_name( $t_row['category_id'], false );
 	$t_bug_data['email_category'] = $t_category_name;
 
+	if( count( $t_tag_rows = tag_bug_get_attached( $p_bug_id ) ) !== 0 && access_compare_level( $t_user_access_level, config_get( 'tag_view_threshold' ) ) ) {
+		foreach( $t_tag_rows as $t_tag ) {
```
vboctor: I would use implode() with ', ' (comma + space) as the separator instead of the for loop.

vboctor:
The body of the commit message should include "Fixes #20857".

core/email_api.php
```diff
@@ -1635,6 +1640,12 @@ function email_build_visible_bug_data( $p_user_id, $p_bug_id, $p_message_id ) {
 	$t_category_name = category_full_name( $t_row['category_id'], false );
 	$t_bug_data['email_category'] = $t_category_name;
 
+	if( count( $t_tag_rows = tag_bug_get_attached( $p_bug_id ) ) !== 0 && access_compare_level( $t_user_access_level, config_get( 'tag_view_threshold' ) ) ) {
```
dregad: Move the assignment out of the if statement

core/email_api.php
dregad: Not sure this can be done, since he's calling tag_get_link() on each tag.

vboctor:
Replace with #770 due to several issues in this implementation.

PR #771, opened 2016-04-30
Do not enable compress handler if errors occured

dregad:
Prevents 'Content Encoding Error', shows the actual output instead.

Fixes [#20865](https://www.mantisbt.org/bugs/view.php?id=20865)

### DIFF
--- a/core/compress_api.php
+++ b/core/compress_api.php
@@ -85,6 +85,13 @@ function compress_handler_is_enabled() {
  * @access public
  */
 function compress_start_handler() {
+	# Do not start compress handler if we got any output so far, as this
+	# denotes that an error has occurred. Enabling compression in this case
+	# will likely cause a Content Encoding Error when displaying the page.
+	if( ob_get_length() ) {
+		return;
+	}
+
 	if( compress_handler_is_enabled() ) {
 		# Before doing anything else, start output buffering so we don't prevent
 		# headers from being sent if there's a blank line in an included file


core/compress_api.php
```diff
@@ -83,6 +83,13 @@ function compress_handler_is_enabled() {
  * @access public
  */
 function compress_start_handler() {
+	# Do not start compress handler if we got any output so far, as this
+	# denotes that an error has occurred. Enabling compression in this case
+	# will likely cause a Content Encoding Error when displaying the page.
+	if( ob_get_contents() ) {
```
atrol: I didn't try,  but would expect that checking ( ob_get_length() > 0 ) is better in terms of performance.

Do we need this check?
There is one single place where compress_start_handler() is called (core.php) where we do check already for blank output. 
```php
if( ( $t_output = ob_get_contents() ) != '' ) {
	echo 'Possible Whitespace/Error in Configuration File - Aborting. Output so far follows:<br />';
	var_dump( $t_output );
	die;
}
unset( $t_output );

# Start HTML compression handler (if enabled)
require_api( 'compress_api.php' );
compress_start_handler();
```

Of course, there might be output from require_api( 'compress_api.php' );

So, maybe remove the check in core.php and have it just in compress_start_handler is an option.

dregad: > would expect that checking ( ob_get_length() > 0 ) is better in terms of performance.

Good point, will change.

> Of course, there might be output from require_api( 'compress_api.php' );

That's exactly the behavior I'm trying to address here - any errors/warnings/notices that are thrown as we start initialize the core APIs. This single require_api() call actually includes a dozen other APIs

So I'm afraid this check *is* needed.

core/compress_api.php
```diff
@@ -83,6 +83,13 @@ function compress_handler_is_enabled() {
  * @access public
  */
 function compress_start_handler() {
+	# Do not start compress handler if we got any output so far, as this
+	# denotes that an error has occurred. Enabling compression in this case
+	# will likely cause a Content Encoding Error when displaying the page.
+	if( ob_get_length() ) {
```
vboctor: I don't think performance is an issue given that ob_get_length() will typically be 0.  If I'm missing something and there is a performance issue, would @ob_start() fix the error surfacing?  Would setting a flag in trigger_error() and checking it here help?

dregad: I believe @atrol's point about performance was due to my using `ob_get_contents()`, when the buffer's contents is not actually needed. 

> would @ob_start() fix the error surfacing?

No, the error occurs because we have some output prior to starting output buffering with the compression handler, it generates mixed contents that the browser can't interpret.

>  Would setting a flag in trigger_error() and checking it here help?

IMO it's a better approach to avoid triggering the error in the first place - disabling the compression handler in the (rare) case where initializing MantisBT core generates output efficiently resolves the problem IMO.

atrol: > I believe @atrol's point about performance was due to my using ob_get_contents() in the initial commit, when the buffer's contents is not actually needed. 

Right

PR #772, opened 2016-05-01
Fix FILTER_TYPE_MULTI_INT type for plugin filter

cproensa:
When treating filter options keys as strings, META_FILTER_ANY has to
be used as a string, to not trigger a type mismatch error

Also, when using check_selected() to print selected options, "strict"
option cannot be used because not always the option keys can be kept
as strings.
This happens when the options returned by the plugin filter object
uses integers for keys. Even if the keys are set as string type, PHP
will translate them to integer keys, thus triggering a type mismatch
error if strict check is used.

Eg:
array( '1' => 'x' ) is parsed by PHP as:
array( 1 => 'x' )


vboctor:
Merged via f5a278d6aa5c368e6bd89c1fc28123132b91a1eb

PR #773, opened 2016-05-02
Fix error when category default handler is deleted

vboctor:
- Fix the error in edit category page.
- Don't assign to default handler on reporting an issue unless user exists.

Fixes #20867

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -504,7 +504,7 @@ class BugData {
 			$t_result = db_query( $t_query, array( $this->category_id ) );
 			$t_handler = db_result( $t_result );
 
-			if( $t_handler !== false ) {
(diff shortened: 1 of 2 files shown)


atrol:
In addition to this change, wouldn't it be good to delete the entry in category table in function user_delete to prevent this situation?

vboctor:
I'm thinking that it might be helpful for admin to still have user9 next to appropriate categories and replace with the replacement person. 
I'm also planning another change to block deletion of a user if they have issues, notes, history, or categories associated with them.  And advise administrators to disable such users instead. 






On Mon, May 2, 2016 at 11:34 PM -0700, "Roland Becker" <notifications@github.com> wrote:












In addition to this change, wouldn't it be good to delete the entry in category table in function user_delete to prevent this situation?

PR #774, opened 2016-05-06
Improve documentation for Admin Guide > Configuration

evilchewits:
Fixes #20880

### DIFF
--- a/docbook/Admin_Guide/en-US/Configuration.xml
+++ b/docbook/Admin_Guide/en-US/Configuration.xml
@@ -5,6 +5,59 @@
 <chapter id="admin.config">
 	<title>Configuration</title>
 
+	<para>MantisBT is highly customizable through the web interface and
+		configuration files. Configuration options can be set globally as well as
+		customized for a specific project or user (except for options listed in
+		<emphasis>$g_global_settings</emphasis>).
+	</para>
+
+	<para>Configuration options can be set in
+		<emphasis>config_inc.php</emphasis> and in the
+		<emphasis>database</emphasis> (using the various manage pages).
+		Values stored in the database take precedence over values defined in
+		<emphasis>config_inc.php</emphasis>. Options stored in the database
+		can also be viewed and updated on
+		"Manage > Manage Configuration > Configuration Report" page.
+	</para>
+
+	<para>The sequence below describes how MantisBT determines which value
+		to use if an option was overridden:
+	</para>
+
+	<orderedlist>
+		<listitem>
+			<para>Look for the specified configuration option in the
+				database for current user and current project.
+			</para>
+		</listitem>
+		<listitem>
+			<para>Look for the specified configuration option in the
+				database for current user and all projects.
+			</para>
+		</listitem>
+		<listitem>
+			<para>Look for the specified configuration option in the
+				database for all users and current project.
+			</para>
+		</listitem>
+		<listitem>
+			<para>Look for the specified configuration option in the
+				database for all users and all projects.
+			</para>
+		</listitem>
+		<listitem>
+			<para>Look for the specified configuration option in
+				<emphasis>config_inc.php</emphasis> (where you might have some
+				code to distinguish between projects and/or users).
+			</para>
+		</listitem>
+		<listitem>
+			<para>Look for the specified configuration option in
+				<emphasis>config_defaults_inc.php</emphasis>.
+			</para>
+		</listitem>
+	</orderedlist>
+
 	<xi:include href="config/database.xml" xmlns:xi="http://www.w3.org/2001/XInclude"/>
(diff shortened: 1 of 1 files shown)


docbook/Admin_Guide/en-US/Configuration.xml
```diff
@@ -5,6 +5,57 @@
 <chapter id="admin.config">
 	<title>Configuration</title>
 
+	<para>MantisBT is highly customizable through the web interface and
+		configuration files. Configuration options can be set globally as well
+		as customized for a specific project or user (except options listed in
```
dregad: except *for*

docbook/Admin_Guide/en-US/Configuration.xml
```diff
@@ -5,6 +5,57 @@
...
+		<emphasis>$g_global_settings</emphasis>).
+	</para>
+
+	<para>Configuration options can be set in
+		<emphasis>config_inc.php</emphasis> and in
```
dregad: in *the* database

docbook/Admin_Guide/en-US/Configuration.xml
```diff
@@ -5,6 +5,57 @@
...
+		<emphasis>database</emphasis> (using the various manage pages).
+		Values stored in database take precedence over values in
```
dregad: in *the* database take precedence over values *defined* in

docbook/Admin_Guide/en-US/Configuration.xml
```diff
@@ -5,6 +5,57 @@
...
+		<emphasis>config_inc.php</emphasis>. Options stored in database
```
dregad: *the* database

docbook/Admin_Guide/en-US/Configuration.xml
```diff
@@ -5,6 +5,57 @@
...
+		can also be viewed and updated on
+		"Manage > Manage Configuration > Configuration Report" page.
+	</para>
+
+	<para>That's how MantisBT determines which value to use if an option was overridden:</para>
```
dregad: *The sequence below describes how* Mantisbt...

docbook/Admin_Guide/en-US/Configuration.xml
```diff
@@ -5,6 +5,57 @@
...
+
+	<orderedlist>
+		<listitem>
+			<para>Look for the specified configuration option in database
+				for current user and current project.
+			</para>
+		</listitem>
+		<listitem>
+			<para>If not found, look for the specified configuration option in
```
dregad: I don't think it's necessary to repeat *If not found* for each step, it's implicit

dregad:
Typo in commit message: documenta*t*ion

dregad:

docbook/Admin_Guide/en-US/Configuration.xml
evilchewits: If so, maybe to add one more phrase before the enumeration like this one:
"If the value is not found at a certain step it goes to the next. Otherwise, stops searching and starts using the found value."
You won't object?

evilchewits:
Commit has been updated forcibly following @dregad remarks

docbook/Admin_Guide/en-US/Configuration.xml
```diff
@@ -5,6 +5,59 @@
...
+			</para>
+		</listitem>
+		<listitem>
+			<para>Look for the specified configuration option in the
+				database for all users and all projects.
+			</para>
+		</listitem>
+		<listitem>
+			<para>Look for the specified configuration option in
+				<emphasis>config_inc.php</emphasis> (where you might have some
```
cproensa: > (where you might have some code to distinguish between projects and/or users)

I wouldn't encourage that usage of config_inc. Is that suggested in other place of documentation?

atrol: My post in the forum has been written for troubleshooting of this specific issue, not as some kind of documentation or howto.

I think it's better to understand if there is something similar to  what I wrote in the post:

> I recommend database configuration if you want to set options for a special project but not all projects, or a special user, but not all users.

Specific might be a better word than special for it.
I am no native English speaker, that's why I hesitate to write documentation ;-)

dregad: Based on @cproensa and @atrol's feedback, I would suggest to drop that phrase `(where you...`

Also, to avoid repetition, rewording of the opening paragraph `The sequence below` and remove `Look for the specified configuration option in` from the individual list items.

evilchewits:
My next attempt is in a new commit. Is it better or not? Unfortunately, I'm not a native English speaker, too :)

dregad:
Hi @evilchewits I merged your contribution with a few adjustments, see e98a8fa387e6270df59bf4472264b8e4f077f41b.

PR #776, opened 2016-05-11
Do not display the access level in bugnotes when it is 0 (ANYBODY)

atrol:
Prior to this, the access level display in bugnotes was (@0@) if the
user who added the note did not have access to the project. This
situation can happen when an issue is moved to a private project, or
when a user's rights to a private project are revoked.
This is similar to the fix for #11923, but affects email notification.

Fixes #20897

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -1549,11 +1549,13 @@ function email_format_bug_message( array $p_visible_bug_data ) {
 			$t_time_tracking = '';
 		}
 
+		$t_access_level_string = ' - ';
 		if( user_exists( $t_bugnote->reporter_id ) ) {
 			$t_access_level = access_get_project_level( $p_visible_bug_data['email_project_id'], $t_bugnote->reporter_id );
-			$t_access_level_string = ' (' . get_enum_element( 'access_levels', $t_access_level ) . ') - ';
-		} else {
-			$t_access_level_string = '';
+			# Only display access level when higher than 0 (ANYBODY)
+			if( $t_access_level > ANYBODY ) {
+				$t_access_level_string = ' (' . get_enum_element( 'access_levels', $t_access_level ) . ') - ';
+			}
 		}
 
 		$t_string = ' (' . $t_formatted_bugnote_id . ') ' . user_get_name( $t_bugnote->reporter_id ) . $t_access_level_string . $t_last_modified . "\n" . $t_time_tracking . ' ' . $t_bugnote_link;


core/email_api.php
```diff
@@ -1549,11 +1549,13 @@ function email_format_bug_message( array $p_visible_bug_data ) {
 			$t_time_tracking = '';
 		}
 
+		$t_access_level_string = ' - ';
```
vboctor: Looks like the ' - ' should be part of the string concatenation below and not the access level string.  I know it wasn't part of your change.  But before we used to skip this ' - ' in some cases, now you seem to always have it.

core/email_api.php
```diff
@@ -1549,11 +1549,13 @@ function email_format_bug_message( array $p_visible_bug_data ) {
...
 		if( user_exists( $t_bugnote->reporter_id ) ) {
 			$t_access_level = access_get_project_level( $p_visible_bug_data['email_project_id'], $t_bugnote->reporter_id );
-			$t_access_level_string = ' (' . get_enum_element( 'access_levels', $t_access_level ) . ') - ';
-		} else {
-			$t_access_level_string = '';
+			# Only display access level when higher than 0 (ANYBODY)
```
vboctor: I wonder if we should show a string that user as no access, e.g. 'no access' or something similar.  This will make it explicit that such user no longer has access and hence won't receive notifications about future updates for the issue or even @ mentions.

vboctor:
My main comment is about showing an explicit string that indicates that user has not access to the issue and won't get any change notifications.

dregad:
> showing an explicit string that indicates that user has not access to the issue

+1, good idea.

core/email_api.php
atrol: > But before we used to skip this ' - ' in some cases, now you seem to always have it.

This was an intended change as this fixes another minor issue.
I was too lazy to enter a new issue and create a separate PR for it.

If the user has been deleted, user name and date display are not separated.
Before the change, you got something like
(0000597) user22016-05-11 12:38
now you get
(0000597) user2 - 2016-05-11 12:38

atrol:
> showing an explicit string that indicates that user has not access to the issue

Not sure, shouldn't this be a separate task?

Option 1)

Introduce a new helper function to deliver the access level string.
Use this function also in bugnote_view_inc.php where we have at the moment
```php
				$t_access_level = access_get_project_level( null, (int)$t_bugnote->reporter_id );
				# Only display access level when higher than 0 (ANYBODY)
				if( $t_access_level > ANYBODY ) {
					echo '(', get_enum_element( 'access_levels', $t_access_level ), ')';
				}
```
Something like
```php
function get_access_level_string ( $p_access_level )
{
	if( $p_access_level > ANYBODY ) {
		$t_access_level_string = get_enum_element( 'access_levels', $p_access_level );
	} else {
		$t_access_level_string = lang_get( 'no_access' );
	}
	return ( $t_access_level_string );
}
```

Option 2)

Remove special treatment for access level ANYBODY, just set
```php
$g_access_levels_enum_string = '0:No access,10:viewer,25:reporter,40:updater,55:developer,70:manager,90:administrator';
```
and $s_access_levels_enum_string in language files

Of course, this would introduce side effects.
E.g. you can set access level of a user to "No Access" as you will see the entry in drop down lists.

dregad:
> Not sure, shouldn't this be a separate task?

It can certainly be treated as a separate issue if you want to.

With regards to the 2 proposed options, 2 has a certain elegance to it, but I'm concerned about the potential for regressions. 1 is cleaner and safer.

vboctor:
:+1: for option 1.

core/email_api.php
cproensa: > I wonder if we should show a string that user as no access

Thas is a nice idea. I'd say, place a notation next to the user name, that can be a string as you suggested, or a colored asterisk with a popup text (for a leaner ui)

This notice should be showed on most user strings printed in the page: user for notes, reporter user, etc.

In my instance i have a plugin that prints a warning in the bug page when the reporter has no access, or has access but cant write notes. This is needed for me because bugs are moved between private projects, and you may not notice that the reporter will no longer have access in some cases.

Can we open a separate feature request?

atrol: I opened new PR https://github.com/mantisbt/mantisbt/pull/778 where I introduced a string for users with no access.

atrol:
Closing this one, opened new PR https://github.com/mantisbt/mantisbt/pull/778 to implement Option 1

PR #777, opened 2016-05-16
Clean-up account pages

cproensa:
Some minor adjustments:
- Remove some old code after implementation of new verify page:  d7b8d33e79250e67888e2d6e82ee1e5f2bfd091f
- Add "account menu" in tokens page

### DIFF
--- a/account_page.php
+++ b/account_page.php
@@ -102,32 +102,18 @@ $u_email = user_get_email( $u_id );
 # If the password is the default password, then prompt user to change it.
 $t_reset_password = $u_username == 'administrator' && auth_does_password_match( $u_id, 'root' );
 
-# note if we are being included by a script of a different name, if so,
-# this is a mandatory password change request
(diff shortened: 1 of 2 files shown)


dregad:
> Remove some old code after implementation of new verify page: d7b8d33

Reference the relevant issue ID from #735 in the commit message

> Add "account menu" in tokens page

This refers to #685, so I would suggest at least adding `Issue #17766` in the commit message, or even better considering that the feature was released in rc1, open a separate, related issue for tracking.

Otherwise, tested OK, +1 from me

cproensa:
commit texts rewritten, to include issue ids

dregad:
Fixed in 73f2cf8c171848c2f007228374cde87032ff8ff2. I created [#20943](https://www.mantisbt.org/bugs/view.php?id=20943) to track the account menu, I think it's cleaner to track this as a separate bug fix.

PR #778, opened 2016-05-19
Enhance access level display

atrol:
Prior to this, the access level display in bugnotes was (@0@) if the
user who added the note did not have access to the project. This
situation can happen when an issue is moved to a private project, or
when a user's rights to a private project are revoked.
This is similar to the fix for #11923, but
- affects also email notification
- displays "no access" for existing users instead of displaying nothing

Fixes #20897

### DIFF
--- a/core/access_api.php
+++ b/core/access_api.php
@@ -695,3 +695,17 @@ function access_get_status_threshold( $p_status, $p_project_id = ALL_PROJECTS ) 
 		}
 	}
 }
+
+/**
+ * Given a access level, return the appropriate string for it
+ * @param integer $p_access_level
+ * @return string
+ */
+function access_level_get_string( $p_access_level ) {
+	if( $p_access_level > ANYBODY ) {
+		$t_access_level_string = get_enum_element( 'access_levels', $p_access_level );
+	} else {
+		$t_access_level_string = lang_get( 'no_access' );
+	}
+	return $t_access_level_string;
+}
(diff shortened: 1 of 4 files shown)


atrol:
@dregad, the thumbs up emoji looks nice and makes it easy for other users to add one more thumbs up.
The downside is, that you get no email notifications for it.
So I prefer a +1 at the moment.

dregad:
+1 :grin:

core/access_api.php
```diff
@@ -693,3 +693,18 @@ function access_get_status_threshold( $p_status, $p_project_id = ALL_PROJECTS )
...
 	}
 }
+
+/**
+ * Given a access level, return the appropriate string for it
+ * @param integer $p_access_level
+ * @return string
+ */
+function access_level_get_string ( $p_access_level )
+{
```
vboctor: Move to same line as function signature.

core/access_api.php
```diff
@@ -693,3 +693,18 @@ function access_get_status_threshold( $p_status, $p_project_id = ALL_PROJECTS )
...
+	if( $p_access_level > ANYBODY ) {
+		$t_access_level_string = get_enum_element( 'access_levels', $p_access_level );
+	} else {
+		$t_access_level_string = lang_get( 'no_access' );
+	}
+	return ( $t_access_level_string );
```
vboctor: The () are not necessary.

vboctor:
👍

core/access_api.php
dregad: And while you're at it, no space before opening `(`

PR #781, opened 2016-05-27
Do not wrap inline items for custom field value lists

cproensa:
Avoid wrapping of each individual item that is part of a custom
fields value list, displayed as inline inputs.

Fixes #0020964

### DIFF
--- a/core/cfdefs/cfdef_standard.php
+++ b/core/cfdefs/cfdef_standard.php
@@ -321,10 +321,11 @@ function cfdef_input_checkbox( array $p_field_def, $p_custom_field_value ) {
 	for( $i = 0; $i < count( $t_values ); $i++ ) {
 		$t_input_id = 'custom_field_' . $p_field_def['id'] . '_value_' . $i;
 		$t_input_name = 'custom_field_' . $p_field_def['id'] . '[]';
-		echo '<input id="$t_input_id" ' . helper_get_tab_index() . ' type="checkbox" name="' . $t_input_name . '" value="' . string_attribute( $t_values[$i] ) . '"';
+		echo '<label class="nowrap">';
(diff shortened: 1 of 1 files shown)


dregad:
+1

Any particular reason why you removed the *for* attribute on the labels ?

core/cfdefs/cfdef_standard.php
```diff
@@ -321,10 +321,11 @@ function cfdef_input_checkbox( array $p_field_def, $p_custom_field_value ) {
 	for( $i = 0; $i < count( $t_values ); $i++ ) {
 		$t_input_id = 'custom_field_' . $p_field_def['id'] . '_value_' . $i;
 		$t_input_name = 'custom_field_' . $p_field_def['id'] . '[]';
-		echo '<input id="$t_input_id" ' . helper_get_tab_index() . ' type="checkbox" name="' . $t_input_name . '" value="' . string_attribute( $t_values[$i] ) . '"';
+		echo '<label class="nowrap">' ;
```
dregad: extra space before `;`

dregad:
Missing `Fixes #20964` reference in commit message

cproensa:
> Any particular reason why you removed the for attribute on the labels ?

Seems like having the input inside the label block, makes the label to apply the nowrap property to the whole content. (in the example of the checkbox list)
With the "for" syntax, the label and input are rendered as independent flowed blocks, and most of the times the checkbox is placed in a different line that the label text, if line wrap occurs.

(*) in chrome there's a small chance that the checkbox is separated from the label, when no part of the label text fits in the current line. Firefox, however, works better in that situation, keeping the checkbox always attached to the label text...

cproensa:
> extra space before ;
> Missing Fixes #20964 reference in commit message.

fixed
thanks.

dregad:
Fixed in 1d3af0c53509061800c001f25ed2ffabc8998ef0

PR #782, opened 2016-05-31
Do not enforce related thresholds when sponsorship is OFF

dregad:
When $g_enable_sponsorship = OFF, we should not enforce related
thresholds ($g_handle_sponsored_bugs_threshold and
$g_assign_sponsored_bugs_threshold) when updating issues.

Fixes [#21030](https://www.mantisbt.org/bugs/view.php?id=21030)


atrol:
There are similar changes needed in bug_change_status_page.php.

dregad:
@atrol thanks for the heads up, I missed that one.

atrol:
+1 (just code review, no test)

cproensa:
by the way, asigning with bug actiongroup does not validate anything about sponsorship

atrol:
>  asigning with bug actiongroup does not validate anything about sponsorship

same for SOAP API

PR #784, opened 2016-06-04
Support $g_debug_email being set to OFF

vboctor:
For backward compatibility support this config option set to OFF.   Also update the documentation of the configuration option to reflect that '' is the new way of reflecting that debug emails are off.

### DIFF
--- a/docbook/Admin_Guide/en-US/config/logging.xml
+++ b/docbook/Admin_Guide/en-US/config/logging.xml
@@ -31,12 +31,12 @@
 			<term>$g_debug_email</term>
 			<listitem>
 				<para>Used for debugging e-mail notifications.
-					When it is OFF, the emails are sent normally.
+					When it is '', the emails are sent normally.
 					If set to an e-mail address, all messages are sent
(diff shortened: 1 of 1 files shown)


core/email_api.php
```diff
@@ -1144,7 +1144,9 @@ function email_send( EmailData $p_email_data ) {
 	$t_mail->LE         = "\r\n";
 	$t_mail->Encoding   = 'quoted-printable';
 
-	if( !empty( $t_debug_email ) ) {
+	# String '' or value OFF (backward compatibility with old configs) are considered as debugging
+	# is not enabled.
+	if( !empty( $t_debug_email ) && $t_debug_email !== OFF ) {
```
dregad: As per [my note ~52738](https://www.mantisbt.org/bugs/view.php?id=20684#c52738), this change is really not necessary, since in PHP, `'' == 0`.

The first predicate will be only be true when $t_debug_email != [0, '' or false], so the 2nd is either not evaluated or true.

Suggest revert to original code.

PR #786, opened 2016-06-12
User @ mentions improvements

dregad:
This PR covers the following changes

- add new test case for, and fix [#21083](https://www.mantisbt.org/bugs/view.php?id=21083) 
- refactoring of unit tests to use a more efficient and easier to maintain data provider function approach
- code cleanup
- replace deprecated `<strike>` tag

The mentions parser in *mention_get_candidates()* has been simplified by going back to a regex approach which hopefully covers all the cases. This should also improve performance.

There is one failing unit test case: *MentionWithInvalidChars*; I did not fix it because I feel that the assertion is incorrect (i.e. IMO it should not fail), as discussed in https://github.com/mantisbt/mantisbt/commit/a9a20393d0ed765a72d69f0f68e536458748b894#commitcomment-17835719.

Even though the test cases are OK from a phpUnit standpoint, there are still issues with the regex, as it does not play nice with some of them when used in a bugnote.

So this is still a work-in-progress, but I thought I'd share and get some feedback.

### DIFF
--- a/core/mention_api.php
+++ b/core/mention_api.php
@@ -168,10 +168,18 @@ function mention_format_text( $p_text, $p_html = true ) {
 		$t_formatted_mentions[$t_mention] = $t_mention_formatted;
 	}
 
-	$t_text = str_replace(
-		array_keys( $t_formatted_mentions ),
-		array_values( $t_formatted_mentions ),
-		$p_text
+	# Replace the mentions, ignoring existing anchor tags (otherwise
+	# previously set mailto links would be processed as mentions,
+	# corrupting the output
+	$t_text = string_process_exclude_anchors(
+		$p_text,
+		function( $p_string ) use ( $t_formatted_mentions ) {
+			return str_replace(
+				array_keys( $t_formatted_mentions ),
+				array_values( $t_formatted_mentions ),
+				$p_string
+			);
+		}
 	);
 
 	return $t_text;


dregad:
> there are still issues with the regex, as it does not play nice with some of them when used in a bugnote.

Actually I think this may not be caused by the regex, but by mantion_format_text() trying to process previously formatted mailto links. Will look into it tomorrow.

vboctor:
The key here is to have a comprehensive list of test cases.  Whenever we are not sure, I would be consistent with github.  The failing test case seems to be inconsistent with github, hence, I'm ok to remove it.

dregad:
Added commits to
- change the expected return value for the `@vboctor%%%` test case as agreed
- fix problem with mention API processing existing anchor tags, corrupting mailto links

With this, the PR should be ready for merge. Let me know your feedback.

core/mention_api.php
```diff
@@ -168,12 +168,20 @@ function mention_format_text( $p_text, $p_html = true ) {
...
+			return str_replace(
+				array_keys( $t_formatted_mentions ),
+				array_values( $t_formatted_mentions ),
+				$p_string
+			);
+		}
 	);
 
-	return $t_text;
+	return $t_text ;
```
vboctor: Remove space before semi-colon.

vboctor:
:+1:

core/mention_api.php
dregad: Fixed

PR #787, opened 2016-06-14
Bump the required MantisCore version in bundled plugins

syncguru:
This fix is needed for plugins to function properly with 2.0.0

Please note that we may need to followup with another PR once the plugin and MantisCore versions are finalized.


vboctor:
Thanks @syncguru - I'm closing this in favor of #788 where I incorporated your commit and added another change to update versions of the actual plugins.

PR #788, opened 2016-06-14
Fix core plugin versions

vboctor:
Update core plugins as follows:
- Support/require MantisCore 2.0.0.
- Use MANTIS_VERSION for core plugin versions.

I've noticed that plugins used to require 1.3.0 and now 2.0.0 and that matches pre-releases of such release numbers.  We can look at that later and see if we want to keep this behavior or change it.


syncguru:
Looks good.

dregad:
+1

PR #789, opened 2016-06-15
Fix numerous validation issues

badfiles:


manage_proj_cat_edit_page.php
```diff
@@ -68,7 +68,7 @@
 print_manage_menu( 'manage_proj_cat_edit_page.php' );
 ?>
 
-<div class="col-md-12 col-xs-12" xmlns="http://www.w3.org/1999/html">
+<div class="col-md-12 col-xs-12" xmlns="http://www.w3.org/1999/xhtml">
```
dregad: @syncguru why do we need an xml namespace here ?

view_user_page.php
```diff
@@ -75,7 +75,7 @@
 
 layout_page_begin();
 ?>
-<div class="col-md-12 col-xs-12" xmlns="http://www.w3.org/1999/html">
+<div class="col-md-12 col-xs-12" xmlns="http://www.w3.org/1999/xhtml">
```
dregad: same as above

manage_proj_cat_edit_page.php
badfiles: Because there is no http://www.w3.org/1999/html namespace.
xhtml != xml

syncguru: I don't think xmlns attribute is needed. You can remove it. I don't recall if I added this and for what reason.

view_user_page.php
syncguru: Same as above.

core/layout_api.php
```diff
@@ -397,14 +397,14 @@ function layout_navbar() {
 
 	# mobile view
 	echo '<div class="hidden-sm hidden-md hidden-lg">';
-	echo '<nav class="navbar-menu pull-left navbar-collapse collapse" role="navigation" style="height: auto;">';
```
syncguru: I don't think we need the <nav> element anymore. It used to be part of the UI theme in early version but does not seem to be the case anymore.

http://responsiweb.com/themes/preview/ace/1.4/index.html

dregad: @syncguru So you're saying it should be removed entirely ?

syncguru: @dregad 
No. I would switch "<nav" with "<div" instead. I just tried it and it seems to work just fine.

view_all_inc.php
```diff
@@ -223,7 +223,7 @@ function write_bug_rows( array $p_rows ) {
 <?php
 		if( $g_checkboxes_exist ) {
 			echo '<label class="inline">';
-			echo '<input class="ace check_all" type="checkbox" id="bug_arr_all" name="bug_arr_all" value="all" class="check_all input-small" />';
+			echo '<input class="ace check_all input-small" type="checkbox" id="bug_arr_all" name="bug_arr_all" value="all" />';
```
syncguru: What is 'input-small'? did you mean 'input-sm'?

badfiles: Maybe, but as you can see I just merged 'classes'. Should it be 'input-sm' here?

syncguru: yes

PR #791, opened 2016-06-16
Restore password page should not ask for a username

badfiles:


dregad:
Yes it does, because we may have multiple user Ids using the same e-mail.

badfiles:
I disagree, as we have '$g_email_login_enabled'.
Maybe it should be conditional on  $g_email_login_enabled?

PR #792, opened 2016-06-16
Unifiy 'echo' usage

badfiles:


dregad:
> echo with multiple parameters should be faster

Please provide evidence of faster execution time as applied to Mantis in a real-life scenario (i.e. I'm not interested in a 'lab' test running comparing 4 million iterations of `echo $x . $y;` vs `echo $x , $y;`).

Unless we see a significant difference, I don't see much point in this change.

vboctor:
I agree with @dregad -- I'm not a fan of churn that doesn't necessarily add much user value.  I suspect our performance bottleneck will be mostly around access to db and not echo statements.

badfiles:
Anyway, we should decide we use it one way or another.
Now both ways are used, sometimes mixing them up.
'echo $a, $b;' is synthetically faster, no extra space.
'echo $a . $b;' is better for copy-paste to '$c = $a . $b;' and does not give $a in case function $b has an error.

dregad:
Based on feedback from @vboctor and @syncguru, I'm closing this PR without merging it.

Feel free to reopen if you can provide evidence of a significant benefit for implementing it.

badfiles:
OK, I agree, no significant benefit. (though I never said there was significant memory or cpu benefit)
Let's forget about it.

PR #794, opened 2016-06-17
Restore code that has been lost during modern-ui merge.

badfiles:


syncguru:
👍

account_page.php
```diff
@@ -191,7 +191,7 @@
 			} ?>
 			<tr>
 				<td class="category">
-					<span class="required"><?php if( $t_force_pw_reset ) { ?> * <?php } ?></span> <?php echo lang_get( 'password' ) ?>
+					<span class="required"><?php if( $t_force_pw_reset ) { ?> * <?php } ?></span> <?php echo lang_get( 'new_password' ) ?>
```
vboctor: This doesn't seem right.  This should be password.

vboctor: Nevermind, I was looking at a slightly older code.

vboctor:
Merged via 0be20bdbe6bef6ff2a3317e43c8d075c98878aa4

PR #795, opened 2016-06-18
Fix news layout

badfiles:
+remove buggy '?>' from news_update.php

### DIFF
--- a/news_update.php
+++ b/news_update.php
@@ -71,9 +71,9 @@ echo '<div class="space-20"></div>';
 echo '<div class="alert alert-success">';
 echo lang_get( 'operation_successful' );
 echo '<br /></div>';
-echo '<div class="pull-left padding-right-2">';
 
 print_button( 'news_edit_page.php?news_id=' . $f_news_id . '&action=edit', lang_get( 'edit_link' ) );
+echo '&nbsp;&nbsp;&nbsp;&nbsp;';
 print_button( 'news_menu_page.php', lang_get( 'proceed' ) );
 
 echo '<br /><br />';
--- a/news_update.php
+++ b/news_update.php
@@ -70,8 +70,8 @@ layout_page_begin( 'main_page.php' );
 echo '<div class="space-20"></div>';
 echo '<div class="alert alert-success">';
 echo lang_get( 'operation_successful' );
-
-echo '<br />';
+echo '<br /></div>';
+echo '<div class="pull-left padding-right-2">';
 
 print_button( 'news_edit_page.php?news_id=' . $f_news_id . '&action=edit', lang_get( 'edit_link' ) );
 print_button( 'news_menu_page.php', lang_get( 'proceed' ) );
@@ -79,7 +79,5 @@ print_button( 'news_menu_page.php', lang_get( 'proceed' ) );
 echo '<br /><br />';
 
 print_news_entry( $f_headline, $f_body, $t_row['poster_id'], $f_view_state, $f_announcement, $t_row['date_posted'] );
-
-echo '</div>';
 
 layout_page_end();
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -502,16 +502,21 @@ function print_news_entry( $p_headline, $p_body, $p_poster_id, $p_view_state, $p
 	$t_body = string_display_links( $p_body );
 	$t_date_posted = date( config_get( 'normal_date_format' ), $p_date_posted );
 
-	$t_news_css = VS_PRIVATE == $p_view_state ? 'hidden' : '';
+	$t_news_css = VS_PRIVATE == $p_view_state ? 'widget-color-red' : 'widget-color-blue2';
 	?>
 
 	<div class="space-10"></div>
-	<div class="widget-box widget-color-blue2">
-		<div class="widget-header widget-header-small <?php echo $t_news_css; ?>">
+	<div class="widget-box <?php echo $t_news_css ?>">
+		<div class="widget-header widget-header-small">
 			<h4 class="widget-title lighter">
 				<i class="ace-icon fa fa-edit"></i>
 				<?php echo $t_headline ?>
 			</h4>
+			<div class="widget-toolbar">
+				<a data-action="collapse" href="#">
+					<i class="ace-icon fa fa-chevron-up bigger-125"></i>
+				</a>
+			</div>
 		</div>
 
 		<div class="widget-body">


news_update.php
```diff
@@ -69,9 +69,8 @@
 
 echo '<div class="space-20"></div>';
 echo '<div class="alert alert-success">';
-echo lang_get( 'operation_successful' ) ?>
-
-echo '<br />';
+echo lang_get( 'operation_successful' );
+echo '<br /></div><div class="pull-left padding-right-2">';
```
vboctor: Where is the end for the div opened here?

news_update.php
```diff
@@ -69,17 +69,15 @@
...
 echo '<div class="space-20"></div>';
 echo '<div class="alert alert-success">';
-echo lang_get( 'operation_successful' ) ?>
-
-echo '<br />';
+echo lang_get( 'operation_successful' );
+echo '<br /></div>';
 
 print_button( 'news_edit_page.php?news_id=' . $f_news_id . '&action=edit', lang_get( 'edit_link' ) );
+echo '&nbsp;&nbsp;&nbsp;&nbsp;';
```
vboctor: 4 spaces seem strange.  Should we be using some css style or equivalent instead?

badfiles: Everything in modern UI seems strange. I just follow the present style.

atrol: > I just follow the present style.

What do you mean with "present style"?
It did not find similar echoing of &nbsp in code.

atrol: Thanks, got it.

syncguru: @badfiles It would be good to more descriptive about what is strange.

badfiles: To start with, on what terms do we use Ace-Responsive? I could not find it's license in Mantis files.
How should we update it? It sure does and will have bugs.

syncguru: 1- Mantishub has extended license
2- Admin themes are very good fit for bug-tracker app. Ace was top seller back then when this effort started (around May 2014) 
3- We have full source code control. Also the author is reachable on twitter. 

@badfiles None of your questions are either technical or point to anything strange.  Saying "everything is strange" is counter productive. I am happy to answer any question you may have but let's keep the conversation productive.

atrol: Bad news
> Ace admin template is retired and not for sale 
https://wrapbootstrap.com/theme/ace-responsive-admin-template-WB0B30DGR

@syncguru would it be good to have the non minified versions of the files to fix issues?
If so, they should be included in our source and we would have to add a step to our build process to create the minified versions.

grangeway: Roland, Badfiles - I've been in discussion with the FSF Licensing help desk for the past week regarding the addition of this template. I've sent the following email to the mantisbt-dev list:

Hello,

I have been in discussion with the GPL/FSF licensing team for the last week regarding a recent commit to mantisbt.

The following files were added to the open source mantisbt project which is GPL licensed:

1) https://github.com/mantisbt/mantisbt/blob/master/js/ace-elements.min.js
2) https://github.com/mantisbt/mantisbt/blob/master/js/ace-extra.min.js
3) https://github.com/mantisbt/mantisbt/blob/master/js/ace.min.js
4) https://github.com/mantisbt/mantisbt/blob/master/css/ace.min.css
5) https://github.com/mantisbt/mantisbt/blob/master/css/ace-ie.min.css
6) https://github.com/mantisbt/mantisbt/blob/master/css/ace-rtl.min.css

These files are licensed under a commercial license ( https://wrapbootstrap.com/theme/ace-responsive-admin-template-WB0B30DGR ) for which a fee may be payable by the end user.

The license options are available at http://support.wrapbootstrap.com/knowledge_base/topics/usage-licenses. 

Item 6 of the licenses state that "If the item was created using materials which are the subject of a GNU General Public License (GPL), your use of the item is subject to the terms of the GPL in place of the foregoing conditions (to the extent the GPL applies)."

A copy of the GPL license is available at https://www.gnu.org/licenses/gpl-3.0.en.html.

My understanding and suggested by the licensing@fsf.org team (although they are not lawyers) is that would be a violation of the GPL to add these files in this way.

The advice received from their team in instances like this was to have a look at http://www.gnu.org/licenses/gpl-violation.html, and reach out to the GPL project, and try to resolve the issue within the community. If that fails, their advice is to contact http://softwarefreedom.org/ who may offer free legal services to open developers.

My reason for a number of years for contributing to the MantisBT project was it was an open source project, and the original authors when i joined had said would not make money from the project. From a personal point of view, I did not want to spend my time to aid a commercial company, so that was an important goal.

Whilst I've been doing some research into the situation, badfiles asked the question of https://github.com/mantisbt/mantisbt/pull/795#discussion_r67747346 . The reply from syncguru seems to confirm that the files listed in lines 1-6 above, which I believed were from the commercial licensed ace admin template, are indeed from the commercial template, as indicated at https://github.com/mantisbt/mantisbt/pull/795#discussion_r67768649

He states:
"1- Mantishub has extended license
2- Admin themes are very good fit for bug-tracker app. Ace was top seller back then when this effort started (around May 2014) 
3- We have full source code control. Also the author is reachable on twitter."

This would seem to confirm that the code is indeed commercial, and equally that the license is owned by a separate commercial company "Mantishub". 

My questions to the FSF and their responses suggested that the first step is to reach out in a friendly manner and explain the problem - hopefully the above achieve this.

They then suggest proposing a solution, and that if that fails the next step should be contacting the Software Freedom Law Center, which provides pro-bono legal services to developers of Free, Libre, and Open Source Software (http://softwarefreedom.org/).

I'd like to firstly provide the option for the developers of the mantisbt project to provide an explanation if I've misunderstood anything in the GPL in relation to this, but having asked various linux licensing groups/people over the last week, I suspect this is unlikely.

In terms of workable solutions, I see two options:

a) The files listed in 1-6 above (and there are a couple of extra files which i've not yet evaluated) are removed from the mantisbt project. 

b) Any contributions that I (and any other contributor that feels the same) have made to the mantisbt project are removed, such that they no longer have any code in the project.

At this stage, I'm going to await other responses, and do nothing else apart from send this email and reply as appropriately.

I've been informed of the following: "However if you and others have GPL'd contributions in the project then you have the legal power to enforce the terms of the software and refuse to let anyone distribute your code linked to proprietary/non-free software."

On that basis, and in case users read this and are concerned - I'd like to make clear that at this point, that the released versions of mantisBT only contain GPL compatible code, and therefore there is no issues with distribution of that code. The code i'm emailing about only exists in the github repository.

Thanks
Paul

badfiles:
@syncguru, technically, strange != bad, strange == (strange to me).
Everything != (literally everything), but hey, in my native language half of Latin letters are strange, so that is still questionable. Joking aside, a lot is strange.
1. License issues with the project itself and with it's further usage do exist.
2. How MU PR could be merged while the core team keeps very high integrity level is unclear.
3. I think, technically strange moments should be discussed in corresponding PR's.

Nonetheless, it is admirable work to give an old-school project such a great appearance.

news_update.php
badfiles: @grangeway, may I also suppose the mentioned files are distributed under GPL as the part of  GPL-software, while inside none of them, or anywhere in the project, other copyright and/or license notice present?
As a fair contributor and/or user, should I investigate derivation of the code in the files, and if found to be proprietary, accept the corresponding non-free license and be somehow obliged to the owner?
It is very sad the things have turned this way.
I hope an unharmful solution will be found, after all the author decided to retire Ace Admin, he or she might release it under a GPL-compatible license.

atrol: IANAL and I am not a native English speaker.
I did not see a problem after reading this in the license terms.
Maybe I am getting something wrong.

> If the item was created using materials which are the subject of a GNU General Public License (GPL), your use of the item is subject to the terms of the GPL in place of the foregoing conditions (to the extent the GPL applies).

syncguru:
@atrol good suggestion. Will submit the non-minified files.

news_update.php
badfiles: @atrol, please read the whole license at http://support.wrapbootstrap.com/knowledge_base/topics/usage-licenses
The item here is not Mantis, but Ace Admin template, and if it was based on a GPL code, it would be under GPL license, but so far we know it is not.
We have a completely different situation here, Ace Admin template (which itself might be under GPL, if it had a GPL basis) is non-exclusively and non-transferably incorporated into MantisBT according to the license preamble.

badfiles: As I understood, Mantishub LLC, represented here by @syncguru, purchased Extended License of   Ace Admin template (AAT) and incorporated it into the opensource MantisBT project, making AAT subject to GPL, which AAT's authors would not approve and sure might kindly ask to remove copyrighted materials from the MantisBT project.

grangeway: @badfiles I think you've summarized quite well. If you wanted to redistribute the code in your repository of mantisbt for instance, I think you'd then be unable to without purchasing a license yourself (of the admin template).

My understanding was that you couldn't include propriety code for this reason and was why a) a number of projects 'dual license' technology for scenarios like this. b) why people suggest that sometimes the MIT licenses should be used over the GPL.

From my point of view, it may be I need to contact the software freedom center with an initial email and see what advice they give. I spent over 10 years on the mantisbt project on that basis that the code would always be free to use for anyone in any way.

grangeway: @badfiles although in this case, the situation may also be somewhat different - AAT authors may or may not approve with the way there template is used.  I'm querying as a contributor to MantisBT whether the inclusion by Mantishub LLC of commercial code in code that I have an interest in is allowed. 

The drupal project ( https://www.drupal.org/about/licensing#q1 ) is licensed under the GPL. They state:

Does the license cover just PHP, or everything?

> 
> We require that all files (PHP, JavaScript, images, Flash, etc.) hosted on Drupal.org be under the GPL. If it's in Git, then it is under the same license (GPL version 2 or later). That way there is no confusion about what license a given file is under.
> 
> Who holds the copyright for Drupal?
> 
> Drupal and contributed modules and themes include code written by thousands of different people. All Drupal contributors retain copyright on their code, but agree to release it under the same license as Drupal. If you are unable or unwilling to contribute a patch under the GPL version 2 or later, do not submit a patch.

This would seem to support my analysis - so from a drupal standpoint - and I suspect as a larger project they have probably checked out the situation properly, the patch that includes the AAT should be licensed under the GPL or not submitted - so the question here may be - is the AAT available under GPL2?

dregad:
@badfiles, @syncguru after all this discussion, what should be done about the original submitted change ?

Regarding submission of non-minified ACE-template files, we have agreement that this will be done, and this is tracked in [#21603](https://www.mantisbt.org/bugs/view.php?id=21603).

If still valid, @badfiles please rebase the branch as it now has conflicts with latest master.

dregad:
This was merged in bfbb8de080b6a98de234c430f9dd76a524f25c16

PR #796, opened 2016-06-18
Fix config parsing issues and improve unit tests

dregad:
Covers the following points

- fixed regressions introduced by the config parser ([21124](https://www.mantisbt.org/bugs/view.php?id=21124))
  - *No more tokens* exception with empty values (`''`, `0`, etc)
  - *Unexpected token "-"* exception with negative numbers (and the same with "+")
- fixed config value corruption when editing a value containing newlines [21136](https://www.mantisbt.org/bugs/view.php?id=21136)
- improved handling of empty values
- added new unit test cases
- improved and simplified unit tests through use of data providers

### DIFF
--- a/adm_config_report.php
+++ b/adm_config_report.php
@@ -104,9 +104,9 @@ function print_config_value_as_string( $p_type, $p_value, $p_for_display = true 
 			echo (integer)$p_value;
 			return;
 		case CONFIG_TYPE_STRING:
-			$t_value = string_nl2br( string_html_specialchars( config_eval( $p_value ) ) );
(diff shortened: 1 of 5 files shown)


vboctor:
:+1:

atrol:
While looking at the source I found another minor bug in line 180 of Tokenizer.class.php
This line will output $p_value instead of the actual value of $p_value
```php
				'Invalid token: got "' . $this->value() . '", expected "$p_value"'
```

atrol:
:+1: for the current changes

PR #797, opened 2016-06-18
Replace UTF-8 4bytes chars by '�' (U+FFFD) on MySQL

dregad:
Workaround applied to the following fields:

- bug.summary
- bug.description
- bug.steps_to_reproduce
- bug.additional_information
- bugnote.text
- custom fields with type *string* and *text area*

Let me know if you think other fields need to be fixed as well.

Fixes [#21101](https://www.mantisbt.org/bugs/view.php?id=21101)

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -318,11 +318,21 @@ class BugData {
 				}
 				break;
 			case 'summary':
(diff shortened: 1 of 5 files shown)


atrol:
File core/u.php not needed

dregad:
Fixed.

vboctor:
👍

PR #801, opened 2016-06-19
Use "Google Fonts" in cdn mode

badfiles:


core/layout_api.php
```diff
@@ -230,17 +230,20 @@ function layout_head_css() {
...
 		html_css_cdn_link( 'https://maxcdn.bootstrapcdn.com/font-awesome/' . FONT_AWESOME_VERSION . '/css/font-awesome.min.css' );
+
+		# theme text fonts
+		html_css_cdn_link( 'https://fonts.googleapis.com/css?family=Open+Sans' );
 	} else {
 		html_css_link( 'bootstrap-' . BOOTSTRAP_VERSION . '.min.css' );
 		html_css_link( 'font-awesome-' . FONT_AWESOME_VERSION . '.min.css' );
+
+		# theme text fonts
+		html_css_link( 'ace-fonts.css' );
```
syncguru: @badfiles How can we ensure ace-fonts.css is same as google cdn version? Remember that ace-fonts.css can change with new version of the UI theme.

badfiles: It will never be the same, as Google serves different css for different OSes and browsers. If we add other fonts, we should also provide valid cdn links.
GF css version is supposed to be best for one's OS and browser, but not always, of course.

syncguru: Keeping the local file name under ace-fonts.css is confusing. I would rename this file to something like open-sans.css which should be identical to the CDN file. We should not have different files between CDN and local.

dregad:
I have little expertise in this area so please forgive me if this is a silly question, but why are you limiting the font to specific subsets ? What happens if e.g. Kanji characters need to be displayed ?

badfiles:
Rebased + removed subsets limit

vboctor:
Merged via 4d6818db7d1564a9d4908ad5022de6392efa910a

PR #802, opened 2016-06-20
Dropzone fixes and improvements

badfiles:
Fully localize dropzone
Correct spelling
Use existing config property 'allowed_files' in 'acceptedFiles'
Remove multiple inputs as they're no longer necessary


syncguru:
👍

bug_report_page.php
```diff
@@ -624,7 +624,7 @@
 ?>
 	<tr>
 		<th class="category">
-			<label for="file"><?php echo lang_get( $t_file_upload_max_num == 1 ? 'upload_file' : 'upload_files' ) ?></label>
+			<label for="ufile[]"><?php echo lang_get( $t_file_upload_max_num == 1 ? 'upload_file' : 'upload_files' ) ?></label>
```
vboctor: I suggest we just use plural.  The old config option default (of 1) was based on the old UI of adding more than one browse control, but in the new UI, limiting max attachments to 1 doesn't make sense.  I suggest a default value of 10.  To limit a user to upload no more than 10 attachments at a time.

vboctor: As you update the config option description and its default, please also do the matching change to the manual.

dregad:
We need a bugtracker reference for this.

Otherwise, +1 including to increase of max number of files to 10

badfiles:
Rebased, max number of files set to 10

vboctor:
Thanks @badfiles - merged via c73ff8d3538c462bbfeecf55707f44fab288a985

PR #804, opened 2016-06-21
FIX upload file path generation (adding trailing slash if needed)

spidgorny:
It includes changes needed to show menu items from plugins. These can be ignored.


vboctor:
This pull request has debug statements that should be removed.  Has some noise relating to submodules that should be removed as well.

We generally don't normalize paths and assume they end with a slash.  But I'm OK with a model to make the code more forgiving.  The way to do that would be to have config_get_path( $p_config_option ) method that loads a config value and normalized before returning it to the code, then we can call this instead of config_get() when retrieving a path.

vboctor:
Closing this PR for now.  Feel free to re-open once the feedback is applied.

PR #805, opened 2016-06-23
Add Ace theme files to MantisBT under GPL with author's approval

syncguru:


doc/ace-theme-license.txt
```diff
@@ -0,0 +1,42 @@
...
+and use them on a stand-alone basis.
+5. If the item was created using materials which are the subject of a GNU General Public License (GPL), your use of the
+item is subject to the terms of the GPL in place of the foregoing conditions (to the extent the GPL applies).
+
+
+*** MantisBT core team is in possession of the proof of purchase for this license ***
+
+
+
+Author affirmation of MantisBT usage requires extended license:
```
vboctor: I would reward this title as: "Author's confirmation that MantisBT usage is covered by the purchased Extended License"

vboctor:
:+1: Thanks for following up on this.

badfiles:
Thank you for understanding the license issue and following up.
Although my opinion has not been changed:
"MantisBT is not a service and must not hold licenses or possess property. It is an opensource project under GPL, and each contributor agrees their every patch is also under GPL.
Incorporating the item (AAT) with a patch into MantisBT makes AAT under GPL, which means free to download, distribute and use on GPL terms."
The preamble to the license, which you did not include for an unknown reason, states: 'a license grants you a non-exclusive and non-transferable right to use and incorporate...'. That means you are unable to commit the AAT patch, because it changes the AAT end user license to GPL, and as I noticed you have no right to do it, only the author has all rights.
The letter you included is unclear on the matter that the respective author understands and agrees to transfer his work under the GPL terms, which would make his work available to use in other GPL projects, and non-minfied versions be accessible and altered according to the GPL license.

syncguru:
@badfiles please review

grangeway:
There's no need to add a file with the author's approval if he's agreed to license the code under the GPLv2. As the files under the GPLv2 will become free for anyone to use in anyway.

PR #806, opened 2016-06-29
Introduction of js/css build

vboctor:
The goal of this change is to start exploring the introduction of a package management and build steps for MantisBT (see [modern build](https://github.com/vboctor/mantis-specs/blob/master/modern_build.md)).  The goal here is to provide a first step in this direction, and not a complete re-working of all structure of MantisBT and switch over to fully utilizing the benefits of the new tools incorporated here.

As a first step into exploring Mantis build, the output js/css files are also checked into git.  This is likely to change, but I deferred this step until we agree on the full build story.  This change focuses on the most important needs that we lack right now including js concatenation, compiling ES 2016 down to JS supported by browsers via Babel and minification.  This change also adds css compilation from scss.  Hence, MantisBT continues to work out of git with no build step, until the need arises to edit css/js.  As mentioned earlier, like to drop these from git, and require the build step.

**Javascript**
- Javascript code is written under src/js/\* and compiled under js/mantis.min.js
- ES 2016 javascript is compiled into javascript supported by browsers of today.
- Runs jslint on javascript files.

**CSS**
- Styles are authored under src/scss/ using [Sass](http://sass-lang.com/) and are compiled into css under css/default.css.
- Multiple scss files are consolidated into default.css.  Maybe eventually to be renamed to mantis.css.

**NPM**
- npm is used as the package manager for gulp and its components.

**Composer**
- Composer is added but not used yet.  It should be used to pull in php dependencies.

**Gulp**
- [Gulp](http://gulpjs.com/) is the task runner / build that runs the tool chains used above.
- Just running gulp command will compile the js/css and will stay in watch mode to re-compile as source js/scss files are saved.


syncguru:
@vboctor agree with where this is going.
Can you add setup instruction to this PR?
1- What is command line to use to install packages?
2- What is command line to instruct gulp to monitor the project?

Few things I tried:
1- Run "composer install" -> gives missing composer.json
2- Run "npm install" -> installed bunch of packages
3- Run "gulp" to watch the filesystem

dregad:
I'm OK with this on principle. I think it's a great idea, something that I've had on my todo list for a while.

Before this PR gets merged, I would like to seriously play with it; unfortunately, due to the vacation period I will not be able to do so before mid August, so please bear with me.

In response to @syncguru's note, I believe the Developer's guide should be amended with instructions on how to setup the toolchain, and how to use it.

vboctor:
@syncguru I've fixed the composer issue and update the developer's guide.

@dregad I have started with the basics which we need now, but I expect that we will do a lot more in this space.  Since you are agreeing with the direction, I suggest to get this in, and then evolve it as you play with it.  This will enable us to start cleaning up some the js/css assets that we have.

syncguru:
@vboctor How about getting rid of Gulp and just use npm for builds? This is inline with the current trend in web development ... see http://blog.keithcirkel.co.uk/how-to-use-npm-as-a-build-tool/

vboctor:
@syncguru I'm done some research and I'm OK just using npm.  If we need gulp (or alternative) later we can re-add.

rombert:
Overall this looks good to me, a welcome change.

Do we plan to remove the submodules setup and pull in the current PHP dependencies via composer?

vboctor:
@rombert yes, the plan is to move our dependencies to packages via composer.  But I'll leave this to @dregad to do.

vboctor:
@dregad just a reminder to play with this PR :)

package.json
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/mantisbt/mantisbt.git"
+  },
+  "author": "",
+  "license": "ISC",
```
dregad: GPL ?

dregad:
@vboctor I had just a quick look so far, not really "played" just yet. A few comments:

- you said you were OK with using just NPM, yet there are still several references to Gulp (e.g. in Docbook, gulpfile.js still present)
- one thing I don't really like about the choice of NPM (other than the fact that I don't know much about this tool), is that we're moving away from the PHP ecosystem; did you consider using a PHP-based tool such as phing, which could be added as a composer dev library ? If so, why did you reject it ?
- I don't think _composer.phar_ belongs in our repo, that's an external tool which should be installed separately
- Why introduce a new layout for CSS with the closing bracket on the same line as the last attribute ? I prefer having it on the next line as we did before; can it be changed back ?
- Build.xml: 
  - shorten text lines to ~100 so it's not necessary to scroll left/right when reviewing on Github
  - as mentioned before, guide should be amended with instructions on how to setup the toolchain, and explain how to use it for the builds

About switching from submodules to composer packages, I did a quick check and not all our libraries exist in Composer:

library | composer
--- | ---
adodb       | https://packagist.org/packages/adodb/adodb-php
disposable  | not found
phpmailer   | https://packagist.org/packages/phpmailer/phpmailer
rssbuilder  | not found
utf8        | not found
secureimage | https://packagist.org/packages/dapphp/securimage

_rssbuilder_ and _utf8_ are very old libraries, so we'll probably need to look for alternatives, or publish them on Packagist ourselves (or keep them as submodules ?)

As a side note, during my vacation, I started playing with PHP CodeSniffer. I created a ruleset to try and match our coding guidelines, including a few custom Sniffs for our specific style. This is not quite ready yet, but when it is it should integrate quite nicely with this build.

vboctor:
Thanks @dregad for the feedback.

> you said you were OK with using just NPM, yet there are still several references to Gulp (e.g. in Docbook, gulpfile.js still present)

I don't want to make changes until we have a specific goal that we agree on.  But I'm ok with changing the PR to only use npm rather than npm + gulp.

> did you consider using a PHP-based tool such as phing

I haven't looked at phing.  I chose NPM because it is the packaging system with the strongest momentum.  It will help with node-js based tools that are used to compile code as well as javascript libraries.  It can easily integrate with php ecosystem tools like composer.

> I don't think composer.phar belongs in our repo

I can remove it.

> Why introduce a new layout for CSS with the closing bracket on the same line as the last attribute

This is the output of the CSS compiler.  Ideally we won't even check-in the compiler output files in git.  I just checked them in to make it still possible to run mantisbt just out of git without a build step.  But I think we will move away from this and require git clone + run build to run MantisBT locally.  For releases, this will be part of the packaging process and hence users won't have to worry about this.

> build.xml 100 characters

I can do that.

> Update guide with steps to setup the build tools

I've already added this via https://github.com/mantisbt/mantisbt/pull/806/commits/a4a56ecdf8c135891ff16d3a33d908eb57ce1235 -- however, I expect this to change as we do more changes to this PR.

> composer libraries

I believe there is a way to pull libraries directly from github even if it is not explicitly published or at least if it has the right meta-data files.  We can switch the libraries that support composer to composer and leave the others as sub-modules until they are updated or replaced.  disposable library can be published to composer since we control it.

> php code sniffer

I would rather not have such a tool enforced in a way that requires us to do major churn in the code to fix previous violations.  This introduces bugs, takes time, and breaks pull requests and patches.  If you have a way to do this for new work, that's OK.  Otherwise, it is a lot of pain for a cosmetic thing.

dregad:
> changing the PR to only use npm rather than npm + gulp

I would recommend to do it then.

> phing vs npm

I was just checking, I'm fine either way.

>  Ideally we won't even check-in the compiler output files in git [...] require git clone + run build to run MantisBT locally

That would be preferable IMO.

>> Update guide with steps to setup the build tools

>I've already added this via a4a56ec

You're right, sorry  I missed it.

>> composer libraries

> I believe there is a way to pull libraries directly from github 

I suppose you're referring to [this](https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository).

For the record, rssbuilder and utf8 are currently NOT submodules, and AFAIK do not exist on Github, hence my suggestion to replace them with something more recent and supported, if possible. 

Until then, as you said we can simply leave the code in the repo as-is.

> php code sniffer

It is inevitable to cause a little bit of churn as we enforce the standard, but it's a one-time cost that we can minimize by not going too far in terms of the rules we decide to apply. 

I believe that would be an acceptable cost compared to the benefit of being able to automate the coding guidelines checks, and not have to worry about doing it manually as part of reviewing PRs.

sync1:

l266509:

vboctor:
I'll open a new PR based on the plan here once I'm have it ready.  Will close this PR.

PR #807, opened 2016-06-29
Change look & feel of links, borders and label background

atrol:
I would like to introduce a few small CSS changes to
- have some optimizations in look & feel (at least in my opinion)
- be easily able to distinguish between 1.2.x and 1.3.x when users are posting screen shots

**Changed CSS**

Links:
- change and harmonize colors
- don't display changed link color for visited links
- don't use text decoration underline for links
- change color when hovering over links

Borders:
- change and harmonize colors, only use grey, remove black (to much
  contrast, stresses eyes)

Background:
- replace blueish color by another blueish color at various places

Margin:
- reduced at one place

Fixes #21171

Before
viewold
After
viewnew

### DIFF
--- a/css/default.css
+++ b/css/default.css
@@ -11,12 +11,15 @@ p.compact		{ margin: 0;}
 
 pre				{ margin-top: 0px; margin-bottom: 0px; }
 
-a					{}
+a					{ color: royalblue; }
 a:active 			{}
-a:link    			{}
+a:link    			{ text-decoration: none; }
+a:hover 			{ color: red; }
 a:visited 			{}
+/*
 a.subtle			{ color: blue; text-decoration: none; }
-a.resolved			{ text-decoration: line-through underline; }
+*/
+a.resolved			{ text-decoration: line-through; }
 a > img				{ border: none; }
 
 form				{ display: inline; }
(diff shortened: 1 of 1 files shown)


vboctor:
@atrol please share before/after pictures.

atrol:
Pictures are attached at https://www.mantisbt.org/bugs/view.php?id=21171

cproensa:
> be easily able to distinguish between 1.2.x and 1.3.x when users are posting screen shots

good idea

another suggestion: shange the background greys in the alternated rows into a lighter shade.
(that is one complain i usually hear from users, the visual style get overcharged from so many grey rows...)

and these ones:
- maybe it can be tweaked by #ids at css: change the background  in the main content fields: summary, description, additional, etc, Dont get the grey background from the alternate rows, use a cleaner background to differentiate from the mass of fields.

- Regarding content fields: summary, description, etc. Keep them spatially differentiated, by moving the "tags" and "custom fields" rows up, with the rest of fields.

atrol:
@cproensa, you are right. There is some UI optimization possible in 1.3.x without huge changes.

I made only few small changes just in CSS, not in PHP code as I don't want
- to introduce regressions or risk stability
- add lot of delay to releasing final 1.3.0
- to start long running discussions, as even my small changes need to be discussed, see https://www.mantisbt.org/bugs/view.php?id=21171

dregad:
:+1: 

To go along the lines of @cproensa's suggestion and minimize the changes, maybe just use a lighter shade of gray for alternate rows ?

atrol:
@dregad, @cproensa 
I am not sure if I understand right what you mean.

I tried a lighter gray instead of the darker one, so that the differnence between the two shade of grays is less than before.
I did not like it, as the colors are no longer easy to distinguish, which is a bit confusing.
You even might ask yourself when seeing it: Are the colors different? Maybe it's just a bug and someone made just a typo when entering rgb values? 

After that I tried to use just the darker gray everywhere in Mantis.
Have a look at first screen shot after this change.
I like it.
Not sure if other users will like it, especially in tables like "Issue History"
Have a look at second screen shot after this change.
viewnewonegray
viewnewonegrayhistory

cproensa:
i mean, shift both greys to the white

this is a sample:
alternated rows is using greys:  #e8e8e8 and #d8d8d8
by shifting them halfway its distance to white, i tried: E8 and EB

This is the default: 
#e8e8e8, #d8d8d8

seleccion_017


modified grey levels:
#f3f3f3, #ebebeb

seleccion_016


probably other grey should be adjusted too, see the attachments block, monitoring, etc.

atrol:
This one uses the gray levels proposed by cproensa #f3f3f3, #ebebeb

myviewlightergray

vboctor:
I like the purple color much better than the new blue color, specially when the text itself is light blue, like the filter box field titles.

I'm not sure I really understand the important of differentiating versions by screenshots, but if you really have to, then maybe lighter grays is a better option.

The rest of the tweaks makes sense, so 👍 for the others.

atrol:
Pushed https://github.com/mantisbt/mantisbt/commit/1612662d1cd2998090bae2bc18a8ada63131178e removed the new blue color.

PR #810, opened 2016-07-04
Modern Charts based on Chartjs

vboctor:
This change updates our MantisGraph core plugin to match the modern UI.  This change includes the following:

- Drops dependency on jpgraph or ezc libraries (bye bye 1990s).
- Leverages [chartjs](http://www.chartjs.org/)
- Graphs are now javascript based rather than image based, hence animated and interactive.
- Provides a simpler model to add / remove graphs that are easier to understand by users.
- Removes config page and options for the plugin since they are no longer needed.

Fixes #12825

See the video below for the new experience:
http://www.screencast.com/t/KjEzJ1AjtQF

Here are some samples of the old experience:
http://www.screencast.com/t/KJ7rX9fryv
http://www.screencast.com/t/5CN0pWjHV
http://www.screencast.com/t/I5llI4QOQ9l


rombert:
+1 on the overall approach, looks great.

Did not test

atrol:
I did not have a look at the source, but tried on one of my testing environments using a copy of real life data.
It worked out of the box, got no crash or errors and the displayed charts are better than the old ones at first sight.

I am really happy to see that there is a great chance that I will no longer have to deal with users in forum that are not able to setup the environment to get working charts (font installation, jpgraph, ...)

PR #811, opened 2016-07-06
move "auto_set_status_to_assigned" logic into bug_api

cproensa:
Create a bug api function to evaluate the resulting status after an
assignment action. This function account for the option
"auto_set_status_to_assigned", current handler and status, and requested
new handler and status.

The behaviour of assignments, when this option is enabled, is slightly
modified:
Previously, inconcistent logic allowed to transition a bug into assigned
status by changing handler user, even if the transition is not allowed
by workflow configuration.
Additionally, the assign action was not available to the user if the
assigned status is not reachable by workflow, even when the assign
action would not result in a status change.

Now, the assign action is always available (if allowed by permissions),
and the status change is performed only if the conditions are met,
otherwise, the handler change is always performed.

The conditions needed for automatic change to 'bug_assigned_status' are:
- current handler is empty
- new handler is not empty
- status is not already part of the bug modification
- current status is lower than 'bug_assigned_status'
- 'bug_assigned_status' is reachable by workflow

Fixes #21203

### DIFF
--- a/bug_actiongroup.php
+++ b/bug_actiongroup.php
@@ -151,11 +151,7 @@ foreach( $f_bug_arr as $t_bug_id ) {
 			break;
 		case 'ASSIGN':
 			$f_assign = gpc_get_int( 'assign' );
-			if( ON == config_get( 'auto_set_status_to_assigned' ) ) {
-				$t_assign_status = config_get( 'bug_assigned_status' );
-			} else {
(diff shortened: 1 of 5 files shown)


dregad:
+1, didn't test

vboctor:
:+1: - merged via 5de0133c1f890cdb9641263bee56587f77ec8649

PR #812, opened 2016-07-08
Maintain consistent coding styles between different editors and IDEs

syncguru:
See editorconfig.org for details

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = tab
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true


.editorconfig
```diff
@@ -0,0 +1,10 @@
+# editorconfig.org
+root = true
+
+[*]
```
atrol: This could introduce unwanted side effects.
I would start a bit more conservative and restrict it to something like [*.{php,css,js,sh,md}]
At least UTF-8 should be restricted, maybe even more.
e.g. what will happen if you store a file like api/soap/mantisconnect.wsdl that comes with ISO-8859-1 encoding?

dregad: > [*.{php,css,js,sh,md}]

[Trailing whitespace can be significant in markdown](https://daringfireball.net/projects/markdown/syntax#p) (indicating a hard line break)

dregad:
I don't really like this approach.

In my opinion, it would be much better to rely on code inspection (e.g. using codesniffer) as part of a either a git precommit hook or via continuous integration builds (or both), rather than having an editor forcefully changing files potentially without the developer being aware of it.

atrol:
Right, changes should not happen without beeing aware that something has changed.
Instead of this we could add the config file using a different file name to our repository, so developers could manually run tools like https://github.com/jedmao/eclint if they want.

syncguru:
@dregad I think you should give it a try first.
It is good to unify some key elements in every IDE used by MantisBT contributors. I recall I had to switch to tabs from spaces in early version of ModernUI PR which was painful.

"editor forcefully changing files potentially without the developer being aware of it"
Too many assumptions, @dregad. If this was the case, then this PR will be the entire project. 

The concept it really good and solves a problem common in projects with diverse set of contributors, like MantisBT.

dregad:
> Too many assumptions, @dregad. If this was the case, then this PR will be the entire project. 

Don't understand what you mean.

In any case, I stand by my position.

I prefer an approach of validating a user's contribution via a tools like codesniffer which can alert of diverging coding standards in a much more detailed manner (i.e. not limited to just whitespace).

syncguru:
Both approaches are not mutually exclusive. 

@dregad regarding your approach, are you planning to submit a pull request with the implementation? I am happy to review.

PR #813, opened 2016-07-08
Only prepend 'css/' when given a filename without path

dregad:
Commit 1819bbdf8c2d629798fa48537f9bb167e8d33005 introduced new
html_css_link() function to include CSS files, but made the assumption
that these would always be in the css/ directory.

This lets the admin specify $g_css_include_file with a path, allowing
them to store custom CSS in a different location within the MantisBT
root.

Fixes [#21194](https://www.mantisbt.org/bugs/view.php?id=21194)

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -388,7 +388,11 @@ function html_css() {
  * @return void
  */
 function html_css_link( $p_filename ) {
-	echo "\t", '<link rel="stylesheet" type="text/css" href="', string_sanitize_url( helper_mantis_url( 'css/' . $p_filename ), true ), '" />' . "\n";
+	# If no path is specified, look for CSS files in default directory
+	if( $p_filename == basename( $p_filename ) ) {
+		$p_filename = 'css/' . $p_filename;
+	}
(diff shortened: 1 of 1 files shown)


vboctor:
👍

syncguru:
Looks good

atrol:
I agree, this should fix the issue (didn't test) but isn't it some kind of dirty solution?
We had no hardcoded string in 1.2, now there is one ('css') depending on the value of a parameter.

Another approach could start by changing 

```php
$g_css_include_file = 'default.css';
```
to
```php
$g_css_include_file = 'css/default.css';
```
There are some more changes needed after this, but wouldn't it the better solution?

vboctor:
I like the solution by @dregad it allows easily including files from default css folder without hard-coding it in the configs.  But if there is a need to point to a different folder, users can do that.  That approach enables us to potentially move the folder for css with less breaking changes for customer configs.

I'll integrate @dregad 's fix as is.

vboctor:
Merged via 1643e474c5762117d7583eaf571626ea35a5dac9

PR #814, opened 2016-07-08
Tag devide single project - bug. #9716

ilsaul:
Hi,
I'm not the original maker of this solution but I need then I readapt it 
I'm not the original creator of this solution but I need it and so I adapted it and submitted for adding to mantis 1.3.


atrol:
The PR contains changes that are not related to tagging and has conflicts that must be resolved.
It seems also that you forgot some changes, e.g. the original patch contained changes of the database schema
https://www.mantisbt.org/bugs/view.php?id=9716

>  I adapted it and submitted for adding to mantis 1.3.

Your PR wants to merge to  branch master, which means version 2.0.
The right branch for 1.3 would be master-1.3.x, but we don't allow any database schema changes for new features in version 1.3.x.

ilsaul:
Ok I can try to resolve conflict and yes I forget one column add in the scheme.
When I submit in 1.3 I think is more easy because is new and not release jet.
In the end what I can do?

vboctor:
Thanks @ilsaul 

I wouldn't rush into getting this in for the following reasons:

- I think we need to think about the experience all up for managing global and project specific tags.
- If we are to add reporting/graphs based on tags, make sure that there are no complexities.
- Filtering on tags, do we know what tags to show in case All Projects / specific project in the filter box.
- We were deferring schema changes to make it easier to switch back and force for 2.0.0 and 1.3.x for the time being.

and of course the patch needs cleanup.

dregad:
@ilsaul Considering the review comments, the fact that the PR is targeting the wrong branch, contains unrelated changes and conflicts with current head, I'm closing this issue for now. Feel free to submit a new PR taking the remarks into consideration.

PR #815, opened 2016-07-10
Remove leftover php closing bracket from priority graph page

syncguru:
Minor fix. Remove "?>" from priority graph page. See image
screen shot 2016-07-09 at 6 50 01 pm


vboctor:
Merged via df1d844a2406921684d16a4055c719f5b1d6e411

PR #816, opened 2016-07-10
Fix footer layout for users without 'Save Filter' privileges 

syncguru:
Relocate opening div outside of the conditional statement to match closing div

Fixes #20240


vboctor:
I've fixed the commit title to not exceed 50 characters and remove extra white space, then merged to master.

Merged via 70c7d7ad110d3d167c32629bbafbf0603eeff2ca.

PR #817, opened 2016-07-10
Replace reference to undefined classic UI function

syncguru:
Fixes #21130


vboctor:
Merged via 6854bb2e5b6532010d14dbb243358cbba5c0f0d6

PR #819, opened 2016-07-10
Use matching label for max custom field length

syncguru:
Fixes #20182

screen shot 2016-07-10 at 12 49 13 am


vboctor:
Merged via ea745b3463fc460a0aabf44ba876332de8325649

PR #820, opened 2016-07-10
Use CDN-related functions in MantisGraph

badfiles:
- add missing base64 alignment
- add comment on howto calculate hashes


vboctor:
@badfiles thanks, this was on my todo list, but I was waiting for potentially using npm/composer to pull the dependencies for local package.

btw, if the hashes were incorrect due to the missing '=' why wasn't the code failing?  Was it creating warnings in the browser console?

badfiles:
The hashes are correct even without '=', browsers are smart enough to assume there must be alignment.
With '=' is what they should be.

vboctor:
:+1:

syncguru:
👍

atrol:
:+1:

vboctor:
Merged via c3ad66cc459642053b7b4aa59027cfa039a1e886

PR #821, opened 2016-07-17
Prevent TLS if smtp_connection_mode is empty (no encryption)

atrol:
Since version 5.2.10 PHPMailer automatically tries to enable TLS.
This is not what we want if the Mantis administrator decided not to use
encryption (smtp_connection_mode is empty)

Fixes #21293


syncguru:
+1
I can see the need for non-TLS during testing and development. Is there a way to enforce TLS in production?

atrol:
> Is there a way to enforce TLS in production?

Setting $g_smtp_connection_mode = 'tls'; in config_inc.php should be enough for it

atrol:
Messed up my repo, replaced by
https://github.com/mantisbt/mantisbt/pull/839

PR #824, opened 2016-07-22
Hide ‘Report Issue’ when user doesn’t have access

vboctor:
Fixes #21223

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -538,19 +538,29 @@ function layout_navbar_button_bar() {
 		return;
 	}
 
+	$t_can_report_bug = access_has_project_level( config_get( 'report_bug_threshold' ) );
+	$t_can_invite_user = current_user_is_administrator();
+
+	if( !$t_can_report_bug && !$t_can_invite_user ) {
+		return;
+	}
+
 	echo '<li class="hidden-sm hidden-xs">';
   	echo '<div class="btn-group btn-corner padding-right-8 padding-left-8">';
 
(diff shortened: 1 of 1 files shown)


atrol:
Shouldn't we discuss https://www.mantisbt.org/bugs/view.php?id=21223#c53558 before going on to invest time in a feature that might be removed/replaced quite soon?

core/layout_api.php
```diff
@@ -538,13 +538,22 @@ function layout_navbar_button_bar() {
...
+	}
+
 	echo '<li class="hidden-sm hidden-xs">';
   	echo '<div class="btn-group btn-corner padding-right-8 padding-left-8">';
 
-	$t_bug_url = string_get_bug_report_url();
-  	echo '<a class="btn btn-primary btn-sm" href="' . $t_bug_url . '">';
-	echo '<i class="fa fa-edit"></i> ' . lang_get( 'report_bug_link' );
-	echo '</a>';
+  	if( !$t_can_report_bug ) {
```
atrol:
```php
if( $t_can_report_bug )
```

core/layout_api.php
```diff
@@ -538,13 +538,22 @@ function layout_navbar_button_bar() {
 		return;
 	}
 
+	$t_can_report_bug = access_has_project_level( config_get( 'report_bug_threshold' ) );
+	$t_can_invite_user = current_user_is_administrator();
```
atrol: use $t_can_invite_user also at line 558 to replace call of current_user_is_administrator()

PR #827, opened 2016-07-28
Don't show version release date if not set

cproensa:
In roadmap, dont show release date if it's not set.
When the date is not explicitly set, its stored as 1 in database.
Logically it means "date null", but it was being shown as an actual
date: 1970-01-01

When not showing date, also becasue of option $g_show_roadmap_dates set
to OFF, do not show the calendar icon.

Fixes: #12409


atrol:
+1, didn't test

vboctor:
:+1: Merged via 2a8f4c363ffa035ffc1ce88c7dedbe54f835673f

PR #828, opened 2016-07-28
Don't show version release date if not set (target 1.3.x)

cproensa:
In roadmap, dont show release date if it's not set.
When the date is not explicitly set, its stored as 1 in database.
Logically it means "date null", but it was being shown as an actual
date: 1970-01-01

Fixes: #12409


atrol:
+1, didn't test

vboctor:
:+1: merged via 79976d43533129773449ff469eb535eb3160c46e

PR #829, opened 2016-07-29
Don't show empty EVENT_MENU_ISSUE menu items

cproensa:
If the result of event call for EVENT_MENU_ISSUE is null, dont print
bracket with empty content.

Having null as returned value is valid.
For example: a plugin can hook to this event and return either an
actual item or an empty one, based in access level.

Fixes: #21407


vboctor:
:+1:

PR #830, opened 2016-07-29
Don't show empty EVENT_MENU_ISSUE menu items (target 1.3.x)

cproensa:
If the result of event call for EVENT_MENU_ISSUE is null, dont print
bracket with empty content.

Having null as returned value is valid.
For example: a plugin can hook to this event and return either an
actual item or an empty one, based in access level.

Fixes: #21407

### DIFF
--- a/bug_view_inc.php
+++ b/bug_view_inc.php
@@ -271,7 +271,7 @@ foreach ( $t_links as $t_plugin => $t_hooks ) {
 					print_bracket_link( $t_href, $t_label );
 				}
 			}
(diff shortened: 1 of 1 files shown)


atrol:
Commit ff94e46 should not be part of this PR

cproensa:
> Commit ff94e46 should not be part of this PR

Thanks, i stacked on top of other branch.
I have rebased this PR to contain only the correct commit.

vboctor:
:+1:

PR #832, opened 2016-07-30
0020424 csv memory (1.3.x)

cproensa:
Resubmit this branch as new PR from #698 
after rebased on 1.3.x

There is a problem with the export to CSV functionality when retrieving a large list of issues to export.
It is getting the list through filter_api:filter_get_bug_rows(), which pre-process the list loading it into memory, and doing cache of some bug data. Because of this, the memory usage grows linearly and php memory limit may be reached.
Excel export has a work around that uses pagination to get the data in batches, however there is still some redundancy because each time the function filter_get_bug_rows is called, a new query is being composed and executed.

The changes here are:
filter_get_bug_rows is a function which does all in one:
- prepare pagination
- compose query conditions
- execute query and load results in memory, add to cache related data

First modification is refactoring filter_get_bug_rows by separating the outer layers of logic (the core query composer is untouched and still that ugly block of code with hundreds of lines...).
However, now the result set can be accessed before its loaded and cached into memory.
To me it makes sense, to separate the query building and execution, from pagination which is very screen oriented.

Next modifications are the export procedures, which now reads directly from the result set. 
By reading rows in batches, memory used to store data and cache is controlled, and cleared between each batch.

fixes #0020424

### DIFF
--- a/csv_export.php
+++ b/csv_export.php
@@ -108,8 +108,8 @@ define( 'EXPORT_BLOCK_SIZE', 500 );
 $t_end_of_results = false;
 do {
 	# Clear cache for next block
-	unset( $g_cache_bug );
-	unset( $g_cache_bug_text );
+	bug_clear_cache();
+	bug_text_clear_cache();
 
 	# Keep reading until reaching max block size or end of result set
 	$t_read_rows = array();
--- a/csv_export.php
+++ b/csv_export.php
@@ -102,9 +102,6 @@ if( strcmp( $t_first_three_chars, 'ID' . $t_sep ) == 0 ) {
 
 echo $t_header;
 
-
-define( 'EXPORT_BLOCK_SIZE', 500 );
-
 $t_end_of_results = false;
 do {
 	# Clear cache for next block
--- a/csv_export.php
+++ b/csv_export.php
@@ -49,25 +49,27 @@ auth_ensure_user_authenticated();
 
 helper_begin_long_process();
 
-$t_page_number = 1;
-$t_per_page = -1;
-$t_bug_count = null;
-$t_page_count = null;
-
 $t_nl = csv_get_newline();
 $t_sep = csv_get_separator();
 
-# Get bug rows according to the current filter
-$t_rows = filter_get_bug_rows( $t_page_number, $t_per_page, $t_page_count, $t_bug_count );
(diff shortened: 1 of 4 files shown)


cproensa:
Note: commits can also be picked into master without conflicts

vboctor:
I would replace EXPORT_BLOCK_SIZE with the use of $g_default_limit_view configuration option.  This makes sense and will guarantee that a fixed number like 500 isn't too big to fit in PHP memory limit.  i.e. if View Issues page works, then export will work.

cproensa:
Well. EXPORT_BLOCK_SIZE, should be moved into constants, or better, made a global variable?

I don like it to be as low as $g_default_limit_view.
The breaking limits for export sizes are:
- Exceeding memory limit. Memory is used mainly for cached data. I my case the defaults allowed to process 15k bugs before crashing.
- Exceeding DB query limits: Mysql/others, limit is package size. Oracle has a more problematic limit for 1000 elements in "IN" clauses"

My guess is the common low limit is 1000 bugs. Half that to be safe: 500.

Having the block size too low results in worse performance: more queries executed, more cache missed

csv_export.php
```diff
@@ -100,29 +102,74 @@
...
-		if( !$t_first_column ) {
-			echo $t_sep;
-		} else {
-			$t_first_column = false;
+define( 'EXPORT_BLOCK_SIZE', 500 );
+
+$t_end_of_results = false;
+do {
+	# Clear cache for next block
+	unset( $g_cache_bug );
```
vboctor: I would call the proper api to reset its own cache.  For example, bug_cache_reset().  Do the same with other purging of caches.

vboctor:
Let's start off with moving EXPORT_BLOCK_SIZE to constants_inc for now.  If we see a need later, then we can add a config option.

:+1: once feedback is addressed.  Haven't tested.

cproensa:
Done with Victor comments.

dregad:
@vboctor When merging a PR that applies to both stable and development branches, rather than merging the PR into both branches I would recommend to follow these steps instead in the future:

1. merge PR into *master-1.3.x*
2. merge *master-1.3.x* into *master*

This avoids spaghetti in our git history, like

spaghetti

Thanks for your understanding

PR #833, opened 2016-07-30
Add "new status" to EVENT_UPDATE_BUG_STATUS_FORM

cproensa:
Add parameter to EVENT_UPDATE_BUG_STATUS_FORM, to include the new
status.

Fixes: #17923

Can be cleanly picked into master

### DIFF
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -330,7 +330,7 @@ print_recently_visited();
 		}
 	}
 
-	event_signal( 'EVENT_UPDATE_BUG_STATUS_FORM', array( $f_bug_id ) );
+	event_signal( 'EVENT_UPDATE_BUG_STATUS_FORM', array( $f_bug_id, $f_new_status ) );
 
 	if( $f_change_type == BUG_UPDATE_TYPE_REOPEN ) {
 ?>
--- a/docbook/Developers_Guide/en-US/Events_Reference_Bug.xml
+++ b/docbook/Developers_Guide/en-US/Events_Reference_Bug.xml
@@ -197,6 +197,7 @@
 				<itemizedlist>
 					<title>Parameters</title>
 					<listitem><para>&lt;Integer&gt;: Bug ID</para></listitem>
+					<listitem><para>&lt;Integer&gt;: New Status</para></listitem>
 				</itemizedlist>
 			</blockquote>
 		</blockquote>


vboctor:
Should we have an object that has the parameters that can be updated on such page?  For example, duplicate id, note, custom fields, etc?  I even wonder if we should be using the standard bug update events or at least the same parameters we send to bug update events.

cproensa:
The update action itself is already using the bug update events.

The motivation for this parameter, is to know the actual action requested by the user.
As a plugin, you may want to add some inputs to the form, by knowing which status is the requested one. For example, implement a "required field for resolution" type of logic, this allows to mark the fields as required.

vboctor:
Makes sense.  Should we have a associative array of key value pairs instead of a normal parameters?  Makes it easier to add more and check for existence.  Thoughts?

cproensa:
> Should we have a associative array of key value pairs instead of a normal parameters? 

You have suggested that several times, and has its good points. But wouldn't that break the compatibility?
I'd keep it for future in case some plugin/events redesign is made that implies a breaking change by itself.

vboctor:
> But wouldn't that break the compatibility?
I was referring to the added parameter being replaced by an associative array rather than changing the already existing parameter.

dregad:
@vboctor I'm not sure I'm following you here. Are you saying that instead of passing a bug id + a potentially evolving list of fields, we should instead send an array (or object) with a list of several fields (potentially the whole BugData object) ? IE. @cproensa's currently proposed solution
```
event_signal( 'EVENT_UPDATE_BUG_STATUS_FORM', 
    array( $f_bug_id, $f_new_status ) 
);
```
which could later evolve into
```
event_signal( 'EVENT_UPDATE_BUG_STATUS_FORM', 
    array( $f_bug_id, $f_new_status, $other1, $other2, ... ) 
);
```

vs my understanding of your suggestion
```
event_signal( 'EVENT_UPDATE_BUG_STATUS_FORM', 
    array( array( 'id' => $f_bug_id, 'new_status' => $f_new_status, 'other1' => $other1, ... ) ) 
);
```
or, without breaking compatibility (which is something we should avoid IMO)
```
event_signal( 'EVENT_UPDATE_BUG_STATUS_FORM', 
    array( $f_bug_id, array( 'new_status' => $f_new_status, 'other1' => $other1, ... ) ) 
);
```

vboctor:
How about the following?

    event_signal( 'EVENT_UPDATE_BUG_STATUS_FORM', 
        array( 0 => $f_bug_id, 'id' => $f_bug_id, 'new_status' => $f_new_status, 'other1' => $other1, ... ) 
    );

cproensa:
> How about the following?

    event_signal( 'EVENT_UPDATE_BUG_STATUS_FORM', 
    array( 0 => $f_bug_id, 'id' => $f_bug_id, 'new_status' => $f_new_status, 'other1' => $other1, ... ) 
    );

I think that produces the same call we have now, and would implement as:

    function event_hook ( $p_eventm, $p_bug_id, $p_new_status, $other1, ...)
the indexes are lost through the function call.

The proposal to pass all event parameters as an array, is good imho:

    event_signal( 'EVENT', array( array( 'par1' => $val1, 'par2'=> $val2, ... ) ) );

which is then captured as a single array parameter by the plugin.

    function hook ( $p_event, $p_params )

This allows more dynamic changes in events definition, but plugins should add more checks to make sure all expected parameters are present (in that scenario i would add some standard structures that provides the context of various aspect of application runtime, in the idea of BugData structure, but on app level)
Changing it right now for one event, doesn't seem very productive at the moment. Moreso, if not breaking compatibility, the parameter should be a mixed type, like dregad said: _array( old_param, array( new_params... ) )_, which can add confusion to the standard.

To sum up, i think the ideal approach is to have one single associative array as event parameter, but this should be implemented at a stage that justifies possibly breaking compatibility, and standarize it for all events.
At the moment, even if not the prettiest, i'd suggest to keep it as:

    event_signal( 'EVENT_UPDATE_BUG_STATUS_FORM', 
        array( $f_bug_id, $f_new_status ) 
    );

dregad:
Merged in b2455a3fee1482ad29c7bb4e75ed3a783db0a214

PR #834, opened 2016-07-31
Move Bug update events into BugData

cproensa:
(updated PR from #720, targeting current master, and finishing what was left)

EVENT_UPDATE_BUG_DATA and EVENT_UPDATE_BUG
are now triggered in the BugData update function, as suggested in #0020577
A callback queue is created in BugData to be able to perform post-update operations.

Fixes: #20577

 summary of this work:
- EVENT_UPDATE_BUG_DATA and EVENT_UPDATE_BUG should be triggered inside the BugData update method
- All bug modifications should be done through the BugData object. That would guarantee the consistency of the application for core, plugins, and soap, operations. (We'll still keep some direct table updates for some specific core code, that should be under control)

The problem with this approach is that other modifications beside the main bug table may exist (eg: adding notes). These operations have to be done:
- After EVENT_UPDATE_BUG_DATA, because this event can be used for validation to stop the update operation. 
- before EVENT_UPDATE_BUG, because this event should indicate finalization of the update, including all related tasks.
  Since the BugData->update() is an atomic call, the method proposed is to include a callback queue that stores the operations to be performed after "validation" and before "finalization".

A clear example of this is updating custom fields.
Previously, CFs are not updated by the BugData operation, and the changes has to be performed by the caller.
If you update _before_, then if a plugin stops the update (due to some failed validation), the fields are already changed. If you update _after_, a plugin that listens to bug update event dont know the new CF values, since at that moment they are not still committed.
By adding the CF update as post-update callbacks, it is performed after BugData changes are saved, and before the finalization event is triggered, so plugins listening to EVENT_UPDATE_BUG can see the updated custom fields.

### DIFF
--- a/bug_actiongroup.php
+++ b/bug_actiongroup.php
@@ -218,7 +218,7 @@ foreach( $f_bug_arr as $t_bug_id ) {
 						$t_bugdata->add_update_callback( $add_bugnote_func, array( $t_bugnote_params ) );
 						}
 
-					$t_bugdata->update( false, true );
+					$t_bugdata->update( /* update extended */ false, /* bypass mail */ true );
 
 					if( is_blank( $f_bug_notetext ) ) {
 						# If there is a note, bugnote add, send email notification for new note
@@ -306,7 +306,7 @@ foreach( $f_bug_arr as $t_bug_id ) {
 				if( access_has_bug_level( config_get( 'due_date_update_threshold' ), $t_bug_id ) ) {
 					$t_bugdata = bug_get( $t_bug_id );
 					$t_bugdata->due_date = $t_due_date;
-					$t_bugdata->update( false, true );
+					$t_bugdata->update( /* update extended */ false, /* bypass mail */ true );
 				} else {
 					$t_failed_ids[$t_bug_id] = lang_get( 'bug_actiongroup_access' );
 				}
@@ -331,7 +331,7 @@ foreach( $f_bug_arr as $t_bug_id ) {
 				# @todo we need to issue a helper_call_custom_function( 'issue_update_validate', array( $t_bug_id, $t_bug_data, $f_bugnote_text ) );
 				$t_bugdata = bug_get( $t_bug_id );
 				$t_bugdata->sticky = intval( !$f_sticky );
-				$t_bugdata->update( false, true );
+				$t_bugdata->update( /* update extended */ false, /* bypass mail */ true );
 				helper_call_custom_function( 'issue_update_notify', array( $t_bug_id ) );
 			} else {
 				$t_failed_ids[$t_bug_id] = lang_get( 'bug_actiongroup_access' );
--- a/bug_actiongroup.php
+++ b/bug_actiongroup.php
@@ -192,8 +192,9 @@ foreach( $f_bug_arr as $t_bug_id ) {
 			if( access_has_bug_level( config_get( 'update_bug_threshold' ), $t_bug_id ) ) {
 				$f_priority = gpc_get_int( 'priority' );
 				# @todo we need to issue a helper_call_custom_function( 'issue_update_validate', array( $t_bug_id, $t_bug_data, $f_bugnote_text ) );
-				bug_set_field( $t_bug_id, 'priority', $f_priority );
-				email_bug_updated( $t_bug_id );
+				$t_bugdata = bug_get( $t_bug_id );
+				$t_bugdata->priority = $f_priority;
+				$t_bugdata->update();
 				helper_call_custom_function( 'issue_update_notify', array( $t_bug_id ) );
 			} else {
 				$t_failed_ids[$t_bug_id] = lang_get( 'bug_actiongroup_access' );
@@ -204,14 +205,24 @@ foreach( $f_bug_arr as $t_bug_id ) {
 			if( access_has_bug_level( access_get_status_threshold( $f_status, $t_bug->project_id ), $t_bug_id ) ) {
 				if( true == bug_check_workflow( $t_status, $f_status ) ) {
 					# @todo we need to issue a helper_call_custom_function( 'issue_update_validate', array( $t_bug_id, $t_bug_data, $f_bugnote_text ) );
-					bug_set_field( $t_bug_id, 'status', $f_status );
+					$t_bugdata = bug_get( $t_bug_id );
+					$t_bugdata->status = $f_status;
 
 					# Add bugnote if supplied
 					if( !is_blank( $f_bug_notetext ) ) {
-						$t_bugnote_id = bugnote_add( $t_bug_id, $f_bug_notetext, null, $f_bug_noteprivate );
-						bugnote_process_mentions( $t_bug_id, $t_bugnote_id, $f_bug_notetext );
-						# No need to call email_generic(), bugnote_add() does it
-					} else {
+						$add_bugnote_func = function( array $p_bugnote_params ) {
+							$t_bugnote_id = call_user_func_array( 'bugnote_add', $p_bugnote_params );
+							bugnote_process_mentions( $p_bugnote_params[0], $t_bugnote_id, $p_bugnote_params[1] );
+						};
+						$t_bugnote_params = array( $t_bug_id, $f_bug_notetext, null, $f_bug_noteprivate );
+						$t_bugdata->add_update_callback( $add_bugnote_func, array( $t_bugnote_params ) );
+						}
+
+					$t_bugdata->update( false, true );
+
+					if( is_blank( $f_bug_notetext ) ) {
+						# If there is a note, bugnote add, send email notification for new note
+						# if not, notify generic update.
 						email_bug_updated( $t_bug_id );
 					}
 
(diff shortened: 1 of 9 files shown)


cproensa:
This should be implemented also in 1.3

bug_actiongroup.php
```diff
@@ -204,14 +205,24 @@
...
+					$t_bugdata = bug_get( $t_bug_id );
+					$t_bugdata->status = $f_status;
 
 					# Add bugnote if supplied
 					if( !is_blank( $f_bug_notetext ) ) {
-						$t_bugnote_id = bugnote_add( $t_bug_id, $f_bug_notetext, null, $f_bug_noteprivate );
-						bugnote_process_mentions( $t_bug_id, $t_bugnote_id, $f_bug_notetext );
-						# No need to call email_generic(), bugnote_add() does it
-					} else {
+						$add_bugnote_func = function( array $p_bugnote_params ) {
```
vboctor: Indentation seems to be broken here.

bug_actiongroup.php
```diff
@@ -204,14 +205,24 @@
...
+							$t_bugnote_id = call_user_func_array( 'bugnote_add', $p_bugnote_params );
+							bugnote_process_mentions( $p_bugnote_params[0], $t_bugnote_id, $p_bugnote_params[1] );
+						};
+						$t_bugnote_params = array( $t_bug_id, $f_bug_notetext, null, $f_bug_noteprivate );
+						$t_bugdata->add_update_callback( $add_bugnote_func, array( $t_bugnote_params ) );
+						}
+
+					$t_bugdata->update( false, true );
```
vboctor: calls with parameters ( false, true ) is not readable.  Consider either adding /* paramName */ before each parameter, or create a method that calls this update internally but has a name that reflects the semantics of ( false, true ).

bug_update.php
```diff
@@ -380,40 +380,59 @@
...
+	$t_updated_bug->add_update_callback( $add_relatioship_func, array( $t_updated_bug ) );
 }
 
+# Commit the bug updates to the database.
+$t_text_field_update_required = ( $t_existing_bug->description != $t_updated_bug->description ) ||
+								( $t_existing_bug->additional_information != $t_updated_bug->additional_information ) ||
+								( $t_existing_bug->steps_to_reproduce != $t_updated_bug->steps_to_reproduce );
+$t_updated_bug->update( $t_text_field_update_required, true );
+
+
```
vboctor: Remove extra blank lines.

core/bug_api.php
```diff
@@ -781,8 +789,26 @@ function update( $p_update_extended = false, $p_bypass_mail = false ) {
...
+			$t_callback = array_shift( $this->update_callbacks );
+			call_user_func_array( $t_callback['func'], $t_callback['params'] );
+		}
+
+		event_signal( 'EVENT_UPDATE_BUG', array( $t_old_data, $this ) );
+
 		return true;
 	}
+
+	function add_update_callback( callable $p_callable, array $p_params = null ) {
```
vboctor: Add doc to function.

core/bug_api.php
```diff
@@ -631,6 +637,8 @@ function update( $p_update_extended = false, $p_bypass_mail = false ) {
 
 		$t_old_data = bug_get( $this->id, true );
 
+		event_signal( 'EVENT_UPDATE_BUG_DATA', $this, $t_old_data );
```
vboctor: Should we call again self::validate() after the plugins call?  If they can change the data.

core/bug_api.php
```diff
@@ -781,8 +789,26 @@ function update( $p_update_extended = false, $p_bypass_mail = false ) {
 			email_bug_updated( $c_bug_id );
 		}
 
+		# Execute all registered update callbacks
+		while( !empty( $this->update_callbacks ) ) {
+			$t_callback = array_shift( $this->update_callbacks );
+			call_user_func_array( $t_callback['func'], $t_callback['params'] );
+		}
+
+		event_signal( 'EVENT_UPDATE_BUG', array( $t_old_data, $this ) );
```
vboctor: Should we re-load a fresh instance from bug data to make sure that the results of bug data are reflected in the data passed to the plugins?

vboctor:
Is it possible to retire bug_set_field()?  If there are valid scenarios for its usage, then can we propose a model that doesn't allow new code to be written using the old model where field like 'status' is changed using it?

:+1: once comments are addressed, didn't test.

cproensa:
> Is it possible to retire bug_set_field()?
Aside from compatibility issues (as it may be used by 3rd p. code, plugins)...

The usages of bug_set_field remaining in code are in bug_copy().
Which means the BugData::create() method probably needs also to be revised in the same way than "update"

bug_actiongroup.php
cproensa: basically ( false, true ) is to skip email notificacions, on those point where it was processed separately.
i agree to make it clearer.

core/bug_api.php
cproensa: you mean: the fields that have just been updated?
technically, that's not needed... they should be the same as in the current object.
in a defensive scenario: i'd be more concerned about "old_data" being modified on its way through the plugins.

cproensa:
_Reminder to myself:
explain potential implications of the return value of events_

cproensa:
Currently we have EVENT_UPDATE_BUG_DATA defined as "chain". This means that the returned value by the event call is the object expected to be used.

In the case of modifying a BugData object, this is not needed, as the BugData object passed as parameter to the event is a reference, and changes made through this reference are made into the actual (original) object.

Let's see the BugData::update() function after this PR:

inside update(), we call:
event_signal( 'EVENT_UPDATE_BUG_DATA', $this, $t_old_data );
and later, we update the bug with the values from $this.

If the plugin makes changes to the supplied BugData reference, there is no problem, they are saved in the actual object.
But, being a chain event, the returned reference may be altered, to another object not being the original one. This doesnt make much sense, but is possible though. For example:

function plugin_bug_data_hook( $event, $updated_bug, $old_bug ){
    $another_data = new BugData();
    ...
    return $another_data;
}

back to the BugData::update(), we cannot update "self" with another reference, from inside the object.

Hence, my proposal is to change the event type to "execute", and make clear that changes to the BugData object to be updated, must be made through the supplied reference, and the hooked function does not need to return anyting.
This is the cleaner approach, this could be done for master branch, since this may be a breaking change.

As for 1.3.x., to not break current plugin definitions, the fix would be:

inside update():
$updated_bug = event_signal( 'EVENT_UPDATE_BUG_DATA', $this, $t_old_data );
then:
perform all subsequent actions with $updated_bug, instead of $this.


What do you think?

dregad:
> proposal is to change the event type to "execute", and make clear that changes to the BugData object to be updated, must be made through the supplied reference

I agree it's a cleaner solution, but I think we need to consider the questions a plugin author may ask:

- how will I know that something has changed ? 
- what will I need to change in my code ?
- will there be (visible) errors if I execute my plugin unchanged in the next version ? 

Forgive me as I don't have time to check myself at the moment.

cproensa:
> how will I know that something has changed ?

The same way you have done now: comparing old_bugdata with new_bugdata, as both are being provided as parameters

>    what will I need to change in my code ?

In my plugins i always have modified the object provided as "updated_bug" parameter, and at the end of the function, return this reference. In this case, no change is needed.
Worst case is that the plugin, for any reason, returns a completely different object, then would not be compatible with the new event type. 

>    will there be (visible) errors if I execute my plugin unchanged in the next version ?
At first, no errors should arise. In the comented incompatible case, the modified bug is lost and nono of the changes made by the plugin are performed.
If you are really concerned, a warning check may be implemented, to be raised when the returned values is not null.

But, consider that v2 is already a breaking risk for plugins. All my plugins that use plugin pages, are broken with the new UI.

dregad:
I'm not *really* concerned, was just wondering if we can make it easier for plugin authors to adapt to changes, and avoid unexpected behavior (e.g. silent data corruption) down the line.

bug_actiongroup.php
cproensa: @vboctor i see he indentation ok. It's a function definition, if needs to be changed, tell me how indentation should be here.

cproensa:
- Updated cleanup & comments as @vboctor suggested
- fixed update() to avoid early exists that would bypass the final steps
- changed the use of $this to the returned object to comply with curent event fedinition.

For @dregad comments, I'm also infavour of:
- Change behaviour of EVENT_UPDATE_BUG_DATA and EVENT_REPORT_BUG_DATA to type EVENT_TYPE_EXECUTE
- update Developer's Guide.
- add a check, if returned value from event is different from null, issue a warning. (In this case, how to do it exactly?)

However for the current PR, i have left that out, by adapting to current event definition (using the returned value instead of $this).

Once the code is integrated in both v1.3 and v2, then perform the changes that would only apply to master-v2
Additionally, for a future work, EVENT_REPORT_BUG_DATA and EVENT_REPORT_BUG should be also moved inside BugData, to allow for this event to be triggered from soap api, etc.

cproensa:
I am thinking again if there should be new events defined to replace 
EVENT_UPDATE_BUG_DATA
EVENT_REPORT_BUG_DATA
EVENT_REPORT_BUG
EVENT_UPDATE_BUG

With this branch, the events are placed inside BugData, and dont have to be triggered by outside code at every place the object is updated, like is done now.
We can refactor mantis code, but there may be external code that still use this methodology. For example, the plugin EmailReporting creates bugs and call those events by itself. This would make EVENT_REPORT_BUG to be triggered twice for the same bug, which can lead to inconsistent behaviour for plugins that listen to that event.

Having new event names can provide a more compatible scenario, for example:
- Deprecate warning for old events in 1.3. Current plugins would still work. New or updated plugins can hook to the new events for improved functionality
- In v2, make the new events a mandatory replacement.

cproensa:
I have added the changes for triggering EVENT_REPORT_BUG_DATA, EVENT_REPORT_BUG inside BugData.
Still, my previous comment for defining new events, is valid.
I'd like to hear some input on this.

cproensa:
closed in favour of #966

PR #835, opened 2016-08-01
Fix checking of limit_reporters and report_bug_threshold

cproensa:
"report_bug_threshold" was assumed to be an integer in some places
It can also be an array, so, in these cases some errors happens...

Fixes:
https://www.mantisbt.org/bugs/view.php?id=21579
https://www.mantisbt.org/bugs/view.php?id=21580
https://www.mantisbt.org/bugs/view.php?id=21581

More commits for fixing workflow config pages issues:
(added in this PR as they all modify the same set of pages)
https://www.mantisbt.org/bugs/view.php?id=21583
https://www.mantisbt.org/bugs/view.php?id=21582

### DIFF
--- a/changelog_page.php
+++ b/changelog_page.php
@@ -180,9 +180,6 @@ foreach( $t_project_ids as $t_project_id ) {
 	$t_project_name = project_get_field( $t_project_id, 'name' );
 	$t_can_view_private = access_has_project_level( config_get( 'private_bug_threshold' ), $t_project_id );
 
-	$t_limit_reporters = config_get( 'limit_reporters' );
-	$t_user_access_level_is_reporter = ( config_get( 'report_bug_threshold', null, null, $t_project_id ) == access_get_project_level( $t_project_id ) );
-
 	$t_resolved = config_get( 'bug_resolved_status_threshold' );
 
 	# grab version info for later use
@@ -192,6 +189,10 @@ foreach( $t_project_ids as $t_project_id ) {
 	category_get_all_rows( $t_project_id );
 
 	$t_project_header_printed = false;
+
+	$t_limit_reporters = config_get( 'limit_reporters' );
+	$t_report_bug_threshold = config_get( 'report_bug_threshold', null, null, $t_project_id );
+	$t_access_limit_reporters_applies = !access_has_project_level( access_threshold_min_level( $t_report_bug_threshold ) + 1, $t_project_id );
 
 	foreach( $t_version_rows as $t_version_row ) {
 		$t_version_header_printed = false;
@@ -232,8 +233,9 @@ foreach( $t_project_ids as $t_project_id ) {
 
 			# check limit_Reporter (Issue #4770)
 			# reporters can view just issues they reported
-			if( ON === $t_limit_reporters && $t_user_access_level_is_reporter &&
-							!bug_is_user_reporter( $t_row['id'], $t_user_id )) {
+			if( ON == $t_limit_reporters
+					&& $t_access_limit_reporters_applies
+					&& !bug_is_user_reporter( $t_row['id'], $t_user_id ) ) {
 				continue;
 			}
 
--- a/core/access_api.php
+++ b/core/access_api.php
@@ -414,13 +414,10 @@ function access_has_bug_level( $p_access_level, $p_bug_id, $p_user_id = null ) {
 		static $s_thresholds = array();
 		if( !isset( $s_thresholds[$t_project_id] ) ) {
 			$t_report_bug_threshold = config_get( 'report_bug_threshold', null, $p_user_id, $t_project_id );
-			if( !is_array( $t_report_bug_threshold ) ) {
-				$s_thresholds[$t_project_id] = $t_report_bug_threshold + 1;
-			} else if( empty( $t_report_bug_threshold ) ) {
+			if( empty( $t_report_bug_threshold ) ) {
 				$s_thresholds[$t_project_id] = NOBODY;
 			} else {
-				sort( $t_report_bug_threshold );
-				$s_thresholds[$t_project_id] = $t_report_bug_threshold[0] + 1;
+				$s_thresholds[$t_project_id] = access_threshold_min_level( $t_report_bug_threshold ) + 1;
 			}
 		}
 		if( !access_compare_level( $t_access_level, $s_thresholds[$t_project_id] ) ) {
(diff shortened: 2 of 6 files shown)


changelog_page.php
```diff
@@ -232,8 +229,11 @@ function print_project_header_changelog ( $p_project_name ) {
 
 			# check limit_Reporter (Issue #4770)
 			# reporters can view just issues they reported
-			if( ON === $t_limit_reporters && $t_user_access_level_is_reporter &&
-							!bug_is_user_reporter( $t_row['id'], $t_user_id )) {
+			$t_limit_reporters = config_get( 'limit_reporters' );
+			$t_report_bug_threshold = config_get( 'report_bug_threshold', null, null, $t_project_id ); // == access_get_project_level( $t_project_id ) );
+			if( ON == $t_limit_reporters
+					&& !access_has_project_level( access_threshold_min_level( $t_report_bug_threshold ) + 1, $t_project_id )
```
vboctor: Why do we need the +1 here?

changelog_page.php
```diff
@@ -232,8 +229,11 @@ function print_project_header_changelog ( $p_project_name ) {
 
 			# check limit_Reporter (Issue #4770)
 			# reporters can view just issues they reported
-			if( ON === $t_limit_reporters && $t_user_access_level_is_reporter &&
-							!bug_is_user_reporter( $t_row['id'], $t_user_id )) {
+			$t_limit_reporters = config_get( 'limit_reporters' );
+			$t_report_bug_threshold = config_get( 'report_bug_threshold', null, null, $t_project_id ); // == access_get_project_level( $t_project_id ) );
```
vboctor: I don't think the comment here is necessary.

changelog_page.php
```diff
@@ -232,8 +229,11 @@ function print_project_header_changelog ( $p_project_name ) {
 
 			# check limit_Reporter (Issue #4770)
 			# reporters can view just issues they reported
-			if( ON === $t_limit_reporters && $t_user_access_level_is_reporter &&
-							!bug_is_user_reporter( $t_row['id'], $t_user_id )) {
+			$t_limit_reporters = config_get( 'limit_reporters' );
```
vboctor: Wouldn't it be better if we move this outside the loop?  I would also calculate the minimum access level outside the loop.

core/access_api.php
```diff
@@ -709,3 +706,21 @@ function access_level_get_string( $p_access_level ) {
...
+
+/**
+ * Return the minimum access level, as integer, that matches the threshold.
+ * $p_threshold may be a single value, or an array. If it is a single
+ * value, returns that number. If it is an array, return the value of the
+ * smallest element
+ * @param integer|array $p_threshold         Access threshold
+ * @return integer		Integer value for an access level.
+ */
+function access_threshold_min_level( $p_threshold ) {
```
vboctor: Consider handling the empty array case.

core/filter_api.php
```diff
@@ -1236,7 +1236,7 @@ function filter_get_bug_rows( &$p_page_number, &$p_per_page, &$p_page_count, &$p
 
 		foreach( $t_project_ids as $t_pid ) {
 			# limit reporters to visible projects
-			if( ( ON === $t_limit_reporters ) && ( !access_has_project_level( config_get( 'report_bug_threshold', null, $t_user_id, $t_pid ) + 1, $t_pid, $t_user_id ) ) ) {
+			if( ( ON === $t_limit_reporters ) && ( !access_has_project_level( access_threshold_min_level( config_get( 'report_bug_threshold', null, $t_user_id, $t_pid ) ) + 1, $t_pid, $t_user_id ) ) ) {
```
vboctor: Should we have an access check API that handles the limit reporters check rather than duplicating it in so many places?

manage_config_workflow_page.php
```diff
@@ -344,8 +345,11 @@ function access_row() {
 			# 'NEW' status
 			$t_level_project = $t_project_new;
 
-			$t_can_change = ( $g_access >= config_get_access( 'report_bug_threshold' ) );
-			$t_color = set_color_override( $t_file_new, $t_global_new, $t_project_new );
+			# If report_bug_threshold is an array (instead of an integer value), the input is not editable
+			# because it must be configured in manage_config_work_threshold_page.
+			$t_can_change = ( $g_access >= config_get_access( 'report_bug_threshold' ) )
```
vboctor: Should we avoid comparisons like this and use access API to do such checks, so that arrays are handled?  Or maybe just move the !is_array() as the first check.

vboctor:
:+1: didn't test.

changelog_page.php
cproensa: yes... i moved those lines closer to the action... and missed to keep them outside of the loop

changelog_page.php
cproensa: This code is iterating the bugs to add to the list.
IF the option to limit reporters is set
AND we are not (project access level) GREATER THAN reporter 
AND we are not the reporter user in the bug.
THEN do not show this bug

The logic is negated, as if we were higher that a reporter, we could see all bugs

core/access_api.php
cproensa: Right. will fix

core/filter_api.php
cproensa: Yes, probably. Baically the related questions that need to be aswered usually are:
Do i have access to this bug?
Can I see all bugs or just mine?

However, after hitting the limitations of this option (which is the reason i had to use array thresholds, and find the bugs...)
I think it needs to be rethinked, into a more general "visibility" configuration.

manage_config_workflow_page.php
cproensa: config_get_access( 'report_bug_threshold' ) should not be affected with the actual value of the option, is checking "who can change this".
Actually the !is_array is for other reason, as the idea is that a threshold as an array cannot be edited as a plain dropdown select.
This commit is to not not make that field editable if its not a numeric value.

cproensa:
Added the fixes suggested by Victor. (thanks.)

I leave for future work a common implementation for "limit_reporters" related access checks.

vboctor:
@cproensa This doesn't rebase cleanly on master-1.3.x or master.

cproensa:
> This doesn't rebase cleanly on master-1.3.x or master.

right, might be due to having merged two local branches...
Anyway, i have rewritten the branch clean, rebased, and squashed corrections, targeted at 1.3.x

However, i couldn't rebase into master, getting errors on unrelated files (is that right?)

PR #836, opened 2016-08-04
Modify empty and default cases for workflow configuration

cproensa:
Fix several inconsistencies with the treatment of status_enum_workflow
configuration.

(1) Fix showing a non existant status row in workflow config array.
This is interpreted as if that status allows all transitions, but in the
config page, the checks were shown unmarked, which can be confusing.
Now the checks appear marked in this situation.

(2) As consequence of (1), now the message "You cannot move an issue out
of this status" is not showed in that situation.

(3) When a status row is saved with all transitions unchecked, now an
empty enum string is saved, meaning effectively that no transitions is
allowed from that status.
This specific configuration was previously unavailable from the workflow
config page.

Fixes: #20682


vboctor:
:+1: haven't tested.

dregad:
:+1: 

Merged in c2bee49c4bdadb7e7c3eea177331b60f05efca03

PR #73, opened 2013-02-15
Favicon and search box improvements

Wizmaster:
I made two improvements for MantisBT:
- eed56461 allowed the configured favicon to be image/png
- 5ac340d3 modified the browser search plugin to be branded (custom name and favicon)

Enjoy :)

PR #840, opened 2016-08-09
Opensearch improvements

dregad:
This PR follows-up and improves on #73, fixes [#11964](https://www.mantisbt.org/bugs/view.php?id=11964)

- Use standard favicon instead of hardcoded PNG for Mantis search engines
- Define new config for search engine name prefix (defaults to $g_window_title)
- Localize search engine ShortNames and Descriptions

### DIFF
--- a/browser_search_plugin.php
+++ b/browser_search_plugin.php
@@ -36,31 +36,28 @@ require_api( 'gpc_api.php' );
 
 $f_type = gpc_get_string( 'type', 'text' );
 
+$t_path = config_get_global( 'path' );
+$t_searchform = $t_path . 'view_all_bug_page.php';
+
+if( strtolower( $f_type ) == 'id' ) {
+	$t_shortname = 'MantisBT IssueId';
+	$t_description = 'MantisBT Issue Id Search';
+	$t_url = $t_path . 'view.php?id={searchTerms}';
+} else {
+	$t_shortname = 'MantisBT Search';
+	$t_description = 'MantisBT Text Search';
+	$t_url = $t_path . 'view_all_set.php?type=1&amp;temporary=y&amp;handler_id=[all]&amp;search={searchTerms}';
+}
+
 header( 'Content-Type: application/opensearchdescription+xml' );
 ?>
+<?xml version="1.0" encoding="UTF-8" ?>
 <OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/"
                        xmlns:moz="http://www.mozilla.org/2006/browser/search/">
-<?php
-$t_path = config_get_global( 'path' );
-
-if( strtolower( $f_type ) == 'id' ) {
-	echo '<ShortName>MantisBT IssueId</ShortName>';
-	echo '<Description>MantisBT Issue Id</Description>';
-	echo '<InputEncoding>UTF-8</InputEncoding>';
-} else {
-	echo '<ShortName>MantisBT Search</ShortName>';
-	echo '<Description>MantisBT Text Search</Description>';
-	echo '<InputEncoding>UTF-8</InputEncoding>';
-}
-?>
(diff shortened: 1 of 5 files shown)


config_defaults_inc.php
```diff
@@ -851,13 +851,27 @@
...
+ * OpenSearch engine title prefix.
+ * This is used to describe Browser Search entries, and must be short enough
+ * so that when inserted into the 'opensearch_XXX_short' language string, the
+ * resulting text is 16 characters or less, to be compliant with the limit for
+ * the ShortName element as defined in the OpenSearch specification.
+ * @link http://www.opensearch.org/Specifications/OpenSearch/1.1
+ * @see $g_window_title
+ * @global string $g_search_title
+ */
+$g_search_title = '%window_title%';
```
vboctor: Please mark as a public config.

browser_search_plugin.php
```diff
@@ -34,33 +34,32 @@
...
-$f_type = gpc_get_string( 'type', 'text' );
+$f_type = strtolower( gpc_get_string( 'type', 'text' ) );
 
-header( 'Content-Type: application/opensearchdescription+xml' );
-?>
-<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/"
-                       xmlns:moz="http://www.mozilla.org/2006/browser/search/">
-<?php
 $t_path = config_get_global( 'path' );
+$t_title = config_get_global( 'search_title' );
```
vboctor: I wouldn't use config_get_global() to retrieve search title.  window_title can be specified in the db, the same should be for search title.

vboctor:
:+1: once comments are applied.

atrol:
:+1: once comments are applied.

config_defaults_inc.php
atrol: A bit off topic.
I don't like to introduce one more of this '%var%' stuff.
I would like to discuss removal of recursion in defining configuration variables.
I don't see that much value in it and would expect a better performance if config_eval would be removed and
config_get_global and config_get would be simplified.

dregad: > I don't see that much value in it

It makes configuration easier for admins, by avoiding the need to duplicate declaration of several related options.  Consider the case of paths, or `$g_cookie_prefix` for example. Without the '%%' feature, the admin would have to manually set each of the cookies names in their config_inc.php.

You will probably and rightly argue that we do not need to make the cookies names configurable. Of course by refactoring the code (in the above case, getting rid of these config options), we can implement alternative that works without the recursion.

The questions are, whether this is worth the effort and what kind of performance improvement we can obtain by removing this feature.

> expect a better performance if config_eval would be removed and config_get_global and config_get would be simplified.

If this is important to you, I would suggest to open an issue in the tracker, and evaluate how much we'd gain with your proposed approach.

dregad:
> Please mark as a public config.

Done

> I wouldn't use config_get_global() to retrieve search title

Fixed (also for *favicon_image*).

config_defaults_inc.php
atrol: > evaluate how much we'd gain with your proposed approach.

Quick and dirty removal of config_eval
https://github.com/atrol/mantisbt/tree/config_eval_remove

There is a bit less memory usage (removed the caching of the evaled options) and there is about 7% better performance in 2nd benchmark, see results below.
Not outstanding, but not that bad for such a small part of the code.

Do you think it's worth to think more about it (obsolete cookie seetings, ...) and target it for a version > 2.0?

My View 2nd call
------------------

Original master-1.3.x
Page execution time: 0.3540 seconds
Memory usage: 10,411 KiB

Removed config_eval
Page execution time: 0.3499 seconds
Memory usage: 10,388 KiB

View Issues 2nd call
--------------------

Original master-1.3.x
Page execution time: 0.2916 seconds
Memory usage: 10,514 KiB

Removed config_eval
Page execution time: 0.2705 seconds
Memory usage: 10,441 KiB

dregad: > about 7% better performance in 2nd benchmark

Not bad indeed, I did not expect that much actually. I'm still concerned about the loss of "just in time" evaluation of configs however, especially for paths that are quite likely to have been changed by admins - we need to be very careful not to cause regressions. Consider this example:

_config_defaults_inc.php_
```
$g_path = (dynamically built from $_SERVER) // e.g. http://internal-hostname.example.com/mantisbt/
$g_icon_path = $g_path . 'images/';
```
_config_inc.php_
```
$g_path = 'http://example.com/mantisbt/'; # required e.g. due to use of reverse proxy
```

In this case, `$g_icon_path == 'http://internal-hostname.example.com/mantisbt/images/`, which is invalid in the reverse-proxy scenario; this would force the admin to manually define all variables referencing $g_path (formerly %path% in their config_inc.php (and knowing that they have to do so). Not very user-friendly, and likely to cause support issues when upgrading.

> Do you think it's worth to think more about it (obsolete cookie seetings, ...) 

Yes definitely. And get others' opinions too.

> target it for a version > 2.0?

For sure a 2.x target, if we decide to move forward.

PR #841, opened 2016-08-11
Fix error for excel export page

cproensa:
Fix an error introduced by commit f233869f, that makes
excel_xml_export.php throw an error when exporting issues.


dregad:
:+1:

dregad:
Fixed in 84f43e0f32cdc8246f261a2254b2a0468555f944

PR #842, opened 2016-08-12
Add new password string in russian language

ParseLab:


dregad:
Translations are managed via TranslateWiki. Kindly registere am account and submit your contribution there. 

See https://www.mantisbt.org/wiki/doku.php/mantisbt:translationshowto

PR #844, opened 2016-08-13
Align 'prune accounts' button in manage user page.

syncguru:
Fixes #21114

screen shot 2016-08-13 at 2 09 39 pm


dregad:
Just tested, confirm this fixes the problem. Thanks !

:+1:

PR #848, opened 2016-08-14
Use multiple tokens for action links in same page

cproensa:
Previously, due to performance issues with session store, only one security token was used for several action links in the same page
The side effect is that only one of those actions can be performed from
the specific originating page.
This is especially a drawback when the actions are links that can be open
in new browsers tabs.

Reverting to separate tokens allows to perform all the actions without
having to reload the originating page.

Note, the original performance issue was solved at the session layer
with 8092c3d.

### DIFF
--- a/bug_revision_view_page.php
+++ b/bug_revision_view_page.php
@@ -100,11 +100,10 @@ if( $f_bug_id ) {
  */
 function show_revision( array $p_revision ) {
 	static $s_can_drop = null;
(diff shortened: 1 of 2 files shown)


vboctor:
:+1: not tested.

vboctor:
Merged via f160663f2035497847786400694befb524bc3eb5 and 77db0389feed47af5896f8455b2ca13d262b5b35

PR #854, opened 2016-08-15
Implement bugnotes_count, attachment_count in CSV, Excel export (2)

cproensa:
Resubmit of PR #696 
rebased to current master-1.3.x

Original PR found a problem where the queries used to cache bugnotes may run out of limits for long IN clauses.
This should not happen after #832 was implemented, since bugs are treated in batches of controlled size

### DIFF
--- a/core/csv_api.php
+++ b/core/csv_api.php
@@ -480,3 +480,35 @@ function csv_format_due_date( BugData $p_bug ) {
 function csv_format_sponsorship_total( BugData $p_bug ) {
 	return csv_escape_string( $p_bug->sponsorship_total );
 }
+
+/**
+ * return the attachment count for an issue
+ * @param BugData $p_bug A BugData object.
+ * @return string
+ * @access public
+ */
+function csv_format_attachment_count( BugData $p_bug ) {
+	# Check for attachments
+	$t_attachment_count = 0;
+	if( file_can_view_bug_attachments( $p_bug->id, null ) ) {
+		$t_attachment_count = file_bug_attachment_count( $p_bug->id );
+	}
+	return csv_escape_string( $t_attachment_count );
+}
+
+/**
+ * return the bug note count for an issue
+ * @param BugData $p_bug A BugData object.
+ * @return string
+ * @access public
+ */
+function csv_format_bugnotes_count( BugData $p_bug ) {
+	# grab the bugnote count
+	$t_bugnote_stats = bug_get_bugnote_stats( $p_bug->id );
+	if( $t_bugnote_stats ) {
+		$t_bugnote_count = $t_bugnote_stats['count'];
+	} else {
+		$t_bugnote_count = 0;
+	}
+	return csv_escape_string( $t_bugnote_count );
+}
--- a/core/excel_api.php
+++ b/core/excel_api.php
@@ -252,15 +252,6 @@ function excel_format_reporter_id( BugData $p_bug ) {
 }
 
 /**
- * Gets the formatted number of bug notes.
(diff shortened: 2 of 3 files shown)


vboctor:
The logic to calculate the bugnote count and attachment count based on user access is repeated in 4 places.  columns api for view issues, columns api for print issues, csv, and Excel.  How about we add a method for each in the appropriate API that is used from the 4 locations.

cproensa:
> The logic to calculate the bugnote count and attachment count based on user access is repeated in 4 places.

that was talked in previous PR.
lets see if it can be added here

cproensa:
@vboctor 
regarding the duplicated logic:
- get attachment count: this should be placed in BugData, which already has "attachment_count" and "get_attachment_count()", and those are currently unused.
- get bugnote count: equally, BugData has "bugnotes_count" and "get_bugnotes_count()", also unused.

The problem with those BugData methods/properties:
get_bugnotes_count: uses an ad-hoc query, which should be take advance of cache(*)
both those properties: they are public, but are not calculated at object creation. At which point should they be calculated? probably an improved getter should fill them as needed.
(*) We still dont have a solid way to cache BugData objects. We only cache bug_attr rows, which has the core attributes, but not these calculated ones.

I have some PR in work and planning regarding BugData:
- update events (submitted)
- creation events (planned)
- improve BugData properties management. This would include improved getter/setters, and cache ability for all fields. To progressively move more logic, like that discussed here, into BugData.

I'd prefer you open both requests as issues so i can track them in the future?

vboctor:
@cproensa ok, please open the issue(s) to track this.  :+1:

cproensa:
> please open the issue(s) to track this.

open as #21629, #21629

vboctor:
@cproensa I was in the process of merging this and found the following:

- Found db_param_push() without a db_param_pop() in the bug_get_bugnote_stats() API.  Shouldn't the caller does the push / pop if needed?  -- @dregad we should clarify in the comments for db_param_push() on where it should be used.  That is kind of complexity I was worried about.

- Why did you stop using bug_get_bugnote_stats_array() which takes into consideration access level?

- # @todo - optimise - max(), count() -- why didn't you handle this todo, seems simple enough.

- It doesn't merge with latest on master, not a big deal.  But you may want to rebase on top of master while you are at it.

cproensa:
I'll have a look, but cant until next week

cproensa:
>   It doesn't merge with latest on master, not a big deal. But you may want to rebase on top of master while you are at it.
>   Why did you stop using bug_get_bugnote_stats_array() which takes into consideration access level?

right, this was previous to #856. I will rebase and adapt the changes.

>  why didn't you handle this todo, seems simple enough.
>       @todo - optimise - max(), count() --

That @todo disppeared in said PR: as visible bugnotes must be evaluated per project, an optimization for count or max at query level is not feasible.

>    Found db_param_push() without a db_param_pop() in the bug_get_bugnote_stats() API. Shouldn't the caller does the push / pop if needed?

afaict, there is no problem there

dregad:
:+1:

cproensa:
I have rebased to current head, and cleared all conflicts

dregad:
Should this go in 1.3.1 ?

cproensa:
> Should this go in 1.3.1 ?

i'd prefer not to include it as a last time change. That will give time until 1.3.2 to find any unexpected error, if there is any...

dregad:
Merged in 1279ac3a7f71788bbbf093d92d15d037a7fe6f6f

PR #855, opened 2016-08-16
Optimize vertical space in "My View" page

syncguru:
Fixes #21123

Before:
screen shot 2016-08-15 at 7 32 27 pm

After:
screen shot 2016-08-15 at 7 30 27 pm

### DIFF
--- a/css/ace-mantis.css
+++ b/css/ace-mantis.css
@@ -126,8 +126,7 @@
     background: none repeat scroll 0 0 #393939;
 }
 
-#filter .widget-toolbar > .widget-menu > a,
-#bug-update .widget-toolbar > .widget-menu > a {
+.widget-toolbar > .widget-menu > a {
     display: inline-block;
     font-size: 11px;
     padding: 0 5px;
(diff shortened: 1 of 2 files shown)


vboctor:
How about adding a link icon after the issue count?

Here is the URL for the icon
http://fontawesome.io/icon/link/

dregad:
I'm fine with removing the table's footer which does not add any value, but is it really necessary to add a burger icon with a menu that only provides a link which is already existing ?

I would propose to remove the menu, leaving only the link on the table's title. If you do add the link icon per victor's suggestion, IMO it would be better to have it next to the link (i.e. *before* the issue count, not after)

atrol:
What about using the icon on the left top of the box (change fa-list-alt to fa-link) for the link?

syncguru:
@vboctor - too busy for table header
@atrol - that's anti-pattern

syncguru:
Updated the PR, here is how it looks now:

screen shot 2016-08-21 at 12 31 44 pm

vboctor:
:+1:

vboctor:
Merged via 5de037c326d550535afad3288c72bf9876439630

PR #856, opened 2016-08-16
 Count only visible notes

cproensa:
Count only visible notes
    
    Fix showing the number of notes to not show those that are not visible
    according to user access levels (private notes).
    Bugnote count is built alongside bugnote statistics.
    This commit rewrites the statistics generation to account for user access
    level for each bug and each project.
    
    Fixes: #0006939
    
    Removes duplicate logic (sql queries) from:
        bug_api.php: function bug_get_bugnote_stats
        filter_api.php: function filter_get_bug_rows

Force storing empty bugnote statistics
    
    (1) When a bug has no bugnotes, or visible ones, force storing a 'false'
    value as bugnote stats in bug cache, to not query the database again
    later.
    
    (2) Additionally, when the function bug_cache_database_result is called with
    a stats array parameter, and the bug has already been cached, force the
    creation of the stats index.
    Previously this wasn't possible if the bug was already cached without
    the bugnote stats.
    
    Fixes #0020121

Reworked after previous PR #650

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -790,16 +790,24 @@ $g_cache_bug_text = array();
 
 /**
  * Cache a database result-set containing full contents of bug_table row.
- * @param array $p_bug_database_result Database row containing all columns from mantis_bug_table.
- * @param array $p_stats               An optional array representing bugnote statistics.
+ * $p_stats parameter is an optional array representing bugnote statistics.
+ * This parameter can be "false" if the bug has no bugnotes, so the cache can differentiate
+ * from a still not cached stats registry.
+ * @param array $p_bug_database_result  Database row containing all columns from mantis_bug_table.
+ * @param array|boolean|null $p_stats   Optional: array representing bugnote statistics, or false to store empty cache value
  * @return array returns an array representing the bug row if bug exists
  * @access public
  */
-function bug_cache_database_result( array $p_bug_database_result, array $p_stats = null ) {
+function bug_cache_database_result( array $p_bug_database_result, $p_stats = null ) {
 	global $g_cache_bug;
 
 	if( !is_array( $p_bug_database_result ) || isset( $g_cache_bug[(int)$p_bug_database_result['id']] ) ) {
-		return $g_cache_bug[(int)$p_bug_database_result['id']];
+		if( !is_null($p_stats) ) {
+			# force store the bugnote statistics
+			return bug_add_to_cache( $p_bug_database_result, $p_stats );
+		} else {
+			return $g_cache_bug[(int)$p_bug_database_result['id']];
+		}
 	}
 
 	return bug_add_to_cache( $p_bug_database_result, $p_stats );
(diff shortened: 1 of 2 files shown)


cproensa:
My branch is based on master-1.3.x, but submitted this PR against master unintendedly. However, seems like there are no conflicts.

core/bug_api.php
```diff
@@ -795,11 +795,16 @@ function update( $p_update_extended = false, $p_bypass_mail = false ) {
  * @return array returns an array representing the bug row if bug exists
  * @access public
  */
-function bug_cache_database_result( array $p_bug_database_result, array $p_stats = null ) {
+function bug_cache_database_result( array $p_bug_database_result, $p_stats = null ) {
```
dregad: You should also update the phpDoc block to reflect the change in function signature.

core/bug_api.php
```diff
@@ -1545,6 +1550,93 @@ function bug_get_newest_bugnote_timestamp( $p_bug_id ) {
 }
 
 /**
+ * For a list of bug ids, return array of bugnote stats:
```
dregad: phpDoc block should start with a 1-line summary of what the function does, followed by a period or empty line.

core/bug_api.php
```diff
@@ -1545,6 +1550,93 @@ function bug_get_newest_bugnote_timestamp( $p_bug_id ) {
...
+	if ( null === $p_user_id ) {
+		$t_user_id = auth_get_current_user_id();
+	}
+	else {
+		$t_user_id = $p_user_id;
+	}
+
+	db_param_push();
+	$t_params = array();
+	$t_dbparam = array();
```
dregad: I would suggest a different name than $t_db_param to avoid confusion with $t_param. Maybe $t_in_clause_elements or something similar.

core/bug_api.php
```diff
@@ -1545,6 +1550,93 @@ function bug_get_newest_bugnote_timestamp( $p_bug_id ) {
...
+		. ' ORDER BY b.project_id, n.bug_id, n.last_modified';
+	# perform query
+	$t_result = db_query( $t_query, $t_params );
+	$t_query_row = db_fetch_array( $t_result );
+	$t_counter = 0;
+	$t_stats = array();
+	# We need to check for each bugnote if it has permissions to view in respective project.
+	# bugnotes are grouped by project_id and bug_id to save calls to config_get
+	if( $t_query_row ) {
+		do {
```
dregad: Not sure why you use `if( do { ... } )` here ? 
It seems simpler to do `while( $t_query_row = db_fetch_array( $t_result ) ) { ... }` instead

core/bug_api.php
```diff
@@ -1545,6 +1550,93 @@ function bug_get_newest_bugnote_timestamp( $p_bug_id ) {
...
+			$c_bug_id = (int)$t_query_row['bug_id'];
+			if( 0 == $t_counter || $t_current_project_id !== $t_query_row['project_id'] ) {
+				# evaluating a new project from the rowset
+				$t_current_project_id = $t_query_row['project_id'];
+				$t_user_access_level = access_get_project_level( $t_query_row['project_id'], $t_user_id );
+				$t_private_bugnote_visible = access_compare_level( $t_user_access_level, config_get( 'private_bugnote_threshold', null, $t_user_id, $t_query_row['project_id'] ) );
```
dregad: Split long line

core/bug_api.php
```diff
@@ -1545,6 +1550,93 @@ function bug_get_newest_bugnote_timestamp( $p_bug_id ) {
...
+			}
+			if( 0 == $t_counter || $t_current_bug_id !== $c_bug_id ) {
+				# evaluating a new bug from the rowset
+				$t_current_bug_id = $c_bug_id;
+				$t_note_count = 0;
+				$t_last_submit_date= 0;
+			}
+			$t_note_visible = $t_private_bugnote_visible || $t_query_row['reporter_id'] == $t_user_id || ( VS_PUBLIC == $t_query_row['view_state'] );
```
dregad: Split long line

dregad:
:+1: on principle following code review, but haven't tested yet

core/bug_api.php
cproensa: done

cproensa:
Updated with @dregad suggestions

dregad:
Merged in c7bb7f7a5611e2d99d14c2322cbdb4505b4825ba

PR #858, opened 2016-08-17
Don't show "Upgrade Your Installation" when switching admin bar tabs

syncguru:
Fixes #21602

### DIFF
--- a/admin/email_queue.php
+++ b/admin/email_queue.php
@@ -25,6 +25,8 @@
  */
 
 require_once( dirname( dirname( __FILE__ ) ) . '/core.php' );
+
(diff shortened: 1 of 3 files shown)


dregad:
I think adding a comment indicating why schema.php has to be included in each of these files would be useful, since it is *a priori* unrelated to the task at hand.

syncguru:
@dregad Comment added.

dregad:
Thanks. :+1:

vboctor:
:+1:

vboctor:
Merged via 2a64d17673329e0897acecc3f56e21ffcc9405b5

PR #859, opened 2016-08-17
Provide API for plugins to update CSP header

vboctor:
Provide an API to manage the Content-Security-Header and change the Gravatar plugin to use it.

This fixes a bug where Gravatar plugin was generating its own header and hence overwriting the default headers.

Fixes #21263

### DIFF
--- a/core/http_api.php
+++ b/core/http_api.php
@@ -154,6 +154,11 @@ function http_content_headers() {
 function http_csp_add( $p_type, $p_value ) {
 	global $g_csp;
 
+	if ( $g_csp === null ) {
+		# Development error, headers already emitted.
+		trigger_error( ERROR_GENERIC, ERROR );
+	}
+
 	if ( isset( $g_csp[$p_type] ) ) {
 		if ( !in_array( $p_value, $g_csp[$p_type] ) ) {
 			$g_csp[$p_type][] = $p_value;
@@ -170,6 +175,11 @@ function http_csp_add( $p_type, $p_value ) {
 function http_csp_value() {
 	global $g_csp;
 
+	if ( $g_csp === null ) {
+		# Development error, headers already emitted.
+		trigger_error( ERROR_GENERIC, ERROR );
+	}
+
 	$t_csp_value = '';
 
 	foreach ( $g_csp as $t_key => $t_values ) {
@@ -179,6 +189,17 @@ function http_csp_value() {
 	$t_csp_value = trim( $t_csp_value, '; ' );
 
 	return $t_csp_value;
+}
+
+/**
+ * Send header for Content-Security-Policy.
+ * @return void
+ */
+function http_csp_emit_header() {
+	header( 'Content-Security-Policy: ' . http_csp_value() );
+
+	global $g_csp;
+	$g_csp = null;
 }
 
 /**
@@ -209,8 +230,7 @@ function http_security_headers() {
 			http_csp_add( 'style-src', "'unsafe-inline'" );
 		}
 
-		# Set CSP header
-		header( 'Content-Security-Policy: ' . http_csp_value() );
+		http_csp_emit_header();
 
 		if( http_is_protocol_https() ) {
 			header( 'Strict-Transport-Security: max-age=7776000' );
--- a/core/http_api.php
+++ b/core/http_api.php
@@ -30,6 +30,12 @@
 require_api( 'config_api.php' );
 
 /**
+ * The Content-Security-Policy settings array.  Use http_csp_add() to update it.
+ * @var array
+ */
+$g_csp = array();
+
+/**
  * Checks to see if script was queried through the HTTPS protocol
  * @return boolean True if protocol is HTTPS
  */
@@ -139,6 +145,43 @@ function http_content_headers() {
 }
 
 /**
+ * Add a Content-Security-Policy directive.
+ *
+ * @param  string $p_type  The directive type, e.g. style-src, script-src.
+ * @param  string $p_value The directive value, e.g. 'self', https://ajax.googleapis.com
+ * @return void
+ */
+function http_csp_add( $p_type, $p_value ) {
+	global $g_csp;
+
+	if ( isset( $g_csp[$p_type] ) ) {
+		if ( !in_array( $p_value, $g_csp[$p_type] ) ) {
+			$g_csp[$p_type][] = $p_value;
+		}
+	} else {
+		$g_csp[$p_type] = array( $p_value );
+	}
+}
+
+/**
+ * Constructs the value of the CSP header.
+ * @return string CSP header value.
+ */
+function http_csp_value() {
+	global $g_csp;
+
+	$t_csp_value = '';
+
+	foreach ( $g_csp as $t_key => $t_values ) {
+		$t_csp_value .= $t_key . ' ' . implode( ' ', $t_values ) . '; ';
+	}
+
+	$t_csp_value = trim( $t_csp_value, '; ' );
+
+	return $t_csp_value;
+}
+
+/**
  * Set security headers (frame busting, clickjacking/XSS/CSRF protection).
  * @return void
  */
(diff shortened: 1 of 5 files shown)


vboctor:
This is to replace PR #740.

vboctor:
Once this is checked in, I'll port to master with some changes to enable few more sources and re-enabling CSP there.

core/http_api.php
```diff
@@ -139,6 +145,59 @@ function http_content_headers() {
...
+/**
+ * Constructs the value of the CSP header.
+ * @return string CSP header value.
+ */
+function http_csp_value() {
+	global $g_csp;
+
+	$t_csp_value = '';
+
+	foreach ( $g_csp as $t_key => $t_values ) {
```
dregad: This will trigger a PHP warning if called when $g_csp is null, maybe add the same error check as in `http_csp_add()` ?

core/http_api.php
```diff
@@ -139,6 +145,59 @@ function http_content_headers() {
...
+
+	return $t_csp_value;
+}
+
+/**
+ * Send header for Content-Security-Policy.
+ * @return void
+ */
+function http_csp_emit_header() {
+	header( 'Content-Security-Policy: ' . http_csp_value() );
```
dregad: Consider not resending the CSP header if it has already been emitted ($g_csp === null).

core/http_api.php
```diff
@@ -147,32 +206,26 @@ function http_security_headers() {
...
-			$t_script_src .= " $t_cdn_url";
+			http_csp_add( 'style-src', $t_cdn_url );
+			http_csp_add( 'script-src', $t_cdn_url );
 		}
 
 		# Relaxing policy for roadmap page to allow inline styles
 		# This is a workaround to fix the broken progress bars (see #19501)
 		if( 'roadmap_page.php' == basename( $_SERVER['SCRIPT_NAME'] ) ) {
-			$t_style_src .= " 'unsafe-inline'";
+			http_csp_add( 'style-src', "'unsafe-inline'" );
```
dregad: I don't understand why you're initializing the CSP headers in `http_security_headers()`, and sending the actual header in a separate function. 

Wouldn't it make more sense to have a single API (e.g. `http_csp_headers()`) that deals with both aspects of initializing the CSP ? Or did you have a specific purpose in mind when you created a separate function ?

plugins/Gravatar/Gravatar.php
```diff
@@ -104,24 +104,14 @@ function config() {
...
-	 *
-	 * @return void
-	 */
-	function csp_headers() {
-		# Policy for images: Allow gravatar URL
 		if( config_get( 'show_avatar' ) !== OFF ) {
 			# Set CSP header
-			header( "Content-Security-Policy: img-src 'self' " .
-				self::getAvatarUrl() );
+			http_csp_add( 'img-src', self::getAvatarUrl() );
```
dregad: I would recommend to keep the `csp_headers()` function as a separate event hook, like it was before. I understand we have a problem of sequence in the core, i.e. by the time *EVENT_LAYOUT_RESOURCES* is signaled `http_security_headers()` has already been called, but I don't think it's conceptually correct for the `hooks()` method itself to update the CSP.

To be clean, I think we should declare a new event to allow plugins to properly hook their code. I can think of 2 approaches here:

1. Use a CSP-specific event (e.g. *EVENT_CSP_INIT* or something similar), to be signaled from `http_security_headers()` before issuing the actual header; simple, but not reusable for other purposes.

2. Go for a more generic approach, i.e. an event allowing plugins to add other headers. Here the challenge is to decide where to signal the event... For CSP it must occur *before* we send headers (i.e. at beginning of the if statement in `http_all_headers()`), but then plugins won't be able to override Mantis default headers (assuming we want to let them do it) reducing the event's . In that case maybe we would need 2 events, like *EVENT_HTTP_HEADERS_PREPARE* *EVENT_HTTP_HEADERS_SENT* ?

My preference would be for 1.

dregad:
Thanks for the PR @vboctor, see my review comments above.

dregad:
> port to master with some changes to enable few more sources

What do you mean, *few more sources* ?

vboctor:
@dregad by "few more sources", I mean white list more CDNs for fonts, bootstrap, etc.

plugins/Gravatar/Gravatar.php
cproensa: Sorry i can't provide much feedback here, since i am no expert. But, as i understand, the treatment of CSP directives probably fits in it's own functionality, i.e, separated from other header declarations.

- Is the needed action, for plugins, always to "add"?
i found that sometines, i tried to do something that gets blocked by mantis CSP defaults. In those cases, do i need to add a new directive, or do i actually need to "remove" one of the mantis ones?

dregad:
I was also wondering if we should not define $g_csp as a global config with default settings, to let admins customize it based on their individual needs, rather than hardcode the policy in the API.

> Is the needed action, for plugins, always to "add"? i found that sometines, i tried to do something that gets blocked by mantis CSP defaults. In those cases, do i need to add a new directive, or do i actually need to "remove" one of the mantis ones?

That's a good question.

From a Mantis admin standpoint, It's a bit of a dangerous game to let 3rd party code alter your security policy, but on the other hand it's necessary for us to let plugins add some sites to the CSP; however, that already lets a plugin do something like ```http_csp_add( 'script-src', "'unsafe-inline'" );```...

That makes me think that I didn't test what happens when adding contradicting keywords to a directive, e.g. ```http_csp_add( 'script-src', "'none'" );``` when other elements exist

vboctor:
I suggest we split the ideas here into:

1. What is needed to fix the current bug enabling plugins with adding their CSP white list entries without breaking security.
2. What can be done in this space to make it even better.

I think the PR was focused on 1.  Please open issues for comments that relate to 2, and let me know the minimum bar for 1.  We may also want to clarify what we expect in 2.0 only vs 1.3 + 2.0.

vboctor:
@dregad ping.

core/http_api.php
vboctor: Will add a check.

core/http_api.php
vboctor: The call to http_csp_value() will error out (after the change) if the header was already emitted.

core/http_api.php
vboctor: The http_security_headers() is using a lower level API that manages $g_csp.  This method doesn't handle that.  It uses the http_csp_*() apis to manage and emit the header.  This method also does another header, so it is not just CSP.

vboctor:
I've added a dedicated events for the headers.  All feedback is now addressed.

vboctor:
Will merge this PR before 1.3.1 and 2.0.0-beta.2 tomorrow morning (in about 12 hours or so).  That is to enable CSP in the new releases. @dregad @atrol @cproensa @syncguru

atrol:
Looks good at first sight.
I have hardly any time to have a deeper look or tests the next two days.

dregad:
Please don't merge until I've had time to review and test. Will do so tonight after kids are in bed.

atrol:
@dregad
Would be great if you could try if using $g_custom_headers works to enable Google Analytics
https://www.mantisbt.org/forums/viewtopic.php?f=3&t=24032

dregad:
Thanks @vboctor for taking comments into consideration. I tested the updated PR, and as far as I can tell it's working fine and does address the original issue.

On my dev box, CSP is generated as follows (output from Firebug):
```
img-src http://www.gravatar.com/ 'self'; default-src 'self'; frame-ancestors 'none'; style-src 'self'; script-src 'self'
```

@atrol I tested with ` $g_custom_headers = array( 'Content-Security-Policy: ' );` in my config_inc.php, Firebug tells me there is no content-security-policy header, so I am effectively disabling CSP; I assume this is what you wanted to know ?

With regards to my suggestion for improvement (https://github.com/mantisbt/mantisbt/pull/859#issuecomment-240694660), I will open an issue in the tracker to follow up.

dregad:
Merged in b3511d2feb47eaee41feb5f69cf3c8a2c9acd229

atrol:
> I tested with $g_custom_headers = array( 'Content-Security-Policy: ' ); in my config_inc.php, Firebug tells me there is no content-security-policy header, so I am effectively disabling CSP; I assume this is what you wanted to know ?

No, the question was if it can be used to add headers (e.g. to allow Google Analytics) without disabling the whole CSP protection.

dregad:
> No, the question was if it can be used to add headers (e.g. to allow Google Analytics) without disabling the whole CSP protection.

Any header defined using $g_custom_headers will actually *replace* whatever was previously set.

So to answer your question, yes it is possible to *add* something to CSP, but in this case the admin must set the *whole* CSP, i.e. manually maintaining the one defined by MantisBT (if this is what they want).

This is not ideal, but without a method to let admins customize CSP with a config option, this is the only way.

PR #860, opened 2016-08-18
Fire EVENT_LAYOUT_PAGE_HEADER right after rendering navbar

syncguru:
Fixes #21113

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -347,27 +347,6 @@ function html_print_logo( $p_logo = null ) {
 	}
 }
 
-
-
-/**
- * Print a user-defined banner at the top of the page if there is one.
- * @return void
- */
-function html_top_banner() {
-	$t_page = config_get( 'top_include_page' );
-	$t_logo_image = config_get( 'logo_image' );
-
-	if( !is_blank( $t_page ) && file_exists( $t_page ) && !is_dir( $t_page ) ) {
-		include( $t_page );
-	} else if( !is_blank( $t_logo_image ) ) {
-		echo '<div id="banner">';
-		html_print_logo( $t_logo_image );
-		echo '</div>';
-	}
-
-	event_signal( 'EVENT_LAYOUT_PAGE_HEADER' );
-}
-
 /**
  * A function that outputs that an operation was successful and provides a redirect link.
  * @param string $p_redirect_url The url to redirect to.
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -425,6 +425,8 @@ function layout_navbar() {
 
 	echo '</div>';
 	echo '</div>';
+
+	event_signal( 'EVENT_LAYOUT_PAGE_HEADER' );
 }
 
 /**


dregad:
:+1:

atrol:
@dregad is [this](https://www.mantisbt.org/bugs/view.php?id=21113#c53381) no longer a problem for you? If so, we have to obsolete option top_include_page

dregad:
Oops good point I forgot about that one. 

@syncguru how would you recommend we integrate the _top_include_page_ feature in the new UI ? Or do did you not fix it because you believe it should be deprecated ?

Should I open a separate issue to track this ?

vboctor:
If we have a good understanding of what scenarios top include is useful for, then we can have a better hook for integrating such functionality.

- Logo - I don't think we should support this beyond the login page in case of the modern UI.
- Banner - Should we enable config / plugins to show a banner? In a way that is integrated in the UI, rather than just injecting random HTML?

So I'm in favor of deprecating the include and considering better handling for specific scenarios.  Note that also there can be a template plugin that enables developer to inject whatever html they want.

atrol:
> If we have a good understanding of what scenarios top include is useful for

We use it at the moment at https://www.mantisbt.org/bugs to display the message

`Currently signup for new accounts is disabled due to spam attacks! Use our support channels to request new accounts. We apologize for the inconvenience. `

cproensa:
> Logo - I don't think we should support this beyond the login page in case of the modern UI.

In corporate usage of mantis this may be a requeriment, (it is for my main instance, if you need real cases)

> Banner - Should we enable config / plugins to show a banner? In a way that is integrated in the UI, rather than just injecting random HTML?

I suggest to support current scheme of inserting arbitrary HTML, either from the included file, or the event.
If there are concerns about conflicting with UI layout, then we could enclose that html in a div that would be placed in the current locations of logo/site name.
This div can be modeled to accomodate wihout conflict with the elements at the right side of header bar

sync1:

l266509:

JWPlatt:
We need $g_top_include_page and $g_bottom_include_page to create a consistent site-wide banner and menu navigation and access status across all pages/subdomains.  I was about to upgrade from 1.3 to 2.1, but now I can't.

Here is our specific example:
http://bugs.openuru.org

cproensa:
Adding a simple div,
- inserted div gets under the nav-bar
- page body has a margin, after the inserted div.

seleccion_146

seleccion_147

cproensa:
- By removing class: navbar-fixed-top, we get a non-floating nav-bar
- And placing the event before rendering the nav-bar
The result is then more satisfactory
(Still, the sidebar is floating and leaving a top-margin when scrolling the page)

Do we really need a floating navigation bar?
the visual is nice but it has side effects for layout, another example is #21675
At least, the feature could be disabled in some way?

cproensa:
@syncguru 
any update for this?
or alternatives?

cproensa:
@syncguru 
I have been thinking more about this.

From what i can see, which may not be the most proper perspective, since i'm not that expert with these bootstrap css:
- Including a header, like old style banner that spanned all the screen wide at the top, is very difficult with current navbar and sidebars design. Since they are made all-time visible, it's not feasible without javascript tricks.

- Possiblity (1) is to include the custom header area, and add JS to make navbar and sidebar to stick below that, and float as usual when page is scrolled. As this example:
http://mhkeller.github.io/floating-scroll/anchor-behavior.html
(dont get confused that "floatingscroll" with the other PR, these are both coincident in naming)
Honestly i think this option is too much complexity for the actual needs here.

- Possibility (2) is to include the custom header area within the "main-content" div. Similarly to what EVENT_LAYOUT_CONTENT_BEGIN does now (it's placed between breadcrumbs bar, and the actual contents). the suggestion here would be lacing EVENT_LAYOUT_PAGE_HEADER (and _top_include_page_, why not), inside "main-content" div, and just before breadcrumbs div.
This has the advantage, that "main-content" div is already layed out with left-margin that fits the side bar. So if the user includes html there, it will easily fit with the current layout.

- Possibility (3) is to discontinue both EVENT_LAYOUT_PAGE_HEADER and top_include_page, since they seem to not fit with the current intentions for the UI look and feel. In this case, i suggest to provide a new method to include user code _within the navbar_. this can be a new event EVENT_LAYOUT_PAGE_NAVBAR. The rendering of this content can be controlled in some way, or simply set some guidelines in documentation on how to correctly mark it up, with some preset classes, or bootstrap layout classes...

At the moment my preferred option is (3), i think...
Option (2) can be an addition to (3). I would actually like to have a insertion point _above_ the breadcrumbs, not only _below_.
Option (1), i wouldnt go for that, because it has the risk of creating a header area too big for small screens.

FYI:
In my case, i'd just want to add a logo and a bit of text, in compliance with company branding. The ideal place would be the navbar.

cproensa:
Additionally, seeing how said top/bottom includes, and header/footer events, are a recurring use case for many mantis users to wrap a mantis site as part of another site integration, I put again the possibility:

Adding an option to enable a "static/legacy" layout (off by default), where the navbar and sidebar are not fixed on top.
When this option is enabled, the old style includes and events for header and footer are used, inserting the raw contents, as previously happened.
This allows to not break the UI, at the cost of not having the fixed elements feature. This is a trade-off that those who still need the includes, or are not willing to work around the complexities that the new UI implies, would probably assume.

vboctor:

estomagado:
As @cproensa commented in 2016 (3 years ago!), Corporate Branding policies in many companies (mine included) enforce some level of local branding on all internal systems, so we need some sort of customisation features for that. g_window_title (which still works) is not enough.

Is there any way to reopen this?

keessonnema:
> As @cproensa commented in 2016 (3 years ago!), Corporate Branding policies in many companies (mine included) enforce some level of local branding on all internal systems, so we need some sort of customisation features for that. `g_window_title` (which still works) is not enough.
> 
> Is there any way to reopen this?

I'm looking at this right now. Sadly I can't re-open the PR

PR #861, opened 2016-08-18
Show report bug menu entry if allowed for any project

cproensa:
Create api function to check a threshold in several projects

```
Add an api function to check the current user's access against the given
threshold, in each of the provided projects, and return true if the
user's access is equal to or higher in any of the projects, false otherwise.

The provided threshold can be the config literal to be evaluated in each
project's context.
```

Show report bug menu entry if allowed for any project

```
Previously, the "report bug" menu entry was not displayed when:
1. current project is ALL PROJECTS and user has not global report bug access
2. current project is a specific one, and user has not report access in
  that project

(1) is confusing for users, when they have access to report in any of
the specific projects they are assigned to, but still cant find the menu
entry.
Note that for globally allowed users, when trying to report in ALL
PROJECTS, a page appears that promts the user to select a valid project

(2) can be also confusing like (1), when the user is positioned in a
project where is effectively a "viewer", want to report a bug and cant
find the menu entry. This forces the user to select a valid project
before being allowed to use the report menu.

The fix implemented here checks if the user has permissions to report in
any of his assigned projects, then show the menu entry.
If the user cant report in current project once the menu entry is
selected, show the project selection page to choose a valid project.

This fixes (1) which is a common complaint from new users
and (2), which also provides a more intuitive way to use the report
option, as first choose the action, later choose the project.

Fixes: #14268
```

### DIFF
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -151,6 +151,12 @@ if( $f_master_bug_id > 0 ) {
 		print_header_redirect( 'login_select_proj_page.php?ref=bug_report_page.php' );
 	}
 
+	# Check for bug report threshold
+	if( !access_has_project_level( config_get( 'report_bug_threshold' ) ) ) {
+		# If can't report on current project, show project selector if there is any other allowed project
+		access_ensure_any_project_level( 'report_bug_threshold' );
+		print_header_redirect( 'login_select_proj_page.php?ref=bug_report_page.php' );
+	}
 	access_ensure_project_level( config_get( 'report_bug_threshold' ) );
 
 	$f_build				= gpc_get_string( 'build', '' );
--- a/core/access_api.php
+++ b/core/access_api.php
@@ -309,9 +309,9 @@ function access_get_project_level( $p_project_id = null, $p_user_id = null ) {
 /**
  * Check the current user's access against the given value and return true
  * if the user's access is equal to or higher, false otherwise.
- * @param integer      $p_access_level Integer representing access level.
- * @param integer      $p_project_id   Integer representing project id to check access against.
- * @param integer|null $p_user_id      Integer representing user id, defaults to null to use current user.
+ * @param integer|array $p_access_level Threshold representing an access level.
+ * @param integer       $p_project_id   Integer representing project id to check access against.
+ * @param integer|null  $p_user_id      Integer representing user id, defaults to null to use current user.
  * @return boolean whether user has access level specified
  * @access public
  */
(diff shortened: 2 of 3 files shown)


cproensa:
Coding this i found that access_api - access_has_any_project() is at risk of not functioning correctly, as probably, to check an access level on several projects, the config option has to be evaluated for each individual project.

Thats the reason that i wrote access_has_any_project_level() as a replacement.
We may evaluate said function usage, and then deprecate or replace that in favour of the proposed logic?

core/access_api.php
```diff
@@ -333,6 +333,69 @@ function access_has_project_level( $p_access_level, $p_project_id = null, $p_use
...
+/**
+ * Check the current user's access against the given value, in each of the provided projects,
+ * and return true if the user's access is equal to or higher in any of the projects, false otherwise.
+ * @param integer|array|string  $p_access_level Parameter representing access level threshold, may be:
+ * - integer: for a simple threshold
+ * - array: for an array threshold
+ * - string: for a threshold option which will be evaluated for each project context
+ * @param array                 $p_project_ids  Array of project ids to check access against, default to null to use all user accesible projects
+ * @param integer|null          $p_user_id      Integer representing user id, defaults to null to use current user.
+ * @return boolean whether user has access level specified
```
dregad: Wrap long lines.
Bullets for $p_access_level should be indented below the param's description

core/access_api.php
```diff
@@ -333,6 +333,69 @@ function access_has_project_level( $p_access_level, $p_project_id = null, $p_use
...
+	if( is_string( $p_access_level ) ) {
+		$t_default = config_get( $p_access_level, null, $p_user_id, ALL_PROJECTS );
+		if( null === $t_default ) {
+			$t_default = config_get_global( $p_access_level );
+		}
+	}
+
+	$t_check_level = $p_access_level;
+	$t_has_access = false;
+	foreach( $p_project_ids as $t_id ) {
```
dregad: Suggest to rename the for variable `$t_project_id`

dregad:
Reviewing your PR made me realize that a few years ago I started working on a fix for this issue, but got sidetracked and never finalized it :confused:  

I followed an approach similar to the one you propose, although mine was less generic as the API function I created only covered the report_bug_threshold case, so I think your fix is better.

bug_report_page.php
```diff
@@ -151,6 +151,12 @@
 		print_header_redirect( 'login_select_proj_page.php?ref=bug_report_page.php' );
 	}
 
+	# Check for bug report threshold
+	if( !access_has_project_level( config_get( 'report_bug_threshold' ) ) ) {
+		# If can't report on current project, show project selector if there is any other allowed project
+		access_ensure_any_project_level( 'report_bug_threshold' );
+		print_header_redirect( 'login_select_proj_page.php?ref=bug_report_page.php' );
```
dregad: Nice touch. That's the bit that was missing in my half-finished attempt at fixing this.

dregad:
:+1:

dregad:
One more thing, looking at my branch I see that I had also applied the logic on the *bug copy* group action - remove the entry from the selection list if the user has no access. You may want to consider this additional improvement  too.

cproensa:
updated with @dregad suggestions

also fixed that $t_check_level wasn't being used correctly, now it is ok.
which leads to: 
_access_compare_level( 25, 'report_bug_threshold' )_ returns true...
this is an invalid parameter (threshold as string), at least it should return false in that case
or, should issue a warning/error?

> One more thing, looking at my branch I see that I had also applied the logic on the bug copy group action - remove the entry from the selection list if the user has no access. You may want to consider this additional improvement too.

That is one of the few places where where "access_has_any_project()" is used.
I think this method is flawed, as it does not evaluate the actual threshold for each project.
The best solution may be:
A) keep that function but rewritting with the new logic i used here (but i don't like that user id would still be 2nd parameter)
B) make it an alias of the new "access_has_any_project_level() ", and deprecate warning the original one. 

then fixing all (4) usages to the new logic, probably would solve some bugs:

bug_group_action_api.php
  199:  access_has_any_project( config_get( 'report_bug_threshold', null, null, $t_project_id ) ) ) {
columns_api.php
1014:  if( access_has_any_project( config_get( 'report_bug_threshold', null, null, $p_bug->project_id ) ) ||
filter_api.php
2320:  if( !access_has_any_project( config_get( 'view_handler_threshold' ) ) ) {
html_api.php
  933:  if( access_has_global_level( $t_show_access ) || access_has_any_project( $t_show_access ) ) {

cproensa:
@dregad 
Cleaning up current usages of access_has_any_project() is more involved that it seems at first. There are problems with threshold operations that sould also be fixed

i have opened a separate issue for that task:
https://www.mantisbt.org/bugs/view.php?id=21654

dregad:
Merged in 5d2f72a95c0397a2abc936736f0437bc0e553166

PR #783, opened 2016-06-01
Manage and edit stored filters

cproensa:
Stored filter edit feature
A lot of clean up in filter_api to allow for code reuse.
Some bug fixes that came in the way

The new pages may need some tuning, but the underlying changes make it easy to work from here on

PR #862, opened 2016-08-18
Manage and edit stored filters

cproensa:
Resubmit of original PR #783, updated and rebased to master-1.3.x.
(see there for history of comments)

Stored filter edit feature
A lot of clean up in filter_api to allow for code reuse.
Some bug fixes that came in the way

The new pages may need some tuning, but the underlying changes make it easy to work from here on

Tracking changes and related issues in:
https://www.mantisbt.org/bugs/view.php?id=21935


dregad:
In manage_filter_edit_page.php, the labels at the bottom of the frame need some tweaking, see screenshot:

selection_001

view_all_set.php
```diff
@@ -374,11 +94,6 @@
 	$f_per_page = config_get( 'default_limit_view' );
```
dregad: While testing, I got a `SYSTEM NOTICE: 'Undefined variable: f_per_page' in 'view_all_set.php'` at line 93

cproensa: marked to remove in next clean-up commit.
that code s not needed as it is moved to filter_gpc_get()

cproensa:
> In manage_filter_edit_page.php, the labels at the bottom of the frame need some tweaking, see screenshot:

Also accepting suggestion for what options to show, and how to phrase it,
The alignement may be an easy css style.
But the presentation of both options may be discussed:
- change the "public" status of a filter
- Project visibility for the filter: whether if its available for "all projects" , or a specific one.
In case of a specific project, you can choose to change project to another one (current global project)

dregad:
@cproensa What is the status of this PR ? it is not clear from your previous message, if you actually need something from me, and if so, what. 

For now, I would suggest to just apply necessary CSS tweaks to address the display issues so we can get this in, and the rest can be dealt with in follow-up issues/PRs.

cproensa:
> For now, I would suggest to just apply necessary CSS tweaks to address the display issues so we can get this in, and the rest can be dealt with in follow-up issues/PRs.

Thats fair... i'll do then.
however, my concern is that no other dev has positioned for this changes, and having new pages, is a significant one. Thus, i get to a point that i stop spending time without having certainty that it's useful.
Note that this will require tweaking for v2 too, since the filter_api has embedded html, and will not merge cleanly.

dregad:
> my concern is that no other dev has positioned for this changes, and having new pages, is a significant one. 

Considering how long this has been open, I think everyone had a chance to voice their opinions, so I wouldn't let this concern block merging this useful new feature.

manage_filter_page.php
```diff
@@ -0,0 +1,157 @@
...
+require_api( 'filter_api.php' );
+require_api( 'helper_api.php' );
+require_api( 'html_api.php' );
+require_api( 'lang_api.php' );
+require_api( 'print_api.php' );
+require_api( 'rss_api.php' );
+
+auth_ensure_user_authenticated();
+
+html_page_top( 'MANAGE FILTER PAGE' );
```
vboctor: The 'manage filter page' is missing the footer.

core/filter_api.php
```diff
@@ -3186,6 +3205,10 @@ function filter_draw_selection_inputs( $p_filter, $p_for_screen = true, $p_stati
  *      These functions could also be shared by view_filters_page.php
  */
 
+/**
+ * Print the current value of this filter field, as visible string, and as a hidden form input.
+ * @param type $p_filter	Filter array
```
vboctor: Should you replace "type" with the parameter type?  Applies to multiple methods.

core/filter_api.php
```diff
@@ -3186,6 +3205,10 @@ function filter_draw_selection_inputs( $p_filter, $p_for_screen = true, $p_stati
...
+ */
 function print_filter_values_reporter_id( $p_filter ) {
```
vboctor: For methods that return no value, add "@return void".  Applies to multiple methods.

core/filter_api.php
```diff
@@ -4513,52 +4547,146 @@ function print_filter_plugin_field( $p_field_name, $p_filter_object ) {
 }
 
 /**
- * print custom fields
- * @param integer $p_field_id Custom field identifier.
- * @return void
+ * Print the current value of custom field, as visible string, and as a hidden form input.
+ * @param type $p_filter	Filter array to use
```
vboctor: Types are not set in phpdoc here too.

cproensa:
@vboctor 

> phpdoc issues
> The 'manage filter page' is missing the footer.
> delete filter missing redirect

fixed

> Testing - This will require a fair bit of testing.

i know....

> Modern Version we will need a version ported to master right after getting this into master.

I think i am finally reimplementing drawing the filter box, to reduce the ad-hoc html in 
source. Filter fields would be sequentially drawn, avoiding empty gaps, and make the adding of new filter fileds an easier task in the future.
And mainly, will make it easier to merge in modern-ui

> 'Undefined variable: f_per_page' in '/home/ubuntu/workspace/view_all_set.php' line 93

fixed

> view_filters_page.php. [ Simple Filters ] vs. [ Advanced Filters ] to switch filter type seems confusing

At this moment i have tried not to change that much the ui for filter box and surrounding actions. 
What is your suggestion for this point, and improve it?

> manage_filters_page.php

do you think this is better?
seleccion_057

I tested three buttons (for apply, edit, and delete) under the actions column and looked too cluttered,
Linking the filter name to go to detailed page, it's the usual logic we already have in other manage pages.
Note also that, depending on your user and filter owner, you cant always edit a filter. If not having edit permissions, you will go "view_filters_page.php", where you can also edit the filter values, but the only action available is "use as current"

> updating filter definiton
we remember the active filter. going into manage filters and editing the filter definition doesn't reflect on the issues. So we show the name of the filter, but it is no longer consistent with the applied filter. So after updating a filter, if it is current, then we should make sure we refresh the result set.

I think i have reproduced what you mean, will try a fix. 
The idea then is that when the current filter is a "named filter", it should always be retrieved from the stored filter array, instead of the current active filter array (which is stored separately and specifically for the user)
The second implication is that any modification the user may do to the active filter, will mean its no longer linked to that "named filter"

> Save current filter - ideally when saving a filter, we provide the option of saving to an existing filter, and not just new filter. At the moment after editing we reset the current filter to none, and when user clicks save they have to select new name. This expects users to know about manage filters and discover that the link edits, do the changes there instead, etc. I think save current to new or existing filter will be much more discoverable

I dont quite understand what you mean there.
Note that saving a filter is still using the old query_store_page.php. This is still the only way to create a new stored filter. In later iteration, that page should be improved as you say to also allow selection of an existing filter.
I havent done that because it was not essential to get the edit feature working (i'd rather settle all these refactors first)
ALso note that the new filter edit pages don't mess with the current active filter. You can view and edit other filters, without altering the one in use for view_all page (that was one of the motivations for the filter api rewrite, decoupling api from that limitation)

therefore, i'd say this point is not blocking for this PR.

> If we can select a filter, update it in View Issues page, and then save it, then why would we need a filter view/edit pages? We seem to have two places to do exactly the same thing. So I suggest one of the following:
1. Have the viewing and editing happen on the View Issues page.
2. Have View Issues page just show the filter critera (the active fields) - e.g. "Assigned To: [myself], Hide Stats: Resolved" and then when click Edit, go the edit filte page to update the filter. This reduces clutter and reduces the real estate needed on View Issues page, with a very easy way to know the set fields / active filter in a glance.

That can be considered. But note that the effort in this PR was to allow the reuse of filter api and encapsulating the filter presentation, and be able to use it outside the context of view_all page.
I still think that having a dedicated page for filter edit is a must, since there are options that don't quite have a place in an embedded filter box in view_all page (change private/public status, delete, etc).

imo, not blocking for this PR

cproensa:
> query_*.php - should these be deleted?

query_store*.php are still in use.
the rest, have been marked as deprecated. However, they still work, because the filter structure and storage has not been modified

cproensa:
Note, again, that the main work of this PR is to refactor filter api, to allow reusability and decoupling from the previous limitations of only accounting for view_all page and active filter.

The new pages provide the minimal functionality for editing, and to make sure that all works as it should

Further improvements in the workflow can be made, like victor suggestions,
Also, for example, manage_filter_page, only shows filter for current project. Another page may be useful to show all filters owned by the user, etc

cproensa:
>>    updating filter definiton
    we remember the active filter. going into manage filters and editing the filter definition doesn't reflect on the issues. So we show the name of the filter, but it is no longer consistent with the applied filter. So after updating a filter, if it is current, then we should make sure we refresh the result set.

> I think i have reproduced what you mean, will try a fix.
The idea then is that when the current filter is a "named filter", it should always be retrieved from the stored filter array, instead of the current active filter array (which is stored separately and specifically for the user)
The second implication is that any modification the user may do to the active filter, will mean its no longer linked to that "named filter"

fixed

vboctor:
> Note that saving a filter is still using the old query_store_page.php. This is still the only way to create a new stored filter. In later iteration, that page should be improved as you say to also allow selection of an existing filter. ... therefore, i'd say this point is not blocking for this PR.

That's fine, we can do this in a follow up iteration.

> manage_filters_page.php - do you think this is better?

I think we should be consistent with the model we use elsewhere, e.g. in the Configuration Report page, we have 3 buttons under actions, Edit, Clone, Delete.  I also think there is no need to have the captions as "Use Filters" and "Delete Filters" with the word "Filters" being redundant.

> Note also that, depending on your user and filter owner, you cant always edit a filter

We should show the buttons that make sense to the user access level.

> I think i am finally reimplementing drawing the filter box, to reduce the ad-hoc html in 
source. Filter fields would be sequentially drawn, avoiding empty gaps, and make the adding of new filter fileds an easier task in the future.  And mainly, will make it easier to merge in modern-ui

That's great.

cproensa:
@vboctor 
so do you mean this:

seleccion_058

or..

seleccion_059

I think the reason i placed the "apply" button on the second column, is to keep it next to filter name, as this is the most usual operation that all users can do.
Would it be better to remove the "delete" button, and place checkboxes to operate on multple selection for operations: delete, make public, make private, etc?

vboctor:
@cproensa The second screenshot is what i meant.  So it is missing the Edit action.  If user clicks the hyperlink of the filter name, we should go to the view issue page (or edit if we user can edit the filter).  If you want the actions closer to the filter name, then you can move the actions column next to the filter names column.

cproensa:
@vboctor 
> The second screenshot is what i meant. ... . If you want the actions closer to the filter name, then you can move the actions column next to the filter names column.

now is updated. I dont mind the position, for the moment is kept to the right

> So it is missing the Edit action. If user clicks the hyperlink of the filter name, we should go to the view issue page (or edit if we user can edit the filter)

This was already implemented.

cproensa:
I've updated rewriting of filter box generation.
Created a new file "filter_form_api.php" to keep the code better organized.
Some filter fields has been reordered, (see 5e1aff1)
Rebased to current master-1.3.x

This is previous filter box:
seleccion_061

This is the new filter box:
seleccion_062

The difference is that there are no gaps for unused fields, so it's slighlty more compact.
At this point, creating or reordering filter fileds is an easier task.

core/filter_form_api.php
```diff
@@ -0,0 +1,2601 @@
...
+
+/**
+ * Returns HTML for each filter field, to be used in filter form.
+ * $p_filter_target is a field name to match any of "the print_filter_..." functions,
+ * excluding those related to custom fields and plugin fields.
+ * When $p_show_options is enabled, the form inputs are returned to allow selection,
+ * if the option is disabled, returns the current value and a hidden input for that value.
+ * @param array $p_filter Filter array
+ * @param string $p_filter_target Filter field name
+ * @param boolean $p_show_options Whether to return a visible form input or a text value.
```
atrol: Should be $p_show_inputs

core/filter_form_api.php
```diff
@@ -0,0 +1,2601 @@
...
+
+/**
+ * Print the current value of this filter field, as visible string, and as a hidden form input.
+ * @param array $p_filter			Filter array
+ * @param string $p_field_name		Field name
+ * @param object $p_filter_object	Filter object
+ * @return void
+ */
+function print_filter_values_plugin_field( array $p_filter, $p_field_name, $p_filter_object ) {
+	$t_filter = $p_filter;
```
atrol: Line can be removed. $t_filter is not used.

core/filter_form_api.php
```diff
@@ -0,0 +1,2601 @@
...
+ * @param array $p_filter Filter array
+ * @return void
+ */
+function print_filter_custom_field( $p_field_id, array $p_filter = null ) {
+	global $g_filter;
+	if( null === $p_filter ) {
+		$p_filter = $g_filter;
+	}
+
+	$t_cfdef = custom_field_get_definition( $p_field_id, $p_filter );
```
atrol: Remove 2nd parameter as custom_field_get_definition has just one parameter

core/filter_form_api.php
```diff
@@ -0,0 +1,2601 @@
...
+
+	# get default values from filter structure
+	$t_sort_fields = explode( ',', $p_filter[FILTER_PROPERTY_SORT_FIELD_NAME] );
+	$t_dir_fields = explode( ',', $p_filter[FILTER_PROPERTY_SORT_DIRECTION] );
+	if( !isset( $t_sort_fields[1] ) ) {
+		$t_sort_fields[1] = '';
+		$t_dir_fields[1] = '';
+	}
+
+	# if there are fi
```
atrol: The previous four lines can be removed as $t_tdclass is not used.

css/default.css
```diff
@@ -66,6 +66,7 @@ td.print-bottom		{ border-bottom: 1px solid #999; }
 td.print-spacer		{ background-color: #ffffff; color: #000000; font-size: 1pt; line-height: 0.1; padding: 0px;}
 td.print > a:link.user { text-decoration: none; color: black; }
 
+
```
atrol: Remove blank line

cproensa:
Thanks @atrol. Now is updated with your fixes

I have also included a rewrite of the behaviour of filter properties: "show status", and "hide status".
The mixing of both, and the addition of simple/advanced filter type, resulted in some erratic results.

Refer to commits, for details:
- Rewrite logic for filter query, show/hide status
- Resolve conflicts between show and hide status

vboctor:
@cproensa some comments on how the new filter box looks:

- Why did the order of the fields change? e.g. before we used to start by Reporter, etc.  I would rather maintain the order.  It is ok to re-order some fields explicitly, like the last row makes sense to have these fields together.
- Why is Tags on its own line?
- Is the line for "dias inactivo" for custom fields?

atrol:
If ordering is changed, what about some more changes concerning ordering of the fields?
Group user related fields, so that we have "Reporter:", "Assigned To:", "Note By:", "Monitored By:"

I would like to prevent double review, testing and merging to 2.0.x.
Do you think this PR should / could be included https://github.com/mantisbt/mantisbt/pull/897 ?

cproensa:
> Why did the order of the fields change?

That's explained in the commit:

```
Change the position of some fileds in the filter box:
- (1) For main fields, keep first the most common fields.
- (2) For main fields, move to the end fields that depends on project configuration to
  be hidden or shown. This way fixed fields wont change their position
  across different projects.
- Filter related fields (per-page, sort, etc), are shown with a
  separator to other field groups. This also helps with the styling since their
  orientation is horizontal.
- "per-page" has been moved to the filter-related fields group
- "note-by" has been moved to the main fields group.
```

See how (2) is important to avoid some fields jumping places across projects, and try not to mess with visual memory from the user.
And (1), Once that some fields must be reordered anyway, i grouped them in some order of supposedly frequent usage. 
_This is open to suggestions, of course, so please share an alternate field order to see all pros and cons_

> Why is Tags on its own line?

Tags is defined with a colspan of 5, so it didn't fit in the second row, and was pushed to third.

> Is the line for "dias inactivo" for custom fields?

Those are plugin fields.
Plugin fields and custom fields start a row for each group respectively. We could discuss if any of those group of fields can be appended to the main fields sequentially, to get an even more compact box, or keep it separated as they are now.

> Do you think this PR should / could be included #897 ?

A new filter field can be defined, however that PR should be rewritten as most of the supporting api being modified has been refactored . I have some issues on the specifics of that PR, but i can communicate them to make his changes included here. It wont add much complexity because they are already touching the refactored code.

> I would like to prevent double review, testing and merging to 2.0.x.

I foresee a bit of merge hell... however, i'm optimistic, in that the conflicts with anything changed for new UI will finally be reduced to:
- update some very localized html-related code, (which this branch have helped to reduce). Moving most of the html related code out from filter_api to filter_form_api, probably helps too.
- update styles to new pages.

cproensa:
Note that the html_api code to allocate fields in the filter box can be also used to arrange bug fields in bug-view-page, to optimize empty spaces.
@atrol , are there any related requests/issues?
Obviously, not part of this PR, but take it as a possibility.

core/columns_api.php
```diff
@@ -275,6 +302,50 @@ function column_is_extended( $p_column ) {
...
+		return $t_plugin_obj->sortable;
+	}
+
+	#standard fields: define exceptions here
+	switch( $p_column ) {
+		case 'selection':
+		case 'edit':
+		case 'bugnotes_count':
+		case 'attachment_count':
+		case 'selection':
```
atrol: Two times case 'selection"

core/filter_form_api.php
```diff
@@ -0,0 +1,2592 @@
...
+			echo lang_get( 'after_date' ) . '<br>';
+			echo $t_start;
+			break;
+		case CUSTOM_FIELD_DATE_ONORAFTER:
+			echo lang_get( 'on_or_after_date' ) . '<br>';
+			echo $t_start;
+			break;
+	}
+	# print hidden inputs
+	$tcf = $p_filter['custom_fields'][$p_field_id];
```
atrol: $tcf is not used

cproensa:
The build is failing, only for PHP 5.3, but the error doesn't show any useful info...

cproensa:
Added some fixes for sorting fields, and column sorting.
Includes some improvement for column sorting via column headers, and preliminary support for more than 2 sort fields

cproensa:
relevant, discussing order of filter fields: https://www.mantisbt.org/bugs/view.php?id=18045

atrol:
I get SYSTEM NOTICE: 'Undefined variable: t_settings_serialized' in '/var/www/html/master-1.3.x/view_all_set.php' line 252 when clicking one of the links from top of the boxes on "My View" page, same for the links on "Summary" page

cproensa:
@atrol, that's fixed now (see 649136d)

atrol:
Unfortunately GitHub is not able to show the diff of filter_api.php (too large), thus I can't comment inline
Around line 526 there is three times config_get( 'view_filters' )
I would use something like  $t_config_view_filters = config_get( 'view_filters' );
```php
	$f_default_view_type = 'simple';
	if( ADVANCED_DEFAULT == config_get( 'view_filters' ) ) {
		$f_default_view_type = 'advanced';
	}
	$t_view_type = $p_filter_arr['_view_type'];
	if( ADVANCED_ONLY == config_get( 'view_filters' ) ) {
		$t_view_type = 'advanced';
	}
	if( SIMPLE_ONLY == config_get( 'view_filters' ) ) {
		$t_view_type = 'simple';
	}
```

atrol:
> @atrol, that's fixed now (see 649136d)

Thanks, confirmed

cproensa:
@atrol regarding config_get calls in ensure_valid_filter. These are the configs used:
```
default_show_changed
view_filters
default_limit_view
show_sticky_issues
enable_profiles
enable_projection
enable_eta
enable_product_build
enable_sponsorship
```
I think the unlogged scenario is not an issue for config_get(), as it can deal with that case.
Other discussion, is if any of those configs make sense to be defined per user/project.

return_dynamic_filters.php
```diff
@@ -135,19 +93,25 @@ function return_dynamic_filters_prepend_headers() {
 $t_function_name = 'print_filter_' . utf8_substr( $f_filter_target, 0, -7 ); # -7 for '_filter'
 if( function_exists( $t_function_name ) ) {
 	return_dynamic_filters_prepend_headers();
-	call_user_func( $t_function_name );
+	call_user_func( $t_function_name, $t_filter );
```
MrBricodage: This function is used to call function print_filter_do_filter_by_date( $p_hide_checkbox = false, array $p_filter = null )

==> call_user_func( $t_function_name, false, $t_filter ); 

Without it, no "use date filter" checkbox during date filter declaration and start and end dates are not init properly.

cproensa: right...

change the function call to this one, which is defined in filter_form_api.php:
`function filter_form_get_input( array $p_filter, $p_filter_target, $p_show_inputs = true )`
that function was made to skip the special cases like thar

like this, i think:
`call_user_func( 'filter_form_get_input', array( $t_filter, filter_field_name  ) );`

are you working on this, so i wait for you to have it fixed?

MrBricodage: yes. That will be included in my PR.

cproensa: great. thanks!

atrol:
> @atrol regarding config_get calls in ensure_valid_filters

I assume you are talking about https://www.mantisbt.org/bugs/view.php?id=21678#c53998

> I think the unlogged scenario is not an issue for config_get(), as it can deal with that case.

Right, it can deal with it and uses ALL_USERS.
But this is still wrong when using this function in upgrade process.
e.g. you want to validate and correct a stored filter of user U1, but the function will use the setting of ALL_USERS for it. 

> Other discussion, is if any of those configs make sense to be defined per user/project.

Certainly not part of this PR
There is another issue in our tracker where similar discussion started. I don't find it at the moment.
First step would be to have two properties for each option, something like
can_be_set_per_project, can_be_set_per_user to cover the four possible cases
ALL_USERS ALL_PROJECTS
ALL_USERS PER_PROJECT
PER_USER ALL_PROJECTS
PER_USER PER_PROJECT

cproensa:
I have included the merge of @MrBricodage commits to add new field for "last updated date"
Rebased the branch to current master-1.3.x to resolve conflicts

cproensa:
@atrol 
> But this is still wrong when using this function in upgrade process.
Yes, i understand you...
I wonder too if calling to ensure_valid_filter is even needed at all

cproensa:
I want to move for getting this branch into deeper testing to surface bugs and fixes and get clorer to merge...
I plan to put his on work on my prod instances for real usage test. But before that, i need to resolve some presentation issues, to have them fixed now and not altering the UI more than necessary.

I need an agreement on how to show the filter fields in the filter box. This includes:
1) Order of fields in the form:
The fields are drawn sequentially (from left to right), only those that are enabled by config, so no gaps appear (except those not fitting due to colspan requirements).
- Some reordering is needed, to group similar fields together
- My suggestion included placing first those that are always-shown, to minimize displacement across projects.
2) Plugin fieds, custom fields. At the moment they are placed on their own rows (like the old code did). This could be changed, to place them without separation following the main fields, then having a more compact filter box. 
3) Width of some fields: Some fileds have a different colspan, due to the nature of their content:
- tags: this was set to 5, but i suggest lowering to 3
- date fields: they are set now at 2.

cproensa:
@atrol @dregad @vboctor 
can you say your final thoughts on field ordering?

vboctor:
I think with the update to change the layout we have traded one problem for another.  We have traded gaps where we had full control of fields being grouped together that make sense, with no gaps, but wrapping can happen anywhere.  For example, Date Submitted and Last Updated should ideally be next to each other, but it is hard to deterministically do that.

As I'm thinking of the order, I'm struggling between 3 aspects:

- Habit - what I'm used to from before
- Nice - what order makes sense (e.g. all filters related to a user grouped together).
- Pragmatic - what fields are likely to be used the most (may seem random, no logic other than frequency, and each company's frequency could be different).

I wouldn't optimize so much for different projects have different set of fields, and I would work around these by having such fields be replaced by blanks at the end of the rows rather than in the middle.  There will be some level of change at least for custom fields and plugin fields anyway.

So assuming the algorithm is updated to enable inserting blanks at end of rows.  Here is one proposal of how it would look like

Ordering - Row 1: (7 entries + 1 blank)

- Reporter By
- Assigned To
- Monitored By
- Note By
- Priority + Severity
- View Status
- Blank

Ordering - Row 2 (8 entries)

- Category + Tags (4)
- Status + Hide status + Resolution
- Blank

Ordering - Row 3 (8 entries or 4 if profile/platform disabled + 4 blank)

- Date Submitted (2) + Last Updated (2)
- Profile + Platform + OS + OS Version OR 4 blanks if disabled.

Ordering - Row 4 (5 entries + 3 blank)

- Product Version + Fixed in Version + Target Version
- Relationships
- Sticky

Rest:

- I like having a group for filter attributes like sorting, highlight hours, etc.
- I would combine custom fields + plugins, but separate them from native fields.

For column spans:

- Tag going from 5 to 3 makes sense.  But if we need more space to fit fields that make sense together, we could even consider 2.

Bug: (haven't test much, but found the bug below) 

- Show sticky filter doesn't work.  Unchecking the checkbox doesn't persist.

cproensa:
> So assuming the algorithm is updated to enable inserting blanks at end of rows

it's supported, by having each row to be a different "bucket" always <= table column width.
Also, in your scheme, the table columns is fixed. How would this option be honored, or ignored: $g_filter_custom_fields_per_row?

Take in mind, this is probably a transitional change, since in a sort term we aim to have a filter interface that only shows the fields that are in use

atrol:
"Use Filter for Last Updated Date" does not work. Date UI elements remain insensitive after enabling the check box.

cproensa:
@vboctor 
> Show sticky filter doesn't work. Unchecking the checkbox doesn't persist.

Fixed, thanks!

cproensa:
@atrol 
> "Use Filter for Last Updated Date" does not work. Date UI elements remain insensitive after enabling the check box

i cant reproduce your issue.
notie that the js was changed, maybe you are using a cached one by the browser?

atrol:
> maybe you are using a cached one by the browser?

Thanks, seems you are right. I am no longer able to reproduce the issue after clearing all caches.

cproensa:
Starting from @vboctor suggestion, made some changes to field ordering.
Tried to allow for having 3 rows when some options are inactive.

- Row 1
 - reporter user
 - handler user
 - monitor user
 - note user
 - priority
 - severity
 - view status
 - show sticky

(8 fields)

- Row 2
 - status
 - hide status *
 - resolution
 - category
 - submit date (2)
 - update date (2)
 - project id *

(8 fields, as * are complementary)

- Row 3
 - profile
 - platform
 - os
 - os version
 - build
 - version
 - target version
 - fixed in version
 - relationship
 - tags (3)

(if profiles or versions are not active, all filts in one row, otherwise it can overflows to a 4th row)

- Plugin fields + custom fields

- Filter properties

vboctor:
@cproensa please provide updated screenshots.

cproensa:
Filterbox with profiles disables, 3 main rows + combined plugin and custom fields

seleccion_086

profiles enabled, has 4 main rows:

seleccion_087

atrol:
Isn't "show sticky" some kind of filter property and should be in the last row?

cproensa:
> Isn't "show sticky" some kind of filter property and should be in the last row?

Probably yes.
That's bad luck that then we have 5 fields for the "meta" part of filter.

Not to include in this PR, but "show" and "highlight changed", should not be part of the filters, not even stored within them.
I see them more as a user preference than a filter parameter.

vboctor:
Looks much better.

For second row, how about starting with: Category, Hide Status, Status, Resolution?  This way we keep the logical grouping and have the most used ones at the beginning.  Also keep status and resolution together.

> Not to include in this PR, but "show" and "highlight changed", should not be part of the filters, not even stored within them.  I see them more as a user preference than a filter parameter.

I agree with this.  Maybe we should leave show sticky where it is for now, and then have a separate issue to move these fields to user preferences.

dregad:
Thanks Carlos, it looks good to me (with Victor's suggestions).

> Isn't "show sticky" some kind of filter property and should be in the last row?

I think so too.

> > Not to include in this PR, but "show" and "highlight changed", should not be part of the filters, not even stored within them.  I see them more as a user preference than a filter parameter.
> 
> I agree with this.  Maybe we should leave show sticky where it is for now, and then have a separate issue to move these fields to user preferences.

+1

cproensa:
> with Victor's suggestions

I placed status first to keep current position in the 2nd row.
If the visual memory is not an issue, this can be changed to:
 Category, Hide Status, Status, Resolution

>  Maybe we should leave show sticky where it is for now, and then have a separate issue to move these fields to user preferences.

Actually not totally "move" them to preferences. They are closely tied to the filter functionality to be shown in the filter section. My core suggestion is to not store/treat them as part of the filter clauses.
They could have a place in preferences, as defaults, but also they are handy to be changed ad-hoc.
These nuances should be discussed with care in the separate issue to track this

atrol:
Some comments to filter_api.php

Parameter $p_view_type is not used in
`function filter_get_default_property( $p_filter_property, $p_view_type = null )`



PHPdoc of function filter_serialize
` * @param type $p_filter_array Filter array to be serialized`
should be
` * @param array $p_filter_array Filter array to be serialized`




function filter_get_bug_rows
What happens at this place (currently line 1209) if $t_rows is not set because db_fetch_array did not deliver any row?
	`return filter_cache_result( $t_rows, $t_bug_id_array );`



Function filter_get_bug_rows_query_cluase
$t_report_bug_threshold and $t_where_param_count are set but not used
```
	$t_report_bug_threshold = config_get( 'report_bug_threshold' );
	$t_where_param_count = 0;
```


Function filter_cache_result
$b is set but not used
`	$b = $t_stats[$t_row['id']];`



```
PHPdoc
 * @param type $p_filter_id	Filter id
 * @param type $p_user_id	User id
 * @return boolean	true if the filter is accesible by the user
 */
function filter_is_accessible( $p_filter_id, $p_user_id = null ) {
```
$p_filter_id and $p_user_id are type integer

core/filter_form_api.php
```diff
@@ -0,0 +1,2758 @@
...
+		<input type="text" name="<?php echo FILTER_PROPERTY_TAG_STRING;?>" id="<?php echo FILTER_PROPERTY_TAG_STRING;?>" size="40" value="<?php echo string_attribute( $t_tag_string )?>" />
+		<select <?php echo helper_get_tab_index()?> name="<?php echo FILTER_PROPERTY_TAG_SELECT;?>" id="<?php echo FILTER_PROPERTY_TAG_SELECT;?>">
+			<?php print_tag_option_list();?>
+		</select>
+		<?php
+}
+
+/**
+ * Print the current value of this filter field, as visible string, and as a hidden form input.
+ * @param type $p_filter	Filter array
```
atrol: array $p_filter

cproensa:
@atrol thanks:

> Parameter $p_view_type is not used in
> function filter_get_default_property( $p_filter_property, $p_view_type = null )

i fixed its usage, as it's necessary.

> PHPdoc of function filter_serialize
> * @param type $p_filter_array Filter array to be serialized
> should be
> * @param array $p_filter_array Filter array to be serialized
 
ok!

> function filter_get_bug_rows
> What happens at this place (currently line 1209) if $t_rows is not set because db_fetch_array did not deliver any row?
> return filter_cache_result( $t_rows, $t_bug_id_array );

this wasnt modified here, but i guess it would throw a warning for using a foreach with null, within filter_cache_result()
i added the initialization.

> Function filter_get_bug_rows_query_cluase
> $t_report_bug_threshold and $t_where_param_count are set but not used
> 
>     $t_report_bug_threshold = config_get( 'report_bug_threshold' );
>     $t_where_param_count = 0;

this function was not changed. In any case, this function will benefit from a clean up in a later stage, i'm leaving those as is until then.

> Function filter_cache_result
> $b is set but not used
> $b = $t_stats[$t_row['id']];

right, this slipped from a previous PR. I have actually fixed for an on-comoing PR, so i'll leave as is to be fixed as part of the rebase/merge

> PHPdoc
>  * @param type $p_filter_id Filter id
>  * @param type $p_user_id   User id
>  * @return boolean  true if the filter is accesible by the user
>  */
> function filter_is_accessible( $p_filter_id, $p_user_id = null ) {
> 
> $p_filter_id and $p_user_id are type integer

ok!

atrol:
Changes in lang/strings_french.txt shouldn't be part of this PR as translations are managed via Translatewiki.

cproensa:
> Changes in lang/strings_french.txt shouldn't be part of this PR as translations are managed via Translatewiki.

You're right. Lang files has been reverted

vboctor:
Closing this since this work has been merged into master and will be part of 2.1.0.

PR #863, opened 2016-08-20
Fix querying zero projects

cproensa:
Fix the new api functions, to support the case when a user has not any
accesible project. In this case the filter query can't be
built, so fix returning empty values.

The affected functions were introduced in relation to Issue #20424
and Isssue #21072

### DIFF
--- a/core/category_api.php
+++ b/core/category_api.php
@@ -265,7 +265,6 @@ function category_remove_all( $p_project_id, $p_new_category_id = 0 ) {
 	$t_category_ids = join( ',', $t_category_ids );
 
 	# update bug history entries
-	db_param_push();
 	$t_query = 'SELECT id, category_id FROM {bug} WHERE category_id IN ( ' . $t_category_ids . ' )';
 	$t_result = db_query( $t_query );
 
--- a/core/custom_field_api.php
+++ b/core/custom_field_api.php
@@ -731,7 +731,6 @@ function custom_field_get_ids() {
 	global $g_cache_cf_list, $g_cache_custom_field;
 
 	if( $g_cache_cf_list === null ) {
-		db_param_push();
 		$t_query = 'SELECT * FROM {custom_field} ORDER BY name ASC';
 		$t_result = db_query( $t_query );
 		$t_ids = array();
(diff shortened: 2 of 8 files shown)


vboctor:
The reason for the two calls for "db_param_pop()" is unclear to me.  Please comment to clarify why such calls are needed.

cproensa:
> The reason for the two calls for "db_param_pop()" is unclear to me. Please comment to clarify why such calls are needed.

I have gone further, since i think there is something wrong by using the same (parametrized) query string for subsequent executions (remember, each plain db_query, pops the param stack)

I'd like @dregad to have a look here.

Meanwhile, i am trying to test whether the actual error is reproducible in some before/after test with pgsql... note that this situation probably existed before all these filter_api changes.

cproensa:
Note 2: offtopic, but related rant:
Current db_param() functionality is flawed and can be seen with this "filter query parts" building process.
with mysql, the parameters used for db_query must be ordered exaclty as their appearance in the SQL string, which may not be the order in which they where created (building the sql string as concatenation of unordered parts).
This is why you cant use db_params in a JOIN clause for a MantisFilter object: it is created after standard WHERE params, but then its positioned before them in the actual query string.

This also makes that the same query may work for pgsql, and fail with mysql, becasue pgsql parameters are numbered, so they are immune to the "appearance order" issue

i had a sketch of a db_param object, that is self-contained, and db_query refactor that makes the substiution for actual ADO params at exec-time. which would solve these issues.

cproensa:
I had problems reproducing the stack error, because of the extra db_param_push in custom_field_get_ids (2a5ce47)
which balanced the extra param-pop made by both db_querys in filter_api: get-bug-count and get-bug-rows

i reproduced it in release-1.3.0, which is older that these filter_api changes, so i think with these fixes, it's ok now.

vboctor:
@dregad can you please review and sign-off on this since @cproensa believes it is necessary for 1.3.1.

dregad:
:+1:

dregad:
Merged in 22c421e9b2ce623b84b901b0fea6c0753e1865c3

PR #865, opened 2016-08-21
Show text search bar when the filter is collapsed

syncguru:
Fixes #20102

Filter box collapsed:
screen shot 2016-08-21 at 3 34 37 pm

Filter box expanded:
screen shot 2016-08-21 at 3 34 50 pm

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -2315,36 +2315,73 @@ function filter_draw_selection_area2( $p_page_number, $p_for_screen = true, $p_e
 				<?php echo lang_get( 'filters' ) ?>
 			</h4>
 			<div class="widget-toolbar">
+				<?php
+					$f_switch_view_link = (config_get('use_dynamic_filters')) ? 'view_all_set.php?type=6&amp;view_type=' : 'view_filters_page.php?view_type=';
+					$t_view_filters = config_get('view_filters');
+					if ( ( ( SIMPLE_ONLY != $t_view_filters ) && ( ADVANCED_ONLY != $t_view_filters ) ) ||
+						access_has_project_level( config_get( 'create_permalink_threshold' ) ) ) { ?>
+					<div class="widget-menu">
+						<a href="#" data-action="settings" data-toggle="dropdown">
+							<i class="ace-icon fa fa-bars bigger-125"></i>
+						</a>
+						<ul class="dropdown-menu dropdown-menu-right dropdown-light-blue dropdown-caret dropdown-closer">
+							<?php
+							if ( ( SIMPLE_ONLY != $t_view_filters ) && ( ADVANCED_ONLY != $t_view_filters ) ) {
+								echo '<li>';
+								if ('advanced' == $t_view_type) {
+									echo '<a href="' . $f_switch_view_link, 'simple"><i class="ace-icon fa fa-filter"></i> ' . lang_get('simple_filters') . '</a>';
+								} else {
+									echo '<a href="' . $f_switch_view_link, 'advanced"><i class="ace-icon fa fa-filter"></i> ' . lang_get('advanced_filters') . '</a>';
+								}
+								echo '</li>';
+							}
+							if ( access_has_project_level( config_get( 'create_permalink_threshold' ) ) ) {
+								echo '<li>';
+								echo '<a href="permalink_page.php?url=' . urlencode( filter_get_url( $t_filter ) ) . '">';
+								echo '<i class="ace-icon fa fa-link"></i> ' . lang_get( 'create_filter_link' );
+								echo '</a>';
+								echo '</li>';
+							}
+							?>
+						</ul>
+					</div>
+			 	<?php } ?>
 				<a id="filter-toggle" data-action="collapse" href="#">
 					<i class="1 ace-icon fa bigger-125 <?php echo $t_block_icon ?>"></i>
 				</a>
 			</div>
-			<div id="filter-search-bar" class="widget-toolbar no-border" style="display: <?php echo $t_collapse_block ? 'block' : 'none' ?>">
-				<div class="widget-menu">
-					<input type="text" size="16" name="search" class="input-xs" value="<?php echo string_html_specialchars( $t_filter['search'] ); ?>" />
-					<input type="submit" name="filter" class="btn btn-primary btn-white btn-round btn-xs" value="<?php echo lang_get( 'filter_button' ) ?>" />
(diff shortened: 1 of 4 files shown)


vboctor:
It would be great if we can have "Search" next to the text box and also provide a way to select from the saved filters.  Here is the collapsed version from the classic UI.

filter_collapsed

syncguru:
@vboctor the classic view is way too busy for a collapsed filter. I've a simpler collapsed filter bar for modern ui:

screen shot 2016-08-22 at 9 06 25 pm

screen shot 2016-08-22 at 9 18 43 pm

atrol:
While you are on it, some more ideas:

- Add a "Reset Filter" button
- Remove the [Reset Filter] entry from drop down list
- Remove the "Use Filter" button (no longer needed as we agreed some while ago that JavaScript must be enabled for MUI)

"Save Current Filter" and "Create Permalink" are similar operations.
It's a bit confusing to have one of them as a button and the other one as a menu entry.
What about removing the buttons "Save Current Filter" and "Manage Filters" and add them to the menu?

dregad:
:+1: with atrol's suggestions

vboctor:
In the case where the user has put together an adhoc filter, what would be shown in the filter name dropdown?  Suggestions to avoid confusion?

dregad:
`Unsaved filter` maybe ?

syncguru:
@atrol I did try to work through your suggestions, here is what I ended up with:
1- "Use Filter" button is gone. Now, form submission is triggered on dropdown change
2-  Removing [Reset Filter] option and switching that to a button felt like undoing (1) .. having no button to control filter selection is much better
3- Moved the 'manage filter' to menu
4- Left "Save current Filter" as is for couple of reasons:
a. In the menu, saving was hard to find.
b. Saving filter requires changes to filter JS logic which should be its own issue/PR. I think (a) was really bad so (b) is not very relevant.

screen shot 2016-08-25 at 7 19 28 pm

screen shot 2016-08-25 at 7 19 46 pm

vboctor:
:+1:

atrol:
Issue 1:
Try collapsed filter without having a stored filter.
Resetting filter is not that intuitive.
You have to select the empty row in the drop down and change to "Reset Filter" after that.

Issue 2:
Try collapsed filter with one or more stored filters.
Select a filter from drop down.
The drop down does not show the chosen filter after that.

Issue 3:
The "Search" in text input field is not complete visible in collapsed mode.
Seems to be just a Safari issue.
searchcollapsed

Issue 4:
Strings entered in the search field are not displayed after applying the filter.
You get "Search" again.

syncguru:
Rebased and fixed #3 & #4 in @atrol list

atrol:
There are changes in the PR (verify.php) that are not related to search bar.

syncguru:
@atrol Fixed the PR. As a quick recap of your list of issues:
#1 dropdown is hidden when there are no stored filters
#2 fixed
#3 fixed
#4 fixed

core/filter_api.php
```diff
@@ -2315,31 +2316,76 @@ function filter_draw_selection_area2( $p_page_number, $p_for_screen = true, $p_e
...
-					}
-				}
-				?>
-				<?php
-				if( access_has_project_level( config_get( 'create_permalink_threshold' ) ) ) {
-					echo '<a class="btn btn-xs btn-white btn-primary btn-round" href="permalink_page.php?url=', urlencode( filter_get_url( $t_filter ) ), '">';
-					echo lang_get( 'create_filter_link' );
-					echo '</a>';
-				} ?>
+			<?php if ( count( $t_stored_queries_arr ) > 0 ) { ?>
```
atrol: There should be no blank between if and (

core/filter_api.php
```diff
@@ -3613,24 +3659,16 @@ function filter_draw_selection_area2( $p_page_number, $p_for_screen = true, $p_e
 	echo '<div class="btn-toolbar pull-right">';
 	echo '<div class="btn-group">';
 
-	$t_stored_queries_arr = filter_db_get_available_queries();
-
-	if( access_has_project_level( config_get( 'stored_query_create_threshold' ) ) ) { ?>
+	if ( access_has_project_level( config_get( 'stored_query_create_threshold' ) ) ) { ?>
```
atrol: Same as above

core/filter_api.php
```diff
@@ -3613,24 +3659,16 @@ function filter_draw_selection_area2( $p_page_number, $p_for_screen = true, $p_e
...
 	}
-	if( count( $t_stored_queries_arr ) > 0 ) { ?>
-
-	<form class="form-inline pull-left" method="post" name="open_queries" action="query_view_page.php">
-			<?php # CSRF protection not required here - form does not result in modifications ?>
-			<input type="submit" name="switch_to_query_button" class="btn btn-primary btn-white btn-sm btn-round"
-				value="<?php echo lang_get( 'open_queries' )?>" />
-		</form>
-		<form class="form-inline pull-left padding-left-8"  method="get" name="list_queries<?php echo $t_form_name_suffix;?>" action="view_all_set.php">
+	if ( count( $t_stored_queries_arr ) > 0 ) { ?>
```
atrol: Same as above

atrol:
Thanks for fixing the issues.
I tested a bit and did not find any obvious bug.

What I still don't like that much is the way to reset the filter if there exist stored filters.
To be sure this is not caused by my testing environment:
After changing the filter you see this
resetfilter1

After opening the menu you get 
resetfilter2


You have to select the empty entry to reset the filter.
Is that what you expected?
If so, I think it's not that intuitive. 

If you still don't like the additional button to reset the filter:
What about displaying [Unsaved Filter] instead of the empty line and selecting this entry
instead of [Reset Filter]?
So you would open the drop down, choose [Reset Filter] to reset and will get [Unsaved Filter]
again.

syncguru:
@atrol - I checked a really simple fix - move [Reset Filter] to the second spot in the dropdown list instead of first. I think this should deliver the expected behavior. Give it a try and let me know.

I prefer not having an additional button for this. MantisBT has been around for a long time and if this was real pain point, it would be been fixed long time ago. I'd like to not optimize for the rare usage.

atrol:
> move [Reset Filter] to the second spot in the dropdown list 

It's better than before, but I still don't like the UI that much (same in 1.3.x).
It does not mean that I want to block this PR as the main reason for it is to show the
text search when the filter is collapsed.
Thus, +1 after having applied my minor requests for the white space issues.

vboctor:
I did some testing and it looks good.  I would just replace the blank element in the dropdown with '[select filter]', but we can do this as a follow up

PR #866, opened 2016-08-22
Cache latest data in function user_cache_database_result

atrol:
The functions did not return a value in all cases.
There is no place in code where the return value of the function is used.

The cache was not updated with latest data if's been set before.

Fixes #20551

### DIFF
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -130,16 +130,12 @@ function user_cache_array_rows( array $p_user_id_array ) {
 }
 
 /**
- * Cache an object as a bug.
+ * Cache a user row
  * @param array $p_user_database_result A user row to cache.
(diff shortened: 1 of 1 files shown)


cproensa:
Is the value returned never used throughout the code?
May the value be used by external code, eg plugins?

atrol:
> Is the value returned never used throughout the code?

There is no place in code where the return value of the function is used.

> May the value be used by external code, eg plugins?

None of the plugins hosted on https://github.com/mantisbt-plugins uses this function.
The function did not return a value in all cases, so someone who uses the return value might get random results.

vboctor:
:+1:

dregad:
:+1:

PR #868, opened 2016-08-26
Highlight due date when the date has passed

syncguru:
Fixes #21642

screen shot 2016-08-25 at 7 52 11 pm


vboctor:
Did you check the case where the due date is visible on the View Issues page as a column?  I would compare the behavior with the classic UI to make sure we covered all the cases.

syncguru:
@vboctor Yes, I looked at that case. It is the same as in the issue details page. The table cell is in view issues page decorated with 'overdue' css class when overdue which will be highlighted as shown in the image above.

vboctor:
👍

dregad:
@syncguru in terms of UI, I was wondering since you removed colored backgrounds pretty much everywhere (replacing them with colored squares), why you didn't follow the same approach here.

PR #873, opened 2016-08-28
Do not use FILTER_PROPERTY_HIDE_STATUS for timeline

cproensa:
To match previous behaviour of Timeline, do not hide bugs based on any
status, even if the default empty filter is configured to do so.

Fixes: #21652


atrol:
👍

dregad:
:+1:

PR #874, opened 2016-08-31
Dont break complex threshold in workflow config page

cproensa:
The configuration option "set_status_threshold" may contain complex
thresholds defined as arrays instead of integer values.
In this case, on manage_config_workflow_page, show those fields as not
editable.

Fixes: #21655, #21656


dregad:
:+1: 

Code looks OK, but did not actually test.

dregad:
Merged in 6e0801d328fe6d52a5c3c55e400de8a17584dac1

PR #876, opened 2016-09-02
Issue #21659 fix for PHP Parse error

jenolan:
syntax error, unexpected 'version' (T_STRING) in browser_search_plugin.php on line 56

### DIFF
--- a/browser_search_plugin.php
+++ b/browser_search_plugin.php
@@ -52,8 +52,8 @@ if( $f_type == 'id' ) {
 }
 
 header( 'Content-Type: application/opensearchdescription+xml' );
+echo '<?xml version="1.0" encoding="UTF-8" ?>';
 ?>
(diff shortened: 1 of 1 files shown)


syncguru:
👍

atrol:
👍

vboctor:
It is unclear to me why the changed code is different from the old code and why it fixes the issue?  It seems that we are outputing exactly the same output but form PHP instead of inline code.

grangeway:
Victor, you can use <? as "Short tags" in php ( the short_open_tag config option), so in this case, php is trying to parse the string as php code

vboctor:
Merged via 1b32685bfd2a82fe53848686dd393b46276ae622 in 1.3.x and will be merged into master as part of regular merges.  Thanks @jenolan

PR #877, opened 2016-09-03
Upgrade to Ace theme 1.4 & include unminified files

syncguru:
Fixes #21603


dregad:
Thank you Rafik.

Just a note regarding the file header that you added to the ACE template files. Shouldn't this reference "ACE Template" or whatever it's called, and show copyright information for the original author, instead of making it look like it's an integral part of MantisBT ?

syncguru:
@dregad I switched to using generic gnu GPL license wording as agreed with the theme author at: https://github.com/mantisbt/mantisbt/blob/master/doc/ace-theme-license.txt

syncguru:
@dregad, @vboctor: Added author info, can we get this in?

grangeway:
This looks like you are adding author information to the file as opposed to the original author.

vboctor:
Merged via 6c082930e8e3b98870cfe70b6c8e2e7d88e56501
Thanks @syncguru

PR #880, opened 2016-09-04
fix: Repeat 302 after the project appears to be disabled

wapznw:
fix: Repeat 302 after the project appears to be disabled


vboctor:
Seems like a fix for https://www.mantisbt.org/bugs/view.php?id=21157

wapznw:
It seems to be, but the solution is too cumbersome.

vboctor:
Merged via 9a7ae21655bce7f1346cd81a3f635d301c8844fd in both 1.3.x and master.

PR #881, opened 2016-09-06
Fix CSP errors in MantisGraph plugin

vboctor:
Fixes #21653

### DIFF
--- a/plugins/MantisGraph/MantisGraph.php
+++ b/plugins/MantisGraph/MantisGraph.php
@@ -70,7 +70,21 @@ class MantisGraphPlugin extends MantisPlugin  {
 	 */
 	function csp_headers() {
 		if ( config_get_global( 'cdn_enabled' ) == ON ) {
-			http_csp_add( 'script-src', 'https://cdnjs.cloudflare.com/ajax/libs/Chart.js/' );
+			http_csp_add( 'script-src', 'https://cdnjs.cloudflare.com' );
+		}
+
+		# Enable inline scripts for MantisGraph plugin pages unless the need for inline
+		# scripts is removed.
+		$t_page = gpc_get_string( 'page', '' );
+		if ( !is_blank( $t_page ) ) {
+			$t_pos = stripos( $t_page, '/' );
+
+			if ( $t_pos !== false ) {
+				$t_page = substr( $t_page, $t_pos + 1 );
+				if ( $_SERVER['REQUEST_URI'] == plugin_page( $t_page ) ) {
+					http_csp_add( 'script-src', "'unsafe-inline'" );
+				}
+			}
 		}
 	}
 


dregad:
:-1: 

Allowing inline scripts is defeating one of the main purposes of using CSP in the first place, effectively disabling the protection it gives against XSS attacks.

I did not look at the graph pages source so I'm not sure how complex a task it would be, but the code should be rewritten to reference external javascript files instead of inlining it.

vboctor:
This PR is about fixing the broken functionality, rather than having no reporting at all for 2.0.0-beta.2.  These pages are much less critical than the view page which is affected by the Dropzone inline script.  There are follow up bugs for fixing the root cause.

https://www.mantisbt.org/bugs/view.php?id=21669
https://www.mantisbt.org/bugs/view.php?id=21651

I think we should get this in, then fix the above.

atrol:
> I think we should get this in, then fix the above.

I would agree if you agree that severity of https://www.mantisbt.org/bugs/view.php?id=21669 should be  blocking and we don't release any release candidate until this is fixed.

vboctor:
I disagree that we should stop beta/rc releases on CSP issues.  Also note that whatever we are doing is implementing improvements.

2.0.0-beta.1 - CSP disabled
2.0.0-beta.2 - CSP enabled, with white list for view page, charts broken.
2.00-beta.3 - CSP enabled, charts not broken, white lists for view/charts pages.
2.0.0-... - CSP enabled, no white lists.

The above are not exact, but gives you an idea.  Also note that 1.3.x has a bug_update_page.php whitelist due to jscalendar - https://www.mantisbt.org/bugs/view.php?id=20040

Hopefully we fix this all soon, but I don't see these as blockers for releases.

atrol:
>  but I don't see these as blockers for releases.

I didn't tell that I would like to see this as blockers in general for all releases, I just asked for release candidates. Beta versions are Ok for me.

BTW, wouldn't it be good to add @todo comments?

vboctor:
We could use @todo comments, but having an issue should to the trick.

There are very few things that I would block a release on where the release is better and more stable than the previous one.  Having a graphs page white list CSP inline scripts (vs broken) is not one of them.  In the past, releases took years because we had such things act as blockers.  Everyone is willing to identify blockers but not necessary have the time to fix them.  Specially blockers that are mainly "engineering" but not "functional" blockers.

I would like to continue to shoot for monthly releases for both stable / development branches.  And have a short cycle for moving from one 1.3.x to 2.0.x, 2.0.x to 2.1.x, etc.

Release often is the main order bit.  Everyone is welcome to fix what they think is important, but we will keep releasing.

atrol:
> Everyone is willing to identify blockers but not necessary have the time to fix them. 

Blockers are blockers, you should not release if no one has the time to fix it.
The question is: What is a blocker?
We could distinguish between blockers for beta versions and blockers for release candidates.

2.0.0-beta.1 introduced a regression, it is worse than 1.3.x in terms of security of the graph functionality.
It is Ok for me that 2.0.0-beta.3 would be just a functional improvement compared to beta.2.
But the reduced security could be considered as a blocker for the first release candidate.

> 2.0.0-... - CSP enabled, no white lists.

Isn't this what I am asking for?
Just set target version to 2.0.x and severity to blocking
https://www.mantisbt.org/bugs/view.php?id=21669

dregad:
> This PR is about fixing the broken functionality, rather than having no reporting at all for 2.0.0-beta.2

I understand that. My point is that the fix is not the correct one, and I don't really understand why we'd need an intermediate step that degrades security, rather than implement the proper solution right away.

If this is an issue because of MantisHub, then I suggest the workaround proposed in this PR is applied locally.

> These pages are much less critical than the view page which is affected by the Dropzone inline script. 

I didn't realize there was an inline script there too. If I had, I would have had the same reaction.

> There are very few things that I would block a release on where the release is better and more stable than the previous one. 

IMO, degraded security is one of these.

> Everyone is willing to identify blockers but not necessary have the time to fix them. 

Sure it's easier to say that something is broken than go about and fix it, but at the same time it is also the responsibility of the person who implements a feature to make sure it works without introducing issues.

> 2.0.0-beta.1 - CSP disabled

That was a regression from 1.3, and should *not* have happened in the first place

> 2.0.0-beta.2 - CSP enabled, with white list for view page, charts broken.

Restoring security the way it should always have been, allowing identification of a mistake in the development of charts (i.e. inlining scripts)

> 2.00-beta.3 - CSP enabled, charts not broken, white lists for view/charts pages.
> 2.0.0-... - CSP enabled, no white lists.

My preference would be to go straight to fixing the root cause (i.e. remove the inline scripts), but I'm willing to compromise if the 2 follow-up issues you mentioned are set as blocking for the next non-beta release

scalastic:
I've been working on the DropZone CSP inline script error.

One way to deal with is to generate dynamically the JavaScript with PHP and call this PHP script directly from the header page as Javascript source with function like `html_javascript_link()` from `html_api.php`.

I have implemented one `fileupload_js.php` script that generates the JavaScript and which is included with a new function `html_javascript_dynamic_link( 'fileupload_js.php' );`.

Does this nomenclature suit you ?
Waiting your feedback to pull request.

PR #882, opened 2016-09-08
Don't show UI to attach tags if user has no access rights for it

atrol:
Fixes #21679

### DIFF
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -203,9 +203,6 @@ $t_show_resolution = in_array( 'resolution', $t_fields );
 $t_show_status = in_array( 'status', $t_fields );
 $t_show_tags =
 	in_array( 'tags', $t_fields ) &&
-	access_has_global_level( config_get( 'tag_view_threshold', /* default */ null, /* user */ null, $t_project_id ) );
-$t_can_attach_tag =
-	$t_show_tags &&
 	access_has_project_level(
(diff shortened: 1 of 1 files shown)


bug_report_page.php
```diff
@@ -201,11 +201,8 @@
 $t_show_os_version = $t_show_profiles && in_array( 'os_version', $t_fields );
 $t_show_resolution = in_array( 'resolution', $t_fields );
 $t_show_status = in_array( 'status', $t_fields );
-$t_show_tags =
-	in_array( 'tags', $t_fields ) &&
-	access_has_global_level( config_get( 'tag_view_threshold', /* default */ null, /* user */ null, $t_project_id ) );
 $t_can_attach_tag =
```
vboctor: I would rename this variable to $t_show_tags since it doesn't necessarily ability to add tags but whether they should be displayed or not.

bug_report_page.php
```diff
@@ -545,8 +542,8 @@
 <?php
 	}
 
-	# Display tags fields
-	if( $t_show_tags ) { ?>
+	# Display tags fields if user is allowed to attach tags
```
vboctor: I would remove the added comment.  Again it is not just about being allowed, but about the fields configured to show in the report page.

vboctor:
:+1: with the minor changes proposed.

dregad:
:+1:

PR #884, opened 2016-09-10
Standardize "operation successful" messages

dregad:
Adding new API functions to handle display of confirmation messages, and replace custom HTML in various files to display the same by calls to these functions.

Fixes [#21683](https://www.mantisbt.org/bugs/view.php?id=21683)

### DIFF
--- a/bug_report.php
+++ b/bug_report.php
@@ -323,10 +323,11 @@ if( !is_blank( $f_tag_string ) || $f_tag_select != 0 ) {
 	}
 }
 
-html_operation_confirmation( array(
+$t_buttons = array(
 	array( string_get_bug_view_url( $t_bug_id ), sprintf( lang_get( 'view_submitted_bug_link' ), $t_bug_id ) ),
 	array( 'view_all_bug_page.php', lang_get( 'view_bugs_link' ) ),
-) );
+);
+html_operation_confirmation( $t_buttons, CONFIRMATION_TYPE_SUCCESS );
 
 if( $f_report_stay ) {
 ?>
--- a/bug_report.php
+++ b/bug_report.php
@@ -323,7 +323,7 @@ if( !is_blank( $f_tag_string ) || $f_tag_select != 0 ) {
 	}
 }
 
-html_operation_successful_buttons( array(
+html_operation_confirmation( array(
 	array( string_get_bug_view_url( $t_bug_id ), sprintf( lang_get( 'view_submitted_bug_link' ), $t_bug_id ) ),
 	array( 'view_all_bug_page.php', lang_get( 'view_bugs_link' ) ),
 ) );
--- a/bug_report.php
+++ b/bug_report.php
@@ -322,16 +322,11 @@ if( !is_blank( $f_tag_string ) || $f_tag_select != 0 ) {
 		}
 	}
 }
-?>
 
-<div class="col-md-12 col-xs-12">
-	<div class="space-10"></div>
-	<div class="alert alert-success">
-
-<?php
-echo '<p>' . lang_get( 'operation_successful' ) . '</p><br />';
-print_button( string_get_bug_view_url( $t_bug_id ), sprintf( lang_get( 'view_submitted_bug_link' ), $t_bug_id ) );
-print_button( 'view_all_bug_page.php', lang_get( 'view_bugs_link' ) );
+html_operation_successful_buttons( array(
+	array( string_get_bug_view_url( $t_bug_id ), sprintf( lang_get( 'view_submitted_bug_link' ), $t_bug_id ) ),
+	array( 'view_all_bug_page.php', lang_get( 'view_bugs_link' ) ),
+) );
 
 if( $f_report_stay ) {
 ?>
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -373,7 +373,8 @@ function html_top_banner() {
  * @param array  $p_buttons      Array of (URL, label) pairs used to generate
  *                               the buttons; if label is null or unspecified,
  *                               the default 'proceed' text will be displayed.
- * @param string $p_message      Message to display to the user.
+ * @param string $p_message      Message to display to the user. If none is
+ *                               provided, a default message will be printed
  * @param string $p_type         One of the constants CONFIRMATION_TYPE_SUCCESS,
  *                               CONFIRMATION_TYPE_WARNING, CONFIRMATION_TYPE_FAILURE
  * @return void
@@ -382,16 +383,16 @@ function html_operation_successful_buttons( array $p_buttons, $p_message = '', $
 	switch( $p_type ) {
 		case CONFIRMATION_TYPE_FAILURE:
 			$t_alert_css = 'alert-danger';
-			$t_message = lang_get( 'operation_failed' );
+			$t_message = 'operation_failed';
 			break;
 		case CONFIRMATION_TYPE_WARNING:
-			$t_message = lang_get( 'operation_warnings' );
+			$t_message = 'operation_warnings';
 			$t_alert_css = 'alert-warning';
 			break;
 		case CONFIRMATION_TYPE_SUCCESS:
 		default:
 			$t_alert_css = 'alert-success';
-			$t_message = lang_get( 'operation_successful' );
+			$t_message = 'operation_successful';
 			break;
 	}
 
@@ -401,8 +402,10 @@ function html_operation_successful_buttons( array $p_buttons, $p_message = '', $
 	echo '<div class="alert ' . $t_alert_css . ' center">';
 
 	# Print message
-	if( !is_blank( $p_message ) ) {
-		echo $p_message . '<br />';
+	if( is_blank( $p_message ) ) {
+		$t_message = lang_get( $t_message );
+	} else {
+		$t_message = $p_message;
 	}
 	echo '<p class="bold bigger-110">' . $t_message  . '</p><br />';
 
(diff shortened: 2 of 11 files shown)


core/html_api.php
```diff
@@ -369,27 +369,98 @@ function html_top_banner() {
...
- * A function that outputs that an operation was successful and provides a redirect link.
- * @param string $p_redirect_url The url to redirect to.
- * @param string $p_message      Message to display to the user.
+ * Outputs an operation successful message with multiple redirection buttons.
+ * @param array  $p_buttons      Array of (URL, label) pairs used to generate
+ *                               the buttons; if label is null or unspecified,
+ *                               the default 'proceed' text will be displayed.
+ * @param string $p_message      Message to display to the user. If none is
+ *                               provided, a default message will be printed
+ * @param string $p_type         One of the constants CONFIRMATION_TYPE_SUCCESS,
```
atrol: integer $p_type

core/html_api.php
```diff
@@ -369,27 +369,98 @@ function html_top_banner() {
...
  */
-function html_operation_successful( $p_redirect_url, $p_message = '' ) {
+function html_operation_successful_buttons( array $p_buttons, $p_message = '', $p_type = CONFIRMATION_TYPE_SUCCESS ) {
+	switch( $p_type ) {
+		case CONFIRMATION_TYPE_FAILURE:
+			$t_alert_css = 'alert-danger';
+			$t_message = 'operation_failed';
+			break;
+		case CONFIRMATION_TYPE_WARNING:
+			$t_message = 'operation_warnings';
```
atrol: For better readability and to be consistent with other cases.
Insert this line after `$t_alert_css = 'alert-warning';`

core/html_api.php
```diff
@@ -369,27 +369,98 @@ function html_top_banner() {
...
+ *                               the buttons; if label is null or unspecified,
+ *                               the default 'proceed' text will be displayed.
+ * @param string $p_message      Message to display to the user. If none is
+ *                               provided, a default message will be printed
+ * @param string $p_type         One of the constants CONFIRMATION_TYPE_SUCCESS,
+ *                               CONFIRMATION_TYPE_WARNING, CONFIRMATION_TYPE_FAILURE
  * @return void
  */
-function html_operation_successful( $p_redirect_url, $p_message = '' ) {
+function html_operation_successful_buttons( array $p_buttons, $p_message = '', $p_type = CONFIRMATION_TYPE_SUCCESS ) {
```
atrol: html_operation_confirmation_buttons might be a better name for the function.
It's a bit strange to to have the word successful for failures and warnings, e.g.
```php
function html_operation_failure( $p_redirect_url, $p_message = '' ) {
	html_operation_successful_buttons(...
```

manage_user_create.php
```diff
@@ -131,16 +131,12 @@
 layout_page_header( null, $t_redirect_url );
 
 layout_page_begin( 'manage_overview_page.php' );
-?>
-
-<div class="space-20"></div>
-<div class="alert alert-success">
-<?php
-$t_access_level = get_enum_element( 'access_levels', $f_access_level );
```
atrol: Don't remove this line as you use $t_access_level to build $t_message

manage_user_reset.php
```diff
@@ -79,31 +79,24 @@
...
-echo '<div class="col-md-12 col-xs-12">';
-echo '<div class="space-10"></div>';
-echo '<div class="alert alert-success">';
-
-echo '<p class="bigger-110">';
 if( $t_reset ) {
 	if( false == $t_result ) {
 		# PROTECTED
-		echo lang_get( 'account_reset_protected_msg' );
+		# @TODO this should be a failure message !
```
atrol: So why don't you use html_operation_failure?

manage_user_update.php
```diff
@@ -226,12 +226,12 @@
...
 
-if( $f_protected && $t_old_protected ) {				# PROTECTED
-	echo '<div class="failure-msg">';
-	echo lang_get( 'manage_user_protected_msg' ) . '<br />';
-	print_button( $t_redirect_url, lang_get( 'proceed' ) );
-	echo '</div>';
-} else if( $t_result ) {					# SUCCESS
+if( $f_protected && $t_old_protected ) {
+	# PROTECTED
+	# @TODO this should actually be a warning message
```
atrol: So why don't you use html_operation_warning?

atrol:
👍 in general, just reviewed, didn't test.

core/html_api.php
dregad: >integer $p_type

Right, forgot to change the phpdoc when I moved the strings to constants

core/html_api.php
dregad: > Insert this line after `$t_alert_css = 'alert-warning';`

Ok.

core/html_api.php
dregad: > html_operation_confirmation_buttons might be a better name for the function.

Good point. Initially, this function was only to extend the existing `html_operation_successful()` function, then I realized there were use cases for warning/failure but forgot to change the name. Will do.

manage_user_create.php
dregad: Oops...

manage_user_reset.php
dregad: I was planning to (hence the @TODO), then forgot about it...

manage_user_update.php
dregad: > So why don't you use html_operation_warning?

Same comment as above

dregad:
Thanks @atrol for the review, I made the necessary corrections.

bug_report.php
```diff
@@ -322,16 +322,11 @@
...
-	<div class="alert alert-success">
 
-<?php
-echo '<p>' . lang_get( 'operation_successful' ) . '</p><br />';
-print_button( string_get_bug_view_url( $t_bug_id ), sprintf( lang_get( 'view_submitted_bug_link' ), $t_bug_id ) );
-print_button( 'view_all_bug_page.php', lang_get( 'view_bugs_link' ) );
+html_operation_confirmation( array(
+	array( string_get_bug_view_url( $t_bug_id ), sprintf( lang_get( 'view_submitted_bug_link' ), $t_bug_id ) ),
+	array( 'view_all_bug_page.php', lang_get( 'view_bugs_link' ) ),
+) );
```
syncguru: Should this be a call to html_operation_successful instead?
On a side note, I find this function call with array of arrays input to be rather complex. I would suggest separating the buttons build up logic first. Also using static arrays if possible http://php.net/SplFixedArray

news_update.php
```diff
@@ -69,17 +69,10 @@
...
-print_button( 'news_edit_page.php?news_id=' . $f_news_id . '&action=edit', lang_get( 'edit_link' ) );
-echo '&nbsp;&nbsp;&nbsp;&nbsp;';
-print_button( 'news_menu_page.php', lang_get( 'proceed' ) );
-echo '</div>';
-echo '</div>';
-echo '</div>';
+html_operation_confirmation( array(
+	array( 'news_menu_page.php' ),
+	array( 'news_edit_page.php?news_id=' . $f_news_id . '&action=edit', lang_get( 'edit_link' ) ),
+) );
```
syncguru: Same here.

syncguru:
Apart from comments, this PR looks good.

bug_report.php
dregad: > Should this be a call to html_operation_successful instead?

html_operation_successful() only allows a single button, so to maintain existing functionality it is required to use html_operation_confirmation().

> I find this function call with array of arrays input to be rather complex. 
> I would suggest separating the buttons build up logic first. 

So you're saying I should do this instead ?
```
$t_buttons = array(
	array( string_get_bug_view_url( $t_bug_id ), sprintf( lang_get( 'view_submitted_bug_link' ), $t_bug_id ) ),
	array( 'view_all_bug_page.php', lang_get( 'view_bugs_link' ) ),
);
html_operation_confirmation( $t_buttons );
```
If that's the case, I don't really see the benefit.

> Also using static arrays if possible http://php.net/SplFixedArray

Considering that in 100% of current use cases the array size is <= 2, and that it's not reused after the function call, I'm not sure there is a real benefit for fixed-size arrays here.

news_update.php
dregad: Same reply...

bug_report.php
syncguru: The code snippet is more clear than the original.

syncguru:
👍

dregad:
@syncguru thanks for the feedback. For the record I did not use SplFixedArray because I don't think it makes sense in this context, it just adds unnecessary overhead.

I just rebased the branch to fix the conflict on _lost_pwd.php_

cproensa:
Looks good, but i could only test lightly so far.

By the way, do you plan to port it to 1.3?

dregad:
> do you plan to port it to 1.3?

I was not planning to initially, but I guess it would make sense

lang/strings_english.txt
```diff
@@ -103,6 +103,8 @@ $s_actiongroup_error_issue_is_readonly = 'Issue is readonly.';
 $s_all_projects = 'All Projects';
 $s_move_bugs = 'Move Issues';
 $s_operation_successful = 'Operation successful.';
+$s_operation_warnings = 'Operation completed with warnings.';
+$s_operation_failed = 'Operation did not complete successfully.';
```
vboctor: Why not "Operation failed."?  Do we or should we include reasons when it fails?  Not a big deal.

jllano:
@dregad 
1. I have had checkout the master branch, able to see your changes
2. I tried to replicate the issue on #894 , and got this below:

checkstay_broken

dregad:
Opened [#21710](https://www.mantisbt.org/bugs/view.php?id=21710) to track this

PR #885, opened 2016-09-11
Cleanup html api in 1.3

atrol:
Should also be a very small performance improvement as there is no longer basename( $_SERVER['SCRIPT_NAME'] for each request.

No need to merge to master as those functions do no longer exist.


vboctor:
:+1:

syncguru:
👍

PR #886, opened 2016-09-12
Obsolete configuration option icon_path

atrol:
After using FontAwesome for icons there is some dead code in Mantis.
After removing the dead code, there is no longer any access to setting
icon_path.

Fixes #21689


vboctor:
:+1:

syncguru:
👍

PR #887, opened 2016-09-12
Fix CSP errors with Dropzone JavaScript

scalastic:
Fix CSP restriction on inline scripts when Dropzone’s JavaScript loads


scalastic:
I've been working on the DropZone CSP inline script error.

One way to deal with is to generate dynamically the JavaScript with PHP and call this PHP script directly from the header page as Javascript source with function based on `html_javascript_link()` from `html_api.php`.

I have implemented one `fileupload_js.php` script that generates the JavaScript and which is included with a new function `html_javascript_dynamic_link( 'fileupload_js.php' );`.

Does this nomenclature suit you ?

vboctor:
I was thinking about a different approach.  The approach would be for the PHP code of the page to create a hidden element with json of the configs with it.

    <div class="dropzone-json hidden">{ ...json... }</div>

Then include a static regular javascript file that would replace the above element with a call to javascript code with the parameters parsed from the json text.

Hence, standard javascript caching would work fine, and there won't be a need to load all Mantis internals to get the PHP working and trigger the generation of the javascript file.

In the case of graphs, there may be more than one graph on the page.  In such case, there will be a div per graph, and the javascript will process all these divs that have the appropriate class name.

Thoughts?

dregad:
Thanks @scalastic for your contribution.  I haven't looked into the details, but off-hand I would give preference to @vboctor's approach. 

On the latter, instead of the hidden element hack, maybe you could consider using [data attributes](https://developer.mozilla.org/en/docs/Web/Guide/HTML/Using_data_attributes) instead ?

scalastic:
Then I'll give a try dataset and static javascript ;)

vboctor:
Closing in favor of #931

PR #888, opened 2016-09-16
Update account verify page for modern ui

cproensa:
Fixes: #21684

### DIFF
--- a/verify.php
+++ b/verify.php
@@ -84,7 +84,8 @@ extract( $t_row, EXTR_PREFIX_ALL, 'u' );
 
 $t_can_change_password = helper_call_custom_function( 'auth_can_change_password', array() );
 
-layout_login_page_begin();
+layout_page_header();
+layout_page_begin();
 
 ?>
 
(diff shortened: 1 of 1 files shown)


cproensa:
@syncguru 
I'd want this page to show only the main form, without sidebar, navigation, etc
If i use layout_login_page_begin(), the main form is not displayed
I can't find a way to do that properly, please have a look.

dregad:
I'm not the resident modern UI expert here, but as reference _login_page.php_ has the kind of layout you are looking for, so maybe you can adapt it to your purposes.

syncguru:
@cproensa - The login pages uses a different layout helpers function as @dregad pointed out. You would want to use layout_login_page_begin() & end() instead. Examples to look at are login_page.php, signup.php, lost_pwd_page.php ...

cproensa:
If i use layout_login_page_begin() it's not working 100%...
note that those pages you said are not using the "blue widget stuff" css, (the styles are different)

if i use: layout_page_header() + layout_login_page_begin()
there is a margin at left
seleccion_051


if i use only layout_login_page_begin(), the form-div is not rendered:
seleccion_050

jllano:
Hi,

I'm adding a reference with my PR on the issue as well. 
https://github.com/mantisbt/mantisbt/pull/895

vboctor:
Closing this in favor of #895

PR #890, opened 2016-09-17
helper_array_transpose should always return an array

dregad:
The function now triggers an error if the input array is not bi-dimensional.

This fixes the issues introduced by my earlier attempt, which was inadvertently pushed to the core repository (see commit eaa7286) and later reverted (in dd39f53).

Fixes [#17325](https://www.mantisbt.org/bugs/view.php?id=17325)

The related PHPUnit tests have been modified to take this change into account. 

As a prerequisite for this, the Tests configuration script was modified to disable MantisBT's error handler so that errors can be captured as exceptions by PHPUnit (see [#21696](https://www.mantisbt.org/bugs/view.php?id=21696)).

### DIFF
--- a/core/helper_api.php
+++ b/core/helper_api.php
@@ -89,13 +89,11 @@ function helper_alternate_class( $p_index = null, $p_odd_class = 'row-1', $p_eve
  * @return array|mixed transposed array or $p_array if not 2-dimensional array
  */
 function helper_array_transpose( array $p_array ) {
-	if( !is_array( $p_array ) ) {
-		return $p_array;
-	}
 	$t_out = array();
 	foreach( $p_array as $t_key => $t_sub ) {
 		if( !is_array( $t_sub ) ) {
-			return $p_array;
+			# This function can only handle bidimensional arrays
+			trigger_error( ERROR_GENERIC, ERROR );
 		}
 
 		foreach( $t_sub as $t_subkey => $t_value ) {
--- a/tests/Mantis/HelperTest.php
+++ b/tests/Mantis/HelperTest.php
@@ -37,41 +37,135 @@ require_mantis_core();
 class MantisHelperTest extends PHPUnit_Framework_TestCase {
 
 	/**
-	 * Tests helper_array_transpose()
-	 * @dataProvider providerArrayTranspose
+	 * Custom assertion to evaluate whether the given exception was a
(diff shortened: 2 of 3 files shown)


dregad:
Without feedback for over a week, should I assume everyone is OK with this ? I'll merge the changes tonight or tomorrow.

atrol:
I don't see a relationship between the changes in helper_array_transpose and the changes in core/version_api.php

dregad:
There's none. I have no idea why Github is showing the 1st two commits in the PR, since they are [already existing in master-1.3.x branch](https://github.com/mantisbt/mantisbt/commits/595136f00a3e1463f1b3f362d8f18aa1752aa824), and my feature branch is based on top of 595136f00a3e1463f1b3f362d8f18aa1752aa824. 

git log

I'll try to rebase the branch, hopefully the problem will go away.

dregad:
That did the trick apparently. Weird.

core/helper_api.php
```diff
@@ -89,13 +89,11 @@ function helper_alternate_class( $p_index = null, $p_odd_class = 'row-1', $p_eve
  * @return array|mixed transposed array or $p_array if not 2-dimensional array
  */
 function helper_array_transpose( array $p_array ) {
-	if( !is_array( $p_array ) ) {
```
vboctor: It is unclear to me why this check was removed.

tests/Mantis/HelperTest.php
```diff
@@ -37,41 +37,136 @@
...
 	/**
-	 * Returns test array
-	 * @return array
+	 * Tests helper_array_transpose() with invalid values.
+	 *
+	 * @param mixed $p_in  Input value.
+	 * @return void
+	 *
+	 * @dataProvider providerArrayTransposeInvalid
+	 *
```
vboctor: Remove extra blank comment line.

core/helper_api.php
dregad: It is not necessary - thanks to the `array` type hint $p_array MUST be an array.

http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration

tests/Mantis/HelperTest.php
dregad: Will do.

PR #891, opened 2016-09-17
Better visual marker for selected projects & subprojects

syncguru:
Fixes #21121

Before:
screen shot 2016-09-17 at 12 17 18 pm

After:
screen shot 2016-09-17 at 12 17 36 pm

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -661,7 +661,7 @@ function layout_navbar_subproject_option_list( $p_parent_id, $p_project_id = nul
 		check_disabled( $t_id == $p_filter_project_id || !$t_can_report );
 		echo '>';
 		echo str_repeat( '&#160;', count( $p_parents ) * 4 );
-		if( strlen( $p_project_id ) >= strlen( $t_full_id ) && 0 == strpos ( $p_project_id,  $t_full_id ) ) {
+		if( strlen( $p_project_id ) >= strlen( $t_full_id ) && 0 == strpos( $p_project_id, $t_full_id ) ) {
 			echo '<i class="ace-icon fa fa-check"></i> ';
 		} else {
 			echo '<i class="ace-icon fa fa-fw"></i> ';
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -599,9 +599,9 @@ function layout_navbar_projects_list( $p_project_id = null, $p_include_all_proje
 			check_selected( $p_project_id, ALL_PROJECTS, false );
 		}
 		if( ALL_PROJECTS == $p_project_id ) {
-			echo '><i class="ace-icon fa fa-dot-circle-o"></i> ';
+			echo '><i class="ace-icon fa fa-check"></i> ';
 		} else {
-			echo '><i class="ace-icon fa fa-circle-o"></i> ';
+			echo '><i class="ace-icon fa fa-fw"></i> ';
 		}
 		echo lang_get( 'all_projects' ) . ' </a></li>' . "\n";
 		echo '<li class="divider"></li>' . "\n";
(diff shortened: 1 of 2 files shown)


cproensa:
I see some issues, see image:
- all projects in the same level (even from different parents), are marked
- alignment doesnt look right... same level projects looks like subprojects, due to both a small misalignent to the selected (same level project) and the visual weight of the mark on the left of project caption

seleccion_049

Still, i dont see the use case for also marking the parent projects

cproensa:
I'd also suggest a stronger wight for the selected project caption
For example: style the text bolder.

Also, a standard HTML selector will have the selected option highlighted when first unfolded. Consider if this behaviour is worth implementing

cproensa:
> I didn't test the behavior with many projects (which was the initial point of #21121

right, this does not solve the main issue: project list is not scrollable.

The visual presentation, which is being addressed here, was added as secondary note.
Let me know if you need it to be opened as separate issue, if the actual #21121 is not being treated here..

syncguru:
@cproensa, @dregad Fair enough. Rethinking marking of active project in the menu:

screen shot 2016-09-17 at 4 28 01 pm

cproensa:
now, as simple as it is, i think it's satisfactory. 

Have a look at the "compact" view, the highlighted current project now results in white text over white background.

syncguru:
@cproensa - Good catch!

screen shot 2016-09-17 at 8 21 55 pm

core/layout_api.php
```diff
@@ -613,15 +609,11 @@ function layout_navbar_projects_list( $p_project_id = null, $p_include_all_proje
 			$t_can_report = access_has_project_level( $t_report_bug_threshold, $t_id, $t_user_id );
 		}
 
-		echo '<li><a href="' . helper_mantis_url( 'set_project.php' ) . '?project_id=' . $t_id . '"';
+		echo 0 == strcmp( $t_id, $p_project_id ) ? '<li class="active">' : '<li>';
```
vboctor: Why not $t_id == $p_project_id?

core/layout_api.php
```diff
@@ -655,19 +647,12 @@ function layout_navbar_subproject_option_list( $p_parent_id, $p_project_id = nul
 			$t_full_id = $t_id;
 		}
 
-		echo '<li>';
+		echo 0 == strcmp( $p_project_id, $t_full_id ) ? '<li class="active">' : '<li>';
```
vboctor: same here.

cproensa:
also ok for me
but don't mark #21121 as solved, as the main issue there is different.

syncguru:
@cproensa I cannot repro the main issue which is scale to high number of projects. It seems to be working fine and does as well as standard dropdown box like the classic version.

cproensa:
@syncguru add enough projects so that the list overflows your screen
then you cant select the last ones because the list is not scrollable.

syncguru:
@cproensa please try again
The funny thing is that I could not repro on local dev box (http) but was able to repro the issue when I used c9.io IDE (https). Quite strange, but the fix forces scroll-ability of the projects dropdown menu.

atrol:
The scrollbar issue is no longer reproducible on my test system after latest changes in commit 337c343

cproensa:
the scroll issue is fixed for me too

core/layout_api.php
syncguru: @vboctor $t_id is int while $p_project_id is a string. The == check will succeed in pass for the top parent project (int comparison - skips text ;xx) and the selected subproject (string comparison). The correct behavior is single selection thus we need to enforce string comparison.

core/layout_api.php
syncguru: Not needed here since $p_project_id & $t_full_id are strings - but I wanted to keep it consistent.

vboctor:
Merged via 35b21c348853ac05cf917f80b3d810493ae7a1f9

PR #892, opened 2016-09-20
Clearer distinction between private and public notes

jllano:
Use a different color for private notes when being edited and viewed.

Fixes #21697

Before:
before1
before2

After:
after2
after1

### DIFF
--- a/css/default.css
+++ b/css/default.css
@@ -31,12 +31,10 @@ span.dependency_upgrade		{ color: orange; }
 
 tr.bugnote .bugnote-note { background-color: #e8e8e8; color: #000000; width: 75%; vertical-align: top; }
 
-tr.bugnote > td.bugnote-note.bugnote-private { background-color: #fcf8e3;}
-
 .bugnote-private { background-color: #fcf8e3 !important;}
 
 .nowrap
-{ 
+{
 	white-space: nowrap;
 }
 
--- a/css/default.css
+++ b/css/default.css
@@ -31,6 +31,8 @@ span.dependency_upgrade		{ color: orange; }
 
 tr.bugnote .bugnote-note { background-color: #e8e8e8; color: #000000; width: 75%; vertical-align: top; }
 
+tr.bugnote > td.bugnote-note.bugnote-private { background-color: #fcf8e3;}
+
 .nowrap
 {
 	white-space: nowrap;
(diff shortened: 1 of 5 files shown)


js/common.js
```diff
@@ -335,6 +335,15 @@ $(document).ready( function() {
 	$('a.click-url').bind("click", function() {
 		$(this).attr("href", $(this).attr("url"));
 	});
+
+	$('input[name=private].ace').bind("click", function() {
+		if ($(this).is(":checked")){
+			$('textarea[name=bugnote_text]').css("background-color", "#fcf8e3");
```
cproensa: i'd rather change a class in the textarea/container, than harcode color codes in javascript
Then in css, this class would define background colors

jllano: I like your idea, that wd be best practice and standard. I'll update my code.

css/default.css
```diff
@@ -31,8 +31,12 @@ span.dependency_upgrade		{ color: orange; }
 
 tr.bugnote .bugnote-note { background-color: #e8e8e8; color: #000000; width: 75%; vertical-align: top; }
 
+tr.bugnote > td.bugnote-note.bugnote-private { background-color: #fcf8e3;}
```
dregad: This rule is overqualified, and anyway redundant considering the _important_ flag on the more generic rule below. Please remove it.

css/default.css
```diff
@@ -31,8 +31,12 @@ span.dependency_upgrade		{ color: orange; }
...
+
+.bugnote-private { background-color: #fcf8e3 !important;}
+
 .nowrap
-{
+{
```
dregad: Trailing whitespace

css/default.css
```diff
@@ -31,8 +31,12 @@ span.dependency_upgrade		{ color: orange; }
 
 tr.bugnote .bugnote-note { background-color: #e8e8e8; color: #000000; width: 75%; vertical-align: top; }
 
+tr.bugnote > td.bugnote-note.bugnote-private { background-color: #fcf8e3;}
+
+.bugnote-private { background-color: #fcf8e3 !important;}
```
dregad: I guess it's a debatable question of taste, but I'm not that keen on this yellowish background. 

IMO, one of the improvements of the Modern UI was to remove the many colors used throughout Mantis 1.3 and earlier, and this feels like going back in this direction. Maybe use another shade of gray, or some other variation of a color already present in the interface ?

@syncguru your thoughts on this would be appreciated.

jllano: The color was suggested by @syncguru.

syncguru:
@dregad Private comments should really be rare, thus alert the user if there is any in the page is the purpose of this design. 
I am not hang up on specific color. Feel free to pick a different color, post a snapshot and reach consensus with the rest o the team.

vboctor:
Merged via d62f601b28034ad8d223000f21544164129a08b4

PR #895, opened 2016-09-20
Account verify page is broken

jllano:
-Added required modern api and styles
-Ensure to wrap modern container and layouts
-Ensure buttons and icons are consistent with other

Fixes #21684

Before looks:
previos

After:
latest

### DIFF
--- a/verify.php
+++ b/verify.php
@@ -106,7 +106,7 @@ layout_login_page_begin();
 
 		<?php
 			if( $t_can_change_password ) {
-				echo '<div id="reset-passwd-msg" class="alert alert-sm alert-danger ">';
+				echo '<div id="reset-passwd-msg" class="alert alert-sm alert-warning ">';
 				echo lang_get( 'verify_warning' ) . '<br />';
 				echo lang_get( 'verify_change_password' );
 				echo '</div>';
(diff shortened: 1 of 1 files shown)


cproensa:
This one has better looks than #888.
Notice that i used the streamlined manage form styles from mantis ui (which seems to have some issues if the full page, sidebars etc are not initialized)
Yours looks more like the "disconnected" pages (login-page, etc) which is also a valid option. In this case it still seems that the style is different that standard pages: login, lost passwd, create account, etc
Should it comply more with that style then?

syncguru:
I cloned this pull request locally and made few tweaks to get the page to look like the snapshot below:

screen shot 2016-09-20 at 8 48 15 pm

I added comment on the code for the changes needed.

jllano:
Looks great! Could you commit your changes?

dregad:
> I added comment on the code for the changes needed.

@syncguru I don't see your changes anywhere, did you actually push them ?

cproensa:
You can also send a pull rquest to his branch with your changes so they are integrated in this PR.

Also, confirm if this page styling is the desired for this page, so i can close then my other related PR #888

@syncguru if you have the time, i'd like to sort out my issue in #888, where i cant use the main ui styles without the sidebar, header, etc.
Maybe it's necessary to create a lighter version of layout_page_header()?

jllano:
I already merged your changes @syncguru

dregad:
> Maybe it's necessary to create a lighter version of layout_page_header()?

As a follow-up action to this PR, I believe it would make sense to do this. It seems like a lot of manual HTML code used here (and in login_page.php) could (and should) be dealt with in Mantis core.

> On a minor note: Id have a look if displaying the username as: `username (xxxx)` is the correct one.

I agree with @cproensa here, I think this part of the layout can be improved. 

I would suggest, instead of having the username on its own line, to display it in the form's legend, e.g. `Edit Account - <username>`.

syncguru:
👍

vboctor:
I've applied the comment about how we present user id + merged this via 67d860e52e1e5750563a4d3a3a301f4d92f170aa

PR #896, opened 2016-09-20
plugin_get() fails to use the default argument

val-kulkov:
Fixes a bug that prevents defaulting to current plugin if plugin_get() is called without arguments.

The bug is fairly obvious, given that $t_current is assigned either the function argument or the default value and then the value of $t_current is not used at all in the function's return value.

Fixes [#21707](https://www.mantisbt.org/bugs/view.php?id=21707)

### DIFF
--- a/core/plugin_api.php
+++ b/core/plugin_api.php
@@ -117,7 +117,7 @@ function plugin_get( $p_basename = null ) {
 		trigger_error( ERROR_PLUGIN_NOT_REGISTERED, ERROR );
 	}
 
-	return $g_plugin_cache[$p_basename];
+	return $g_plugin_cache[$t_current];
 }
 
 /**


dregad:
Thanks for the patch.

Out of curiosity, do you have any real-life scenario where we actually hit this bug ?

val-kulkov:
I hit the bug when trying to call a class method on a handle of my plugin's class:
```
$myplugin = plugin_get();
$myplugin->push_menu_item($menuitem);
```

dregad:
I opened [#21707](https://www.mantisbt.org/bugs/view.php?id=21707) to track this.

dregad:
Fixed in 06b44c90576e200c18dc2aa00ee9faeee8f740c3

PR #897, opened 2016-09-21
Fixed #6823 : Date filter should work with "last update"

MrBricodage:


dregad:
@MrBricodage Thanks for the pull request. I have not had time to look into it, but I just wanted mention that there is another pending PR proposing significant changes around filters API (see #862). I do not know if it conflicts with yours, but regardless I suggest you discuss with @cproensa who has done a lot of work in this area lately, to try and coordinate the efforts.

cproensa:
This work has caught me off guard, sorry..

From what i see, this is adding a new filter property for "last update date".
Also its renaming current "date" property to a more correct "submit date" (this is cosmetic, but note that it can break current stored filters)

We could argue if the new property for "last update" should behave as a separate filter field, to keep consistency with a 1:1 relation. 
What happens when both date types are informed, are they ANDed or ORed?

Regarding my related PR #862, there's a lot of refactor of the filter core code, which actually allows an easier addition of new properties/fields. However it is stalled, at the moment, as it adds ui pages, that are waiting feedback.
Probably time to give it a push again @dregad ?

MrBricodage:
> From what i see, this is adding a new filter property for "last update date".
Right
> Also its renaming current "date" property to a more correct "submit date" 
Right.
> (this is cosmetic, but note that it can break current stored filters)
Right, that's the dark side of this PR

Others modifications : 
- remove checkbox "Use Date Filters" in view_filter_page, not usefull anymore with one checkbox for each date type.
- display months as number (don't remember exactly where (filter view or filter edit mode) but lang was forced to english, even if an other language was used in Mantis)



[...]
>What happens when both date types are informed, are they ANDed or ORed?
ANDed

cproensa:
@MrBricodage 
As was noted: as PR #862 is on the way, which is a big effort for reviewing and testing, and as most of the code changed here has been refactored in that PR, we could include your changes there to join the reviewing process (as suggested also by other team members)

However most will need to be rewritten, but hopefully, will need less code to be implemented.

For the changes proposed, i suggest:
- Create the needed properties for new field, but keep the current ones for "date_submitted" untouched. To ensure existing filters will work.
- Show "last update date" as a separated filter field. There is not any interaction for both date fields that is not supported by the standard field aggregation. Also, this helps in keeping it simpler for layout (as current filter layout may be transitional, with new-ui and redesign proposals)

I you want, we can coordinate a pull request to my branch, where i can help you through the new api structure
...or i can do it myself if you dont have the time.

MrBricodage:
@cproensa
Thanks for your response. I'll do anything needed to add "last updated date" filter in Mantis.

> Create the needed properties for new field, but keep the current ones for "date_submitted" untouched. To ensure existing filters will work.

Right

>Show "last update date" as a separated filter field. There is not any interaction for both date fields that is not supported by the standard field aggregation. Also, this helps in keeping it simpler for layout (as current filter layout may be transitional, with new-ui and redesign proposals)

Could you please confirm that your request is to use separate filters like this : 
proposal2

I just tested your branch that still use "Use Date Filter" as filter label for creation date bug but there is no checkbox : 
capture2
How users will disable filter if needed? Using the same date for start and end is a workaround but not really "user friendly". Is it a mistake or a new behaviour?

FYI, Mantis 1.3.2 provide a submenu with checkbox for creation date :  
capture3


Regards

cproensa:
@MrBricodage 
The existing date filter was not modified. I still see the check to enable or disable it. I'm not sure why you dont have it there...

> Could you please confirm that your request is to use separate filters like this : 
Your sketch is good.

Soon when i get the time I will summarize the places to put your code, to get you started on it. Thanks!

cproensa:
@MrBricodage , to integrate in my branch:
these steps will be the same, i think:
create property constants for new field
create lang strings for new field and related texts

These steps have been reallocated in several functions:
**in filter_api.php:**
_filter_get_default_array()_ to create default values
_filter_gpc_get()_ to parse GET/POST
_filter_get_bug_rows_query_clauses()_ to build the sql query clauses
**in filter_form_api.php:**
_print_filter_xxx()_ to create the filter form inputs
_print_filter_values_xxx()_ to get the filter values for non-editing filter form
_filter_form_get_input()_ as a mapping to print_filter.... functions
_filter_form_draw_inputs()_ to draw a new box in filter form

possibly, tweak (or duplicate) the javascript for the "use date filter" check. Don't think hard here, as those are ugly and would need a rewrite in future.

other pages:
_view_all_set.php, view_filters_page.php,_ now they should not need modification
_search.php_, (this is used for permalinks) i still haven't refactored that yet, so replicate current date logic.

and do a search for existing date property constants to see if any other place applies for changes.

reach me on gitter if need help
thanks

MrBricodage:
Asked modifications are present in https://github.com/cproensa/mantisbt/pull/1.
Will be included in #862 if accepted.

I close this PR

PR #894, opened 2016-09-20
Report Stay checkbox shows broken layout on action page

jllano:
Fix Broken layout when "Report Stay" checkbox is checked

Fixes #21704

Previous looks:
stay_before

After:
stay_report

PR #898, opened 2016-09-23
Fix report stay button

dregad:
To leverage the new API functions handling the display of operation confirmation messages, the _Report More Issues_ button is now handled as a single URL query with get parameters, instead of an HTML form.

Fixes [#21704](https://www.mantisbt.org/bugs/view.php?id=21704)

This replaces @jllano's PR #894

### DIFF
--- a/bug_report.php
+++ b/bug_report.php
@@ -327,30 +327,24 @@ $t_buttons = array(
 	array( string_get_bug_view_url( $t_bug_id ), sprintf( lang_get( 'view_submitted_bug_link' ), $t_bug_id ) ),
 	array( 'view_all_bug_page.php', lang_get( 'view_bugs_link' ) ),
 );
-html_operation_confirmation( $t_buttons, '', CONFIRMATION_TYPE_SUCCESS );
 
 if( $f_report_stay ) {
-?>
-	<p>
-	<form method="post" action="<?php echo string_get_bug_report_url() ?>">
-	<?php # CSRF protection not required here - form does not result in modifications ?>
-		<input type="hidden" name="category_id" value="<?php echo string_attribute( $t_bug_data->category_id ) ?>" />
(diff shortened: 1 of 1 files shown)


dregad:
For the record, for those who did not read the commit message: this replaces the HTML form by an URL query with get parameters. This makes for a rather long URL when submitting the next issue, but the advantage is that we minimize code duplication (i.e. leverage the new APIs).

See https://www.mantisbt.org/bugs/view.php?id=21710#c54057,

jllano:
Are we not going to compromise later of exposing the id's in the url vs post submit? What if adding the form back and tweak the styles to make it in-line with the other buttons? @syncguru @vboctor

vboctor:
Tested and it works.

- The operation successful page should auto-redirect to same url as "Report More Issues".
- Not a big fan of the long url, though it works.  I wonder if we can re-use the concept of master id, so we pass in the id of the bug and use the bug id as a template (similar to master id, but for a subset of fields).  In general, we should make sure we are validating that user has view access to master id.

dregad:
> The operation successful page should auto-redirect to same url as "Report More Issues".

Good point, I'll fix that

> Not a big fan of the long url

I don't like it much either, but I thought it was an acceptable compromise to increase code reuse.

> use the concept of master id, so we pass in the id of the bug and use the bug id as a template 

That's an option, but it goes beyond the scope of this PR. I let you open an issue in the tracker to follow-up with that, if you feel it's important.

vboctor:
The important thing is to fix the auto-redirect.  The URL is secondary.  So go ahead and merge once you fix the former.

dregad:
@vboctor I logged [#21723](https://www.mantisbt.org/bugs/view.php?id=21723) to track the redirect with report stay feature, since that did not exist before, and implemented it. 

Branch rebased and ready to merge unless there are further review comments.

PR #899, opened 2016-09-25
Duplicate id field should show up when closing an issue

vboctor:
Fixes #21711


dregad:
I think the idea behind this was that issues should be resolved (duplicate), before being closed as opposed to going straight for closure, but I'm fine with the change.

PR #901, opened 2016-09-27
Handle null tokens when upgrading from 1.2.x

vboctor:
I haven't tested the fix, appreciate if someone who has a repro to test.  I can simulate it if no one has a real repro from a failing db.

Fixes #21573


atrol:
I hesitated because of encoding null values, but looks good as PHP supports it.
http://php.net/manual/en/function.json-encode.php
Didn't test.

PR #904, opened 2016-09-27
Fix missing tooltips on issue Ids

dregad:
This is a partial revert of 4d8ce6ff687ab5bec9af3d635aefe7fe9f24d6f7
(see issue #21112).

Hiding the tooltip in My View page is now done using print_bug_link()'s
$p_detail_info parameter.

Fixes [#21721](https://www.mantisbt.org/bugs/view.php?id=21721)

### DIFF
--- a/my_view_inc.php
+++ b/my_view_inc.php
@@ -360,7 +360,7 @@ for( $i = 0;$i < $t_count; $i++ ) {
 	# -- Bug ID and details link + Pencil shortcut --?>
 	<td class="nowrap width-13 my-buglist-id">
 		<?php
-			print_bug_link( $t_bug->id );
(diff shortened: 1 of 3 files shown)


my_view_inc.php
```diff
@@ -360,7 +360,7 @@
 	# -- Bug ID and details link + Pencil shortcut --?>
 	<td class="nowrap width-13 my-buglist-id">
 		<?php
-			print_bug_link( $t_bug->id );
+			print_bug_link( $t_bug->id, false );
```
syncguru: I think you need to do the same thing in summary_page.php

dregad: Opened [#21728](https://www.mantisbt.org/bugs/view.php?id=21728) to track this

PR #905, opened 2016-09-27
cleanup db_now()

val-kulkov:
db_now() returns a Unix timestamp, not a DATETIME-formatted string as
currently described in the DocBlock. The proposed changes fix the
function description and introduce a new function, db_now_datetime(),
that returns a DATETIME-formatted string.

In addition, the proposed changes introduce an optional parameter,
$p_gmt. When the parameter is set to TRUE, the current time is
interpreted as GMT time.

Signed-off-by: Val Kulkov <val.kulkov@gmail.com>

### DIFF
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -838,9 +838,8 @@ function db_prepare_bool( $p_bool ) {
 }
 
 /**
- * return current timestamp for DB
- * @todo add param bool $p_gmt whether to use GMT or current timezone (default false)
- * @return string Formatted Date for DB insertion e.g. 1970-01-01 00:00:00 ready for database insertion
+ * return current time as Unix timestamp
+ * @return integer Unix timestamp of the current date and time
  */
 function db_now() {
 	return time();


val-kulkov:
The Travis CI build #3164.5 timed out for no apparent reason. Could it be a temporary problem at their server? Should I re-submit this PR?

core/database_api.php
```diff
@@ -838,12 +838,27 @@ function db_prepare_bool( $p_bool ) {
...
+ * return current time as Unix timestamp
+ * @param boolean $p_gmt True: current time is GMT; False: current time is in the current timezone (default)
+ * @return int Unix timestamp of the current date and time
  */
-function db_now() {
-	return time();
+function db_now( $p_gmt = false ) {
+    $t_now = time();
+    if( $p_gmt )
+        $t_now += (int)date('Z');
```
dregad: Use `{ }`

core/database_api.php
```diff
@@ -838,12 +838,27 @@ function db_prepare_bool( $p_bool ) {
 }
 
 /**
- * return current timestamp for DB
- * @todo add param bool $p_gmt whether to use GMT or current timezone (default false)
- * @return string Formatted Date for DB insertion e.g. 1970-01-01 00:00:00 ready for database insertion
+ * return current time as Unix timestamp
+ * @param boolean $p_gmt True: current time is GMT; False: current time is in the current timezone (default)
+ * @return int Unix timestamp of the current date and time
```
dregad: `integer` not `int`

core/database_api.php
```diff
@@ -838,12 +838,27 @@ function db_prepare_bool( $p_bool ) {
...
+/**
+ * return current time as DATETIME-formatted date and time
+ * @param boolean $p_gmt True: current time is GMT; False: current time is in the current timezone (default)
+ * @return string DATETIME-formatted date, such as "1970-01-01 00:00:00"
+ */
+function db_now_datetime( $p_gmt = false ) {
+    $t_now = time();
+    if( $p_gmt )
+        $t_now += (int)date('Z');
+	return date( 'Y-m-d H:i:s', $t_now);
```
dregad: I wouldn't add the new function unless we have a clearly identified need for it.

dregad:
> db_now() returns a Unix timestamp, not a DATETIME-formatted string as
currently described in the DocBlock. 

I think this is a leftover from the days before we switched from DATETIME fields to INTEGERs, storing dates as Unix timestamps. 

> introduce a new function, db_now_datetime(), that returns a DATETIME-formatted string.

I don't think this new API is necessary.

> introduce an optional parameter, $p_gmt. 

Not really necessary IMO (that `@todo` goes back to 2008, no idea what the reasoning was for adding it), but I guess it does not hurt.

> The Travis CI build #3164.5 timed out for no apparent reason. Could it be a temporary problem at their server? Should I re-submit this PR?

That happens sometimes. It's usually enough to just restart the failed task(s). I just did that.

We would appreciate if any PR you submit is linked to an issue in our bugtracker (create one if necessary). Please reference it in the PR, and/or the commit message(s).

val-kulkov:
@dregad: now that you have explained that all dates are supposed to be stored as integers, the DocBlock's @return and @todo values in the DocBlock section to db_now() indeed look like leftovers from the old days.

I am happy to simply remove the @todo values and references to $p_gmt from my pull request. I do not really see any use cases where $p_gmt may be useful. $p_gmt is probably no longer relevant.

As for the db_now_datetime() function, I am using this function in the plugin that I am working on. My plugin uses datetime fields and therefore I need datetime-formatted strings for database inserts. I will be happy to remove db_now_datetime() from this PR if it seems unlikely that someone else can benefit from having this function. If there are reasons to stay away from using datetime fields and using integer fields instead for storing dates in the database, I would much appreciate a comment here or a reference to an earlier discussion or a link to Wiki or some other guideline.

Also, I will be happy to create an issue in the bugtracker once I get some guidance in respect of my questions above. It would be great to specifically recommend creating an issue in the bugtracker to developers before submitting a PR either here:
https://www.mantisbt.org/wiki/doku.php/mantisbt:howto_submit_patches
or here:
https://www.mantisbt.org/wiki/doku.php/mantisbt:contributing
or in both places.

dregad:
The use of integers for dates happened before I joined the project, but I believe the reason was to maximize portability between supported RDBMS. For consistency, I would indeed suggest you use the same approach as MantisBT core, unless you really need datetime fields.

With regards to the bug id reference, it's necessary so any merged changes are properly tracked on the roadmap/changelog. I agree it should be better documented :)

val-kulkov:
db_now() returns a Unix timestamp, not a DATETIME-formatted string as currently described in the DocBlock. The proposed change fixes the function description and removes the outdated @todo that appears to be no longer relevant.

Fixes [#21725](https://www.mantisbt.org/bugs/view.php?id=21725)

@dregad : please review the updated PR and let me know if I should squash two commits into one.

dregad:
OK for me. If you don't mind, please rebase & squash.

val-kulkov:
@dregad : rebasing and squashing done, please review and merge as appropriate. Thank you.

dregad:
Merged in cd74a2ebd022d88cf33c4eb3673a03777c9a3f37 (updated commit message to add reference to [#21725](https://www.mantisbt.org/bugs/view.php?id=21725)

PR #907, opened 2016-09-28
set default timezone during install

giallu:
installing with a recent PHP (tested with 5.6.25) leads to a
system warning about missing the default timezone setting
that blocks the installation. Since we select the
timezone in the first install step, we can use that before
continuing with the second

screenshot from 2016-09-17 09-40-23


dregad:
Merged in d45035f551bf5c739fd4212fca4d913ce9a4dbd5

PR #908, opened 2016-09-28
Add missing </div>

PatrickMassot:
Fixes [#21726](https://www.mantisbt.org/bugs/view.php?id=21726)

A missing ``</div>`` ruins the footer layout of api_tokens_page

### DIFF
--- a/api_tokens_page.php
+++ b/api_tokens_page.php
@@ -152,4 +152,5 @@ if ( count( $t_tokens ) > 0 ) {
 <?php
 }
 
+echo '</div>';
 layout_page_end();


PatrickMassot:
I'm sorry about that. My initial intention was to fill an issue but I was put off by the message: "Currently signup for new accounts is disabled due to spam attacks! Use our support channels to request new accounts. We apologize for the inconvenience." It made it quicker to submit a PR than a bug report...

dregad:
No problem. We should really get around and find a permanent solution for these spamming issues. Do you want me to create an account for you ?

PatrickMassot:
Thank you. Yes I guess it could be useful to have an account in case I see other things. Maybe I can use this opportunity for a bit of general discussion. Right now I'm investigating whether Mantis could be useful for me and other people like me. The context is slightly unusual, I'm a mathematician looking for a light bug tracker with source control integration and a modern look and feel. It seems like the new Mantis 2.x branch with its source integration plugin could be a nice fit. 

The thing that worries me is, if I understand correctly, getting rid of unneeded features or irrelevant fields necessitates hardcoding settings in ``config_inc.php``. Typical use for us would be: I write a an academic paper with a bunch of colleagues and, at the end of the rewriting process or during revision after submitting the paper, we need to keep track of a number of things to correct and discuss. Doing this by emails is tedious and error prone and we want to conveniently refer to git or svn commits. But there is nothing like a relevant OS, reproducibility...And only one version of the paper will ever be published so no target version etc. For me alone it's not a problem to set things in ``config_inc.php`` but it certainly hampers wider adoption.

syncguru:
👍

dregad:
Merged in b325bc90d0fe52ed579ff265eb4bd7f552515522

PR #909, opened 2016-09-29
Show attachments inline with notes

vboctor:
We should show up the attachments inline with the notes. That is specially useful when the user submits a note with one or more attachments, but it is helpful in all cases for attachments.

In the future as we gradually update the layout of the View Issue page, it is expected that history should also be shown inline, but this is outside the scope of this issue.

Here is how it looks like.  The attachments were removed from the issue header details and are now displayed inline.  When a user submits attachments with a note, they appear as one entry.

inline_attachments

I'll do some more testing, but it would be great to get feedback.

**Questions and TODOs**
- _Should we change the title of this section?_  Now the section includes "Notes and Attachments", but I can see history entries being added inline as well (in the future with a revised UI).  Every time we change the title, we lose all translations, so we either keep it or choose a durable name (e.g. Activities).
- _Should deleting a note, delete its associated attachments?_  I would say yes.
- _Should the user be able to delete a specific attachment associated with a note without deleting the note and other attachments associated with the same note?_  I would say for now, we should only allow deleting a row, since if user deletes an attachment associated with a note, they can't add an alternative one.

Fixes #21727


dregad:
I like the approach, but can't test right now. Did you consider attachments linked without a specific note (i.e. when initially reporting the issue) ? How are they handled ?

From memory, I think there was an existing issue with the tracker dealing with note-specific attachments vs "global", issue-wide attachments; don't have time to search for it atm. You might want to take a look at it.

Will look into more details later on.

vboctor:
Issue attachments show up as their own entry.  See the third example in the screenshot.

Issue note attachments are combined with the note in one entry. See first and fourth entries in the screenshot.  At the moment I associate attachments with notes based on timestamp (within 10 seconds) + user id, in the future, we can have a proper link in the database.

dregad:
Great, thanks for the clarification.

Please take into consideration the following cases 

- deleting attachments: I assume that currently the line simply disappears from the bugnote item ? It may be useful to show it as deleted
- updating/replacing attachments: if I wanted to replace e.g. a bad screenshot by the correct one, it would be useful if the updated attachment were attached to the original note.

Also, I don't see a *Delete* button in your screenshots,

dregad:
> From memory, I think there was an existing issue with the tracker dealing with note-specific attachments vs "global", issue-wide attachments

Found it... https://www.mantisbt.org/bugs/view.php?id=7835

vboctor:
Thanks @dregad for your comments.

> Found it... https://www.mantisbt.org/bugs/view.php?id=7835

Reading the description / title, this issue seems unrelated to what I'm doing here.

> I don't see a Delete button in your screenshots,

In case of issue attachments, the Delete button is on the left (similar to notes).  The work that is not done here is what should happen when a note that has attachments is deleted, and enabling deleting of an attachment associated with a note.  One option is to have the Delete button for the entry delete both the note and the attachments associated with it and not have a way to delete a specific attachment that is associated with a note.

> deleting attachments: I assume that currently the line simply disappears from the bugnote item ? It may be useful to show it as deleted

I'm not sure what you mean by "show as deleted".  In the past when we deleted an attachment, it simplify disappeared as well.  Same for tags, relationships, notes, etc.

> updating/replacing attachments: if I wanted to replace e.g. a bad screenshot by the correct one, it would be useful if the updated attachment were attached to the original note.

For now, you can delete the note + attachment and re-add them.  We can improve on this in the future, but not part of this PR.

vboctor:
There are some open questions in the PR description with suggested answers.  It would be great to get feedback on them.

cproensa:
> In case of issue attachments, the Delete button is on the left (similar to notes). The work that is not done here is what should happen when a note that has attachments is deleted, and enabling deleting of an attachment associated with a note. One option is to have the Delete button for the entry delete both the note and the attachments associated with it and not have a way to delete a specific attachment that is associated with a note.

I'd be conservative in the actual linking for notes and attachments. There is no real data to reflect that relation, only a guess based on timestamps.
If we build core functionality based on that, (like, delete files when deleting notes, etc), my fear is having to bear the technical debt of not doing it properly in the first place.

In terms of UI.
At fisrt, i dont feel like there is a clear organization for notes + attachments. confusing as the container for notes and atts is the same
(you said that this could be reviewed in the future), maybe its worth to polish a bit more just for the start.

Also, as a first impression, i feel that not having the attachments at header position is a lost value. You'd have to skim through the notes to get to them (here too the importance of better indentify them visually)
Probably, something can be done to merge both old and new approaches.

Said that, and didnt try hands-on, these are just quick thoughts

cproensa:
> From memory, I think there was an existing issue with the tracker dealing with note-specific attachments vs "global", issue-wide attachments

can be this one:
https://www.mantisbt.org/bugs/view.php?id=10504

atrol:
>  it is expected that history should also be shown inline

Just in case you didn't know, there is a plugin for it https://github.com/mantisbt-plugins/inline-history

vboctor:
@cproensa

> I'd be conservative in the actual linking for notes and attachments.

Here are issues tracking the work to make the tracking of linked and private attachments in the db:

- https://www.mantisbt.org/bugs/view.php?id=21733
- https://www.mantisbt.org/bugs/view.php?id=9802

For the time being we can be conservative by having a small time window to associate a note with an attachment.  I've already wrote the logic that does the association.  Once we change the db schema, we can use the same code for associating previously uploaded notes/attachments.

> maybe its worth to polish a bit more just for the start.

The updated UI is consistent with the existing one, so it is not a regression.  I would rather have the UI revision of the View Issue page as separate work.  However, I'm planning to update the title of the section to reflect that this section is no longer just about notes.

As for the approach I want to go with, I want to reduce the use of boxes.  I also want to avoid grid of field names and values (similar to what you are planning with filter box).  So I expect a cleaner UI that has the issue title, issue description, issue fields, and then issue activities.  Will have to factor in the edit functionalities like add relationships, monitor users, tags, etc.

As for the activities section, below is a couple of examples, but it may be different.  But you can think of it as some sort of timeline.

Option 1:
timeline1

Option 2:
timeline2

@atrol I recall seeing something related.  It seems to be a plugin rather than an issue.  Though it seems like it is focused on inline history rather than inline attachments (and eventually history).

vboctor:
@cproensa I can also create a config option that controls the number of seconds for associating a note with attachments with 0 turning this functionality OFF, and hence the note will be shown in its own row followed by attachments on their own rows.  The default can be 2-3 seconds.

dregad:
> create a config option that controls the number of seconds for associating a note with attachments with 0 turning this functionality OFF

I think that's a good approach.

I'm still not sure about the *delete* part. I am concerned about potential unwanted / unexpected removal of attachment when a note is deleted, as well as the lack of ability to remove an individual attachment from an existing note which I believe is a necessary feature.

I also agree with @cproensa, that it is (at least for me) quite useful to be able to see all attachments in a single place, without having to read all notes to find them. Maybe an option to only show notes having attachments could be a viable alternative; this could even be expanded to a wider concept of bugnotes filtering (e.g. by date, author, keywords, public/private, etc) but that's a different topic.

cproensa:
@vboctor 
I wrote in [#21733: Attachments should be linkable to notes in db](https://www.mantisbt.org/bugs/view.php?id=21733)

I still think the file-note relation should be solid before doing anything more than layout changes

dregad:
> create a config option that controls the number of seconds for associating a note with attachments with 0 turning this functionality OFF

I think that's a good approach, if you truly want to do this without a hard link between note and attachment.

I'm still not sure about the *delete* part however. I am concerned about potential unwanted / unexpected removal of attachment when a note is deleted, as well as and more importantly the lack of ability to remove an individual attachment from an existing note which I believe is a necessary feature.

I also agree with @cproensa, that it is (at least for me) quite useful to be able to see all attachments in a single place, without having to read all notes to find them. Maybe an option to only show notes having attachments could be a viable alternative; this could even be expanded to a wider concept of bugnotes filtering (e.g. by date, author, keywords, public/private, etc) but that's a different topic.

cproensa:
Should we differentiate bug attachments from note attachments?
I'm in favour of note specific attachments, which should be placed in the note, and bug attachments that still are shown at the main bug info.
Probably in the new workflow, most attachments will be note-level, but developers can still attach a bug-level file in the classic way.
In that scenerio, note-level files should use the note properties and operations:
- private status inherited from note
- be deleted alongside the whole note, (but individual note-file deletion should be available too)

Optionally, attach a file to an existing note?

vboctor:
Based on feedback, I've implemented the following:

- Called the section 'Activities' instead of 'Notes'.
- Added trashcan icon to delete a note when associated with a note.
- Added a configuration option to set the threshold in seconds for associating an attachment with a note or 0 to disable the feature.  Default is 3 seconds.

I've also done the following:

- Refactored logic into bug_activity_api.php

Remaining TODOs:

- Deleting a note should also delete associated attachments, right now, it doesn't.  The delete icon shows a confirmation page.  I can explore providing a button to delete the note only or a note + attachments.

Here is the updated preview:

inline_attachments_2

atrol:
- At the moment "Activities" are displayed in reversed order of what we had as default display for "Notes".
The "Notes Sort Order" (Renamed to "Activities Sort Order") of "Account Preferences" should be considered when displaying "Activities".

- If there are a lot of activities you have to scroll down to add a comment, so you don't see the latest
activity when commenting.
As an enhancement, display the "Add Note" box before the "Activities" box if "Notes Sort Order" is set to "descending".

- You could completely remove the "Upload Files" box if it would be allowed to add empty notes.

config_defaults_inc.php
```diff
@@ -4458,6 +4458,7 @@
 	'html_valid_tags_single_line',
 	'html_valid_tags',
 	'impersonate_user_threshold',
+	'issue_activity_note_attachments_seconds_threshold',
```
atrol: should be placed after 'inline_file_exts'

core/bug_activity_api.php
```diff
@@ -0,0 +1,252 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Bug Activity API
+ *
+ * @package CoreAPI
+ * @subpackage AccessAPI
```
atrol: ActivityAPI

core/bug_activity_api.php
```diff
@@ -0,0 +1,252 @@
...
+ * @copyright Copyright 2000 - 2002  Kenzaburo Ito - kenito@300baud.org
```
atrol: We do no longer add this line for new files.

core/bug_activity_api.php
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * Sort bugnotes and attachments by timestamp then user_id.  If two entries have
+ * the same timestamp and user id, then the note should be before the attachment.
+ *
+ * @param array $p_entries The array of entries.  The array will be updated.
+ * @return void
+ */
+function bug_activity_sort( &$p_entries ) {
+	$t_order = config_get( 'bugnote_order' );
```
atrol: should be current_user_get_pref( 'bugnote_order' )

core/print_api.php
```diff
@@ -1883,14 +1883,15 @@ function get_dropdown( array $p_control_array, $p_control_name, $p_match = '', $
 /**
  * Prints the list of visible attachments belonging to a given bug.
  * @param integer $p_bug_id ID of the bug to print attachments list for.
+ * @param string $p_security_token Optional; null (default) or security token string.
```
atrol: it's not optional and has no default value

core/print_api.php
```diff
@@ -1906,23 +1907,20 @@ function print_bug_attachments_list( $p_bug_id ) {
  * Use this to avoid performance issues when loading pages having many calls to
  * this function, such as print_bug_attachments_list().
  * @param array $p_attachment An attachment array from within the array returned by the file_get_visible_attachments() function.
- * @param mixed  $p_security_token Optional; null (default) or security token string.
- * @see form_security_token()
+ * @param string $p_security_token Optional; null (default) or security token string.
```
atrol: it's no longer optional and has no default value

core/print_api.php
```diff
@@ -1948,11 +1946,10 @@ function print_bug_attachment( array $p_attachment, $p_security_token = null ) {
...
```
atrol: same as above

core/bug_activity_api.php
vboctor: Changed to BugActivityAPI

core/bug_activity_api.php
vboctor: Good catch.

vboctor:
@atrol I've applied the review comments mentioned in the inline comments.

> You could completely remove the "Upload Files" box if it would be allowed to add empty notes.

I've removed the box to attach files to issues and enabled submission of notes that have note + files, note only or files only.  A completely empty form will still trigger an error.  That will also encourage associating files with notes and simplifies the page.

> bugnote ordering

This was implemented, but it was keying off the bugnote_order config option instead of user preference.  This is now fixed.

I've also changed "Notes Sort Order" to "Activities Sort Order" in the preferences page.

**Other changes I've done**

- Disabled attachment control when the note is marked as private.  This is to avoid accidental disclosure.
- Rebased on master

**Status**

I think this PR is ready to checkin.  For deleting notes and attachments, the behavior is as follows:

- Specific attachments can be deleted whether they are standalone or part of a note.
- Deleting a note will just delete the note and will unfold the attachments.  We can revisit this, but for this PR will take the conservative approach.

@atrol @dregad @cproensa let me know if you have any issues that need to be addresses before PR is merged.

atrol:
print_bug_page.php should also show attachments inline with notes.

atrol:
Links in column attachment_count of "View Issues" page will no longer work, e.g.
https://www.mantisbt.org/bugs/view.php?id=21731#attachments

vboctor:
Thanks @atrol - good catches.

> print_bug_page.php should also show attachments inline with notes.

I've got rid of print page and added css media styles to tweak View Issue page to be print friendly.  This way we have a single page to maintain.

> Links in column attachment_count of "View Issues" page will no longer work

I've added anchor tag for attachments.  So now both attachments and bugnotes will link to the right place.  I was going to use activities anchor instead, but I reverted that and kept both.  This way we have an intention from incoming links enabling more flexibility in the future if we decide to evolve the layout.

vboctor:
@atrol @dregad @cproensa planning to merge this tomorrow.

cproensa:
> I've got rid of print page and added css media styles to tweak View Issue page to be print friendly. This way we have a single page to maintain.

ok, better on the long run, but the styles seems to need some additional tweaking

some issues, improvements that could be made:
- Remove username, last visited issues, etc info from print page. They are rendered skewed, and may not be actually needed for print.
- Table cells (eg: description) is displaced to next page if vertical text span overflows available space for current page
- The default text sizes for page print look too big.


seleccion_060

vboctor:
@cproensa I couldn't reproduce the description issue and the issue ids.  However, I've hidden the username / navbar / breadcrumbs explicitly.

cproensa:
>  I couldn't reproduce the description issue and the issue ids

Probably because the template show/hide them based on page width.

core/layout_api.php
```diff
@@ -996,7 +996,11 @@ function layout_page_content_end() {
  * @return null
  */
 function layout_breadcrumbs() {
-	echo '<div id="breadcrumbs" class="breadcrumbs">' , "\n";
+	$t_username = user_get_name( auth_get_current_user_id() );
+	$t_protected = current_user_get_field( 'protected' );
+	$t_access_level = get_enum_element( 'access_levels', current_user_get_access_level() );
```
atrol: I (re)moved those three lines in https://github.com/mantisbt/mantisbt/commit/46fa6283d7364a1acd8b0176e127e245a8c2e374
Is there a reason you reintroduced them?

vboctor: This is a merge error, feel free to fix.

PR #910, opened 2016-10-03
Display realname in breadcrumb div

atrol:
Reintroduce behaviour of versions <= 1.3.x
Display realname independent from $g_show_realname

Rearranged code for small performance enhancements for anonymous users

Fixes #21736


atrol:
I didn't change line 1015 as it is < 120 chars.
There are some lines in this module that are a lot longer.
Not sure if we should use a tool to correct the whole source.

PR #911, opened 2016-10-03
Allow users to remove real name from their account

atrol:
Fixes #21737

### DIFF
--- a/account_update.php
+++ b/account_update.php
@@ -83,10 +83,9 @@ $f_password_confirm	= gpc_get_string( 'password_confirm', '' );
 
 $t_redirect_url = 'index.php';
 
-# @todo Listing what fields were updated is not standard behaviour of MantisBT - it also complicates the code.
-$t_update_email = null;
-$t_update_password = null;
-$t_update_realname = null;
+$t_update_email = false;
+$t_update_password = false;
+$t_update_realname = false;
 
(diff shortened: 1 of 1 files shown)


atrol:
I will merge in two days if no one objects.

account_update.php
```diff
@@ -100,8 +99,8 @@
 # Do not update email for a user verification
 if( !( $t_ldap && config_get( 'use_ldap_email' ) )
 	&& !$t_account_verification ) {
-	if( $f_email != user_get_email( $t_user_id ) ) {
-		$t_update_email = $f_email;
+	if( !is_blank( $f_email ) && $f_email != user_get_email( $t_user_id ) ) {
```
vboctor: Why not just have a direct equality check?  Why did you add the blank check?

atrol: Have a look at the discussion at https://www.mantisbt.org/bugs/view.php?id=21737
This is to prevent users from removing their email address.

atrol: This PR is just dealing with the real name and does not change the behavior when changing email address.
Of course, I changed the way it's implemented as I want it the same for all fields.
(false/true instead of null/input value)
So your comments should not block this PR.
You might want to open a new issue to allow users to remove their email address.

vboctor: 👍

PR #912, opened 2016-10-04
Initial markdown support

jllano:
https://www.mantisbt.org/bugs/view.php?id=17920


jllano:
screen shot 2016-10-04 at 11 43 02 am

vboctor:
@jllano A couple of comments:

- One of these commits is unrelated to the change.
- I would rather have my commits rebased on top of latest from master, but not squashed into a single commit.  Then you can add your commits for changes necessary to finish off the work.

I'll do some testing and provide feedback.

vboctor:
@jllano here are some testing results:

- If a line starts with # and issue id, the line is treated as a header instead of an issue reference and the # is omitted form the output.
- I can't seem to add a image reference.
- The styling for elements like table and quoted text seems to assume a white background, and doesn't work as well with gray or yellow backgrounds.  Do we have control on the style?

Code review feedback (high level):
- Add phpdoc to core/markdown_api.php
- Add new config option to manual.

vboctor:
This was closed in favor of #930

PR #913, opened 2016-10-07
Fix documentation of configuration files

atrol:
Fixes #21757


vboctor:
Note that config folder path is by default sub-folder, but it is configurable.  Hence, we shouldn't refer to it as subfolder.  We may also want to mention the ability to configure the folder location.

PR #916, opened 2016-10-14
Remove ‘bug_print_page_fields’ config option

vboctor:
Remove the unused configuration option and mark it obsolete.


atrol:
👍 in general, add `Issue #21727` to the commit message or maybe create another issue for it.

dregad:
> add Issue #21727 to the commit message 

https://www.mantisbt.org/bugs/view.php?id=21727 seems unrelated.

atrol:
> https://www.mantisbt.org/bugs/view.php?id=21727 seems unrelated.

print_bug_page.php has been removed when implementing #21727.
That's why bug_print_page_fields can be removed.

PR #917, opened 2016-10-15
Don't show reset option for disabled or protected users

cproensa:
Don't show reset option for disabled or protected users
    
    Don't show the password reset option for disabled or protected users:
    - Disabled users can't have email sent.
    - Protected users must not have its password changed (and will show an
      error anyway)
    
    Fixes: #21793

Don't send reset email for disabled users
    
    Don't send email for password reset if the user is disabled.
    Adds log messages to several failure conditions.

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -538,11 +538,18 @@ function email_signup( $p_user_id, $p_confirm_hash, $p_admin_name = '' ) {
  * @return void
  */
 function email_send_confirm_hash_url( $p_user_id, $p_confirm_hash ) {
-	if( OFF == config_get( 'send_reset_password' ) ||
-		OFF == config_get( 'enable_email_notification' ) ) {
+	if( OFF == config_get( 'send_reset_password' ) ) {
+		log_event( LOG_EMAIL_VERBOSE, 'Password reset email notifications disabled.' );
 		return;
 	}
-
+	if( OFF == config_get( 'enable_email_notification' ) ) {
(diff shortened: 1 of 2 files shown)


atrol:
👍, didn't test

core/email_api.php
```diff
@@ -538,11 +538,18 @@ function email_signup( $p_user_id, $p_confirm_hash, $p_admin_name = '' ) {
  * @return void
  */
 function email_send_confirm_hash_url( $p_user_id, $p_confirm_hash ) {
-	if( OFF == config_get( 'send_reset_password' ) ||
-		OFF == config_get( 'enable_email_notification' ) ) {
+	if( OFF == config_get( 'send_reset_password' ) ) {
+		log_event( LOG_EMAIL_VERBOSE, 'Password reset email notifications disabled.' );
```
vboctor: I suggest using LOG_EMAIL for all.  In this case we have a single email per action, so not sure we need to use LOG_EMAIL_VERBOSE or LOG_EMAIL_RECIPIENT.

cproensa: LOG_EMAIL_VERBOSE used like in this case:
https://github.com/mantisbt/mantisbt/blob/master/core/email_api.php#L1338

For LOG_EMAIL_RECIPIENT, can be changed to LOG_EMAIL, if you think it fits better.

vboctor: ok, let's use LOG_EMAIL instead of LOG_EMAIL_RECIPIENT.

PR #918, opened 2016-10-16
Prevent db errors for invalid bug/bugnote ids

cproensa:
Some db engines may throw an error if provided with an integer out of
the supported range for field type. As an example: pgsql.
This can happen when the user provides a number high enough as input in
several places. For example:
- Bug and bugnote linking, in texts that are processed by core formatting
  plugin
- Bug jump quick access
- Bug id provided for a relation

The core formatting case is especially bad, because the bug page cannot
be rendered, always ending in a db error.

The fix is applied in bug_exists(), and bugnote_exists() functions,
which should fix most of said situations all through the applcation.

Fixes: #21802

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -985,10 +985,18 @@ function bug_text_clear_cache( $p_bug_id = null ) {
  * @access public
  */
 function bug_exists( $p_bug_id ) {
-	if( false == bug_cache_row( $p_bug_id, false ) ) {
+	$c_bug_id = (int)$p_bug_id;
+
+	# Check for invalid id values
+	if( $c_bug_id <= 0 || $c_bug_id > DB_MAX_INT ) {
 		return false;
-	} else {
+	}
+
+	# bug exists if bug_cache_row returns any value
+	if( bug_cache_row( $c_bug_id, false ) ) {
 		return true;
+	} else {
+		return false;
 	}
 }
 
--- a/core/string_api.php
+++ b/core/string_api.php
@@ -347,13 +347,14 @@ function string_process_bug_link( $p_string, $p_include_anchor = true, $p_detail
 		if( $p_include_anchor ) {
 			$s_bug_link_callback[$p_include_anchor][$p_detail_info][$p_fqdn] =
 				function( $p_array ) use( $p_detail_info, $p_fqdn ) {
-					if( bug_exists( (int)$p_array[2] ) ) {
-						$t_project_id = bug_get_field( (int)$p_array[2], 'project_id' );
+					$c_bug_id = (int)$p_array[2];
+					if( bug_exists( $c_bug_id ) ) {
+						$t_project_id = bug_get_field( $c_bug_id, 'project_id' );
 						$t_view_bug_threshold = config_get( 'view_bug_threshold', null, null, $t_project_id );
-						if( access_has_bug_level( $t_view_bug_threshold, (int)$p_array[2] ) ) {
+						if( access_has_bug_level( $t_view_bug_threshold, $c_bug_id ) ) {
 							return $p_array[1] .
 								string_get_bug_view_link(
-									(int)$p_array[2],
+									$c_bug_id,
 									(boolean)$p_detail_info,
 									(boolean)$p_fqdn
 								);
@@ -364,10 +365,11 @@ function string_process_bug_link( $p_string, $p_include_anchor = true, $p_detail
 		} else {
 			$s_bug_link_callback[$p_include_anchor][$p_detail_info][$p_fqdn] =
 				function( $p_array ) {
-					if( bug_exists( (int)$p_array[2] ) ) {
+					$c_bug_id = (int)$p_array[2];
+					if( bug_exists( $c_bug_id ) ) {
 						# Create link regardless of user's access to the bug
 						return $p_array[1] .
-							string_get_bug_view_url_with_fqdn( (int)$p_array[2] );
+							string_get_bug_view_url_with_fqdn( $c_bug_id );
 					}
 					return $p_array[0];
 				}; # end of bug link callback closure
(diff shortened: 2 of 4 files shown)


core/bug_api.php
```diff
@@ -985,7 +985,14 @@ function bug_text_clear_cache( $p_bug_id = null ) {
...
 function bug_exists( $p_bug_id ) {
-	if( false == bug_cache_row( $p_bug_id, false ) ) {
+	$c_bug_id = (int)$p_bug_id;
+
+	# Prevent db query errors
+	if( $c_bug_id > db_max_uint() ) {
+		return false;
+	}
+
+	if( false == bug_cache_row( $c_bug_id, false ) ) {
```
vboctor: Should this use === ?

core/string_api.php
```diff
@@ -445,13 +448,15 @@ function( $p_array ) use( $p_detail_info, $p_fqdn ) {
...
-						return $p_array[0];
+					$c_bugnote_id = (int)$p_array[2];
+					if( bugnote_exists( $c_bugnote_id ) ) {
+						$t_bug_id = bugnote_get_field( $c_bugnote_id, 'bug_id' );
+						if( $t_bug_id && bug_exists( $t_bug_id ) ) {
+							return $p_array[1] .
+								string_get_bugnote_view_url_with_fqdn( $t_bug_id, $c_bugnote_id );
+						}
 					}
+					return $p_array[0];
```
vboctor: There seems to be a change of logic of when we return $p_array[0].  Are you sure this doesn't break the logic?

core/bug_api.php
cproensa: Not necessarily.
== will match for equivalent values like 0, '', null... etc, Even if the function returned any of those, the condition should be still be matched, since no real bug data was found.
However, "false ==" is misleading, as "false" is not the only value we would want to match.
In this case i prefer the syntax (here inversed, though)

    if( bug_cache_row( $c_bug_id, false ) )

core/string_api.php
cproensa: Moving return $p_array[0] outside of else branches is valid, since this should be the default (unmodified) string. Other positive conditions finalize with its own "return"

But i added a check for bugnote_exists that wasn't there in the first place. I will try to verify that i can find a test case for that code path (not sure right now if i tested it explicitly)

cproensa: ok, tested this case. The construction is functionally equivalent.
by doing:
`$t_bug_id = bugnote_get_field( $c_bugnote_id, 'bug_id' );`
and then 
`if( $t_bug_id ... `
it's implicitly checking the bugnote exists.
I added the explicit check to make it clearer

core/bug_api.php
```diff
@@ -985,10 +985,18 @@ function bug_text_clear_cache( $p_bug_id = null ) {
  * @access public
  */
 function bug_exists( $p_bug_id ) {
-	if( false == bug_cache_row( $p_bug_id, false ) ) {
+	$c_bug_id = (int)$p_bug_id;
+
+	# Prevent db query errors
+	if( $c_bug_id > db_max_uint() ) {
```
atrol: Why didn't you implement the logic in bug_cache_row? (near to the database code)

core/database_api.php
```diff
@@ -1323,3 +1323,23 @@ function db_mysql_fix_utf8( $p_string ) {
...
+ * Adodb maps "I" as "I4". The actual types used for each driver are mapped in
+ * ADODB_DataDict objects, function ActualType.
+ * @return int
+ */
+function db_max_uint() {
+	# Safe assumption is for maximum value of 4 bytes signed integer: 2^31-1
+	# This is the limit for databases that dont support unsigned types, eg pgsl, mssql
+	$t_max = 2147483647;
+
+	if( db_is_mysql() ) {
```
atrol: I would not distinguish between databases
- id's > 2147483647 is not a real use case
- better performance

I would just use a constant for it.

core/bug_api.php
cproensa: I thought of that. My reasoning is:
bug_cache_row doesnt perform input validation. The main use for that function is for known bug ids, and adding additional checks would add logic unneeded (by this premise). 
Note that this function expects an integer (i'd add, expects an existing bug id, within reasonable circumstances)), due to php loose types, that cannot be enforced by declaration.
 
bug_exists() is closer to the input layer, where validation should be performed, and is a good compromise for those pages that take arbitrary input. For example: view page relies on ensure_bug_exists for check the inputted bug_id, which then calls to bug_exists.

Notice how in bug_exists(), i added a cast to (int) for $p_bug_id, so if a garbage string is passed in, at least it's casted to 0 and returns no rows.

core/database_api.php
cproensa: > id's > 2147483647 is not a real use case 

good enough, if we take that compromise, that can be a universal limit to check against.
If we rely in Adodb mapping to INT4 types, worst case is that the DB does not provide unsigned types, so 31bits is a common upper limit.

side-note: 2038 date limit?


> I would just use a constant for it.

probably better for performance. 
Show me one more up for this and i will change to a constant, @vboctor, @dregad  ?

vboctor: I'm OK with using a constant, but in this case I would have it defined in constants_inc.php and use it directly in bug_exists() and bugnote_exists() since it is no longer db dependent. 

You may also want to check that the ids are > 0.  This may not be triggered in this scenario due to regular expression filtering #-1234 out, but it can come from other calls to these same functions.

okdana: I'm not familiar with ADOdb, but doesn't it seem strange, just conceptually, to have a function called `db_max_uint()` that actually returns the max *signed* integer in at least two cases?

cproensa: The reasoning for calling it db_max_uint() was:
The biggest number we can use for the type that WE have defined as "unsigned int"
What whe declare as "I UNSIGNED" is mapped to the closest type existing in database. In the case of pgsql, this is a "4 byte signed int"

Anyway, as per the suggestions, the function will be converted to a constant, which is good enough for our purpose. Calling it still DB_MAX_UINT?

okdana: `DB_MAX_INT` makes sense to me, but i guess i can see why you'd call it `UINT` too

cproensa:
Updated with a rewrite, with the discussed suggestions (using a constant)

@vboctor 
> You may also want to check that the ids are > 0.

Are we sure that ids can't be 0, as part of an auto-sequence?

okdana:
I'm not sure about MySQL/MSSQL, but i know that in Postgres, sequence values can be 0 or even negative — but *by default* the minimum value is 1, so that would never happen unless you explicitly asked for it. SQLite (which i know you don't support, but just for reference) is the same way — a `ROWID` *can* be <1, but only if you deliberately set it that way.

cproensa:
Probably, some stuff may not work properly in mantis if ids<1 get to happen anyway...

PR #919, opened 2016-10-17
Relocate inline JS code in graph plugin to separate file

syncguru:
Fixes #21669

The approach adapted for this fix is to use html data attributes  .. for more details, see https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_data_attributes


plugins/MantisGraph/MantisGraph.php
```diff
@@ -113,6 +113,7 @@ function resources() {
 			html_javascript_link( 'chart-' . CHARTJS_VERSION . '.min.js' );
 			html_javascript_link( 'chart.bundle-' . CHARTJS_VERSION . '.min.js' );
 		}
+		html_javascript_link( '../plugins/MantisGraph/MantisGraph.js' );
```
vboctor: Move the MantisGraph.js file into plugins/MantisGraph/files and include it by getting its location using plugin_file( 'MantisGraph.js' ).  See Snippets plugin for an example.

vboctor:
There seems to be some javascript errors on the login page:

javascript_error_on_login_page

plugins/MantisGraph/MantisGraph.php
cproensa: Should chart js libs, ("chart-xxxx.min.js", etc) also be provided by the plugin?

cproensa:
> There seems to be some javascript errors on the login page

i noticed those when doing the work for verify page.. (but forgot to follow on them)
I don't think they are directly related to this PR

plugins/MantisGraph/MantisGraph.php
vboctor: Currently chart-xxxx.min.js files are in the js folder for core.  They should be moved into plugins/MantisGraph/files as well.

cproensa: > We could use chart-xxxx.min.js for some dashboard functionality (issue trend, ...) on "My View" page. So maybe better not to move.

That was my question about. I don't know if they are used system-wide.
In that case, the inclusion of those js, should be done by core, not by the plugin?

dregad: Unless they are actually used somewhere else, I would rather move all js files to the plugin's file directory.

We can revisit this if we realize later on that some of these scripts are used elsewhere.

One could also argue that a dashboard should be a part of the Graph plugin itself.

dregad:
>There seems to be some javascript errors on the login page

This is addressed in PR #922 (see [#21805](https://www.mantisbt.org/bugs/view.php?id=21805))

syncguru:
@vboctor, @cproensa, @dregad 
+ Moved JS into plugin files directory
+ Removed CSP cdoe

PR #921, opened 2016-10-19
Fix missing dropzone when view state is hidden

vboctor:
Attachment dropzone missing from notes when user doesn't have access to set view state

Fixes #21806

### DIFF
--- a/bugnote_add_inc.php
+++ b/bugnote_add_inc.php
@@ -87,6 +87,7 @@ require_api( 'lang_api.php' );
 		<tbody>
 
 <?php
+	$t_default_bugnote_view_status = config_get( 'default_bugnote_view_status' );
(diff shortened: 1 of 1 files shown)


dregad:
I guess that's a separate issue, but the dropzone is also hidden (i.e. it is not possible to add attachments) when view status is private.

vboctor:
@dregad I've removed ability to attach files as part of private notes by design as part of 53b8da3960ad61a3ecbbfa3ec59c54dcebe77519.  The reason is that the user would assume that both the note and attachment be protected as private, but as per our current capabilities, we only support private notes.  Once we support private attachments, we should re-enable attaching files with private notes.

dregad:
That makes sense, I should have thought about that. Thanks for the clarification.

PR #923, opened 2016-10-20
Relocate dropzone inline js to separate file

syncguru:
Fixes #21651


core/http_api.php
```diff
@@ -233,13 +233,6 @@ function http_security_headers() {
 			http_csp_add( 'img-src', 'ajax.googleapis.com' );
 		}
 
-		# Relaxing policy for view issue page to allow inline scripts.
-		# Should be removed once #21651 is fixed.
-		if( 'view.php' == basename( $_SERVER['SCRIPT_NAME'] ) ||
-			'bug_report_page.php' == basename( $_SERVER['SCRIPT_NAME'] ) ) {
```
vboctor: Do we still need need bug_report_page.php for the case when due_date is enabled and showing the jscalendar control?  If so, then we should only skip if that is the page and due date is enabled.  Hopefully we can remove it completely once we remove that.  I can be wrong, but something to test.

vboctor:
Closing in favor of #931

PR #926, opened 2016-10-20
Add compatibility for v1.3 layout and css

cproensa:
Include layout functions from v1.3 that were removed in v2:
html_page_top()
html_page_bottom()
So that plugins pages can work without errors.

Add compatibility styles for some v1.3 css styles . These are loaded
once that html_page_top() is used.

Fixes: #21706


syncguru:
I thought the reason we are labeling modern UI as v2.0 is to make breaking changes (otherwise we should call it v1.4)
With this change, there is no incentive for plugin authors (users) to upgrade to (push for) the new look and feel. The end result is that we are going to end up with a hybrid solution. Parts of it looks modern while others are classic. I cannot think of worse UI & UX. I would rather have people stick with v1.3 instead and this mixed "app".

-1

cproensa:
This is a plugin page in modern-ui:
seleccion_164

This is when there is a compatible html_page_top() function:
seleccion_166

This is once the compatibility css styles are present:
seleccion_167


Note that the way it is implemented:
1) only when the old html_page_top() is called, is the css inserted
2) css classes are prefixed with m13 container class path to not conflict with the app styles
3) html_page_top inserts a div with class "m13" to allow the legacy styles only in pages that use it.

If you feel more comfortable, we can make html_page_top() raise an error, unless a flag "DEBUG_M13_COMPAT" is present.

syncguru:
@cproensa 
1- That snapshot is quite ugly
2- I am happy to help port this plugin to Modern UI.

cproensa:
Yes, that is ugly table styile as is.
I am not planning porting it right now, not until this mantis instance starts using v2. This is a lot of work to mantain 2 branches atm (this is >10kloc)

With this i am trying to get my first impressions of what a plugin developer may experience when confronting the new version.

This screenshots are a more proper representation of plugin page (without/with css)

seleccion_067

seleccion_066

vboctor:
I would rather if we keep the core focused on modern UI and encouraging most / all plugins to move to the new modern UI rather than just making classic UIs within the modern UI.  If you think for a subset of plugins such compatibility is needed, then I suggest creating classic-ui-compat plugin and have such plugins take a dependency on it.

As for our official bug tracker, I'm expecting we will move it to the modern UI very soon.

plugins/M13PluginCompat/M13PluginCompat.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * MantisBT is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @copyright Copyright 2002  MantisBT Team - mantisbt-dev@lists.sourceforge.net
```
dregad: 2016

plugins/M13PluginCompat/M13PluginCompat.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+class M13PluginCompatPlugin extends MantisPlugin {
+
+	function register() {
+		$this->name = 'Mantis 1.3 plugin pages comptaibility';
```
dregad: *compatibility*

plugins/M13PluginCompat/core/html_compat_api.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * MantisBT is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @copyright Copyright 2002  MantisBT Team - mantisbt-dev@lists.sourceforge.net
```
dregad: 2016

vboctor:
My suggestion about using a plugin was that it will be part of mantisbt-plugins rather than distributed with MantisBT itself given that it is a backwards looking plugin.  Plugins that don't want to provide a native update, can have installing this plugin as part of their install instructions.  But hopefully most of plugins will support the modern UI specially as 2.0.0 (stable) goes out.

PR #927, opened 2016-10-20
Revert breaking changes of print_button()

cproensa:
Bring back the old print_button() function declaration for API compatibility

See https://www.mantisbt.org/bugs/view.php?id=21815

### DIFF
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -2134,5 +2134,7 @@ function print_dropzone_form_data() {
  * @return void
  */
 function print_button( $p_action_page, $p_label, array $p_args_to_post = null, $p_security_token = null ) {
+	error_parameters( __FUNCTION__, 'print_form_button' );
+	trigger_error( ERROR_DEPRECATED_SUPERSEDED, DEPRECATED );
 	print_form_button( $p_action_page, $p_label, $p_args_to_post, $p_security_token );
 }
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -171,7 +171,7 @@ function print_successful_redirect( $p_redirect_to ) {
 		layout_page_begin();
 		echo '<br /><div class="center">';
 		echo lang_get( 'operation_successful' ) . '<br />';
-		print_button( $p_redirect_to, lang_get( 'proceed' ) );
+		print_link_button( $p_redirect_to, lang_get( 'proceed' ) );
 		echo '</div>';
 		layout_page_end();
 	} else {
@@ -1511,7 +1511,7 @@ function print_link( $p_link, $p_url_text, $p_new_window = false, $p_class = '' 
  * @param string  $p_class      The CSS class of the link.
  * @return void
  */
-function print_button( $p_link, $p_url_text, $p_class = '', $p_new_window = false ) {
+function print_link_button( $p_link, $p_url_text, $p_class = '', $p_new_window = false ) {
 	if( is_blank( $p_link ) ) {
 		echo $p_url_text;
 	} else {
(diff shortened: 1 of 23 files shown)


cproensa:
As described in the issue, changing this function resulted in an unnecessary breaking change.
In the other hand, the UI refactor did not account, for example, that html_button and print_button were almost identical methods.
Take in mind, these are core api functions, and very generic. If we encourage plugin developers to make use of our api to improve integration with the standards, then we have to be careful to not break compatibility without a strong reason.

@syncguru do you know if there are more cases like this?

cproensa:
The difference between old print_button and new print_form_button is just adding some css classes
That is not a braking change, imho.
Creating a new function, with different parameters input, collisioning with previous function name, is a breaking change indeed.

vboctor:
If we are to take in this change, then I would rather have print_link_button() be a new method that is added that wraps print_button().  That would be a much smaller change and would make it easier to just remove this legacy function when we no longer need it.  Will discuss on the other PRs the concept of backward compatibility and whether it should be done at all, done as a plugun, or done as part of core.

cproensa:
> If we are to take in this change, then I would rather have print_link_button() be a new method that is added that wraps print_button().
They cant, one prints a form with a button input in it. The new one prints a link without form, but styled like a button.

What i have done here is:
- print_link_button() is the new funcion introduced by v2
- rename all calls to the new function.
- make print_button() as a wrap to print_form_button()

This is a cleaner change in my opinion, and the way it should have been done in the first place at the rewrite.

Changing the functional and definiton of a function is more dangerous than removing it at all, as this now fails silently, and it's difficult to detect.


>Will discuss on the other PRs the concept of backward compatibility and whether it should be done at all,  
done as a plugun, or done as part of core.
Please, discuss that here:  https://www.mantisbt.org/bugs/view.php?id=21706

core/print_api.php
```diff
@@ -2095,3 +2095,19 @@ function print_max_filesize( $p_size, $p_divider = 1000, $p_unit = 'kb' ) {
...
+ * @param string $p_action_page    The action page.
+ * @param string $p_label          The button label.
+ * @param array  $p_args_to_post   Associative array of arguments to be posted
+ * @param mixed  $p_security_token Optional; null (default), OFF or security token string.
+ * @see form_security_token()
+ * @see print_form_button()
+ * @return void
+ */
+function print_button( $p_action_page, $p_label, array $p_args_to_post = null, $p_security_token = null ) {
+	print_form_button( $p_action_page, $p_label, $p_args_to_post, $p_security_token );
```
dregad: Since the function is deprectated, shouldn't it trigger a E_USER_DEPRECATED error ?

cproensa: right.
i will update with that

vboctor:
Doesn't this change break changes to plugins that have been revised to work with 2.0.x releases that use the new API?  We have already acknowledged that there are several changes between 1.3.x and 2.0.x that are breaking.  If so, then plugins that are yet to be ported should adopt the new model or utilize a new method that you can add as part of this PR.   I still disagree with this change.

cproensa:
@vboctor 
> Doesn't this change break plugins that have been revised to work with 2.0.x releases that use the new API?

If this was the situation, they probably changed the code to call print_form_button(), which would still work.
This was proposed before the RC release, change is expected to happen in beta.

>  If so, then plugins that are yet to be ported should adopt the new model 
> We have already acknowledged that there are several changes between 1.3.x and 2.0.x that are breaking

I hope it is well documented, because as i said, this function will fail silently, as it has changed what it does, and the parameters it takes.

vboctor:
I expect that the plugins will have to be tested with 2.0, and even if things slip through they will be found and fixed.  We don't have this level of documentation, but it is something to keep in mind when doing future changes.

cproensa:
I have renamed this PR to more apropiately describe its purpose

I'd like to hear more opinions. @atrol ?

atrol:
I have no strong opinion on that as I am not affected.
Of course, it's always good to be backward compatible as long as there is no good reason to break compatibility.

dregad:
At the moment we have 2 votes for, and 1 against + 1 null.
Unless other arguments/votes are made against the proposal, I'll merge this change by end of week, in time for the release.

dregad:
Merged in fb568b4388d0ce3c7546a1384b084523e422e313

PR #929, opened 2016-10-22
Change view issue menu to allow plugin links

cproensa:
Changed the style of the upper menu in bug view page, to allow the
correct formatting of links provided by plugins with EVENT_MENU_ISSUE.

This event may return a prepared html link, which was being shown as a
bracketed plain link (old style)
The new style introduced by new ui, used link buttons, but the function
provided to print abutton from the event return (print_small_button),
must provide a separated urn and label, which is not possible with an
already prepared html link.

The changed style now uses list items that formats the link html "as
is" in an uniform way with the ui.

Fixes: #21559

### DIFF
--- a/bug_view_inc.php
+++ b/bug_view_inc.php
@@ -268,7 +268,7 @@ foreach ( $t_links as $t_plugin => $t_hooks ) {
 				if( is_numeric( $t_label ) ) {
 					echo '<li>' .  $t_href . '</li>';
 				} else {
-					print_small_button( $t_href, $t_label );
+					echo '<li><a href="' . $t_href . '">' . $t_label . '</a></li>';
 				}
 			}
 		} elseif( !empty( $t_hook ) ) {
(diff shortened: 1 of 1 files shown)


cproensa:
before:
seleccion_073

after:
seleccion_074

vboctor:
Haven't reviewed the change yet.  But the new buttons look disabled?

cproensa:
updated with one case i forgot to change.

>  But the new buttons look disabled?
These are the list styles for manage navigation bar.

The substitution for buttons was made as part of the rewrite, and it can be refined.
I think semantically these links are better suited as list items, and it solves the issues mentioned with the plugin event links, thats a fine combo.

@syncguru could you provide a more stylish class for LI items to be used in this situation?

vboctor:
For plugins, as part of changes done to modern UI we no longer want to accept pre-formatted links.  The plugins provide an associative arrive that provides information like:

- URL
- Label
- Icon (based on font-awesome)

Then the core code does the right thing in terms of formatting to match the core UI.  It is a change from what we had for 1.3 plugins, but it provides more power and flexibility going forward.

cproensa:
I disagree that the UI core should format these as buttons.
They are links, and are more properly rendered as list items. 

>  It is a change from what we had for 1.3 plugins

Then it's not fully implemented, still, as you can see in https://www.mantisbt.org/bugs/view.php?id=21559
the worng format is still being printed

PR #930, opened 2016-10-22
Markdown support

jllano:
This enable markdown support for MantisBT

This would allow users to be able to use markdown when populating issue and notes information. Users could also reference markdown inline images from the internet https://github.com/mantisbt/mantisbt/pull/930/commits/5273330c9c59a62fccb6b0a12a57a9cbfa04f7e0.

Markdown format: http://parsedown.org/tests/ 

Initially, @vboctor added a Parsedown library https://github.com/erusev/parsedown, he enabled the library by adding markdown_api methods and integrated in plugins/MantisCoreFormatting/MantisCoreFormatting.php https://github.com/mantisbt/mantisbt/commit/9e5f3fec6924a6fcd8ff2a7fff08a3267e1fbc29.

I've also added an MantisMarkdown Extension (core/classes/MantisMarkdown.php) class that would overrides default Markdown parsing for Mantis specific scenarios such as Headers, Tables and Quote so far.

For example: "#2" is treated as header markdown, so to make sure "#2" to be treated as bug link (not a header), then we have to change the logic and that whats the MantisMarkdown Extension serves.

I've also added some unit tests for markdown conversion.

Fixes: #17920

Screenshots when markdown is enabled in config [$g_markdown_enabled = ON;]:
screen shot 2016-10-19 at 3 09 07 pm
screen shot 2016-10-19 at 3 09 16 pm
screen shot 2016-10-19 at 3 09 27 pm
screen shot 2016-10-19 at 10 22 05 pm
screen shot 2016-10-19 at 10 42 15 pm
screen shot 2016-10-20 at 9 19 11 pm
screen shot 2016-10-20 at 9 23 58 pm


cproensa:
@vboctor 
can't be this implemented as a plugin?
we already have enough cross-dependency between core->formatterplugin->core apis

as i see on a high level:
markdown settings, logic and libraries, should be bundled in a "markdown formatting plugin"
as an additional improvement: current html output processing should be also included in a similar scheme

are html and markdown compatible? should they be?

core/markdown_api.php
```diff
@@ -0,0 +1,106 @@
...
+ * instead of an issue reference and the # is omitted form the output)
+ *
+ * @return void
+ */
+function markdown_init() {
+    global $g_parsedown;
+
+    if ( $g_parsedown == null ) {
+
+        require_once( dirname( dirname( __FILE__ ) ) . '/core/classes/MantisMarkdown.php' );
```
cproensa: this should be dealt with class loader

core/markdown_api.php
```diff
@@ -0,0 +1,106 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Markdown API
+ *
+ * @package CoreAPI
+ * @subpackage MarkdownAPI
```
cproensa: This api is unncessary as it is now
If you need "firendly" methods and need to track a sigle object instance, then add static methods
MantisMarkdown::markdown_text
MantisMarkdown::markdown_line

There you can deal with the single instance internally to the class, no need for the external code to keep track of a global variable and markdown_init()

cproensa:
Regarding bug linking.

The markdown equivalent to preformatted text, should not translate # and ~ to links, and output them verbatim

config_defaults_inc.php
```diff
@@ -1214,6 +1214,16 @@
...
 
+############
+# Markdown #
+############
+
+/**
+ * Enables markdown support - once enabled all issue titles, text, and notes are formatted based on
+ * markdown syntax.  Default to OFF since it is experimental feature for now.
+ */
+$g_markdown_enabled = OFF;
```
cproensa: Regarding the application global setting of markdown, i suggest that this is icluded within the coreFormatting plugin
Mantis core will not call the markdows formatter directly, but through the coreFormatting as a proxy

cproensa: >  this configuration is no different than others like bug_link_tag and bugnote_link_tag.

Thats what i mean. See:
seleccion_075

Having a global config when this can be done cleanly in the already dedicated functionality, seems counterintuitive to me

core/markdown_api.php
jllano: I also agree of removing the api, it looks an old school approach. 

We could not make the methods static (MantisMarkdown::markdown_text or MantisMarkdown::text()) and so on...as its parent library is non-static. 

I suggest to make the api as a service container instead:

```
class MantisMarkdownApi {
    
    function __construct(){
        load the MantisMarkdown class
    }
    static function text($text){
        return $MantisMarkdown->text($text);    
    }
    
    static function line($text){
        return $MantisMarkdown->line($text);    
    }
}
```

Let me know your thoughts guys?

cproensa:
Thanks @jllano 
with the new commits i think it's getting to more cleaner integration

I still wonder if the Markdown class should be published by mantis core (as it is now) or by the plugin.
At the moment, the only usage from core seems to be: MantisMarkdown::enabled(), only to add csp headers (wasn't that doable by plugins directly, like Gravatar?)

Probably could be revisited later, but my ideal here is that Markdown is a plugin in itself, and is used or not, according to the formatting settings. Mantis core should pass text to the formatting plugin and it returns them as the processed output.
Anyway, some questions that probably should be asked to get the scenario clear:
- is markdown goint to be used directly by core (bypassing the formatter sytem/plugin)?
- if yes (which implies publishing permanently in core), how would another possible markup language ciould be implemented? Allowing integration purely from a plugin allows any implementation to be done by community, otherwise, core updating is needed.
- What is actually the intention of this Markdown integration? into core as the new unique formatting standard? or as a first step into a pluggable markup module system?

Maybe i misunderstood which were the assumed requeriments. In regards to integration method, that was not cleared in the related issue 17920.

vboctor:
@cproensa I've replied to your comment in the MantisBT issue.  @jllano is working on making markdown a core plugin.  So more commits to come.

config_defaults_inc.php
vboctor: @jllano After moving markdown to a plugin do we still need the $g_markdown_enabled config?  I was expecting it will be dropped in favor of installing / uninstalling the plugin.

core.php
```diff
@@ -274,6 +274,7 @@ function __autoload( $p_class ) {
 
 # Set HTTP response headers
 require_api( 'http_api.php' );
+
```
vboctor: It seems that you reverted the code changes here and left with some blanks. Can you revert to original core.php before these changes to remove the noise.

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * @uses Parsedown Library
+ */
+
+require_once( dirname( dirname( __FILE__ ) ) . '../../library/parsedown/Parsedown.php' );
+
+/**
+ * A class that overrides default Markdown parsing for Mantis specific scenarios.
+ */
+class MantisMarkdown extends Parsedown
```
vboctor: I would move this into a sub-folder of the plugin folder, e.g. core or classes.

docbook/Admin_Guide/en-US/config/display.xml
```diff
@@ -194,5 +194,14 @@
 				</para>
 			</listitem>
 		</varlistentry>
+		<varlistentry>
+			<term>$g_markdown_enabled</term>
```
vboctor: If we end up removing this config option, then we should revert the change to this file.

library/parsedown/LICENSE.txt
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
```
vboctor: Move the Parsedown library into a sub-folder of the plugin (e.g. plugin_folder/library/parsedown)

plugins/MantisCoreFormatting/MantisCoreFormatting.php
```diff
@@ -196,4 +196,4 @@ function email( $p_event, $p_string ) {
 
 		return $t_string;
 	}
-}
+}
```
vboctor: Revert the changes to this file.

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -0,0 +1,118 @@
...
+		if ( ON == config_get( 'plugin_MantisCoreFormatting_process_text' ) ) {
+			config_set( 'plugin_MantisCoreFormatting_process_text', OFF );
+		}
+
+		if ( ON == config_get( 'plugin_MantisCoreFormatting_process_urls' ) ) {
+			config_set( 'plugin_MantisCoreFormatting_process_urls', OFF );
+		}
+			
+		#images can be referenced from internet.
+		http_csp_add( 'img-src', "*" );
```
vboctor: You should call this method from a method that gets triggered from EVENT_CORE_HEADERS event.  Check MantisGraph plugin for an example.

tests/Mantis/MarkdownTest.php
```diff
@@ -0,0 +1,152 @@
+<?php
```
vboctor: Move this in a tests folder under the plugin.

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -0,0 +1,128 @@
...
+ * along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @copyright Copyright 2002  MantisBT Team - mantisbt-dev@lists.sourceforge.net
+ */
+
+require_api( 'mention_api.php' );
+require_once( 'core/MantisMarkdown.php' );
+
+/**
+ * Mantis Core Formatting plugin
```
vboctor: Mantis Markdown Plugin rather than Mantis Core Formatting.

vboctor:
@atrol @cproensa I'm planning to merge this tomorrow.

atrol:
I have hardly any time at the moment to have a deeper look at it, maybe Tuesday next week.
I recommend not to merge. It took me just a few minutes to find that
- XSS is possible
- email is not pure text

I would also like to see some comments from @dregad, especially about how to include 3rd party stuff (submodules or not).

jllano:
@atrol I just pushed 2 commits on XSS and pure text issue. Let me know your thoughts @dregad .

atrol:
I still have hardly any time to have a deeper look at it.
If found that @mention functionality does not work the expected way.
Furthermore the current implementation does no longer allow to run 1.3.x and 2.0.x in parallel to the same instance. The plugin changes settings of MantisBT Formatting plugin so that functionality is broken (e.g. # links do no longer work, XSS is introduced, ...).

Unfortunately we are no longer in 2.0.x beta,  so this should be postponed to 2.1.0.

jllano:
@cproensa @vboctor @atrol @dregad I already addressed your comments above.

Let me know your findings gentlemen.

atrol:
> I already addressed your comments above

@jllano
Sure?
@ mentions do still not work and there is still the XSS issue.

Maybe you did not understand what I meant with it.
We want be able to run 1.3.x and 2.0.x in parallel.
After enabling Markdown support, there is a major security issue when using 1.3.x on the same instance.

Open MantisBT Formatting: Configuration page, set all radios to On.
Click Update Configuration
After that, all radios are set to Off, which means that XSS is possible when using 1.3.x.

coreformatting


Getting and and setting the configuration options again and again is also a performance issue.

@cproensa @dregad @syncguru @vboctor
Wouldn't it be better to have markdown as part of the existing core formatting plugin and have just an additional option
Markdown On/Off?

I am not sure if we should backport markdown support to 1.3.x.
If not, I am not shure if we have to target markdown support to 2.1.x.

vboctor:
@atrol I have always thought of markdown as a core feature, so I'm OK having it merge into the core formatting plugin.

We shouldn't back port to 1.3.x in my opinion, we should be releasing 2.0.0 stable soon and markdown will be a new feature there.  Users who want it should upgrade to 2.0.x.  Html emails will also be in 2.0.x to capitalize on markdown support.  I generally want to minimize maintaining features for 1.3.x and 2.0.x, and just focus on 2.0.x.  For smaller features, it is OK to get it into 1.3.x.

For targeting 2.0.x, I think we should get it in when it is ready.  The delta between 2.0.0-rc.1 and 2.0.0 (stable) is not different than the delta between 2.0.0 and 2.1.0, since we don't do beta or rc for each of the dot releases.  Also if we go with markdown as disabled by default, then it is less of a real change.

cproensa:
> Open MantisBT Formatting: Configuration page, set all radios to On.
Click Update Configuration
After that, all radios are set to Off

I dind't catch that change in behaviour.

I was supposing that markdown processing was being added as a complement to current html processing, url processing, bug/note linking processing, Not as a replacement of those.
Thats the reason i suggested to add an entry in coreFormatting plugin to have an ON/OFF there.

Looking at it with perspective, i think that probably this kind of plugin architecture should be revisited with a better approach. This happened too with the avatar plugin, which with no time to dedicate from us to review a best approach, and the urge to put it into release, remains as a uncertain solution. Remember the discussion about distribution of reponsablilities: should the option X be driven by core or plugin?, which parts of the functionality should be in core or in plugin? etc...

For this case: if i have understood correctly:
- currently Markdow plugin implements the processing necessary to replace coreFormatting: it renders/sanitize html, do the markdown stuff, and implements the bug/note linking
- there are problems if coreFormatting and Markdown process sequentially the same text
- for that reason, the approach was to alter the config of coreFormatter to all OFF 
- for that reason, atrol notes the risk of xss if markdown is disabled/uninstalled/not available, and the coreFormatting config has remained to all OFF 

1) I think that disabling the options within coreFormatter is the worst solution. If it has to disable something, then unistall the plugin, or requere the user to do it.
2) Probably, both plugins could be installed, and the conflict resolved by plugin priority, by changing the events to EVENT_TYPE_FIRST types

cproensa:
My thoughts on how this kind of plugins should be architected, is roughly:
- There is a core functionality, that can be implemented in different ways. We want to allow plugins to provide the implementations.
- The functionality management must be done centrally in core. For example, here the functionality is "formatting text". We'd have a "manage formating" page in the admin area.
- "manage formatting" should present the installed formatting plugins, and present options like which one is to be used (exclusively, or combined, etc)
- Internally, the functionality is built around a FormattingInterface:
  - core ask for existing Interface implementations. eg: trigger EVENT_REGISTER_FORMATTING
  - plugin hooks to EVENT_REGISTER_FORMATTING and returns a MarkdownObject that implments FormattingInterface.
  - Core functionality uses the MarkdownObject to do the stuff by calling the methods defined in FormattingInterface.

This is more or less what we already do to provide columns and filters by plugins.

vboctor:
It is important with extensibility to be scenario driven.  Otherwise, we will build a lot of hooks and methods to override, but that won't work for real life scenarios.  The CoreFormatting plugin is an example of a plugin that provides the hooks, but realistically you have to disable it in order to make something interesting.  And hence the existence of the configs to do so.

The formatting is complicated and we need to provide hooks for specific scenarios that we know we can make work.  For example, there can be a specific event to enable plugins to add entity links like what we do with issues and notes.  That we can do pretty well and provide a clean and verify simple interface to have plugins add such functionality with going into the complexity of XSS and making sure their logic won't conflict with some of the processing steps.

So fixing the issues we have with our model is really something that we will have to get to and it will involve identifying the key scenarios and doing breaking changes.

As for markdown support, it may start as off/experimental by default, but it is the future of the core formatting for MantisBT similar to all other tools that our users use.  I expect complementary features will be added like markdown preview, potentially markdown tool, support for embedding images that are attached to issue/notes into their text, enabling these embedded issues in email notifications (like GitHub), etc.

So Markdown should be part of our core formatting (whether that remains a plugin or moves to core).  Then we should expose new events that make sense and provided scenarios that can complement and extend core formatting, rather than replace it.

cproensa:
Let me put my thoughts in a simpler way:
We either:
 - Support a formatting plugin model that works, and that would mean having Markdown as a plugin, as a reference implementation.
 
Or else:
 - Implement Markdown in core AND remove the formatting plugin model, since anything that can be hooked to current events can interfere with Markdown processing (as happens with coreFormatting)


> but it is the future of the **core** formatting for MantisBT

so, is the decision already made?

vboctor:
@cproensa 

> Implement Markdown in core AND remove the formatting plugin model

+1 for implementing Markdown in core.  However, I'm ok with implementing it in CoreFormatting as is for now.  Then we can add in future work new formatting events that can extend rather than replace core formatting.  However, if others are OK with removing the CoreFormatting, then I'm ok with this as well and maybe now would be a good time to do so.  @dregad @atrol can comment here.

> so, is the decision already made?

Yes.  The decision was made by the industry, then by the team as we had the discussion about native markdown support and discussed libraries to use.

jllano:
> @ mentions do still not work (rendering in note)

You are right @atrol, @ mentions doesn't work because of the text process being triggered twice by the plugin both the MantisCoreFormattingPlugin and the MantisMarkdownPlugin.

As for the xss attacks possibility, I dont think its possible now because I retained the current mantis behavior html process.

Let me know guys your final decisions on this, whether that remains a plugin or moves to core.

vboctor:
@jllano let's go ahead and move markdown support to MantisCoreFormatting and add a configuration option then to enable/disable markdown being disabled by default.

jllano:
Ready for test and review @cproensa @dregad @syncguru @vboctor

screen shot 2016-11-25 at 6 31 23 pm

tests/Mantis/AllTests.php
```diff
@@ -52,7 +52,7 @@ public static function suite() {
 		$t_suite->addTestSuite( 'MantisStringTest' );
 		$t_suite->addTestSuite( 'MentionParsingTest' );
 		$t_suite->addTestSuite( 'MantisConfigParserTest' );
-
+
```
atrol: Shouldn't there be the tests from MarkdownTest.php?

atrol:
Missing some kind of README for Parsedown.php, similar to what we have in library/README.md

.gitignore
```diff
@@ -26,6 +26,7 @@ plugins/*
 !plugins/MantisGraph
 !plugins/XmlImportExport
 !plugins/Gravatar
+!plugins/MantisMarkdown
```
atrol: Please revert, not needed after moving functionality to MantisCoreFormatting

config_defaults_inc.php
```diff
@@ -4689,4 +4689,4 @@
  * If a user submits a note with an attachments (with the specified # of seconds)
  * the attachment is linked to the note.  Or 0 for disabling this feature.
  */
-$g_issue_activity_note_attachments_seconds_threshold = 3;
+$g_issue_activity_note_attachments_seconds_threshold = 3;
```
atrol: Not related to this topic

core.php
```diff
@@ -287,4 +287,4 @@ function __autoload( $p_class ) {
 if( !defined( 'PLUGINS_DISABLED' ) && !defined( 'MANTIS_MAINTENANCE_MODE' ) ) {
 	require_api( 'event_api.php' );
 	event_signal( 'EVENT_CORE_READY' );
-}
+}
```
atrol: Not related to this topic

docbook/Admin_Guide/en-US/config/display.xml
```diff
@@ -195,4 +195,4 @@
 			</listitem>
 		</varlistentry>
 	</variablelist>
-</section>
+</section>
```
atrol: Not related to this topic

plugins/MantisCoreFormatting/MantisCoreFormatting.php
```diff
@@ -52,7 +67,88 @@ function config() {
...
+	 */
+	function csp_headers() {
+		static $s_markdown;
+		
+		if( null === $s_markdown ) {
+			$s_markdown = plugin_config_get( 'process_markdown' );
+		}
+
+		if (ON == $s_markdown ) {
+			http_csp_add( 'img-src', "*" );
```
atrol: @mantisbt/core-team do we want to weaken security this way?

dregad: I'm not a security expert, and I'm not sure what kind of implications this may have, but on principle I am also not overly keen on allowing images from anywhere.

dregad:
While this is a highly desirable feature that I would love to have in Mantis as soon as possible, I am also concerned about backwards-compatibility, due to the change in handling user-submitted data, and what would happen e.g. in an evaluation scenario where an admin would run both a 2.0 instance and a 2.x with markdown enabled in parallel. 

Was this fully tested with existing/upgraded databases ? What happens with embedded HTML tags ?

Unfortunately I don't have much time to invest in testing at the moment.

vboctor:
@dregad 

- **Submodule** - makes sense.
- **Evaluating in parallel** - This is not a requirement.  We have done that for a while to support 1.3.x (stable) + 2.0.0 (beta), but we can't use this for all future features making sure they run safely in parallel with older code.  Though anyway, markdown gracefully renders OK to text.
- **Testing w/ data** - By definition markdown will render expected results given some standard text in most cases.  In many cases, the output will be the same or better (e.g. bullets), but in others markdown may wrap lines (removing line break) due to lack of "  " at the end of the line which explicitly indicates that the user wants a new line.  We could accept this and leave it to the users to decide when to edit a block to make it render better with markdown.  The other option is to fiddle with strings as part of an upgrade or pre-processor, but we will break stuff and miss cases anyways.
- **HTML** - The basic support that we had still works.  Though I'm OK with breaking that or users who care about this scenario to not enable markdown.
- **Linking external images** - Do you have specific concerns about enabling images from the internet to be referenced and rendered?  Should we have a flag like $g_markdown_enable_internet_images flag to enable images rather than having CSP disabled all together to enable it?

I would like to get this out and educate the administrators on the benefits and compromises and leave it up to them to decide when to enable / disable markdown.

dregad:
> Evaluating in parallel - This is not a requirement

I agree it's not something that is mandatory, but nevertheless it is a real-life scenario that we should at least consider if feasible, not just for us but for our users too.

> markdown gracefully renders OK to text.

You're right, didn't think about that. So my point is moot, and the only blocker left is the submodule thing.

>  The other option is to fiddle with strings as part of an upgrade or pre-processor, but we will break stuff and miss cases anyways.

I wouldn't go down that route - better not to alter user's data.

Just thinking out loud here, but would it be useful to have a feature to allow users to selectively disable markdown for a given note ?

> HTML - The basic support that we had still works

Shouldn't we deprecate that at some point ? And maybe offer a script that would convert those few tags we allow to the Markdown equivalent ?

> Linking external images - Do you have specific concerns 
> enable reference issue attachments (primarily images), so that is going to be supported then. 

Nothing specific in terms of security, just a precaution. As mentioned, I'm not a security expert. 
With regards to CSP, we should ideally provide a default setting, and let the admin override it should they want a more strict policy.

The main concern is more functional in nature. I strongly believe that allowing references to external attachments is not a good idea, because we have no control over their hosting. If the remote URL becomes unavailable (404) for some reason, e.g. an expired Pastebin, the Mantis issue will likely lose essential information and thus become useless. 

> should we have a flag like $g_markdown_enable_internet_images flag to enable images rather than having CSP disabled all together to enable it? 

Since Markdown is handled by the CoreFormatting plugin, IMO it would be more appropriate to handle this in the plugin's settings, rather than via a Mantis config.

.gitmodules
```diff
@@ -14,3 +14,7 @@
 	path = library/securimage
 	url = https://github.com/mantisbt/securimage.git
 	branch = mantis
+[submodule "library/parsedown"]
+	path = library/parsedown
+	url = https://github.com/erusev/parsedown.git
+	branch = mantisbt
```
dregad: This is incorrect as there is no `mantisbt` branch at https://github.com/erusev/parsedown.

What we normally do for submodules is first fork the repo to mantisbt.org, and create a mantis branch for any specific changes we might have vs the original library. Since you lack privileges to do this yourself, I'll do it for you. Until that's done, you should specify a valid branch in the upstream repository.

jllano: Thanks you @dregad, I assumed we can directly include the submodule from its original location.  

Let me know if you already fork the repo to mantisbt.org.

Thanks again!

dregad: It's technically possible, but using our own fork allows defining of a custom branch where we can e.g. remove unwanted files such as test scripts.

dregad:
@jllano I think this is pretty much good to go now. Could you please rebase your branch on top of latest master, and while you are at it, consider squashing those commits that are not adding any historical value. Thank you !

dregad:
At @vboctor's request, I'll handle the rebase.

library/parsedown/LICENSE.txt
k1631485: MIT License

core/string_api.php
```diff
@@ -144,10 +144,6 @@ function string_nl2br( $p_string, $p_wrap = 100 ) {
  * @return string
  */
 function string_display( $p_string ) {
-	if ( markdown_enabled() ) {
```
k1631485: Code change

library/parsedown/Parsedown.php
```diff
@@ -107,12 +107,6 @@ function setUrlsLinked($urlsLinked)
 
     # ~
 
-    protected $DefinitionTypes = array(
-        '[' => array('Reference'),
-    );
-
```
k1631485: whitespace change

library/parsedown/Parsedown.php
```diff
@@ -169,7 +163,7 @@ private function lines(array $lines)
 
             # ~
 
-            if (isset($CurrentBlock['incomplete']))
+            if (isset($CurrentBlock['continuable']))
```
k1631485: spelling

library/parsedown/Parsedown.php
```diff
@@ -181,12 +175,10 @@ private function lines(array $lines)
                 }
                 else
                 {
-                    if (method_exists($this, 'block'.$CurrentBlock['type'].'Complete'))
+                    if ($this->isBlockCompletable($CurrentBlock['type']))
```
k1631485: spelling

library/parsedown/Parsedown.php
```diff
@@ -1197,7 +1217,7 @@ protected function inlineLink($Excerpt)
             return;
         }
 
-        if (preg_match('/^[(]((?:[^ ()]|[(][^ )]+[)])+)(?:[ ]+("[^"]+"|\'[^\']+\'))?[)]/', $remainder, $matches))
+        if (preg_match('/^[(]((?:[^ ()]|[(][^ )]+[)])+)(?:[ ]+("[^"]*"|\'[^\']*\'))?[)]/', $remainder, $matches))
```
k1631485: Security risk

core/markdown_api.php
```diff
@@ -42,8 +42,8 @@ function markdown_init() {
         require_once( dirname( dirname( __FILE__ ) ) . '/core/classes/MantisMarkdown.php' );
 		
         $g_parsedown = new MantisMarkdown();
-		# set the table class
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -42,8 +42,8 @@ function markdown_init() {
...
-		$g_parsedown->table_class = "table table-nonfluid";
+        # set the table class
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -42,8 +42,8 @@ function markdown_init() {
...
+        $g_parsedown->table_class = "table table-nonfluid";
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -36,25 +36,27 @@
  * @return void
  */
 function markdown_init() {
-	global $g_parsedown;
-	if ( $g_parsedown == null ) {
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -36,25 +36,27 @@
...
+
+    global $g_parsedown;
```
k1631485: global variable

core/markdown_api.php
```diff
@@ -36,25 +36,27 @@
...
+
+    if ( $g_parsedown == null ) {
 
         require_once( dirname( dirname( __FILE__ ) ) . '/core/classes/MantisMarkdown.php' );
-
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -36,25 +36,27 @@
...
 
         require_once( dirname( dirname( __FILE__ ) ) . '/core/classes/MantisMarkdown.php' );
-		
+
         $g_parsedown = new MantisMarkdown();
         # set the table class
         $g_parsedown->table_class = "table table-nonfluid";
         # set the border color of blockquote
         $g_parsedown->inline_style = "border-color:#847d7d";
-	}
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -36,25 +36,27 @@
...
+    }
 }
 
 /**
  * Checked if markdown is enabled from config
  * @return boolean true enabled, false otherwise.
  */
 function markdown_enabled() {
-	return config_get( 'markdown_enabled' ) != OFF;
+    return config_get( 'markdown_enabled' ) != OFF;
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -71,13 +73,14 @@ function markdown_enabled() {
  * @return string html representation generated from markdown
  */
 function markdown_text( $p_text ) {
-	markdown_init();
 
-	global $g_parsedown;
+    markdown_init();
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -71,13 +73,14 @@ function markdown_enabled() {
  * @return string html representation generated from markdown
  */
 function markdown_text( $p_text ) {
-	markdown_init();
 
-	global $g_parsedown;
```
k1631485: global variable

core/markdown_api.php
```diff
@@ -71,13 +73,14 @@ function markdown_enabled() {
...
+    markdown_init();
+
+    global $g_parsedown;
 
-	$t_text = $g_parsedown->text( $p_text );
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -71,13 +73,14 @@ function markdown_enabled() {
...
-	
-	return $t_text;
+    $t_text = $g_parsedown->text( $p_text );
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -95,11 +98,12 @@ function markdown_text( $p_text ) {
  * @return string html representation generated from markdown
  */
 function markdown_line( $p_text ) {
-	markdown_init();
 
-	global $g_parsedown;
+    markdown_init();
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -95,11 +98,12 @@ function markdown_text( $p_text ) {
...
+
+    global $g_parsedown;
```
k1631485: global variable

core/markdown_api.php
```diff
@@ -95,11 +98,12 @@ function markdown_text( $p_text ) {
...
+
+    $t_text =  $g_parsedown->line( $p_text );
 
-	$t_text =  $g_parsedown->line( $p_text );
-	
-	return $t_text;
```
k1631485: whitespace change

core.php
```diff
@@ -274,6 +274,12 @@ function __autoload( $p_class ) {
 
 # Set HTTP response headers
 require_api( 'http_api.php' );
+require_api( 'markdown_api.php' );
+
+if( markdown_enabled() ) {  # if markdown enabled, then images can be referenced from internet.
+	http_csp_add( 'img-src', "*" );
```
k1631485: security risk

core/classes/MantisMarkdown.php
```diff
@@ -38,6 +38,9 @@
 
 require_once( dirname( dirname( __FILE__ ) ) . '../../library/parsedown/Parsedown.php' );
 
+/**
+ * A class that overrides default Markdown parsing for Mantis specific scenarios.
```
k1631485: spelling error

core/classes/MantisMarkdown.php
```diff
@@ -57,23 +60,24 @@ class MantisMarkdown extends Parsedown
...
         # make sure config option bug_link_tag == '#' only
         # check if string start with # symbol followed by numbers
         # hash[numbers] should not be treated as header, then let the app handle it
-        if( '#' == config_get_global( 'bug_link_tag' ) && preg_match_all('/^#\d+$/', $line['text'], $matches) ) {
+        if( '#' == config_get_global( 'bug_link_tag' ) && preg_match_all( '/^#\d+$/', $line['text'], $matches ) ) {
             return;
-        } 
-        # should be treated as headers
+        }
+
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -57,23 +60,24 @@ class MantisMarkdown extends Parsedown
...
+        # Header rules
         /**
          * hash[space][numbers] - treated as header
          * hash[number][*] - treated as header since it is not a pure number
          * hash[letter][*] - treated as header
          * hash[space][letter][*] - treated as header
          */
+
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -85,23 +89,24 @@ protected function blockHeader($line) {
      * @access protected
      * @return void if markdown starts with # symbol | string html representation generated from markdown.
      */
-    protected function blockSetextHeader($line, array $block = NULL) {
-        
-        $block = parent::blockSetextHeader($line, $block);
+    protected function blockSetextHeader( $line, array $block = null ) {
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -85,23 +89,24 @@ protected function blockHeader($line) {
...
         # make sure config option bug_link_tag == '#' only
         # check if string start with # symbol followed by numbers
         # hash[numbers] should not be treated as header, then let the app handle it
-        if( '#' == config_get_global( 'bug_link_tag' ) && preg_match_all('/^#\d+$/', $line['text'], $matches) ) {
+        if( '#' == config_get_global( 'bug_link_tag' ) && preg_match_all( '/^#\d+$/', $line['text'], $matches ) ) {
             return;
-        } 
-        # should be treated as headers
+		}
+
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -85,23 +89,24 @@ protected function blockHeader($line) {
...
+		# Header rules
         /**
          * hash[space][numbers] - treated as header
          * hash[number][*] - treated as header since it is not a pure number
          * hash[letter][*] - treated as header
          * hash[space][letter][*] - treated as header
          */
+
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -131,8 +136,8 @@ private function __doTable($line, $block, $fn) {
      * @access protected
      * @return string html representation generated from markdown.
      */
-    protected function blockTable($line, array $block = null) {
-    	return $this->__doTable($line, $block, __FUNCTION__);
+    protected function blockTable( $line, array $block = null ) {
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -114,9 +119,9 @@ protected function blockSetextHeader($line, array $block = NULL) {
      * @access private
      * @return string html representation generated from markdown.
      */
-    private function __doTable($line, $block, $fn) {
+    private function __doTable( $line, $block, $fn ) {
 
-        if( $block = call_user_func('parent::' . $fn, $line, $block) ) {
+        if( $block = call_user_func( 'parent::' . $fn, $line, $block ) ) {
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -114,9 +119,9 @@ protected function blockSetextHeader($line, array $block = NULL) {
      * @access private
      * @return string html representation generated from markdown.
      */
-    private function __doTable($line, $block, $fn) {
+    private function __doTable( $line, $block, $fn ) {
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -143,8 +148,8 @@ protected function blockTable($line, array $block = null) {
      * @access protected
      * @return string html representation generated from markdown.
      */
-    protected function blockTableContinue($line, array $block) {
-        return $this->__doTable($line, $block, __FUNCTION__);
+    protected function blockTableContinue( $line, array $block ) {
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -156,9 +161,9 @@ protected function blockTableContinue($line, array $block) {
      * @access private
      * @return string html representation generated from markdown.
      */
-    private function __quote($line, $block, $fn) {
+    private function __quote( $line, $block, $fn ) {
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -156,9 +161,9 @@ protected function blockTableContinue($line, array $block) {
...
 
-        if( $block = call_user_func('parent::' . $fn, $line, $block) ) {
+        if( $block = call_user_func( 'parent::' . $fn, $line, $block ) ) {
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -173,8 +178,8 @@ private function __quote($line, $block, $fn) {
      * @access protected
      * @return string html representation generated from markdown.
      */
-    protected function blockQuote($line){
-        return $this->__quote($line, array(), __FUNCTION__);
```
k1631485: whitespace change

core/classes/MantisMarkdown.php
```diff
@@ -185,7 +190,7 @@ protected function blockQuote($line){
      * @access protected
      * @return string html representation generated from markdown.
      */
-    protected function blockQuoteContinue($line, array $block){
-        return $this->__quote($line, $block, __FUNCTION__);
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -36,7 +36,6 @@
  * @return void
  */
 function markdown_init() {
-
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -73,7 +72,6 @@ function markdown_enabled() {
  * @return string html representation generated from markdown
  */
 function markdown_text( $p_text ) {
-
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -98,12 +96,11 @@ function markdown_text( $p_text ) {
  * @return string html representation generated from markdown
  */
 function markdown_line( $p_text ) {
-
```
k1631485: whitespace change

core/markdown_api.php
```diff
@@ -98,12 +96,11 @@ function markdown_text( $p_text ) {
...
     markdown_init();
 
     global $g_parsedown;
 
     $t_text =  $g_parsedown->line( $p_text );
 
     return $t_text;
-}
```
k1631485: whitespace change

config_defaults_inc.php
```diff
@@ -1220,7 +1220,7 @@
 
 /**
  * Enables markdown support - once enabled all issue titles, text, and notes are formatted based on
- * markdown syntax.
+ * markdown syntax.  Default to OFF since it is experimental feature for now.
```
k1631485: Additional space

tests/Mantis/MarkdownTest.php
```diff
@@ -0,0 +1,89 @@
...
+
+class MantisMarkdownTest extends PHPUnit_Framework_TestCase {
+
+	protected static $markdown;
+
+	/**
+	 * Initiazed MantisMarkdown class before a test method run
+	 * @return void
+	 */
+	public static function setUpBeforeClass(){
```
k1631485: whitespace

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
 
 class MantisMarkdownTest extends PHPUnit_Framework_TestCase {
 
-	protected static $markdown;
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
...
+    /**
+     * @var class $markdown
+     */
+    protected static $markdown;
 
-	/**
-	 * Initiazed MantisMarkdown class before a test method run
-	 * @return void
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
...
-	 */
-	public static function setUpBeforeClass(){
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
...
+    /**
+     * Initiazed MantisMarkdown class before a test method run
+     * @return void
+     */
+    public static function setUpBeforeClass() {
         self::$markdown = new MantisMarkdown();
-    }
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
...
+    }
 
-	/**
-	 * Test If string starts with hash character followed by letters
-	 * @return void
-	 */
-	public function testHashLetters() {
-		$this->assertEquals( '<h1>hello</h1>', self::$markdown->text( '# hello' ) );
-		$this->assertEquals( '<h1>hello</h1>', self::$markdown->text( '#hello' ) );
+    /**
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
...
+     */
+    public function testHashLetters() {
+        $this->assertEquals( '<h1>hello</h1>', self::$markdown->text( '# hello' ) );
+        $this->assertEquals( '<h1>hello</h1>', self::$markdown->text( '#hello' ) );
     }
 
     /**
-	 * Test If string starts with hash character followed by number and letters
-	 * @return void
-	 */
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
...
-	public function testHashNumberAny() {
-		$this->assertEquals( '<h1>1abcd</h1>', self::$markdown->text( '# 1abcd' ) );
-		$this->assertEquals( '<h1>1abcd</h1>', self::$markdown->text( '#1abcd' ) );
+     * Test If string starts with hash character followed by number and letters
+     * @return void
+     */
+    public function testHashNumberAny() {
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
...
+        $this->assertEquals( '<h1>1abcd</h1>', self::$markdown->text( '# 1abcd' ) );
+        $this->assertEquals( '<h1>1abcd</h1>', self::$markdown->text( '#1abcd' ) );
     }
 
-	/**
-	 * Test If string starts with hash character followed by letters and numbers
-	 * @return void
-	 */
-	public function testHashLettersAny() {
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
...
-		$this->assertEquals( '<h1>abcd1234</h1>', self::$markdown->text( '# abcd1234' ) );
-		$this->assertEquals( '<h1>abcd1234</h1>', self::$markdown->text( '#abcd1234' ) );
-	}
+    /**
+     * Test If string starts with hash character followed by letters and numbers
+     * @return void
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
...
+        $this->assertEquals( '<h1>abcd1234</h1>', self::$markdown->text( '# abcd1234' ) );
+        $this->assertEquals( '<h1>abcd1234</h1>', self::$markdown->text( '#abcd1234' ) );
+    }
 
-	/**
-	 * Test If string starts with hash character followed by numbers
-	 * @return void
-	 */
-	public function testHashNumbers() {
-		$this->assertEquals( '<p>#1</p>', self::$markdown->text( '#1' ) );
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -40,50 +40,53 @@
...
-	}
-	
-    
-}
+    /**
+     * Test If string starts with hash character followed by numbers
+     * since the class overrides the default Markdown parsing on Header
```
k1631485: whitespace change

tests/Mantis/MarkdownTest.php
```diff
@@ -89,4 +89,82 @@ public function testHashLettersAny() {
...
+
+    /**
+     * Test if table class attribute is defined
+     * @return void
+     */
+    public function testTableClassDefined() {
+        self::$markdown->table_class = "table table-nonfluid";
+        
+        $markdown_table = <<<EOD
+| _header_ 1   | header 2     |
```
k1631485: whitespace

tests/Mantis/MarkdownTest.php
```diff
@@ -89,4 +89,82 @@ public function testHashLettersAny() {
...
+| ------------ | ------------ |
+| _cell_ 1.1   | ~~cell~~ 1.2 |
+| `|` 2.1      | \| 2.2       |
+| `\|` 2.1     | [link](/)    |
+EOD;
+
+        $markdown_table_output = <<<EOD
```
k1631485: whitespace

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+# MantisBT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * MantisMarkdown class
```
k1631485: spelling

core/classes/MantisMarkdown.class.php
k1631485: move into source folder

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * A class that overrides default Markdown parsing for Mantis specific scenarios.
+ */
+class MantisMarkdown extends Parsedown
+{
+	/**
+	 * MantisMarkdown singleton instance for MantisMarkdown class.
+	 */
+	private static $mantis_markdown = null;
```
k1631485: spelling

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+	 * @var string inline style
+	 */
+	private $inline_style = null;
+
+	/**
+	 * MantisMarkdown constructor.
+	 */
+	public function __construct() {
+		# set the table class
+		$this->table_class = 'table table-nonfluid';
```
k1631485: incorrect css for table

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+
+	/**
+	 * @var string table class
+	 */
+	private $table_class = null;
+
+	/**
+	 * @var string inline style
+	 */
+	private $inline_style = null;
```
k1631485: incorrect variable name

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+{
+	/**
+	 * MantisMarkdown singleton instance for MantisMarkdown class.
+	 */
+	private static $mantis_markdown = null;
+
+	/**
+	 * @var string table class
+	 */
+	private $table_class = null;
```
k1631485: incorrect variable name

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+		$this->inline_style = 'border-color:#847d7d';
+	}
+
+	/**
+	 * Check if markdown is enabled.
+	 *
+	 * @return bool true for enabled, false otherwise.
+	 */
+	public static function enabled() {
+		return config_get( 'markdown_enabled' ) != OFF;
```
k1631485: false

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+	 * @access protected
+	 * @return void if markdown starts with # symbol | string html representation generated from markdown.
+	 */
+	protected function blockHeader( $line ) {
+		$block = parent::blockHeader( $line );
+
+		# make sure config option bug_link_tag == '#' only
+		# check if string start with # symbol followed by numbers
+		# hash[numbers] should not be treated as header, then let the app handle it
+		if( '#' == config_get_global( 'bug_link_tag' ) && preg_match_all( '/^#\d+$/', $line['text'], $matches ) ) {
```
k1631485: regex incorrect

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+	 * @access protected
+	 * @return void if markdown starts with # symbol | string html representation generated from markdown.
+	 */
+	protected function blockSetextHeader( $line, array $block = null ) {
+		$block = parent::blockSetextHeader( $line, $block );
+
+		# make sure config option bug_link_tag == '#' only
+		# check if string start with # symbol followed by numbers
+		# hash[numbers] should not be treated as header, then let the app handle it
+		if( '#' == config_get_global( 'bug_link_tag' ) && preg_match_all( '/^#\d+$/', $line['text'], $matches ) ) {
```
k1631485: regex incorrect

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+	 * Add an inline style on a blockquote markdown elements
+	 * 
+	 * @param string $line The Markdown syntax to parse
+	 * @param array $block A block-level element
+	 * @param string $fn the function name to call (blockQuote or blockQuoteContinue)
+	 * @access private
+	 * @return string html representation generated from markdown.
+	 */
+	private function __quote( $line, $block, $fn ) {
+
```
k1631485: whitespace

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+	/**
+	 * Customize the blockQuote method by adding a style attribute
+	 *
+	 * @param string $line The Markdown syntax to parse
+	 * @param array $block A block-level element
+	 * @access protected
+	 * @return string html representation generated from markdown.
+	 */
+	protected function blockQuote( $line ){
```
k1631485: whitespace

core/classes/MantisMarkdown.class.php
```diff
@@ -0,0 +1,244 @@
...
+
+	/**
+	 * Customize the blockQuoteContinue method by adding a style attribute
+	 *
+	 * @param string $line The Markdown syntax to parse
+	 * @param array $block A block-level element
+	 * @access protected
+	 * @return string html representation generated from markdown.
+	 */
+	protected function blockQuoteContinue( $line, array $block ){
```
k1631485: whitespace

core/markdown_api.php
```diff
@@ -1,106 +0,0 @@
...
-/**
- * Initialise the Parsedown library
- * We used our custom markdown (MantisMarkdown) instead of the original library (Parsedown), its because we have
- * our own format settings to follow (e.g If a line starts with # and issue id, the line is treated as a header 
- * instead of an issue reference and the # is omitted form the output)
- *
- * @return void
- */
-function markdown_init() {
-    global $g_parsedown;
```
k1631485: global variable

tests/Mantis/MarkdownTest.php
```diff
@@ -163,8 +145,8 @@ public function testQuoteStyleAttribute() {
 <p>quote</p>
 </blockquote>
 EOD;
-        
-        $this->assertEquals( $markdown_quote_output, self::$markdown->text( $markdown_quote ) );	 
-    }
+
+		$this->assertEquals( $markdown_quote_output, MantisMarkdown::convert_text( $markdown_quote ) );
```
k1631485: whitespace

tests/Mantis/MarkdownTest.php
```diff
@@ -129,28 +113,26 @@ public function testTableClassDefined() {
...
 
 indented:
-   > quote
+	> quote
 
 no space after `>`:
 >quote
 EOD;
 
-        $markdown_quote_output = <<<EOD
```
k1631485: whitespace

tests/Mantis/MarkdownTest.php
```diff
@@ -163,8 +145,8 @@ public function testQuoteStyleAttribute() {
 <p>quote</p>
 </blockquote>
 EOD;
-        
-        $this->assertEquals( $markdown_quote_output, self::$markdown->text( $markdown_quote ) );
```
k1631485: whitespace

plugins/MantisCoreFormatting/MantisCoreFormatting.php
```diff
@@ -78,21 +78,13 @@ function text( $p_event, $p_string, $p_multiline = true ) {
...
-				$t_string = string_html_specialchars( $t_string );
-				$t_string = string_restore_valid_html_tags( $t_string, $p_multiline );
-
-				if( $p_multiline ) {
-					$t_string = string_preserve_spaces_at_bol( $t_string );
-					$t_string = string_nl2br( $t_string );
-				}
+			$t_string = string_strip_hrefs( $t_string );
+			$t_string = string_html_specialchars( $t_string );
+			$t_string = string_restore_valid_html_tags( $t_string, $p_multiline );
```
k1631485: security risk with markdown

plugins/MantisCoreFormatting/MantisCoreFormatting.php
k1631485: not reverted

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -26,19 +26,28 @@
 class MantisMarkdownPlugin extends MantisPlugin {
 	
 	/**
-	 * Make sure to turn off Text Processing, causing conflict with markdown
+	 * Initialized any needed methods, api's, etc...
```
k1631485: spelling

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -26,19 +26,28 @@
 class MantisMarkdownPlugin extends MantisPlugin {
 	
 	/**
-	 * Make sure to turn off Text Processing, causing conflict with markdown
+	 * Initialized any needed methods, api's, etc... 
+	 *
+	 * Make sure to turn off Text Processing, causing markdown to not render properly
```
k1631485: spelling

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -26,19 +26,28 @@
...
+	 * @return void
 	 */	
 	function init() {
-		config_set( 'plugin_MantisCoreFormatting_process_text', OFF );
-		config_set( 'plugin_MantisCoreFormatting_process_urls', OFF );
 
+		if ( ON == config_get( 'plugin_MantisCoreFormatting_process_text' ) ) {
```
k1631485: true

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -26,19 +26,28 @@
...
+			config_set( 'plugin_MantisCoreFormatting_process_text', OFF );
```
k1631485: false

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -26,19 +26,28 @@
...
+		}
+
+		if ( ON == config_get( 'plugin_MantisCoreFormatting_process_urls' ) ) {
```
k1631485: true

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -26,19 +26,28 @@
...
+			config_set( 'plugin_MantisCoreFormatting_process_urls', OFF );
```
k1631485: false

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -26,19 +26,28 @@
...
+		}
+			
 		#images can be referenced from internet.
 		http_csp_add( 'img-src', "*" );
 	}
 
 	/**
-	 * Make sure to turn ON Text Processing back once uninstall
-	 * reset back to MantisCoreFormatting position
+	 * Make sure to turn ON Text Processing back on uninstall
```
k1631485: spelling

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -26,19 +26,28 @@
...
+	 * reset it back to MantisCoreFormatting position
```
k1631485: missing word

config_defaults_inc.php
```diff
@@ -4699,4 +4689,4 @@
  * If a user submits a note with an attachments (with the specified # of seconds)
  * the attachment is linked to the note.  Or 0 for disabling this feature.
  */
-$g_issue_activity_note_attachments_seconds_threshold = 3;
```
k1631485: not related to markdown

config_defaults_inc.php
```diff
@@ -1214,16 +1214,6 @@
...
 
-############
-# Markdown #
-############
-
-/**
- * Enables markdown support - once enabled all issue titles, text, and notes are formatted based on
- * markdown syntax.  Default to OFF since it is experimental feature for now.
- */
-$g_markdown_enabled = OFF;
```
k1631485: should be possible to disable markdown via config

core.php
```diff
@@ -274,7 +274,6 @@ function __autoload( $p_class ) {
 
 # Set HTTP response headers
 require_api( 'http_api.php' );
-
```
k1631485: whitespace

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -28,7 +29,7 @@ class MantisMarkdownPlugin extends MantisPlugin {
 	/**
 	 * Initialized any needed methods, api's, etc... 
 	 *
-	 * Make sure to turn off Text Processing, causing markdown to not render properly
```
k1631485: should be turn off

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -40,18 +41,21 @@ function init() {
 		if ( ON == config_get( 'plugin_MantisCoreFormatting_process_urls' ) ) {
 			config_set( 'plugin_MantisCoreFormatting_process_urls', OFF );
 		}
-			
-		#images can be referenced from internet.
-		http_csp_add( 'img-src', "*" );
+		
+		if ( ON == config_get( 'plugin_MantisCoreFormatting_process_buglinks' ) ) {
```
k1631485: true

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -40,18 +41,21 @@ function init() {
...
+			config_set( 'plugin_MantisCoreFormatting_process_buglinks', OFF );
```
k1631485: false

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -40,18 +41,21 @@ function init() {
...
 
 	/**
-	 * Make sure to turn ON Text Processing back on uninstall
+	 * Make sure to turn Text Processing back on uninstall
 	 * reset it back to MantisCoreFormatting position
 	 */	
 	function uninstall() {
 		config_set( 'plugin_MantisCoreFormatting_process_text', ON );
 		config_set( 'plugin_MantisCoreFormatting_process_urls', ON );
+		config_set( 'plugin_MantisCoreFormatting_process_buglinks', ON );
```
k1631485: true

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -79,10 +83,19 @@ function register() {
 	function hooks() {
 		return array(
 			'EVENT_DISPLAY_FORMATTED'	=> 'markdown',		# Formatted String Display
+			'EVENT_CORE_HEADERS' => 'csp_headers'
```
k1631485: invalid event name

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -79,10 +83,19 @@ function register() {
...
 		);
 	}
 
 	/**
+	 * Add img-src directives to enable to referenced images from internet.
+	 * @return void
+	 */
+	function csp_headers() {
+		http_csp_add( 'img-src', "*" );
```
k1631485: too wide - images should only be allowed from trusted hosts

plugins/MantisMarkdown/core/MantisMarkdown.php
```diff
@@ -36,7 +36,7 @@
  * @uses Parsedown Library
  */
 
-require_once( dirname( dirname( __FILE__ ) ) . '../../library/parsedown/Parsedown.php' );
+require_once( dirname( dirname( __FILE__ ) ) . '/library/Parsedown.php' );
```
k1631485: incorrect directory seperator

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -22,7 +22,7 @@
 require_once( 'core/MantisMarkdown.php' );
 
 /**
- * Mantis Core Formatting plugin
```
k1631485: this is a new file so should not reference core formatting

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -125,4 +127,39 @@ function markdown( $p_event, $p_string, $p_multiline = true ) {
...
+
+	/**
+	 * Email markdown processing.
+	 * @param string $p_event  Event name.
+	 * @param string $p_string Unformatted text.
+	 * @return string Formatted text
+	 */
+	function email( $p_event, $p_string ) {
+		
+		$t_string = $this->markdown( $p_event, $p_string );
```
k1631485: email should not include html

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -125,4 +127,39 @@ function markdown( $p_event, $p_string, $p_multiline = true ) {
...
+		$t_string = string_strip_hrefs( $t_string );
+		$t_string = string_html_specialchars( $t_string );
+		$t_string = string_restore_valid_html_tags( $t_string );
+		$t_string = string_nl2br( $t_string );
+		$t_string = string_insert_hrefs( $t_string );
+		$t_string = string_process_bug_link( $t_string, true, false, true );
+		$t_string = string_process_bugnote_link( $t_string, true, false, true );
+		$t_string = mention_format_text( $t_string, /* html */ true );
+
+		return $t_string;
```
k1631485: rss should not include html

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -112,6 +112,14 @@ function markdown( $p_event, $p_string, $p_multiline = true ) {
 		
 		$t_string = $p_string;	
 		
+		$t_string = string_html_specialchars( $t_string );
+		$t_string = string_restore_valid_html_tags( $t_string, $p_multiline );
+
+		# We need to enabled quote conversion
+		# "> quote or >quote" is part of an html tag
+		# Make sure to replaced the restored tags with ">"
+		$t_string = str_replace( "&gt;", ">", $t_string );
```
k1631485: This allows XSS

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -112,14 +112,9 @@ function markdown( $p_event, $p_string, $p_multiline = true ) {
...
-		$t_string = string_html_specialchars( $t_string );
-		$t_string = string_restore_valid_html_tags( $t_string, $p_multiline );
-
-		# We need to enabled quote conversion
-		# "> quote or >quote" is part of an html tag
-		# Make sure to replaced the restored tags with ">"
-		$t_string = str_replace( "&gt;", ">", $t_string );
-
+		# Filter any possible XSS attacks
+		$t_string = preg_replace('#</*(?:applet|b(?:ase|gsound|link)|embed|frame(?:set)?|i(?:frame|layer)|l(?:ayer|ink)|meta|object|s(?:cript|tyle)|title|xml|input)[^>]*+>#i', '', $t_string);
```
k1631485: incomplete XSS validation string

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -112,22 +112,36 @@ function markdown( $p_event, $p_string, $p_multiline = true ) {
...
+			# "> quote or >quote" is part of an html tag
+			# Make sure to replaced the restored tags with ">"
+			$t_string = str_replace( "&gt;", ">", $t_string );
+
 			$t_string = MantisMarkdown::convert_text( $t_string );
-		} else {
-			$t_string = MantisMarkdown::convert_line( $t_string );
 		}
-
+
```
k1631485: whitespace

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -112,22 +112,36 @@ function markdown( $p_event, $p_string, $p_multiline = true ) {
...
+		# Process mention
 		$t_string = mention_format_text( $t_string, true );
 
-		# Markdown processing
+		# Process markdown with multiline enabled (description, notes, etc)
 		if ( $p_multiline ) {
+			# restore and allows img tag to display
+			$t_string = preg_replace_callback( "/&quot;|'/",
+				function ( ) {
+					return "";
```
k1631485: empty function call

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -112,22 +112,36 @@ function markdown( $p_event, $p_string, $p_multiline = true ) {
...
 		# Process bug links
 		$t_string = string_process_bug_link( $t_string );
 		$t_string = string_process_bugnote_link( $t_string );
-		
+
+		# Process mention
 		$t_string = mention_format_text( $t_string, true );
 
-		# Markdown processing
+		# Process markdown with multiline enabled (description, notes, etc)
```
k1631485: spelling

plugins/MantisMarkdown/MantisMarkdown.php
```diff
@@ -112,22 +112,36 @@ function markdown( $p_event, $p_string, $p_multiline = true ) {
...
-		
+		# Keep mantis current formatting behavior 
+		$t_string = string_strip_hrefs( $t_string );
+		$t_string = string_html_specialchars( $t_string );
+		$t_string = string_restore_valid_html_tags( $t_string, $p_multiline );
+
 		# Process bug links
 		$t_string = string_process_bug_link( $t_string );
 		$t_string = string_process_bugnote_link( $t_string );
-
```
k1631485: whitespace

plugins/MantisCoreFormatting/MantisCoreFormatting.php
```diff
@@ -44,6 +45,20 @@ function register() {
...
 
 	/**
+	 * Event hook declaration.
+	 * @return array
+	 */
+	function hooks() {
+		return array(
+			'EVENT_DISPLAY_TEXT'		=> 'text',			# Text String Display
+			'EVENT_DISPLAY_FORMATTED'	=> 'formatted',		# Formatted String Display
+			'EVENT_DISPLAY_RSS'			=> 'rss',			# RSS String Display
```
k1631485: whitespace

plugins/MantisCoreFormatting/MantisCoreFormatting.php
```diff
@@ -94,7 +188,7 @@ function text( $p_event, $p_string, $p_multiline = true ) {
 	/**
 	 * Formatted text processing.
 	 *
-	 * Performs plain text, URLs and bug links processing
+	 * Performs plain text, URLs, bug links, markdown processing
```
k1631485: missing and

.gitmodules
```diff
@@ -14,3 +14,7 @@
 	path = library/securimage
 	url = https://github.com/mantisbt/securimage.git
 	branch = mantis
+[submodule "library/parsedown"]
+	path = library/parsedown
+	url = https://github.com/erusev/parsedown.git
```
k1631485: this repo does not include a mantisbt branch

.gitmodules
```diff
@@ -16,5 +16,5 @@
 	branch = mantis
 [submodule "library/parsedown"]
 	path = library/parsedown
-	url = https://github.com/erusev/parsedown.git
-	branch = mantisbt
\ No newline at end of file
+	url = https://github.com/mantisbt/parsedown.git
```
k1631485: it's preferable to go to original source for updates

library/README.md
```diff
@@ -10,7 +10,7 @@ directory       | project         | version   | status
 adodb           | adodb           | 5.20.9    | unpatched [1]
 disposable      | disposable      | 2.1.1     | unpatched [1]
 parsedown       | parsedown       | 1.6.1     | unpatched [1]
-phpmailer       | PHPMailer       | 5.2.15    | unpatched [1]
+phpmailer       | PHPMailer       | 5.2.21    | unpatched [1]
```
k1631485: this is not a change to markdown

dregad:
Closing this in favor of #989, which is the rebased code with cleaned up history.

PR #931, opened 2016-10-22
Relocate dropzone inline js to separate file

syncguru:
Fixes #21651

@vboctor in this PR, I adapted the same approach used in #919


core/html_api.php
```diff
@@ -222,6 +222,8 @@ function html_css() {
 	foreach( $g_stylesheets_included as $t_stylesheet_path ) {
 		html_css_link( $t_stylesheet_path );
 	}
+
+	html_css_link( 'dropzone.css' );
```
vboctor: Should we have dropzone css/js only on pages that need it?

dropzone_inc.php
```diff
@@ -0,0 +1,43 @@
...
+ * This include file prints out the fileupload widget
+ *
+ * @package MantisBT
+ * @copyright Copyright 2000 - 2002  Kenzaburo Ito - kenito@300baud.org
+ * @copyright Copyright 2002  MantisBT Team - mantisbt-dev@lists.sourceforge.net
+ * @link http://www.mantisbt.org
+ *
+ */
+?>
+	data-force-fallback="<?php echo config_get( 'dropzone_enabled' ) ? 'false' : 'true' ?>"
```
vboctor: Let's move this to core/print_api.php as print_dropzone_data() instead of being a separate file.

badfiles:
It looks like dz parameters are not being picked up, in my case 
`dictRemoveFile: $(this).data('remove-file')` is 'undefined', although
'data-remove-file is present in \<form\> context.

dregad:
`$(this)` does not refer to the form anymore - see #972 for a proposed fix.

PR #932, opened 2016-10-22
Add missing sidebar trigger button for midsize screen 

syncguru:
+ Also fix consistency issues between mobile screen menu vs. larger screens

Fixes #21715

Before:
image

image

Note: Missing trigger button for sidebar menu
screen shot 2016-10-22 at 10 13 08 am

After:

Note: the two buttons up right are trigger buttons. They show/hide sidebar and navbar menus
image

image

Note: Only one trigger button is needed on larger screen - the sidebar toggler
screen shot 2016-10-22 at 10 10 05 am

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -409,7 +409,6 @@ function layout_navbar() {
 
 	echo '</div>';
 
-	echo '<div class="hidden-xs">';
 	echo '<div class="navbar-buttons navbar-header pull-right navbar-collapse collapse">';
 	echo '<ul class="nav ace-nav">';
 	if (auth_is_user_authenticated()) {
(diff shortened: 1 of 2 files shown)


vboctor:
A couple of questions:

- Why do we have 2 avatar placeholders on the small screen case?
- Should we have a real avatar rather than a placeholder avatar on the small screen as well?  Given that we should the real avatar in other places like notes, we might as well use it on the top bar.

vboctor:
I had an conversation with @syncguru about this.  Following are the changes to be made:

- Avoid having two lines in case of tablet view.
- Have the hamburger button on the top left instead of top right.
- Use a down array icon in small screen to show the second line which includes the user menu and project selector.

syncguru:
Made changes as per conversation with @vboctor.
The down arrow icon was not used however, for couple of reasons: (1) It is rather tricky to implement, (2) I prefer avatar as it personalizes the experience

screen shot 2016-10-24 at 5 48 24 pm

screen shot 2016-10-24 at 5 47 36 pm

screen shot 2016-10-24 at 5 48 57 pm

PR #934, opened 2016-10-30
Disable ‘X-Priority’ header for priority 0

vboctor:
Honor the documentation and don’t emit the X-Priority header when value is 0.

Default the configuration option to 0, since having this header set increases
spam ranking as per mailtrap.io analysis specially for password reset emails.

Fixes #21829

I would like to deprecate this configuration option completely as per the issue details in MantisBT.  However, in the meantime, this change fixes it to behave as per its documentation and changes default to disabled.

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -581,12 +581,10 @@ $g_show_user_email_threshold = NOBODY;
 $g_show_user_realname_threshold = NOBODY;
 
 /**
- * If use_x_priority is set to ON, what should the value be?
  * Urgent = 1, Not Urgent = 5, Disable = 0
- * Note: some MTAs interpret X-Priority = 0 to mean 'Very Urgent'
  * @global integer $g_mail_priority
  */
-$g_mail_priority = 3;
+$g_mail_priority = 0;
 
 /**
  * select the method to mail by:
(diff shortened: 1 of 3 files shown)


atrol:
Looks good, didn't test.

config_defaults_inc.php
```diff
@@ -581,12 +581,10 @@
 $g_show_user_realname_threshold = NOBODY;
 
 /**
- * If use_x_priority is set to ON, what should the value be?
```
dregad: With removal of this line, PHPdoc is missing a description of what this config is about

docbook/Admin_Guide/en-US/config/email.xml
```diff
@@ -266,9 +266,8 @@ $g_notify_flags['new'] = array(
 		<varlistentry>
 			<term>$g_mail_priority</term>
 			<listitem>
-				<para>If use_x_priority is set to ON, what should the value be?
-					Urgent = 1, Not Urgent = 5, Disable = 0 . Default is 3
-					Some MTAs interpret X-Priority = 0 to mean 'Very Urgent'
```
dregad: Same comment as above.

PR #937, opened 2016-11-04
Deprecate ‘mail_priority’ config option

vboctor:
There are several issues with this configuration option:

1. It is not a common practice for services sending emails to set such priority.
2. As per the documentation for this config option, 0 should disable it, but it doesn't.
3. It increases the spam score for some email, specially password reset.

Fixes #21830


dregad:
Duplicate of #938

PR #938, opened 2016-11-04
Deprecate ‘mail_priority’ config option

vboctor:
There are several issues with this configuration option:

1. It is not a common practice for services sending emails to set such priority.
2. As per the documentation for this config option, 0 should disable it, but it doesn't.
3. It increases the spam score for some email, specially password reset.

Fixes #21830

### DIFF
--- a/admin/email_queue.php
+++ b/admin/email_queue.php
@@ -73,7 +73,6 @@ if( $f_mail_test ) {
 	$t_email_data->email = config_get_global( 'webmaster_email' );
 	$t_email_data->subject = 'Testing PHP mail() function';
 	$t_email_data->body = 'Your PHP mail settings appear to be correctly set.';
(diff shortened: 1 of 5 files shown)


dregad:
I'm a bit confused, #934 was targeted at *master* (as [mentioned in the bugtracker](https://www.mantisbt.org/bugs/view.php?id=21829#c54356), I recommended to backport it), but this one is targeted at 1.3 and will therefore likely cause merge conflicts when merged.

That said, I'm fine with the change.

vboctor:
@dregad The fix applied for master was a mitigation until we decide whether to remove the config option or not.  Now that we have decided to remove the config option, I've applied it to master-1.3.x with the intention to also remove from master as part of the merge.

dregad:
Thanks for the clarification.

PR #939, opened 2016-11-05
Remove $g_debug_email in favor of mailtrap.io

vboctor:
Fixes #21860

### DIFF
--- a/admin/check/check_config_inc.php
+++ b/admin/check/check_config_inc.php
@@ -90,11 +90,6 @@ check_print_test_warn_row( 'MantisBT Application Errors should halt execution',
 	array( false => 'Continuing after an error may lead to system and/or data integrity issues. Set $g_display_errors[E_USER_ERROR] = DISPLAY_ERROR_HALT;' )
 );
 
-check_print_test_warn_row( 'Email debugging should be OFF',
-	empty( $g_debug_email ),
-	array( false => 'All notification e-mails will be sent to: ' . $g_debug_email )
-);
-
 check_print_test_row( 'Default move category must exists ("default_category_for_moves")',
 	category_exists( config_get( 'default_category_for_moves' ) ),
 	array( false => 'Issues moved may end up with invalid category id.' )
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -4127,15 +4127,6 @@ $g_show_timer = OFF;
 $g_show_memory_usage = OFF;
 
 /**
- * This is used for debugging the e-mail features in mantis. By default this is blank.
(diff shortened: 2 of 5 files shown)


atrol:
👎 
Maybe I don't understand the benefit of it, but wouldn't it mean that
a) I have to pay for mailtrap.io when using MantisBT in a profit organization
b) I depend on a 3rd party external service in the cloud for debugging purposes
c) email contents go over public lines and are visible for the mailtrap.io provider  (I use debug_email when upgrading internal production systems)

Why should we remove a useful feature?
I see mailtrap.io just as an additional option.

cproensa:
:-1:
I the same reasons than @atrol 

This option does not conflicts in any way with using a "trapping" mail server
Additionally, removing this option will make harder to test with off-the -shelf setups, since you now need to have a system that behaves like that site.

> I see mailtrap.io just as an additional option

My debug methodology is:
set debug_email to "user@localhost"
use a mail client to see local mailbox (eg, mutt)


Additionally:
This is **removing**, not deprectaing
Targeted at 1.3?

vboctor:
@cproensa I should have called it "removing" rather than "deprecating".  Can target this to master instead of master-1.3.x. We have removed mail priority from 1.3.x and 2.0.0 at the same time pretty much.  This is also not a runtime feature, but a plugin feature.

For on-box scenarios:

- [MailCatcher](https://github.com/sj26/mailcatcher)
- localhost as per @cproensa approach.  However, without debug email, user will have to set email addresses to localhost email addresses.

I like the idea of using standard production code during debugging.  However, if the team feels we should keep this debugging functionality, I ok as well.

It may be worth documenting the alternatives for testing email, I have seen people struggle with email setup in general both for development and production scenarios.

cproensa:
> I should have called it "removing" rather than "deprecating"

and still, there is no actual implication to mailtrap.io. So the point discussed here should be strictly the removal of the functionality

> It may be worth documenting the alternatives for testing email

sure, thats a good idea 
My approach is quick and dirty, but good enough for me. A plus is that it works with the standards provided by any linux distribution (eg sendmail), and don't need external connectivity.

> I like the idea of using standard production code during debugging. However, if the team feels we should keep this debugging functionality, I ok as well.

you mean, like, or don't like? anyway i don't see much relation. This is not a compiled executable, so it's not like we can easily release without the debugging logic.
So you maybe mean that this debugging option (and others) should not be present in a production instance? That's another whole point, which seems legit. But right now i don't know how that could be done.

dregad:
I find the debug e-mail is very useful and simple to use for behind-the-firewall scenarios.

While mailtrap is a nice tool, I see it as an additional option, not as an exclusive replacement for the debug emails.

PR #940, opened 2016-11-06
Re-authentication flow should re-use login page

vboctor:
This changes removes a redundant re-authentication page in favor of the standard login page.
This removes redundant code and makes it easier for plugins or custom authentication
schemes to plugin into one place for providing extra functionality.

Fixes #21854

Remove black bar when sign up and sign-in anonymously are disabled.

Fixes #21861

## Login form
login_form

## Login form - wrong password
login_failure

## Reauthenticate form
reauthenticate

## Reauthenticate form - wrong password
reauthenticate_error

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -753,7 +753,7 @@ function auth_set_tokens( $p_user_id ) {
 }
 
 /**
- * Check for authentication tokens, and display re-authentication page if needed.
+ * Check for authentication tokens, and redirect to login page for re-authentication.
  * Currently, if using BASIC or HTTP authentication methods, or if logged in anonymously,
  * this function will always "authenticate" the user (do nothing).
  *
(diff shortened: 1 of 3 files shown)


cproensa:
as i wrote in 
https://www.mantisbt.org/bugs/view.php?id=21854

> I'd add to dregad comment:
> all the complementary elements in the login page should not be used in reuthenticating form, like:
> lost your password
> signup new account
> login anonymous
> Only allow your session to be used from this IP address
> etc
> 
> The core point of this feature should be use the same code to present the login form, to eventually unify login methods/plugins, etc
> 
> And:
> - username should be readonly, imo (as per dregad suggestion)
> - A message should be presented, in the like like current reauthentication does:
> "You are visiting a secure page, and your secure session has expired. Please authenticate yourself to continue."
> 
> I think having the plain login page without modification is confusing.

so:
good on using a single core form
bad on presenting the login page "as is"

vboctor:
@cproensa I've applied the feedback of trimming extra functionality from the re-authentication flow.

cproensa:
I think its better now.

- I still think that username should be readonly
The only scenario that would make sense to allow changing of user is if:
  - we support escalation to enter some admin areas, but then the menus should be available to lower users.
  - the user escalation is temporary and can be reverted to previous user.

which are both not supported at the moment

- Cosmetic: the "login" button text should match the "Authenticate" label.

vboctor:
@cproensa 

> username should be readonly

Don't see the value of doing this.  The username is pre-populated and focus is on password.  If we mark username as disabled, then we have to also handle the fact that form doesn't send this value to action page if username field is disabled.  I don't see an issue with current implementation.

> Cosmetic: the "login" button text should match the "Authenticate" label.

I worry about the form title not being suitable for the button in all translations.  I'm OK with just using Log on all cases and removing the Authenticate form title all together.  We will still have the message that explains why is the user prompted to re-login which is the one that really adds value here.

cproensa:
>  If we mark username as disabled, then we have to also handle the fact that form doesn't send this value to action page if username field is disabled.

adding the username as hidden field

PR #941, opened 2016-11-06
Use selective bug cache for attachment count

cproensa:
Resubmit the changes from PR #654 as the main issue with long IN clauses was solved after refactoring export pages

This is a dramatic improvement in page load when there is a lot of attachments in the db.
See comment: https://github.com/mantisbt/mantisbt/pull/654#issuecomment-153838142

Additionally:
Added some changes to include this file count cache in the clearing procedures at export pages to save on memory.
Fixed some procedures for clearing caches that were wrong.

### DIFF
--- a/core/bugnote_api.php
+++ b/core/bugnote_api.php
@@ -753,10 +753,38 @@ function bugnote_clear_cache( $p_bugnote_id = null ) {
 
 	if( null === $p_bugnote_id ) {
 		$g_cache_bugnote = array();
+		$g_cache_bugnotes = array();
 	} else {
-		unset( $g_cache_bugnote[(int)$p_bugnote_id] );
+		if( isset( $g_cache_bugnote[(int)$p_bugnote_id] ) ) {
+			$t_note_obj = $g_cache_bugnote[(int)$p_bugnote_id];
+			bugnote_clear_bug_cache( $t_note_obj->bug_id );
+			unset( $g_cache_bugnote[(int)$p_bugnote_id] );
+		}
 	}
-	$g_cache_bugnotes = array();
+
+	return true;
+}
+
+/**
+ * Clear the bugnotes related to a bug, or all bugs if no bug id specified.
+ * @param integer $p_bug_id Identifier to clear (optional).
+ * @return boolean
+ * @access public
+ */
+function bugnote_clear_bug_cache( $p_bug_id = null ) {
+	global $g_cache_bugnotes, $g_cache_bugnote;
+
+	if( null === $p_bug_id ) {
+		$g_cache_bugnotes = array();
+		$g_cache_bugnote = array();
+	} else {
+		if( isset( $g_cache_bugnotes[(int)$p_bug_id] ) ) {
+			foreach( $g_cache_bugnotes[(int)$p_bug_id] as $t_note_obj ) {
+				unset( $g_cache_bugnote[(int)$t_note_obj->id] );
+			}
+			unset( $g_cache_bugnotes[(int)$p_bug_id] );
+		}
+	}
 
 	return true;
 }
--- a/core/bugnote_api.php
+++ b/core/bugnote_api.php
@@ -757,8 +757,8 @@ function bugnote_clear_cache( $p_bugnote_id = null ) {
 	} else {
 		if( isset( $g_cache_bugnote[(int)$p_bugnote_id] ) ) {
 			$t_note_obj = $g_cache_bugnote[(int)$p_bugnote_id];
+			# current note id will be unset in the following call
 			bugnote_clear_bug_cache( $t_note_obj->bug_id );
-			unset( $g_cache_bugnote[(int)$p_bugnote_id] );
 		}
 	}
 
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -160,6 +160,24 @@ function file_bug_attachment_count( $p_bug_id ) {
 }
 
 /**
+ * Clear a bug from the cache or all bugs if no bug id specified.
+ * @param integer $p_bug_id A bug identifier to clear (optional).
+ * @return boolean
+ * @access public
+ */
+function file_bug_attachment_count_clear_cache( $p_bug_id = null ) {
+	global $g_cache_file_count;
+
+	if( null === $p_bug_id ) {
+		$g_cache_file_count = array();
+	} else {
+		unset( $g_cache_file_count[(int)$p_bug_id] );
+	}
+
+	return true;
+}
+
+/**
  * Check if a specific bug has attachments
  * @param integer $p_bug_id A bug identifier.
  * @return boolean
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -98,6 +98,52 @@ function file_get_display_name( $p_filename ) {
 }
 
 /**
+ * Fills the cache with the attachement count from a list of bugs
+ * If the bug doesn't have attachments, cache its value as 0.
+ * @global array $g_cache_file_count
+ * @param array $p_bug_ids Array of bug ids
+ * @return void
+ */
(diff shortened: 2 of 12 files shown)


core/bugnote_api.php
```diff
@@ -753,10 +753,38 @@ function bugnote_clear_cache( $p_bugnote_id = null ) {
...
+		}
 	}
-	$g_cache_bugnotes = array();
 
 	return true;
 }
+
+/**
+ * Clear the bugnotes related to a bug, or all bugs if no bug id specified.
+ * @param integer $p_bu_id Identifier to clear (optional).
```
vboctor: Typo in parameter name.

core/bugnote_api.php
```diff
@@ -753,10 +753,38 @@ function bugnote_clear_cache( $p_bugnote_id = null ) {
 
 	if( null === $p_bugnote_id ) {
 		$g_cache_bugnote = array();
+		$g_cache_bugnotes = array();
 	} else {
-		unset( $g_cache_bugnote[(int)$p_bugnote_id] );
+		if( isset( $g_cache_bugnote[(int)$p_bugnote_id] ) ) {
+			$t_note_obj = $g_cache_bugnote[(int)$p_bugnote_id];
+			bugnote_clear_bug_cache( $t_note_obj->bug_id );
+			unset( $g_cache_bugnote[(int)$p_bugnote_id] );
```
vboctor: Would this still be set after clearing the bug and all associated notes?

core/file_api.php
```diff
@@ -98,45 +98,81 @@ function file_get_display_name( $p_filename ) {
 }
 
 /**
- * Check the number of attachments a bug has (if any)
- * @param integer $p_bug_id A bug identifier.
- * @return integer
+ * Fills the cache with the attachement count from a list of bugs
+ * If the bug doesn't have attachments, cache its value as 0.
+ * @param array $p_bug_ids Array of bug ids
  */
```
vboctor: @ return void

core/file_api.php
```diff
@@ -98,45 +98,81 @@ function file_get_display_name( $p_filename ) {
...
-	if( isset( $g_cache_file_count[$p_bug_id] ) ) {
-		return $g_cache_file_count[$p_bug_id];
+	if( empty( $p_bug_ids ) ) {
+		return;
+	}
+
+	$t_ids_to_sarch = array();
+	foreach( $p_bug_ids as $t_id ) {
+		$c_id = (int)$t_id;
+		$t_ids_to_sarch[$c_id] = $c_id;
```
vboctor: Typo in variable name.

core/filter_api.php
```diff
@@ -2225,21 +2225,25 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
...
-		$b = $t_stats[$t_row['id']];
+		$t_bug_ids[] = (int)$t_row['id'];
 		if( array_key_exists( $t_row['id'], $t_stats ) ) {
 			$t_rows[] = bug_row_to_object( bug_cache_database_result( $t_row, $t_stats[$t_row['id']] ) );
 		} else {
 			$t_rows[] = bug_row_to_object( bug_cache_database_result( $t_row ) );
 		}
 	}
+	#cache the attachment count for bug list
+	file_bug_attachment_count_cache( $t_bug_ids );
```
vboctor: Should we be doing this even if the caller has no intention of displaying this information (e.g. View Issues page without attachments column or soap API, etc)?  It feels that this function is not making assumptions about the callers intent.

core/bugnote_api.php
cproensa: what do you mean? having $t_note_obj as reference to the object, even after unsetting the index in the array?

core/bugnote_api.php
```diff
@@ -753,10 +753,38 @@ function bugnote_clear_cache( $p_bugnote_id = null ) {
...
+ * @param integer $p_bug_id Identifier to clear (optional).
+ * @return boolean
+ * @access public
+ */
+function bugnote_clear_bug_cache( $p_bug_id = null ) {
+	global $g_cache_bugnotes, $g_cache_bugnote;
+
+	if( null === $p_bug_id ) {
+		$g_cache_bugnotes = array();
+		$g_cache_bugnote = array();
```
dregad: I never noticed we had 2 distinct variables for bugnotes caching. It is not really clear what the difference between them are, and they are not formally declared at the top of bugnote api, only as global in the functions using them. Do we actually need 2 of them ?

Even though that's not directly related to this PR, I would suggest to declare them, and add a PHPDoc block to explain their purpose and differences.

core/bugnote_api.php
dregad: I think he means that in bugnote_clear_bug_cache(), all entries in $g_cache_bugnote for the corresponding bug_id are already unset (see [line 783](https://github.com/mantisbt/mantisbt/pull/941/files#diff-af4b0457065b17f59cdc20eec8005bc4R783))

core/filter_api.php
dregad: I think he means that in bugnote_clear_bug_cache(), all entries in $g_cache_bugnote for the corresponding bug_id are already unset (see [line 783](https://github.com/mantisbt/mantisbt/pull/941/files#diff-af4b0457065b17f59cdc20eec8005bc4R783))

core/bugnote_api.php
vboctor: My question is whether bugnote_clear_bug_cache() already clears the note from the $g_cache_bugnote, and then you are unsetting it again.  It feels to me that the unset line is redundant.

core/filter_api.php
vboctor: btw, "function is not making assumptions" -> "function is now making assumptions" :)

core/bugnote_api.php
vboctor: I would also rename them to more meaningful names.  For example:

- $g_cache_bugnotes_by_id
- $g_cache_bugnotes_by_bug_id

core/bugnote_api.php
cproensa: I got it now. you're right.
Apart from that... i'm not totally sure it the clearing of one cache should also clear the respective content in the other cache...

core/bugnote_api.php
cproensa: right, those names are self-explaining, way better than current ones.
however i would not rename at this moment that there is other PR rewritting the same api

dregad: For follow-up https://www.mantisbt.org/bugs/view.php?id=21884

core/file_api.php
```diff
@@ -98,45 +98,83 @@ function file_get_display_name( $p_filename ) {
...
+function file_bug_attachment_count_clear_cache( $p_bug_id = null ) {
+	global $g_cache_file_count;
+
+	if( null === $p_bug_id ) {
+		$g_cache_file_count = array();
+	} else {
+		unset( $g_cache_file_count[(int)$p_bug_id] );
+	}
+
+	return true;
```
vboctor: Why do we need a return value for this method?  It is always true.

core/filter_api.php
```diff
@@ -2225,6 +2225,7 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
 
 /**
  *  Cache the filter results with bugnote stats for later use
+ *  also fills bug attachment count cache
```
vboctor: Is this comment still valid?

core/file_api.php
cproensa: The other clear cache methods used this scheme. 
No strong reason to return true, but a more detailed analysis on current usage, for the existing ones, should be done

dregad:
Merged in d47f367fea8dfda1e5bce918419a1bda1dd0e10b

PR #942, opened 2016-11-06
Improve db_fetch_array performance

cproensa:
Improve db_fetch_array performance by caching the result from:
db_is_pgsql()
db_is_oracle()

Based on profiling, the repeated calls were using up to 20% of total time
for the db_fetch_array execution.

Fixes: #21871

### DIFF
--- a/core/constant_inc.php
+++ b/core/constant_inc.php
@@ -624,3 +624,11 @@ define( 'EXPORT_BLOCK_SIZE', 500 );
 # the closest integer (4 bytes) type available. As some DBs dont support unsigned
 # types, 2^31 is a safe limit to be used for all.
 define( 'DB_MAX_INT', 2147483647 );
+
+# Databse functional type identifiers.
+define( 'DB_TYPE_UNDEFINED', 0 );
+define( 'DB_TYPE_MYSQL', 1 );
+define( 'DB_TYPE_PGSQL', 2 );
+define( 'DB_TYPE_MSSQL', 3 );
+define( 'DB_TYPE_ORACLE', 4 );
+define( 'DB_TYPE_DB2', 5 );
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -523,63 +523,57 @@ function db_affected_rows() {
  * @return array Database result
  */
 function db_fetch_array( IteratorAggregate &$p_result ) {
-	global $g_db, $g_db_type;
+	global $g_db_functional_type;
 
 	if( $p_result->EOF ) {
 		return false;
-	}
-
-	# cache these checks to avoid repeated api calls
-	static $t_db_is_pgsql = null;
-	static $t_db_is_oracle = null;
-	if( null === $t_db_is_pgsql ) {
-		$t_db_is_pgsql = db_is_pgsql();
-	}
-	if( null === $t_db_is_oracle ) {
-		$t_db_is_oracle = db_is_oracle();
 	}
 
 	# Retrieve the fields from the recordset
 	$t_row = $p_result->fields;
 
 	# Additional handling for specific RDBMS
-	if( $t_db_is_pgsql ) {
-		# pgsql's boolean fields are stored as 't' or 'f' and must be converted
-		static $s_current_result = null, $s_convert_needed;
+	switch( $g_db_functional_type ) {
 
-		if( $s_current_result != $p_result ) {
-			# Processing a new query
-			$s_current_result = $p_result;
-			$s_convert_needed = false;
-		} elseif( !$s_convert_needed ) {
-			# No conversion needed, return the row as-is
-			$p_result->MoveNext();
-			return $t_row;
-		}
+		case DB_TYPE_PGSQL:
+			# pgsql's boolean fields are stored as 't' or 'f' and must be converted
+			static $s_current_result = null, $s_convert_needed;
 
-		foreach( $p_result->FieldTypesArray() as $t_field ) {
-			switch( $t_field->type ) {
-				case 'bool':
-					switch( $t_row[$t_field->name] ) {
-						case 'f':
-							$t_row[$t_field->name] = false;
-							break;
-						case 't':
-							$t_row[$t_field->name] = true;
-							break;
-					}
-					$s_convert_needed = true;
-					break;
+			if( $s_current_result != $p_result ) {
+				# Processing a new query
+				$s_current_result = $p_result;
+				$s_convert_needed = false;
+			} elseif( !$s_convert_needed ) {
+				# No conversion needed, return the row as-is
+				$p_result->MoveNext();
+				return $t_row;
 			}
-		}
 
-	} elseif( $t_db_is_oracle ) {
-		# oci8 returns null values for empty strings, convert them back
-		foreach( $t_row as &$t_value ) {
-			if( !isset( $t_value ) ) {
-				$t_value = '';
+			foreach( $p_result->FieldTypesArray() as $t_field ) {
+				switch( $t_field->type ) {
+					case 'bool':
+						switch( $t_row[$t_field->name] ) {
+							case 'f':
+								$t_row[$t_field->name] = false;
+								break;
+							case 't':
+								$t_row[$t_field->name] = true;
+								break;
+						}
+						$s_convert_needed = true;
+						break;
+				}
 			}
-		}
+			break;
+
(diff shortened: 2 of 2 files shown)


vboctor:
A couple of comments:

- I would implement the caching in db_is_xxx() methods and not in the place calling them.  Is the cost in the call or in the call eventual call to config_get_global()?

- Is it possible to cache a single value db-type, so calling db_is_mysql() then db_is_oracle() would result in a cache hit?

cproensa:
> I would implement the caching in db_is_xxx() methods and not in the place calling them. Is the cost in the call or in the call eventual call to config_get_global()?

Those methods are already lightweight, only calling to config_get_global, which eventually have it cached too.
This only adds to a measurable amount of spent time when this is called repeatedly, which happens everytime a row is fetched (db_fetch_array is already lightweight internally, like retrieving and moving and iterator). However, for big db results this start to be (relatively) significant in the totals.

Tht's why i decided to cache in the caller, instead of the db_is_xxx(), since is in the caller where the root cause happens.

> Is it possible to cache a single value db-type, so calling db_is_mysql() then db_is_oracle() would result in a cache hit?

1 vs 2 cache miss is not significant in this situation. I used 2 boolean variables, feeling that it's a simpler comparison than having a string value, and not worth defining constants only for this.

core/database_api.php
```diff
@@ -528,11 +528,21 @@ function db_fetch_array( IteratorAggregate &$p_result ) {
 		return false;
 	}
 
+	# cache these checks to avoid repeated api calls
+	static $t_db_is_pgsql = null;
+	static $t_db_is_oracle = null;
```
dregad: Static vars must have `s_` prefix.

I would tend to agree with @vboctor on using a single variable here, I don't think it's more complicated to have `$s_db_type == DB_TYPE_xxx` vs a boolean check. Furthermore, since this is about performance optimization, the single-var+constants approach allows use of a switch statement instead of a series of *elseif*, so the code requires fewer tests: 1 to set the static var, and 1 for the check, instead of 2n (so 4 currently).

cproensa:
If you look at the image posted at 
https://www.mantisbt.org/bugs/view.php?id=21871

my interpretation is that the accumulated time in db_is_xxx() is just the overhead for the nested call to config_get_global() (as this one does not do much work by itself).
These should be negligible in common usage, and only noticeable on big data retrieval: export, summary, etc... when the number of fetches is big enough.

I like dregad suggestion to use constant, and probably redefine the db_is_xxx() to get_db_type(), which returns DB_TYPE_MYSQL, DB_TYPE_PGSQL, etc

atrol:
Just some thoughts, didn't try anything

As there is hardly anyone who sets $g_db_type = '%g_some_other_config_option%';
Wouldn't it be acceptable to use global variable $g_db_type in functions db_is_* ?
@cproensa, maybe you can invest some time to check if the functions do still take a lot of time after such a change? 

Caching of cached values?
config_get_global implements caching ($g_cache_config_eval), so I wouldn't expect that much enhancement.
Maybe there goes something wrong with the caching?

In case this is a config_get_global performance issue, you might remember 
https://www.mantisbt.org/bugs/view.php?id=21604
https://github.com/atrol/mantisbt/commit/48ce794e6953afcd0065907ee2b0e9028de5202d

cproensa:
Implemented the suggestions:
- db_connect() stores the database abstract type, in the global var. They are defined by constants.
- db_is_xxx are now shorthand comparing to that constant
- db_fetch_array doesn't need api calls to perform, the code encased in a switch, ...

I agree this looks cleaner, even if more code modification were involved.

I'd say it is safe to define the db type global at db_connect(), but it could be initialized at the global scope in other case.

core/database_api.php
```diff
@@ -59,6 +59,9 @@
 $ADODB_FETCH_MODE = ADODB_FETCH_ASSOC;
 define( 'ADODB_ASSOC_CASE', ADODB_ASSOC_CASE_LOWER );
 
+# Stores the functional database type based on db driver
+$g_db_functional_type;
```
atrol: missing global?

cproensa: Undefined at that point. But present for the sake of documenting it.
This could be initialized there, or as it is now, from db_connect()

vboctor: Don't we usually initial global variables to some value like null or undefined?

core/constant_inc.php
```diff
@@ -623,4 +623,12 @@
 # Note: mantis ids are defined in schema as "I UNSIGNED", which Adodb maps to
 # the closest integer (4 bytes) type available. As some DBs dont support unsigned
 # types, 2^31 is a safe limit to be used for all.
-define( 'DB_MAX_INT', 2147483647 );
\ No newline at end of file
+define( 'DB_MAX_INT', 2147483647 );
+
+# Databse functional type identifiers.
+define( 'DB_TYPE_UNDEFINED', 0 );
+define( 'DB_TYPE_MYSQL', 1 );
```
vboctor: Should we have all database api except db_get_type() use these constants instead of comparing directly against the config option?  For example, db_query seems to still act directly on the config option.

cproensa: if going the full way, we should identify two types of usage of this config variable:
- check the type of database, in an abstract way: mysql, mssql, etc. 
these usually would use de db_is_xxx methods
- check the specific driver: mysqli, mssql_native, etc
these new to access the config variable, to perform specific operations based on the driver.

cproensa:
I have added initialization of $g_db_functional_type in the global scope.
This is needed on install script, where schema may check this db_is_Xxx before a connection is made.

I tested with the oracle installation (as an example of one that uses the exceptions), which now responds correctly to these checks.

core/database_api.php
```diff
@@ -222,78 +226,75 @@ function db_check_database_support( $p_db_type ) {
...
+		case 'mssql':
+		case 'mssqlnative':
+		case 'odbc_mssql':
+			return DB_TYPE_MSSQL;
+		case 'db2':
+			return DB_TYPE_DB2;
+		case 'oci8':
+			return DB_TYPE_ORACLE;
+		default:
+			return DB_TYPE_UNDEFINED;
```
dregad: This is a quite unlikely scenario, but maybe we could trigger an error in case of unsupported DB ?

atrol:
@cproensa @dregad @vboctor 
what about closing/changing the PR and merge it to master instead of master-1.3.x?

dregad:
For me this is OK to merge. I personally don't really care what branch should be the target, but master makes sense.

core/database_api.php
```diff
@@ -222,78 +226,75 @@ function db_check_database_support( $p_db_type ) {
...
-	return false;
+	global $g_db_functional_type;
+	return( DB_TYPE_MSSQL == $g_db_functional_type );
 }
 
 /**
  * Checks if the database driver is DB2
  * @return boolean true if db2
  */
 function db_is_db2() {
```
vboctor: Do we need db_is_db2()?

core/database_api.php
```diff
@@ -532,43 +533,47 @@ function db_fetch_array( IteratorAggregate &$p_result ) {
...
-		} elseif( !$s_convert_needed ) {
-			# No conversion needed, return the row as-is
-			$p_result->MoveNext();
-			return $t_row;
-		}
+	switch( $g_db_functional_type ) {
+
+		case DB_TYPE_PGSQL:
+			# pgsql's boolean fields are stored as 't' or 'f' and must be converted
+			static $s_current_result = null, $s_convert_needed;
```
vboctor: Should we default $s_convert_needed to false?

core/database_api.php
```diff
@@ -532,43 +533,47 @@ function db_fetch_array( IteratorAggregate &$p_result ) {
...
 
-	} elseif( db_is_oracle() ) {
-		# oci8 returns null values for empty strings, convert them back
-		foreach( $t_row as &$t_value ) {
-			if( !isset( $t_value ) ) {
-				$t_value = '';
+		case DB_TYPE_ORACLE:
+			# oci8 returns null values for empty strings, convert them back
+			foreach( $t_row as &$t_value ) {
+				if( !isset( $t_value ) ) {
```
vboctor: Should this be is_null instead of isset() to be consistent with documentation above?

sync1:

cproensa:
replaced by #1105

PR #943, opened 2016-11-07
Replace jscalendar with bootstrap-datetimepicker library

syncguru:
bootstrap-datetimepicker: https://github.com/eonasdan/bootstrap-datetimepicker

+ Enable datetimepciker on all pages by default
+ Relocate inline JS code into standalone files
+ Remove CSP exceptions for jscalendar
+ Multi-language support

Fixes #20040

Date Picker
screen shot 2016-11-06 at 7 20 54 pm

Time Picker
screen shot 2016-11-06 at 7 23 22 pm

screen shot 2016-11-06 at 7 21 35 pm

screen shot 2016-11-06 at 7 22 07 pm

Bulk Edit
screen shot 2016-11-06 at 7 22 50 pm

French
screen shot 2016-11-06 at 7 32 36 pm


config_defaults_inc.php
```diff
@@ -1137,15 +1137,15 @@
...
- * jscalendar date format string
- * go to http://www.php.net/manual/en/function.date.php
+ * calendar widget (datetimepicker) format string
+ * go to http://momentjs.com/docs/#/displaying/format/
  * for detailed instructions on date formatting
- * @global string $g_calendar_js_date_format
+ * @global string $g_calendar_date_format
  */
-$g_calendar_js_date_format = '\%Y-\%m-\%d \%H:\%M';
+$g_calendar_js_date_format = 'Y-M-D H:m';
```
dregad: For consistency with prior version (month and day with leading 0), the new format should be `Y-MM-DD`.

dregad:
Note: this was just a quick review, I have not yet tested the new functionality.

atrol:
I didn't have a deeper look, but what's wrong using the jquery-ui date-time picker used in PR https://github.com/mantisbt/mantisbt/pull/628/ ?
At first sight I see just that we introduce one more 3rd party dependency and that we have to deliver even more .js files to the client.

syncguru:
@atrol Do you know why we include jquery-ui? 
This PR removes jscalendar library and replace it with new one so net change is zero.

I opened issue #21881 to see if we can get rid of jquery-ui

atrol:
jquery and jquery-ui have been plugins before 1.3.x
https://github.com/mantisbt-plugins/jquery
https://github.com/mantisbt-plugins/jQuery-UI

They have been bundled with MantisBT core to make life of plugin authors a bit easier.

At least this one uses jQuery-UI
https://github.com/mantisbt-plugins/Inline-column-configuration

vboctor:
@syncguru as we replace the date time control, we should do one of two things:

1. Deprecate js calendar configs and add new ones.
2. Do the internal processing to convert existing config options format to the new one, so existing configurations would work.

Please update library/README.md with new libraries used and their version details.

I will do some testing tonight and provide more feedback.

@atrol I haven't researched the components being compared here, but I would rather use a bootstrap based component over a jquery-ui based one.  The component used here has > 4,800 stars.  That is pretty popular.  Will look more.

@syncguru Did you test on small screens?

But generally, I'm happy with finally getting rid of the outdated jscalendar and the inline scripts associated with it.

syncguru:
@dregad 
+ Fixed the date format
+ JS files download are a onetime cost (in addition to cdn support). The logic that existed before to detect the use of the calendar and then include the js/css files is complex, error prone and not plugin friendly. It is hard to imaging a bug tracker without a date/time picker field.  

@vboctor
+ Deprecated js calendar config 
+ There is no new library added under library directory. jscalander did not have an entry in the readme file for me to remove.
+ Yes, the control works on mobile. For MantisBt, it is a shifted to the right as a by product of the table format. This will auto resolve as soon as we optimize tables/forms for mobile.

@atrol 
I strongly think we should remove jquery-ui ... happy to fix broken plugins. It should be very easy fix.

atrol:
> I strongly think we should remove jquery-ui

I don't have a problem with it as I don't use any plugin that uses  jquery-ui, but would like to get also some feedback from others, e.g. @rombert who is the author of the Inline-column-configuration plugin.

atrol:
manage_proj_ver_edit_page.php does not work for me.
bug_update_page.php (Due Date) works when clicking in the text field, but not when clicking on the calendar icon.

badfiles:
I think inputs with datetimepicker should not be auto-completed (` autocomplete="off"`), in case if js code does not already turn off this feature.

dregad:
>> I strongly think we should remove jquery-ui ... happy to fix broken plugins. It should be very easy fix.

> I don't have a problem with it as I don't use any plugin that uses jquery-ui, but would like to get also some feedback from others, e.g. @rombert who is the author of the Inline-column-configuration plugin. 

As far as plugins registered in *mantisbt-plugins* org, [I found several that refer to or use jQuery-UI](https://github.com/search?q=org%3Amantisbt-plugins+jqueryui&type=Code) in one way or another. Mentioning their respective authors to bring their attention to this discussion, and get their opinion:

- [jQuery-UI](https://github.com/mantisbt-plugins/jQuery-UI) (duh)
- [customer-management](https://github.com/mantisbt-plugins/customer-management) by @rombert 
- [Inline-column-configuration](https://github.com/mantisbt-plugins/Inline-column-configuration) by @rombert (as previously mentioned)
- [jsedit](https://github.com/mantisbt-plugins/jsedit) by @Kirill 
- [agileMantis](https://github.com/mantisbt-plugins/agileMantis) by gadiv (@jako87, @jazsch, @jssw) but in this case the plugin bundles jQuery-UI as part of its source

Most of these plugins have not been updated recently, and none have yet been revised by their respective authors to work with MantisBT 2.x.

Since we don't use jQuery-UI ourselves in the core ATM (and it probably does not make sense to do so in the future since we went down the Bootstrap route), I don't have a strong objection against removing it; it would be quite straightforward to update the jQuery-UI plugin, to enable 3rd-party plugins to use that functionality if they need it.

rombert:
Hm, managed to miss the mention. Anyway, I don't see a problem with removing jQuery-UI and having administrators pull in the jQuery-UI plug-in themselves if needed.

core/http_api.php
```diff
@@ -231,17 +232,7 @@ function http_security_headers() {
...
-		# Relaxing policy for bug report page to allow inline scripts.
-		# Should be removed once #20040 is fixed.
-		if( 'bug_report_page.php' == basename( $_SERVER['SCRIPT_NAME'] ) ) {
-			http_csp_add( 'script-src', "'unsafe-inline'" );
-		}
 
-		# The JS Calendar control does unsafe eval, remove once we upgrade the control (see #20040)
-		if( 'bug_update_page.php' == basename( $_SERVER['SCRIPT_NAME'] ) ) {
-			http_csp_add( 'script-src', "'unsafe-eval'" );
 		}
```
syncguru: @dregad fyi: csp exception no longer needed.

syncguru:
@atrol
- Good catch, fixed.
- The calendar icon is not clickable due to some css complexities. This should not block this PR though, will try to fix in a separate PR.

core/http_api.php
dregad: Cheers

core/lang_api.php
```diff
@@ -354,3 +354,20 @@ function lang_get_defaulted( $p_string, $p_default = null, $p_lang = null ) {
 		}
 	}
 }
+
+
+/**
+ * Maps current lang string to moment.js locale https://github.com/moment/moment/tree/develop/locale
+ * @return string
```
vboctor: Document what the return string will contain.

core/lang_api.php
```diff
@@ -354,3 +354,20 @@ function lang_get_defaulted( $p_string, $p_default = null, $p_lang = null ) {
...
+ */
+
```
vboctor: Remove blank line between phpdoc and function.

core/layout_api.php
```diff
@@ -307,11 +313,21 @@ function layout_head_javascript() {
  * @return null
  */
 function layout_body_javascript() {
-	# bootstrap
+
```
vboctor: Remove extra blank line.

syncguru:
@dregad, @vboctor - Addressed your comments.

core/obsolete.php
```diff
@@ -203,4 +203,6 @@
 # changes in 2.0.0dev
 config_obsolete( 'icon_path' );
 config_obsolete( 'bug_print_page_fields' );
+config_obsolete( 'calendar_js_date_format' );
```
atrol: should be
`config_obsolete( 'calendar_js_date_format', 'datetime_picker_format' );`

core/obsolete.php
```diff
@@ -203,4 +203,6 @@
...
+config_obsolete( 'calendar_date_format' );
```
atrol: Should be
`config_obsolete( 'calendar_date_format', 'datetime_picker_format' );`

syncguru:
@atrol - Done
@dregad, @vboctor - can we get this in for today's release?

vboctor:
@dregad Please have a look that your feedback is addressed and merge once ready.

dregad:
Merged in 0253f467996740f7f778db6f80ca7f89ac39749a

PR #944, opened 2016-11-07
Re-implement issue note notifications

vboctor:
Email notifications for notes shouldn't include full issue information
Fixes #21876

Mentioned users in a new note shouldn't receive double notifications
Fixes #21877

Include note attachments information in issue note added notification
Fixes #21875

Explicitly pass issue note id to email notifications rather than getting latest note id
Fixes #21879

Improve issue note caching
Fixes #21878

Simple note added by a reporter
notification_simple_note

Public note with time tracking and attachments
notification_public_note_with_time_attach

Private note with time tracking and attachments
notification_private_note_time_attach

Public note with time tracking and attachments that are not visible to target user for the notification
notification_time_and_attach_not_visible

### DIFF
--- a/core/bugnote_api.php
+++ b/core/bugnote_api.php
@@ -130,20 +130,47 @@ class BugnoteData {
 function bugnote_exists( $p_bugnote_id ) {
 	$c_bugnote_id = (int)$p_bugnote_id;
 
+	global $g_cache_bugnote;
+	if( isset( $g_cache_bugnote[$c_bugnote_id] ) ) {
+		return true;
+	}
+
 	# Check for invalid id values
 	if( $c_bugnote_id <= 0 || $c_bugnote_id > DB_MAX_INT ) {
 		return false;
 	}
 
 	db_param_push();
-	$t_query = 'SELECT COUNT(*) FROM {bugnote} WHERE id=' . db_param();
+	$t_query = 'SELECT b.*, t.note
+			          	FROM      {bugnote} b
+			          	LEFT JOIN {bugnote_text} t ON b.bugnote_text_id = t.id
+						WHERE b.id = ' . db_param();
 	$t_result = db_query( $t_query, array( $c_bugnote_id ) );
+	$t_row = db_fetch_array( $t_result );
 
-	if( 0 == db_result( $t_result ) ) {
+	if( $t_row === false ) {
 		return false;
-	} else {
-		return true;
 	}
+
+	$t_bugnote = bugnote_row_to_object( $t_row );
+	bugnote_cache( $t_bugnote );
+	return true;
+}
+
+/**
+ * Caches the provided bugnote object.
+ *
+ * @param BugnoteData $p_bugnote The bugnote object.
+ * @return void
+ */
+function bugnote_cache( $p_bugnote ) {
+	global $g_cache_bugnote;
+
+	if( !isset( $g_cache_bugnote ) ) {
+		$g_cache_bugnote = array();
+	}
+
+	$g_cache_bugnote[(int)$p_bugnote->id] = $p_bugnote;
 }
 
 /**
@@ -277,7 +304,7 @@ function bugnote_add( $p_bug_id, $p_bugnote_text, $p_time_tracking = '0:00', $p_
 
 	# only send email if the text is not blank, otherwise, it is just recording of time without a comment.
 	if( true == $p_send_email && !is_blank( $t_bugnote_text ) ) {
-		email_bugnote_add( $p_bug_id );
+		email_bugnote_add( $t_bugnote_id );
 	}
 
 	return $t_bugnote_id;
@@ -289,7 +316,7 @@ function bugnote_add( $p_bug_id, $p_bugnote_text, $p_time_tracking = '0:00', $p_
  * @param  int $p_bug_id          The bug id
  * @param  int $p_bugnote_id      The bugnote id
  * @param  string $p_bugnote_text The bugnote text
- * @return void
+ * @return array User ids that received mentioned emails.
  * @access public
  */
 function bugnote_process_mentions( $p_bug_id, $p_bugnote_id, $p_bugnote_text ) {
@@ -302,7 +329,7 @@ function bugnote_process_mentions( $p_bug_id, $p_bugnote_id, $p_bugnote_text ) {
 
 	$t_removed_mentions_user_ids = array_diff( $t_mentioned_user_ids, $t_filtered_mentioned_user_ids );
 
-	mention_process_user_mentions(
+	return mention_process_user_mentions(
 		$p_bug_id,
 		$t_filtered_mentioned_user_ids,
 		$p_bugnote_text,
@@ -389,27 +416,8 @@ function bugnote_get_text( $p_bugnote_id ) {
  * @access public
  */
 function bugnote_get_field( $p_bugnote_id, $p_field_name ) {
-	static $s_vars;
-	global $g_cache_bugnote;
-
-	if( isset( $g_cache_bugnote[(int)$p_bugnote_id] ) ) {
-		return $g_cache_bugnote[(int)$p_bugnote_id]->$p_field_name;
-	}
-
-	if( $s_vars == null ) {
-		$s_vars = getClassProperties( 'BugnoteData', 'public' );
-	}
-
-	if( !array_key_exists( $p_field_name, $s_vars ) ) {
-		error_parameters( $p_field_name );
-		trigger_error( ERROR_DB_FIELD_NOT_FOUND, WARNING );
-	}
-
-	db_param_push();
-	$t_query = 'SELECT ' . $p_field_name . ' FROM {bugnote} WHERE id=' . db_param();
-	$t_result = db_query( $t_query, array( $p_bugnote_id ), 1 );
-
-	return db_result( $t_result );
+	$t_bugnote = bugnote_get( $p_bugnote_id );
+	return $t_bugnote->$p_field_name;
 }
 
 /**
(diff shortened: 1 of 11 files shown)


dregad:
> Include note attachments information in issue note added notification

Would it be worth adding a link to the attachment itself ?

vboctor:
> Would it be worth adding a link to the attachment itself ?

@dregad I don't think so, for the following reasons:

- We have the link to the note, which will have the embedded attachments.
- If user clicks on such links (and they are logged in), it will result in download rather than view.  This can be changed, but it is out of scope here.
- It will clutter the email with a lot of links.  Maybe we could consider when we have html emails, though I think the note link should do the trick.

dregad:
Thanks for the feedback. 

I'm OK with the change on principle, but haven't had time to test yet.

atrol:
> Email notifications for notes shouldn't include full issue information

I like having the whole information in the email notification.
Getting just the latest note forces me to login to Mantis to get an overview of the issue.

At the moment, users who don't like very long e-mails can  set "E-mail Notes Limit" in their account settings to shorten the message.
Introducing some new settings could allow them to get rid of the details, the description and the history if they want.

The shorten-functionality should also be available for other notification types.
At the moment it is a bit strange to get just a very short e-mail when a note is added, but still get the long e-mail for other changes (e.g. status or priority changed)

vboctor:
@atrol

> I like having the whole information in the email notification.
> Getting just the latest note forces me to login to Mantis to get an overview of the issue.

The current approach of dumping the issue and having the user scroll down to find the note that was added is not great (hidden between header information and history).  I've used multiple other systems (e.g. Github, Bitbucket, Intercom, Zendesk, JIRA, etc) and they either show just the note or the note + some header information after it.  I haven't seen the approach of dumping the whole conversation history (or latest N messages from it).  The users can go to Mantis or depending on their mail client conversations support assembling the conversations to see previous related messages.

> At the moment, users who don't like very long e-mails can set "E-mail Notes Limit"
> in their account settings to shorten the message.

That would not solve the issue, they will still have to fish for the note within an email that is longer than necessary.

> Introducing some new settings could allow them to get rid of the details,
> the description and the history if they want.

This would be too much complexity to accommodate for a variety of settings.  I would rather focus on a notification model that focuses on what changed.

> The shorten-functionality should also be available for other notification types.
> At the moment it is a bit strange to get just a very short e-mail when a note is added,
> but still get the long e-mail for other changes (e.g. status or priority changed)

Absolutely.  Notes is just the beginning.  It is a common scenario and it is one of our worst notifications.  After that we can do common scenarios like header information changes, like issue assigned, resolved, closed, etc.  After that we start html support, etc.

cproensa:
-1 This has changed the emails subjects and is inconsistent with the rest of standard notifications.
-1 The format is different to the rest of notifications. I'd expect at least the note block as it is currently formatted in the emails.

I wouldn't change the NOTE emails subject/format in general, and break homogeneity with the rest of workflow-standard notifications. This can confuse users and complicate email rules they may have.
This should be done to all the notifications emails.

The introduction of mention did not follow those rules, but to me it didn't matter much as i see them as special notifications (like previous reminders), which eventually is good to have a differing format (eg, subject line), to break thrugh user mail rules, or just to stand out.

+1 for optimization and clean up
+1 for shortening the note email

Regarding double notifications for mention and notes:
If i see mentions as a notification that must cut through the bulk of standard notifications, it's good to be a separate notification, at least until all the notifications are redesigned.

-1 at targeting such a relevant UX change to 1.3. I'd agree though if the subject/formatting was not altered.

vboctor:
@cproensa thanks for the feedback.  It is all addressed now and the screenshots in the PR description are now updated to reflect latest.

> The format is different to the rest of notifications. I'd expect at least the note block as it is currently formatted in the emails.

I updated the format to be similar the other notifications.  The note block is now consistent (with exception of inline attachments for notes that is already supported in the bug note added notification) and the code is re-used.

> I wouldn't change the NOTE emails subject/format in general, and break homogeneity with the rest of workflow-standard notifications

As was considering the same, since in this case it is not as important as mentions to pop out.  Changed to use standard subject.

> targeting such a relevant UX change to 1.3. I'd agree though if the subject/formatting was not altered.

It is now consistent.  So I'll keep it for 1.3.

atrol:
@vboctor 
> The current approach of dumping the issue and having the user scroll down to find the note that was added is not great

This is no reason to remove all other information. The added note could be shown as the first element of the e-mail.

> I've used multiple other systems ...

Same for me, and I don't like that I have to login to the systems to get the whole story.
Especially when I am not in intranet and have to establish my VPN connection in a first step.
At the moment I can first look at the e-mail on my smart phone and decide if there is a need to boot the laptop, connect via VPN, login to Mantis, ...
Having just the latest note is not enough for this decision, particularly if the issue is a quite old one that has not been updated for a long time.

A bit off topic:
We don't have undo/unerase functionality, so there is a nice side effect of the long e-mails: You have most of the information if someone accidentally deleted an issue.
This was two times my real use case before I decided to set $g_delete_bug_threshold = ADMINISTRATOR;

vboctor:
@atrol

> This is no reason to remove all other information. The added note could be shown as the first element of the e-mail.

Typically the new message which has the new note is part of an email conversation, so the information is there in your email client as part of your conversation view.  I would like to get to a more rich and targeted set of notifications, rather than the lazy logic of sending everything.  Once we have such rich and specific notifications, we can have plugins or even replacement emails to handle them and decide to send all the information as extra information.  Till then I think we should focus and hopefully soon more cosmetically pleasing and powerful email notifications.  The fact that all such tools are using this model means that it has been tested and proven by all these providers, of which some are internet only services and others who also support behind the firewall model.

> Especially when I am not in intranet and have to establish my VPN connection in a first step.

You can argue that systems are put behind a VPN for a reason.  People want to have access to them via the VPN and not work around that.  Having said that, again the email conversation should typically have the issue submitted and its associated notes.

> A bit off topic: We don't have undo/unerase functionality

As you said that is off topic, but emails are not a backup strategy and we will still have the emails to stitch things together.  You can limit deletes as you seem to be doing, backup your db, have a plugin that bcc all emails to a mailbox, etc.

vboctor:
@atrol I'm have been thinking about how to balance ability to improve notifications to provide more context about the change, be more concise and moving to html while still being able to provide the details you are looking for.  The approach I came up with is:

- Short / rich notifications will be the focus going forward and it would be the default experience.
- Provide a config option to use full issue notification similar to the one we have now.  This config option can be set globally and overridden for specific users.
- The full issue notifications (verbose) will be a single template similar to our current system with a hint about the change at the top + history at the bottom.

     $g_email_notifications_verbose = OFF;

We could decide to deprecate this in the future or keep it long term.  But what I really want to avoid is to invest a lot of energy into maintaining two different types of notifications or hybrid type that is in the patch you referred to.  We could enable plugins in the future to provide their own handling of notifications.

Let me know your thoughts.  In the meantime, I will work on this proposal.

vboctor:
@atrol I've pushed the change to support the config option to opt into verbose notification format.

dregad:
@vboctor I haven't tested yet, but think that's your latest proposed change of having by default a short message describing what changed, and an optional bit added after that with full details that can be switched on/off,  is a good transitional approach. 

>  This config option can be set globally and overridden for specific users.

If we're going to keep this moving forward, for convenience to users I would recommend that this setting becomes available as a user preference.

cproensa:
right, i think too that this could be a good moment to introduce a user preference for: short/verbose emails, that is availabe on the user prefs page.

A short type of notification is the idea that victor has proposed (independent of the formatting issues).
for a verbose one, we have two options:
- keep current behaviour for "dump" all the bug history
- do a mixed approach: write first the relevant change (as the shrot notifiacion does), and following, print the full info (the "dump", in lack of a better option)

In my personal case, i appreciate to see the previous notes in the email, to catch on the context of the issue. This helps when i'm receiving tons of emails, some of which may have a summary title very similar to each other. 
On the other side, its a pita to scroll the email to find the new note (which is not marked explicitly as so).  I'm not sure about the option for change sort order, since i still want to have the notes in ascending order in the web page...

vboctor:
@atrol @dregad @cproensa I've implemented the user preference UI on top of the configuration option for opting into full issue details.

Note that full issue details == old notifications, there is no change to make a mix of both and I would rather not add that.

atrol:
@vboctor thanks for the changes.
I hope to find some time for a deeper look and some tests the next days.

Is it a big deal to add the note as the first element of the e-mail also in long format?

vboctor:
Maintaining the two notification types already adds some complexity.  Adding the two types of notifications in one will mean more complexity in terms of code and designing an email template that looks decent.  So for now, I will keep the old model as is.  Once we have revised the rest of the notifications, designed the email templates, we can then consider whether there is a good way to incorporate feedback.  We will also assess the feedback on need for both formats.

I'm hoping to get this in this weekend.  So appreciate you trying the PR out.

core/bugnote_api.php
```diff
@@ -593,6 +613,28 @@ function bugnote_get_all_bugnotes( $p_bug_id ) {
 }
 
 /**
+ * Gets the bugnote object given its id.
+ *
+ * @param $p_bugnote_id The bugnote id.
```
atrol: param integer

core/email_api.php
```diff
@@ -868,12 +876,90 @@ function email_bug_updated( $p_bug_id ) {
 
 /**
  * send notices when a new bugnote
- * @param int $p_bug_id
- * @return null
+ * @param int $p_bugnote_id  The bugnote id.
+ * @param array $p_files The array of file information (keys: name, size)
+ * @param array The id of users to exclude.
```
atrol: array $p_exclude_user_ids

cproensa:
I suggest having this verbose=ON as default, at least for the 1.3 branch
- The new notification concept is not fully defined at this moment. Let this be some "experimental" setting
- Users may be confused on why some emails are short, and others verbose
- For the same reason, having this as ON as default would make sense when the "verbose" content could be removed from all notifications

Also, there's the issue of notes that are part of another action, like resolution, assign, etc, for which still there is no proposed change.

vboctor:
@atrol I've applied your feedback.
@cproensa I've set verbose as ON by default for now.

PR #945, opened 2016-11-08
Use DB_MAX_INT constant

cproensa:
Replace the fix applied for issue: #14014
Use a simpler check against  DB_MAX_INT constant


dregad:
Merged in da64cabab895d24c35e3f53b29e7c7afcc3eb578

PR #946, opened 2016-11-09
Proper fix for gravatar URL '&' encoding

dregad:
Follow-up fix on PR #920.

I did not realize that there was inconsistent output of Gravatars: `print_avatar()` escapes the URL with htmlspecialchars(), while `TimeLineEvent::html_start()` did not.

Fixes [#21844](https://www.mantisbt.org/bugs/view.php?id=21844), [#21804](https://www.mantisbt.org/bugs/view.php?id=21804)

### DIFF
--- a/core/classes/TimelineEvent.class.php
+++ b/core/classes/TimelineEvent.class.php
@@ -80,9 +80,9 @@ class TimelineEvent {
 
 		return sprintf(
 			'<div class="entry"><div class="avatar"><a href="%s"><img class="avatar" src="%s" alt="%s" width="32" height="32" /></a></div><div class="timestamp">%s</div>',
-			$t_avatar->link,
-			$t_avatar->image,
-			$t_avatar->text,
+			htmlspecialchars( $t_avatar->link ),
+			htmlspecialchars( $t_avatar->image ),
+			htmlspecialchars( $t_avatar->text ),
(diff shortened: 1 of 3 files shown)


core/classes/TimelineEvent.class.php
```diff
@@ -80,9 +80,9 @@ public function html_start() {
 
 		return sprintf(
 			'<div class="entry"><div class="avatar"><a href="%s"><img class="avatar" src="%s" alt="%s" width="32" height="32" /></a></div><div class="timestamp">%s</div>',
-			$t_avatar->link,
-			$t_avatar->image,
-			$t_avatar->text,
+			htmlspecialchars( $t_avatar->link ),
+			htmlspecialchars( $t_avatar->image ),
+			htmlspecialchars( $t_avatar->text ),
 			$this->format_timestamp( $this->timestamp )
```
dregad: I'm wondering if we need to htmlspecialchars() the timestamp too.

This is not consistently done in the code, but sometimes we call string_display_line() after formatting the date, and sometimes not.

dregad:
Given the 2 approvals, I'll merge this as-is.

Regarding my earlier question (see https://github.com/mantisbt/mantisbt/pull/946#pullrequestreview-8189232), I'll push a follow-up commit later if you guys think it's needed.

PR #947, opened 2016-11-11
Remove leading newline from email body

dregad:
This extra blank line prevents email clients such as Outlook from
displaying a preview of MantisBT's notification messages in a list view.

The newline does not serve any known purpose (PHPMailer already adds one
to separate the message header from its body), and testing does not show
any regression when notifications are sent sent without it.

This commit removes the extra '\n'.

Fixes [#21896](https://www.mantisbt.org/bugs/view.php?id=21896)


atrol:
👍

PR #950, opened 2016-11-13
Fix the position of parameters in query

cproensa:
The creation of parameters while building a query must happen in the
same order that they appear in the final sql string, and the order in
the array passed to db_query().

Fixes: #21912


cproensa:
@dregad 
> The commit message is missing the Fixes #21912 reference

fixed

dregad:
Merged in 28291bb896d7682ead0f0dfeeb54eda83d9d059b

PR #952, opened 2016-11-16
Bugnote api, use meaningful names and add types

cproensa:
- Rename global variables in bugnote_api used to cache bugnote data, to a
more defining name.
- Declare, initialize and document the cache variables.
- Add type hints to several function parameters

Fixes: #21884

### DIFF
--- a/core/bugnote_api.php
+++ b/core/bugnote_api.php
@@ -60,6 +60,13 @@ require_api( 'mention_api.php' );
 require_api( 'user_api.php' );
 require_api( 'utility_api.php' );
 
+# Cache of bugnotes arrays related to a bug, indexed by bug_id.
+# Each item is an array of BugnoteData objects
+$g_cache_bugnotes_by_bug_id = array();
+
+# Cache of BugnoteData objects, indexed by bugnote id
+$g_cache_bugnotes_by_id = array();
+
 /**
  * Bugnote Data Structure Definition
  */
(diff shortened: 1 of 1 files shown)


core/bugnote_api.php
```diff
@@ -60,6 +60,13 @@
 require_api( 'user_api.php' );
 require_api( 'utility_api.php' );
 
+# Cache of bugnotes arrays related to a bug, indexed by bug_id.
+# Each item is an array of BugnoteData objects
+$g_cache_bugnotes_by_bug_id = array();
+
+# Cache of BugnoteData objects, indexed by bugnote id
```
dregad: Same as above

cproensa:
> Shouldn't this be a PHPDoc block ?

I think currently there is not stablished a PHPdoc routine on global variables

I agree to do it if it's a better approach. What should be the phpdoc block like? (my ide does not have a template for this case)

dregad:
> I think currently there is not stablished a PHPdoc routine on global variables

There used to be one, maybe it was left out in a later version of the standard

cproensa:
> in a later version of the standard

i mean, a routine, in our code. We dont have phpdoc for global vars declaration

dregad:
Well it's currently not consistent... Examples

- Plain comment https://github.com/mantisbt/mantisbt/blob/master/core/html_api.php#L84
- Regular comment with `@global` tag https://github.com/mantisbt/mantisbt/blob/master/core/database_api.php#L43
- Proper PHPDoc block https://github.com/mantisbt/mantisbt/blob/master/core/email_api.php#L83
- and sometimes there's no comment at all.

Let's leave it like that.

dregad:
Merged in ffe67c9e13bc80533f27c65776d69666fbca5a0c

cproensa:
> Let's leave it like that.

https://www.mantisbt.org/bugs/view.php?id=21936

PR #953, opened 2016-11-18
Force logout after verification error

cproensa:
Force the clearing of authentication cookies when the verification data
submitted to account_update page produces an error.
This way, the user cant browse the site as the logged user, if he hasn't
completed yet the verification process.

Fixes #21709

### DIFF
--- a/account_update.php
+++ b/account_update.php
@@ -89,6 +89,9 @@ $t_update_realname = false;
 
 # Do not allow blank passwords in account verification/reset
 if( $t_account_verification && is_blank( $f_password ) ) {
+	# log out of the temporary login used by verification
+	auth_clear_cookies();
+	auth_logout();
 	error_parameters( lang_get( 'password' ) );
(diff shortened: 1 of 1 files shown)


cproensa:
As a side note:
After doing this, i was surprised that auth_logout() would not log out the user in this scenario
I think it's because the login was forced by manipulating cookies in some way.
I'd suggest to have a look at it, and normalize the logic behind. Or at least make auth_logout() to effectively log the user out in all cases.

dregad:
> After doing this, i was surprised that auth_logout() would not log out the user in this scenario
I think it's because the login was forced by manipulating cookies in some way.
I'd suggest to have a look at it, and normalize the logic behind. Or at least make auth_logout() to effectively log the user out in all cases.

@cproensa could you please open an issue to track this ?

dregad:
Merged in 12192f19887e7ebc69d702302f192fe65d8ac9b0

PR #954, opened 2016-11-19
Don't force lowercase for custom field column names

cproensa:
Custom field names are referenced by name in columns definitions.
There were some conversions in place to compare column names by
converting to lowercase.
However, the field name is stored in custom fields table without any
conversion. This causes:
- The configured columns array can be inconsistent with the available
  columns names.
- When the database uses case sensitive matching, the columns are not
  shown properly and are presented as not existant.

So given that custom field names are already stored with unrestricted
names, which include spaces and non-alphanumeric characters, forcing
them to lowercase while treating them is not a significant requirement
at the moment.

Fixes: #17367, #20248

### DIFF
--- a/core/columns_api.php
+++ b/core/columns_api.php
@@ -279,9 +279,7 @@ function column_get_custom_field_name( $p_column ) {
  * @access public
  */
 function columns_string_to_array( $p_string ) {
-	$t_string = utf8_strtolower( $p_string );
-
-	$t_columns = explode( ',', $t_string );
+	$t_columns = explode( ',', $p_string );
 	$t_count = count( $t_columns );
 
 	for( $i = 0; $i < $t_count; $i++ ) {


cproensa:
fyi
I haven't used the manage columns page for a long time, since i use the plugin inline-columns-config (https://github.com/mantisbt-plugins/Inline-column-configuration/pull/2). This plugin bypasses the lowercase conversion. And it has worked so far, so i guess this PR won't impact either.

vboctor:
If there is a configuration where a user has added "Custom1" then added it as a column "custom1", will this change break that?  Would this break already working cases and fix other cases that are not working?

cproensa:
Until now, with a custom fields called "Custom1", the user would have "custom_column1" in his columns config.
- with mysql, view page works, becasue "Column1"=="column1"
- with pgsql, view page cant show that field, becasue "Column1"!="column1", it then shows "@xxx@" and no values

With this fix, he user will be able to save a column config to be stored as "custom_Column1"
On mysql:
- nothing changes on view page. (it was working before)
- lang_get_defaulted() may fail, if the user created a $s_custom1 string, as it now searches for $s_Custom1 (*1)

So, this will not fix immediately a broken configuration. The user needs to modify his columns configuration to have it saved as case sensitive(*2). 

(*1) This could be an additional change, to look always for a lowerstring lang variable. But... given that we could have field names like "aa $@#! fff?", the lang system for custom fields is already flawed....

(*2) However, in that situation, capitalized fields names were never usable in the first place, and probably the user solved it by renaming the field.

cproensa:
The lang_get_defaulted() issue happens also in other places that is used: for example: bug-view page.
If the custom field name is "Custom1", the string variable must be called "$s_Custom1", to work.

So, if users have already made capitalized string variables, to match the capitalized custom fields names, this patch will, if anything, make the column header show the localized text properly.

If user had a lowercase string variable, to work with the column header, then he already had a non-working translation on bug-view page.

I don't think it's worth to add extra conversions atm.

dregad:
> (*1) This could be an additional change, to look always for a lowerstring lang variable. But... given that we could have field names like "aa $@#! fff?", the lang system for custom fields is already flawed....

This is the reason why the documentation requires that valid PHP identifiers are used for CF names (see 1st bullet under https://mantisbt.org/docs/master-1.3.x/en-US/Admin_Guide/html-desktop/#admin.customize.customfields.definitions)

Maybe we should actually enforce this ?

cproensa:
> Maybe we should actually enforce this ?

see: https://www.mantisbt.org/bugs/view.php?id=20512#c53225
The problem is that enforcing this would be a problem with existing invalid CF-names
And fixing that requires more careful steps.

dregad:
Right, I forgot about that discussion... At least I'm consistent ;-)

cproensa:
>  but if there is a place in the manual where we can clarify the recommended syntax and expected behaviors around it in the manual, that would be great.

as commented in #20512:
http://mantisbt.org/docs/master/en-US/Admin_Guide/html-desktop/#admin.customize.customfields.localize

But: 
- Custom field names are not enforced to meet that syntax
- Capitalized names would still meet the syntax indicated in documentation

#20512 should be a place to track a definitive solution to the issues.

PR #955, opened 2016-11-21
Fix broken layout on moving attachments action page

jllano:
- ensure to add styled elements

Fixes #21758


jllano:
Before:
screen shot 2016-11-21 at 11 16 02 pm

After:
screen shot 2016-11-21 at 11 14 16 pm

vboctor:
@jllano Based on the "after" screenshot, the formatting of the "File" column is still broken.  Also should we rename "File" to "File ID".  I can't recall what this field is for, but the name is not descriptive.

jllano:
screen shot 2016-11-22 at 12 20 41 pm

PR #956, opened 2016-11-22
Extend activation url expiry to 7 instead of 1 day

vboctor:
Fixes #21673


vboctor:
Merged into master-1.3.x via 049c6d9c0390e1c8bcdb167155d628d60be7c567

PR #959, opened 2016-11-26
Fix regression in bug_update_page

atrol:
Introduced by commit 1a547899a46467303a60732405c09fba87d68d19


dregad:
Sorry about that

PR #960, opened 2016-11-26
Update system requirements for 2.0

dregad:
Proposal following discussion in [#21841](https://www.mantisbt.org/bugs/view.php?id=21841)


dregad:
Regarding browser requirements, I went for a minimalistic approach. This is fully open to discussion.

- FF 45 because that's what the current ESR version is based upon
- IE 10 because anything to avoid known issues with older versions (was even tempted to set 11 but was not sure if it was necessary to be too restrictive)
- Other browsers did not specify any versions before, not sure if we should define one now

I googled a bit, and it seems that most systems / libraries out there tend to support whatever the "current" version is, as well as "current - 1". That is probably too restrictive for us, but I'm not quite sure what we should recommend as a minimum version.

@syncguru your feedback would be appreciated

vboctor:
@dregad I would merge this and then we can tweak later.  I would go with IE 11.  The issue is that as we use javascript libraries, supporting older browsers can be a pain.

dregad:
@vboctor If there are no objections, I'd suggest to merge this prior to releasing 2.0.0-rc.2

dregad:
X-post :wink:

PR #962, opened 2016-11-27
Fix regression setting version fields on View Issues

cproensa:
There are no entries in bulk operation menu of View Issues page to set:
Product Versions, Target Version or Fixed in Version
Regression in 1.3.4
Caused by d47f367fea8dfda1e5bce918419a1bda1dd0e10b

Fixes: #21963


dregad:
Merged in bfbb8de080b6a98de234c430f9dd76a524f25c16

PR #964, opened 2016-11-29
Allow moving attachments for multiple projects

jllano:
- replaced radio button with checkbox
- minor layout adjustments
- refactor logic in action page to make sure multiple project selection works

Fixes #21927

Before:

screen shot 2016-11-29 at 9 21 20 pm
After:

after

### DIFF
--- a/admin/move_attachments.php
+++ b/admin/move_attachments.php
@@ -257,28 +257,6 @@ function move_attachments_to_disk( $p_type, array $p_projects ) {
 	return $t_moved;
 }
 
-$t_moved = array();
-
-if ( null != $f_project_to_move ) {
-	foreach ( $f_project_to_move as $t_project_to_move) {
-		
-		$t_array = explode( ':', $t_project_to_move );
-
-		if( isset( $t_array[1] ) ) {
-			$f_project_id = $t_array[1];
-
-			switch( $t_array[0] ) {
-				case 'disk':
-					$t_moved[] = move_attachments_to_disk( $f_file_type, array( $f_project_id ) );
-					break;
-				case 'db':
-					$t_moved[] = move_attachments_to_db( $f_file_type, array( $f_project_id ) );
-					break;
-			}
-		}
-	}
-}
-
 form_security_purge( 'move_attachments_project_select' );
 
 # Page header, menu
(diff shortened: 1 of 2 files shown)


admin/move_attachments.php
```diff
@@ -257,20 +257,24 @@ function move_attachments_to_disk( $p_type, array $p_projects ) {
...
-				break;
+$t_moved = array();
+
+if ( null != $f_project_to_move ) {
+	foreach ( $f_project_to_move as $t_project_to_move) {
+		
+		$t_array = explode( ':', $t_project_to_move );
+
+		if( isset( $t_array[1] ) ) {
+			$f_project_id = $t_array[1];
```
vboctor: $f_field is only for form fields and not for temporary / calculated fields.  Please change to $t_project_id.

admin/move_attachments.php
```diff
@@ -257,20 +257,24 @@ function move_attachments_to_disk( $p_type, array $p_projects ) {
...
-			case 'disk':
-				$t_moved = move_attachments_to_disk( $f_file_type, array( $f_project_id ) );
-				break;
-			case 'db':
-				$t_moved = move_attachments_to_db( $f_file_type, array( $f_project_id ) );
-				break;
+$t_moved = array();
+
+if ( null != $f_project_to_move ) {
+	foreach ( $f_project_to_move as $t_project_to_move) {
```
vboctor: Remove space before ( and add a space after $t_project_to_move and before ).

admin/move_attachments.php
```diff
@@ -343,6 +351,7 @@ function move_attachments_to_disk( $p_type, array $p_projects ) {
 		echo '</div>';
 		echo '</div>';
 		echo '</div>';
+		echo "<br/>";
```
vboctor: By default use single quotes unless double quotes are needed.

admin/move_attachments.php
```diff
@@ -291,9 +295,13 @@ function move_attachments_to_disk( $p_type, array $p_projects ) {
 
 # Display results
 if( empty( $t_moved ) ) {
-	echo '<p class="lead">Nothing to do.</p>'. "\n";
+	echo '<div class="alert alert-danger">';
+		echo '<p class="lead"><strong>Opps!</strong> Please select the project you want to move the attachment.</p>';
```
vboctor: The old message indicated that there is no work to do and didn't mention anything about projects not selected.  Which can be the case of there were projects selected that had no attachments to be moved.  The new one seems to indicate that there is no project selected.  If we want the message to indicate that a project is selected, then we should look at project move list, otherwise, I would keep the old message to cover the case where there are no projects selected or projects selected, but no work to be done.

admin/move_attachments.php
```diff
@@ -343,6 +351,7 @@ function move_attachments_to_disk( $p_type, array $p_projects ) {
 		echo '</div>';
 		echo '</div>';
 		echo '</div>';
+		echo "<br/>";
 	}
 }
 echo "<br/>";
```
vboctor: Use single quote unless double quotes are needed.

jllano:
@vboctor I already addressed your comments above.

jllano:
@dregad I already addressed your concern.

PR #968, opened 2016-12-03
fix dropzone localization

badfiles:
Fixes #21925


vboctor:
Thanks @badfiles.

I prefer the old method since it only pays the overhead of fetching the config options from db when they are needed on the page.  Based on data from NewRelic based on MantisHub, the most expensive scripts are the ones that capture a bunch of data and outputs them to javascript/css/etc.  So I would rather see these scripts reduced / removed over time.

most-time-consuming-scripts

Thoughts?

dregad:
Alternative approach, using the correct form object instead of `$(this)` in common.js script, see #972

badfiles:
Close this PR in favor of  #972

PR #971, opened 2016-12-04
Modify history filter range query

cproensa:
To get history rows based on filter, use
SELECT .. FROM history WHERE bug_id IN ( filter query )
instead of
SELECT .. FROM history JOIN ( filter query )

It's a faster join, and generally is better optimized by different
database engines.

Fixes: #21998

### DIFF
--- a/core/history_api.php
+++ b/core/history_api.php
@@ -203,7 +203,7 @@ function history_get_range_result_filter( $p_filter, $p_start_time = null, $p_en
 		return db_empty_result();
 	}
 
-	$t_select_string = 'SELECT DISTINCT {bug}.id ';
+	$t_select_string = 'SELECT {bug}.id ';
 	$t_from_string = ' FROM ' . implode( ', ', $t_query_clauses['from'] );
 	$t_join_string = count( $t_query_clauses['join'] ) > 0 ? implode( ' ', $t_query_clauses['join'] ) : ' ';
 	$t_where_string = ' WHERE '. implode( ' AND ', $t_query_clauses['project_where'] );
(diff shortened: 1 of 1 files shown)


atrol:
Shouldn't be a problem for this use case, I just would like to be sure that you are aware of the Oracle limit of the IN clause: There is a maximum of 1000 values and you will get "ORA-01795: maximum number of expressions in a list is 1000" when exceeding the limit.

cproensa:
>  Oracle limit of the IN clause

I think that only applies to explicit IN values
When doing a subquery, its is treated as a semi-join.

I cant test later with actual queries, just to be sure

dregad:
> I think that only applies to explicit IN values

Correct, more specifically it applies to [*expression lists*](http://docs.oracle.com/database/122/SQLRF/Expression-Lists.htm#SQLRF52099).

core/history_api.php
```diff
@@ -203,7 +203,7 @@ function history_get_range_result_filter( $p_filter, $p_start_time = null, $p_en
 		return db_empty_result();
 	}
 
-	$t_select_string = 'SELECT DISTINCT {bug}.id ';
+	$t_select_string = 'SELECT {bug}.id ';
```
dregad: Out of curiosity, do we get better performance without the *DISTINCT* ?

cproensa: Note that the filter query can return duplicated rows.
Without DISTINCT in the subquery, using a plain JOIN, DISTINCT is needed at top level select.

Looking at that query, the reported cost is slightly higher (oracle, postgres)
and execution time, in some quick tests, this is having ~10k history rows returned (the usual chunk fetched by timeline):
oracle : slightly slower than semi-join
postgres: slightly faster than semi-join
mysql (mariadb): roughly the same semi-join

However, the distinct form scales worse with result size
testing mysq, going into 1 million history rows returned:
select distinct: ~30 secs
semi-join: ~3 secs

cproensa: > do we get better performance without the DISTINCT ?

if you mean
SELECT .. WHERE bug_id IN (SELECT {bug}.id from... )
against
SELECT .. WHERE bug_id IN (SELECT **DISTINCT** {bug}.id from... )

oracle, mysql, plans are identical, so its already detecting the semijoin and optimizing.
postgres, reports doing an inner sort-unique that adds to cost and time.

dregad: Yes, that's what I meant. Thanks for the follow-up !

PR #972, opened 2016-12-05
Fix dropzone localization

dregad:
Fixes [#21925](https://mantisbt.org/bugs/view.php?id=21925)

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -519,6 +519,6 @@ function enableDropzone( classPrefix, autoUpload ) {
 			}
 		});
 	} catch (e) {
-		alert( $(this).data('dropzone-not-supported') );
+		alert( form.data('dropzone-not-supported') );
 	}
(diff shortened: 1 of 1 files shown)


js/common.js
```diff
@@ -471,29 +471,31 @@ function toggleDisplay(idTag)
 Dropzone.autoDiscover = false;
 function enableDropzone( classPrefix, autoUpload ) {
 	try {
-		var zone = new Dropzone( "." + classPrefix + "-form", {
+		var formClass = "." + classPrefix + "-form";
+		var form = $( formClass );
+		var zone = new Dropzone( formClass, {
 			forceFallback: $(this).data('force-fallback'),
```
badfiles: you've missed this line

syncguru: Should we remove this $(this) here too?

PR #974, opened 2016-12-06
Trigger EVENT_BUG_DELETED in core function

badfiles:
Fixes #0022013

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -1351,6 +1351,8 @@ function bug_delete( $p_bug_id ) {
 	# call pre-deletion custom function
 	helper_call_custom_function( 'issue_delete_validate', array( $p_bug_id ) );
 
+	event_signal( 'EVENT_BUG_DELETED', array( $c_bug_id ) );
+
 	# log deletion of bug
 	history_log_event_special( $p_bug_id, BUG_DELETED, bug_format_id( $p_bug_id ) );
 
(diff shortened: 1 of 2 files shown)


cproensa:
This event actually should be triggered in bug_api - bug_delete()

atrol:
> should be triggered in bug_api - bug_delete()

Right, and it should be removed from bug_actiongroup.php

core/bug_api.php
```diff
@@ -1402,6 +1402,8 @@ function bug_delete( $p_bug_id ) {
 	$t_query = 'DELETE FROM {bug} WHERE id=' . db_param();
 	db_query( $t_query, array( $c_bug_id ) );
 
+	event_signal( 'EVENT_BUG_DELETED', array( $c_bug_id ) );
```
cproensa: According to current behaviour, and documentation, this event is triggered before deleting the bug and related items.

In this function, we already have 'issue_delete_validate' and 'issue_delete_notify' custom functions.
I wonder if we should have two events: one for validation, and one for notification, like we have for bug create and update. 
If a plugin must perform some action at bug deletion, it should be done after all other plugins have validated the action.
This scenario is less common, in my opinion,  that the update events one, but still worth considering.

On the other hand, any one of those events, should be triggered before the actual delete of data. Otherwise the information would not be available for plugins to do do their stuff.

For a simple approach here, i'd just place the current event in the same place as 'issue_delete_validate'

cproensa:
@badfiles 
please open a bug to track the change, as this is related, and would solve at least:
- event not called when deleting a project, and its bugs
- event not called when deleting bugs from soap api

badfiles:
ok

vboctor:
@badfiles Please base this PR on master-1.3.x.

badfiles:
I thought it was already based on master-1.3.x.

dregad:
> I thought it was already based on master-1.3.x.

It is (since https://github.com/mantisbt/mantisbt/pull/974#event-884178954)

cproensa:
For me it's ok too

PR #975, opened 2016-12-11
Prevent inline error in RSS output

cproensa:
Disable inline error reporting to not mess with RSS output.

Fixes: #22031


dregad:
At some point we should probably look for a replacement for this library...

PR #976, opened 2016-12-12
Timeline view threshold option added. Closes #592

bastilimbach:
In the PR #592, @schriftgestalt added the functionality to set the visibility of the timeline. Due to [#529 (comment)](https://github.com/mantisbt/mantisbt/pull/592#issuecomment-93680455) the PR was never merged. So I added the fall back to a 1.2.x behaviour of the view, mentioned by @dregad and rebased it to 2.0.

**$g_timeline_view_threshold = VIEWER;**
screen shot 2016-12-12 at 18 29 14

**$g_timeline_view_threshold = NOBODY;**
screen shot 2016-12-12 at 18 29 42

### DIFF
--- a/docbook/Admin_Guide/en-US/config/defaults.xml
+++ b/docbook/Admin_Guide/en-US/config/defaults.xml
@@ -39,6 +39,15 @@
 				</para>
 			</listitem>
 		</varlistentry>
+        <varlistentry>
+            <term>$g_timeline_view_threshold</term>
+            <listitem>
+                <para>Threshold for viewing timeline information. Use NOBODY to turn it off.
+                    If the timeline is turned off, the other widgets are displayed in a two column view.
+                    The default is VIEWER.
+                </para>
+            </listitem>
+		</varlistentry>
 		<varlistentry>
 			<term>$g_default_reminder_view_status</term>
 			<listitem>
--- a/my_view_page.php
+++ b/my_view_page.php
@@ -89,13 +89,16 @@ reset( $t_boxes );
 #print_r ($t_boxes);
 
 $t_project_id = helper_get_current_project();
+$t_timeline_view_threshold_access = access_has_project_level( config_get( 'timeline_view_threshold' ) );
+$t_timeline_view_class = ( $t_timeline_view_threshold_access ) ? "col-md-7" : "col-md-6";
 ?>
-<div class="col-md-7 col-xs-12">
+<div class="col-xs-12 <?php echo $t_timeline_view_class ?>">
 
 <?php
 $t_number_of_boxes = count ( $t_boxes );
 $t_boxes_position = config_get( 'my_view_boxes_fixed_position' );
 $t_counter = 0;
+$t_two_columns_applied = false;
 
 define( 'MY_VIEW_INC_ALLOW', true );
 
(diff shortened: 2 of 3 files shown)


my_view_page.php
```diff
@@ -89,13 +89,17 @@
...
 $t_project_id = helper_get_current_project();
+
+# Add class "col-md-7" only if the timeline is shown.
 ?>
-<div class="col-md-7 col-xs-12">
+<div class="col-xs-12 <?php echo ( access_has_project_level(config_get( 'timeline_view_threshold' )) ) ? "col-md-7" : "col-md-6"; ?>">
 
 <?php
 $t_number_of_boxes = count ( $t_boxes );
+$t_half_number_of_boxes = $t_number_of_boxes / 2;
```
bastilimbach: This variable is unnecessary.

docbook/Admin_Guide/en-US/config/defaults.xml
```diff
@@ -39,6 +39,13 @@
 				</para>
 			</listitem>
 		</varlistentry>
+        <varlistentry>
+            <term>$g_timeline_view_threshold</term>
+            <listitem>
+                <para>The default viewing status for the timeline. The default is VIEWER.
```
vboctor: Revise text to something like: "Threshold for viewing timeline information.  Use NOBODY to turn it off.  The default is VIEWER."

my_view_page.php
```diff
@@ -89,13 +89,17 @@
 #print_r ($t_boxes);
 
 $t_project_id = helper_get_current_project();
+
+# Add class "col-md-7" only if the timeline is shown.
 ?>
-<div class="col-md-7 col-xs-12">
+<div class="col-xs-12 <?php echo ( access_has_project_level(config_get( 'timeline_view_threshold' )) ) ? "col-md-7" : "col-md-6"; ?>">
```
vboctor: It would be better to calculate the the class in line 92, then just use a variable here.  Also review the spacing around ().  For example: 

    access_has_project_level( config_get( 'timeline_view_threshold' ) ) )

vboctor: Cache the config_get( 'timeline_view_threshold' ) in a $t_timeline_view_threshold variable rather than getting it twice within the page.  Here and when including the timeline information below.

vboctor:
@bastilimbach Looks good.  Can you squash into a single commit and refer to the bug in our tracker.  The commit message should look as follows:

"Support turning off timeline from My View page

Fixes #19586"

bastilimbach:
@vboctor Something went wrong with the squash. The master branch worked but the rebase and squash of feature/timeline-threshold didn't. I'll look into that.

bastilimbach:
@vboctor Squash issue should be resolved. Every change is now squashed into one commit.

bastilimbach:
Great thank you @vboctor

vboctor:
Thanks to you @bastilimbach for your contribution.

dregad:
As mentioned in [#19586](http://www.mantisbt.org/bugs/view.php?id=19586#c54879) The 2-column display does not respect the $g_my_view_boxes_fixed_position setting.

bastilimbach:
@dregad I wasn't aware of this setting. It seems as if it had never been used or either documented?! What is the matter of this setting?

tgulacsi:
This is new, to allow disabling the timeline.
My users esp. hate the timeline, just as every change.
I don't know what will I get on Monday for the 2.0.0-rc2...

Sebastian Limbach <notifications@github.com> ezt írta (időpont: 2016. dec.
30., P 19:01):

bastilimbach:
@tgulacsi The functionality of disabling the timeline has nothing to do with the "$g_my_view_boxes_fixed_position" variable. I added this functionality.

dregad:
> I wasn't aware of this setting. It seems as if it had never been used or either documented?! 

@bastilimbach it has been around for ages (see c1ea751dc65650f20d9d7c6c6610d5886b1cb09e), but it's only documented in config_defaults_inc.php. As far as usage is concerned, it is [referenced in my_view_page.php](https://github.com/mantisbt/mantisbt/blob/master/my_view_page.php#L99)

bastilimbach:
@dregad I see. But the functionality got removed so the variable is obsolete. Right?

dregad:
That was my guess, hence suggestion to make the config obsolete, http://www.mantisbt.org/bugs/view.php?id=22096

bastilimbach:
Ok great. How do you handle an obsolete variable? Do I have to create a fall back or just add a warning to the documentary?

dregad:
See https://github.com/mantisbt/mantisbt/blob/master/core/obsolete.php to get an idea of how it's done, but before implementing anything I suggest to wait until a decision is taken whether to remove the config, or reimplement the functionality as it was in 1.3.x. Please continue the discussion in issue [#22096](http://www.mantisbt.org/bugs/view.php?id=22096).

PR #977, opened 2016-12-12
Allow redirect to an absolute path

val-kulkov:
At present, html_meta_redirect() supports redirection to a relative path only. This patch enables handling of both relative and absolute paths by html_meta_redirect().

Closes bug report #22037.

Signed-off-by: Val Kulkov <val.kulkov@gmail.com>

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -254,7 +254,7 @@ function html_css_cdn_link( $p_url ) {
  * $p_time is the number of seconds to wait before redirecting.
  * If we have handled any errors on this page return false and don't redirect.
  *
- * @param string  $p_url      The page to redirect: has to be a relative path.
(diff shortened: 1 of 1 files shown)


cproensa:
$p_sanitize=true conflicts with having a non-relative path to the application's path

I copied the function into a test script, calling it "test_redirect", and print the output:

test_redirect('/download/manual.pdf');
URL=http://localhost:10080download/manual.pdf"

test_redirect('/download/manual.pdf',null, false);
URL=http://localhost:10080/download/manual.pdf"

This means, to properly work in your proposed scenario, $p_sanitize must be false.

val-kulkov:
@cproensa : I agree, the conflict between parameters is intolerable. Please close or delete this PR.

PR #981, opened 2016-12-26
Fix regression for function tag_bug_get_row()

cproensa:
Fix regression caused by changes in function tag_bug_get_row() where parameter
$p_tag_id is no longer used.
The modified function was returning an array of all tag rows for the bug_id,
instead of a single tag row, if it exists.

Fixes: #22074

### DIFF
--- a/core/tag_api.php
+++ b/core/tag_api.php
@@ -709,11 +709,11 @@ function tag_bug_get_row( $p_tag_id, $p_bug_id ) {
 		tag_cache_bug_tag_rows( array( $c_bug_id ) );
 	}
 
-	$t_row = $g_cache_bug_tags[$c_bug_id];
-	if( !$t_row ) {
+	$t_bug_tags = $g_cache_bug_tags[$c_bug_id];
+	if( !$t_bug_tags || !isset( $t_bug_tags[$p_tag_id] ) ) {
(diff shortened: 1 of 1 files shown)


atrol:
-1
Now I am getting
SYSTEM NOTICE: 'Undefined index: user_id' in '/var/www/html/master-1.3.x/core/tag_api.php' line 828

cproensa:
> Now I am getting
SYSTEM NOTICE: 'Undefined index: user_id' in '/var/www/html/master-1.3.x/core/tag_api.php' line 828

I don't see that error now.
What is exactly the action that raises it?

atrol:
User U1 attaches tag T1 to an issue.
I get the notice when user U2 removes tag T1 form the issue.

atrol:
Strange,
don't get it any longer

cproensa:
That was the scenario i tested too.
The error you say is in: tag_api.php' line 828
which is the same as before. I can see all the conditions being evaluated fine in that line...

Unles i am looking at it wrong... Are you sure you applied the commit?

cproensa:
> Strange,
don't get it any longer

ok

atrol:
> cproensa  wants to merge 1 commit into mantisbt:master

I merged to master-1.3.x

PR #982, opened 2016-12-29
Remove jquery_ui due to size & limited use

syncguru:
Jquery UI library size (minified):
js: 240k (largest js file we have today)
css: 30k

No reason to include jquery-ui in core product since it is not used in modern ui.

Fixes: #21881


js/common.js
```diff
@@ -89,25 +89,6 @@ $(document).ready( function() {
         SetCookie("collapse_settings", t_cookie);
     });
 
-    $('input[type=text].autocomplete').autocomplete({
```
vboctor: Looks like there are fields on bug report / update pages that used the auto-complete functionality.  These are platform, OS, and OS_Build.  You can find them by searching for "autocomplete" in the code.  Were these working?  If so, then we should switch them to another approach.

dregad:
I have no objections against removing jQueryUI. I suggest this is done prior to 2.0.0 release. 

In parallel with the merge, the [jQueryUI plugin](https://github.com/mantisbt-plugins/jQuery-UI) should be updated to the latest version of jQueryUI, and tested to make sure it is compatible with Mantis 2.x branch, to offer a fall-back for users who rely on the removed functionality (e.g. for their plugins).

dregad:
Since this was not removed in 2.0.0, it should be targeted at 2.1 release

syncguru:
@vboctor You are right. I totally overlooked autocomplete usage

I replaced jquery-ui autocomplete with typeahead plugin that is bundled with Ace theme which comes already styled. Here is a snapshot of what it looks like now:

screen shot 2016-12-30 at 10 51 37 pm

PR #983, opened 2017-01-01
Don’t show obsolete versions in Changelog

vboctor:
Fixes #11604

### DIFF
--- a/changelog_page.php
+++ b/changelog_page.php
@@ -240,7 +240,7 @@ foreach( $t_project_ids as $t_project_id ) {
 
 	$t_resolved = config_get( 'bug_resolved_status_threshold' );
 
-	# grab version info for later use
+	# grab released versions info for later use, excluding obsolete ones
 	$t_version_rows = version_get_all_rows( $t_project_id, null, false );
 
 	# cache category info, but ignore the results for now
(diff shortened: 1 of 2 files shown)


changelog_page.php
```diff
@@ -241,7 +241,7 @@ function print_project_header_changelog( $p_project_name ) {
 	$t_resolved = config_get( 'bug_resolved_status_threshold' );
 
 	# grab version info for later use
-	$t_version_rows = version_get_all_rows( $t_project_id, null, false );
+	$t_version_rows = version_get_all_rows( $t_project_id, /* released */ null, /* obsolete */ false );
```
dregad: Instead of cluttering the code with inline comments describing each parameter, it would be better to have
```
# grab released versions info for later use, excluding obsolete ones
```

atrol: There are some other places in code where it's done in a similar way.
Introducing some defines (INCLUDE_RELEASED_VERSIONS, EXLUDE_RELEASED_VERSIONS, ...) could also be a way to enhance readability.

dregad: That's a good option too.

vboctor: @dregad I disagree that this is code clutter.  I think it is the a good way to provide annotation of parameters where it is not clear.  My other alternative would be to assign these values to variables and then use them.

@atrol I also thought about the use of constants, but I think it will cause too many constants and will have the risk of being re-used in a context where the value doesn't match the target function expectation, hence, I would rather use defines for regular constants and enums, rather than as a way to provide a friendly name to values that are null, true, and false.

Having said that, I have used @dregad suggest for this one.  But we should revisit this.

PR #987, opened 2017-01-07
Fix negative values in Issue trends graph

dregad:
In mantisbt.org's bugtracker (and probably in other ancient trackers
running since the early days of MantisBT as well), there are issues
having a creation date newer than the last_updated date.

It is not confirmed whether this is a glitch, or a consequence of the
software's evolution. To avoid incorrect display in the Issue Trend
graph, we apply a simple fix while building the data set.

Fixes [#22065](http://www.mantisbt.org/bugs/view.php?id=22065).

### DIFF
--- a/plugins/MantisGraph/core/graph_api.php
+++ b/plugins/MantisGraph/core/graph_api.php
@@ -431,12 +431,12 @@ function create_cumulative_bydate() {
 	$t_specific_where = helper_project_specific_where( $t_project_id, $t_user_id );
 
 	# Get all the submitted dates
-	$t_query = 'SELECT date_submitted FROM {bug} WHERE ' . $t_specific_where . ' ORDER BY date_submitted';
+	$t_query = 'SELECT date_submitted, last_updated FROM {bug} WHERE ' . $t_specific_where . ' ORDER BY date_submitted';
 	$t_result = db_query( $t_query );
 
 	while( $t_row = db_fetch_array( $t_result ) ) {
 		# rationalise the timestamp to a day to reduce the amount of data
-		$t_date = $t_row['date_submitted'];
+		$t_date = min( $t_row['date_submitted'], $t_row['last_updated'] );
 		$t_date = (int)( $t_date / SECONDS_PER_DAY );
 
 		if( isset( $t_calc_metrics[$t_date] ) ) {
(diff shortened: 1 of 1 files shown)


vboctor:
Since this is not a common issue and is likely to impact other reports / logic, should we wait until we do a schema change and add a step to fix issues where date submitted > date modified?

dregad:
I guess it does make sense to fix the data.

Since the fix is ready, is quite low risk and improves performance, I would propose to merge the change anyway, and I'll submit a follow-up PR for the data fix.

plugins/MantisGraph/core/graph_api.php
```diff
@@ -431,13 +431,25 @@ function create_cumulative_bydate() {
 	$t_specific_where = helper_project_specific_where( $t_project_id, $t_user_id );
 
 	# Get all the submitted dates
-	$t_query = 'SELECT date_submitted, last_updated FROM {bug} WHERE ' . $t_specific_where . ' ORDER BY date_submitted';
+	# Calculate the minimum value between submitted and last updated dates to
+	# account for data inconsistencies in very old versions of Mantis, where
+	# issues have an updated date older than their creation date (see #22065).
+	# We also rationalise the timestamp to a day to reduce the amount of data
+	$t_query = 'SELECT FLOOR(min_date / ' . SECONDS_PER_DAY. ') AS date_submitted
```
k1931547: invalid sql

vboctor:
@dregad I would rather just go with the data fix for the following reason:

- If we checkin this temp fix, then it would make sense to undo it once the proper data fix is in.  Which is unnecessary churn.
- The checkin + revert make sense if there is a demand from users (beside us, since our data may have been messed up by some process or script that devs or dev code ran at one point).  You can even argue that we should just fix our data until we hear the same issue from users.

Following are the options:

1. Just fix our data and wait for users to report the issue before adding an upgrade step.
2. Fix the data via upgrade step (to be added as we open db schema changes).
3. Checkin this temp fix and revert it as part of the fix the data upgrade step change.  I like this model when real users are broken and need a workaround until a more comprehensive fix is available.  Not sure this applies here though.

I would start with 1, add 2 as needed.  But I can be convinced otherwise (not end of the world).

dregad:
I would personally have gone for 3, but I guess you may be right that this issue may be specific to mantisbt.org due to its history. So I'll write a quick SQL script to patch our data, and solve this issue without any code change for now.

PR #991, opened 2017-01-11
Bump the required MantisCore version in bundled plugins

atrol:


k2152262:
@vboctor @giallu @daryn @rombert @atrol @grangeway @dregad @dentxinho @bieli @tomcarlson @fengkuok @syncguru @diegovice @caribe @fager @pikiel @nikoutel @6run0 @AgnidykhBayky  @sayyidsh @pixelstreamAu @johnms @mikeching @g-taki @benv8nam @stew721 @sayazyi @mamoniem @lezouretalice @suriyaakudolsc @solovets @kacmem @xiaojiangang @chenyongze @myu9882 @markmarco16 @mgretler @lincolnwung @zedxxx @krime @scottmatheina @faviogalvis @vand3rson @like9729 @badfiles @zenithtekla @sureshinde @peterdavehello @netsf @csatwood @miao606 @bitlin @ioroid @supertassu @ulkuderner @vbathke @hersonpc @jrtorres42 @thejuanwhocode @cyuankai @mantianxingxing @boby75 @lordshell @burningwind  @nospam1961 @planetbrett @gantha @pbcreativesolutions @bastilimbach @ltpgt @modir  +1 I agree

atrol:
@dregad not sure you are aware what has been changed
This is not the version of the plugin.
This version is always in sync as we have
$this->version = MANTIS_VERSION;

This is the dependence to MantisCore which is not met, thus all plugins are broken.

dregad:
Yes I realized that after posting. I just merged & updated the tracker.

vboctor:
Why should avoid having the need to always update this.

- Can we remove the dependency since these are plugins that are distributed with core, so no need to check the version?
- Can we have the dependency to be > 2.0.0 rather than exactly 2.0.0?

vboctor:
Looks like the discussion is happening at http://www.mantisbt.org/bugs/view.php?id=22171

dregad:
> Can we have the dependency to be > 2.0.0 rather than exactly 2.0.0?

It is. The problem is that the plugin API silently adds a `< x.y+1 ` so the effective version requirement is `(>=) 2.0.0, < 2.1` in this case. This was added due to BC issues between 1.2 and 1.3. As per discussion in in [#22171](http://www.mantisbt.org/bugs/view.php?id=22171), we should change the logic so the plugin API does `< x+1`, reflecting the fact that BC changes should on principle only happen in major releases.

vboctor:
@dregad sounds reasonable.

PR #993, opened 2017-01-12
Markdown: Enable line breaks by default

jllano:
- this will enable newlines to insert line break <br>

Fixes #22172


jllano:
Before:
screen shot 2017-01-12 at 11 52 25 pm

After:
screen shot 2017-01-12 at 11 52 43 pm

PR #995, opened 2017-01-16
Plugin requirements

dregad:
This PR addresses the issue of plugins getting disabled when bumping the MANTIS_VERSION to 2.1.x
[#22171](http://www.mantisbt.org/bugs/view.php?id=22171).

It also contains the following changes and improvements
- PHPUnit tests for plugin_dependency()  function -- I may have missed test cases, so your **suggestions for additional test cases are welcome**
- Refactor plugin_version_check() to use PHP's version_compare() function instead of custom code
- Proper handling of "dev" versions (with [version_compare()](http://php.net/manual/en/function.version-compare.php), 2.1.0-dev is < than 2.1.0)
- Reset MANTIS_VERSION to 2.1.
- Revert MantisCore version requirements in bundled plugins
- Fix PHP notice, align code with documentation [#22205](http://www.mantisbt.org/bugs/view.php?id=22205)
- Improved plugin documentation in Developers guide [#22206](http://www.mantisbt.org/bugs/view.php?id=22206)

### DIFF
--- a/core/plugin_api.php
+++ b/core/plugin_api.php
@@ -447,88 +447,20 @@ function plugin_is_loaded( $p_base_name ) {
 }
 
 /**
- * Converts a version string to an array, using some punctuation and
- * number/lettor boundaries as splitting points.
- * @param string $p_version Version string.
- * @return array Version array
- */
-function plugin_version_array( $p_version ) {
-	$t_version = preg_replace( '/([a-zA-Z]+)([0-9]+)/', '\1.\2', $p_version );
-	$t_version = preg_replace( '/([0-9]+)([a-zA-Z]+)/', '\1.\2', $t_version );
-
-	$t_search = array(
-		',',
-		'-',
-		'_',
-	);
-
-	$t_replace = array(
-		'.',
-		'.',
-		'.',
-	);
-
-	$t_version = explode( '.', str_replace( $t_search, $t_replace, $t_version ) );
-
-	return $t_version;
-}
-
-/**
- * Checks two version arrays sequentially for minimum or maximum version dependencies.
- * @param array   $p_version1 Version array to check.
- * @param array   $p_version2 Version array required.
+ * Checks two versions for minimum or maximum version dependencies.
+ * @param string  $p_version  Version number to check.
+ * @param string  $p_required Version number required.
  * @param boolean $p_maximum  Minimum (false) or maximum (true) version check.
  * @return integer 1 if the version dependency succeeds, -1 if it fails
  */
-function plugin_version_check( array $p_version1, array $p_version2, $p_maximum = false ) {
-	while( count( $p_version1 ) > 0 && count( $p_version2 ) > 0 ) {
-
-		# Grab the next version bits
-		$t_version1 = array_shift( $p_version1 );
-		$t_version2 = array_shift( $p_version2 );
-
-		# Convert to integers if possible
-		if( is_numeric( $t_version1 ) ) {
-			$t_version1 = (int)$t_version1;
-		}
-		if( is_numeric( $t_version2 ) ) {
-			$t_version2 = (int)$t_version2;
-		}
-
-		# Check for immediate version differences
-		if( $p_maximum ) {
-			if( $t_version1 < $t_version2 ) {
-				return 1;
-			} else if( $t_version1 > $t_version2 ) {
-				return -1;
-			}
-		} else {
-			if( $t_version1 > $t_version2 ) {
-				return 1;
-			} else if( $t_version1 < $t_version2 ) {
-				return -1;
-			}
-		}
-	}
-
-	# Versions matched exactly
-	if( count( $p_version1 ) == 0 && count( $p_version2 ) == 0 ) {
-		return 1;
-	}
-
-	# Handle unmatched version bits
+function plugin_version_check( $p_version, $p_required, $p_maximum = false ) {
 	if( $p_maximum ) {
-		if( count( $p_version2 ) > 0 ) {
-			return 1;
-		}
+		$t_operator = '<';
 	} else {
-		if( count( $p_version1 ) > 0 ) {
-			return 1;
-		}
+		$t_operator = '>=';
 	}
-
-	# No more comparisons
-	return -1;
+	$t_result = version_compare( $p_version, $p_required, $t_operator );
+	return $t_result ? 1 : -1;
 }
 
 /**
@@ -593,8 +525,6 @@ function plugin_dependency( $p_base_name, $p_required, $p_initialized = false ) 
 			}
 		}
 
-		$t_version_installed = plugin_version_array( $t_plugin_version );
-
 		foreach( $t_required_array as $t_required ) {
 			$t_required = trim( $t_required );
 			$t_maximum = false;
@@ -612,9 +542,7 @@ function plugin_dependency( $p_base_name, $p_required, $p_initialized = false ) 
 				}
 			}
 
-			$t_version_required = plugin_version_array( $t_required );
-
-			$t_check = plugin_version_check( $t_version_installed, $t_version_required, $t_maximum );
+			$t_check = plugin_version_check( $t_plugin_version, $t_required, $t_maximum );
 
 			if( $t_check < 1 ) {
 				return $t_check;
--- a/core/plugin_api.php
+++ b/core/plugin_api.php
@@ -458,6 +458,11 @@ function plugin_version_check( $p_version, $p_required, $p_maximum = false ) {
 		$t_operator = '<';
 	} else {
 		$t_operator = '>=';
+
+		# Special handling for dev versions: we consider them as meeting the
+		# requirement for their baseline, i.e. 2.0.0-dev matches 2.0.0.
+		preg_match( '/^(.*?)([._+-]?dev.*)?$/', $p_version, $t_matches );
+		$p_version = $t_matches[1];
 	}
 	$t_result = version_compare( $p_version, $p_required, $t_operator );
 	return $t_result ? 1 : -1;
--- a/core/plugin_api.php
+++ b/core/plugin_api.php
@@ -494,38 +494,20 @@ function plugin_dependency( $p_base_name, $p_required, $p_initialized = false ) 
 
 		$t_required_array = explode( ',', $p_required );
 
-		# If the plugin's minimum dependency for MantisCore is unspecified or
-		# lower than the current release (i.e. the plugin does not specifically
-		# list the current core version as supported) and the plugin does not
-		# define a maximum dependency, we add one with the current version's
-		# minor release (i.e. for 1.3.1 we would add '<1.3').
-		# The purpose of this is to avoid compatibility issues by disabling
-		# plugins which have not been updated for a new Mantis release; authors
-		# have to revise their code (if necessary), and release a new version
-		# of the plugin with updated dependencies.
-		# To indicate compatibility (e.g. with release 1.3), one can either:
-		# 1. update the minimum required version (i.e. plugin only works with
-		#    1.3; 1.2-compatible code is maintained separately):
-		#    $this->requires = array( 'MantisCore' => '1.3' );
-		# 2. add the new release as a 2nd minimum version (i.e. the plugin is
-		#    compatible with both versions):
-		#    $this->requires = array( 'MantisCore' => '1.2, 1.3' );
-		# 3. add a maximum version higher than the new release:
-		#    $this->requires = array( 'MantisCore' => '1.2, <2.0' );
-		#    Note that this may cause the plugin to face compatibility issues
-		#    if and when a version 1.4 is released.
+		# Set maximum dependency for MantisCore if none is specified.
+		# This effectively disables plugins which have not been specifically
+		# designed for a new major Mantis release to force authors to review
+		# their code, adapt it if necessary, and release a new version of the
+		# plugin with updated dependencies.
 		if( $p_base_name == 'MantisCore' && strpos( $p_required, '<' ) === false ) {
-			$t_version_core = substr(
-				$t_plugin_version,
-				0,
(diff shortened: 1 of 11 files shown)


cproensa:
While changing the plugin doc, you may consider adding this to plugin documentation, which i had some trouble with:
when a plugin uses shcema, basename is used as part of the table name, so if the plugin is planned to be published and portable, it must be accounted that some databases (eg. oracle) limits the table name length.

tests/Mantis/PluginTest.php
```diff
@@ -0,0 +1,151 @@
...
+	const REQ_130 = '1.3.0';
+	const REQ_131 = '1.3.1';
+	const REQ_13_20 = '1.3, <2.0';
+	const REQ_13_21 = '1.3, <2.1';
+	const REQ_13_30 = '1.3, <3.0';
+	const REQ_20 = '2.0';
+	const REQ_21 = '2.1';
+	const REQ_210 = '2.1.0';
+	const REQ_210_DEV = '2.1.0-dev';
+	const REQ_210_A1 = '2.1.0-beta.1';
```
atrol: 2.1.0-alpha.1

docbook/Developers_Guide/en-US/Plugins_Building.xml
```diff
@@ -90,23 +101,216 @@ class ExamplePlugin extends MantisPlugin {
...
+							is 2.1:
+							<itemizedlist>
+								<listitem>
+									<para>Old release without a maximum version
+										specified
+										<programlisting>
+$this->requires = array( 'MantisCore' => '1.3.1' );
+										</programlisting>
+										The plugin is compatible with MantisBT
+										>= 1.3.1 and &lt; 2(.0) - note the maximum
```
vboctor: Why "2(.0)" vs. "2.0.0" in the example here?

docbook/Developers_Guide/en-US/Plugins_Building.xml
```diff
@@ -90,23 +101,216 @@ class ExamplePlugin extends MantisPlugin {
...
+										different branch).
+									</para>
+								</listitem>
+								<listitem>
+									<para>Only specify a maximum version
+										<programlisting>
+$this->requires = array( 'MantisCore' => '&lt; 3.1' );
+										</programlisting>
+										The plugin is compatible up to MantisBT
+										3.1 (not included).
```
vboctor: "not included" - "not inclusive"?

docbook/Developers_Guide/en-US/Plugins_Building.xml
```diff
@@ -90,23 +101,216 @@ class ExamplePlugin extends MantisPlugin {
...
+							See <literal>requires</literal> above
+							for details on how to specify versions.
+						</para>
+					</listitem>
+				</varlistentry>
+
+				<varlistentry>
+					<term>author</term>
+					<listitem>
+						<para>Your name, or an array of names.</para>
```
vboctor: For versions we use comma separation.  Should we do the same for names?  I'm OK either way, just thought I would point the inconsistency out.

docbook/Developers_Guide/en-US/Plugins_Building_Source.xml
```diff
@@ -37,9 +37,10 @@ class ExamplePlugin extends MantisPlugin {
         $this->page = '';           # Default plugin page
 
         $this->version = '1.0';     # Plugin version string
-        $this->requires = array(    # Plugin dependencies, array of basename => version pairs
-            'MantisCore' => '1.2.0, &lt;= 1.2.0',  #   Should always depend on an appropriate version of MantisBT
-            );
+        $this->requires = array(    # Plugin dependencies
+            'MantisCore' => '1.3, &lt; 3.0',  # Should always depend on an appropriate
```
vboctor: Let's use "2.0, < 3.0" for the example, since typically the same plugin code won't support 1.3 and 2.0.  I know this is just an example, but it can confuse developers.

manage_plugin_page.php
```diff
@@ -135,7 +135,7 @@ function ( $p_p1, $p_p2 ) {
 		$t_name = '<a href="' . string_attribute( plugin_page( $t_page, false, $t_basename ) ) . '">' . $t_name . '</a>';
 	}
 
-	if( !is_blank( $t_author ) ) {
+	if( !empty( $t_author ) ) {
```
vboctor: Curious about reason for moving is_blank() to empty()?

tests/Mantis/PluginTest.php
```diff
@@ -0,0 +1,151 @@
...
+			array( '2.0.0-beta.1', self::REQ_20, true ),
+
+			array( '2.1.0', self::REQ_13, false ),
+			array( '2.1.0', self::REQ_13_20, false ),
+			array( '2.1.0', self::REQ_13_21, false ),
+			array( '2.1.0', self::REQ_13_30, true ),
+			array( '2.1.0', self::REQ_20, true ),
+			array( '2.1.0', self::REQ_21, true ),
+			array( '2.1.0', self::REQ_210_DEV, true ),
+			array( '2.1.0', self::REQ_210_B1, true ),
```
vboctor: Should we have tests for RC versions as well?

tests/Mantis/PluginTest.php
dregad: Thanks, fixed

core/plugin_api.php
```diff
@@ -447,88 +447,25 @@ function plugin_is_loaded( $p_base_name ) {
...
 	} else {
-		if( count( $p_version1 ) > 0 ) {
-			return 1;
-		}
-	}
+		$t_operator = '>=';
 
-	# No more comparisons
-	return -1;
+		# Special handling for dev versions: we consider them as meeting the
```
dregad: IIRC from my tests, only *dev* was causing issues but I'll double-check what happens with alpha/beta/rc.

docbook/Developers_Guide/en-US/Plugins_Building.xml
```diff
@@ -32,55 +32,66 @@
...
+
+			<para>
+				Because of how <classname>MantisPlugin</classname> declares the
+				<function>register()</function>	method as <literal>abstract</literal>, our
+				plugin must implement that method before PHP will find it semantically
+				valid.  This method is meant for one simple purpose, and should never be
+				used for any other task: setting the plugin's information properties
+				including the plugin's name, description, version, and more.
+				Please refer to <xref linkend="dev.plugins.building.properties" />
+				below for  details about available properties.
```
dregad: Done

docbook/Developers_Guide/en-US/Plugins_Building.xml
```diff
@@ -90,23 +101,216 @@ class ExamplePlugin extends MantisPlugin {
...
+							This is used to create a link to the specified page
+							on Mantis' manage plugin page.
+						</para>
+					</listitem>
+				</varlistentry>
+
+				<varlistentry>
+					<term>version</term>
+					<listitem>
+						<para>Your plugin's version string.
```
dregad: Although we use SemVer for Mantis versions, this is not necessarily the case for 3rd-party plugins. We could recommend SemVer if you want to, but can't really enforce its use. 

In any case it does not really matter much, as version_compare should be able to handle much anything.

docbook/Developers_Guide/en-US/Plugins_Building.xml
dregad: I wrote it like that because technically the code effectively adds `<2`, and that implies the additional .0 (or even .0.0).

docbook/Developers_Guide/en-US/Plugins_Building.xml
dregad: That would be me reaching the limits of my English language skills...

docbook/Developers_Guide/en-US/Plugins_Building.xml
dregad: Currently if an array of names is provided, the code collapses it into a comma-delimited string, so that's pretty much the same thing, but I believe keeping the array makes sense 
1. for backwards-compatibility, and
2. if in the future we decide to allow specifying additional details about the authors (e.g. individual e-mails address), like [Composer does](https://getcomposer.org/doc/04-schema.md#authors)

docbook/Developers_Guide/en-US/Plugins_Building_Source.xml
dregad: I intentionally used `1.3, < 3.0` range to illustrate the case of a plugin spanning 2 major Mantis versions. 

I guess it makes sense to use a more common use case for the example. Will change that.

manage_plugin_page.php
dregad: http://www.mantisbt.org/bugs/view.php?id=22205

dregad: If you want I can change that to `!empty( trim( $t_author ) )` to have exactly the same behavior with strings as with an is_blank() call.

tests/Mantis/PluginTest.php
dregad: It does not cost much to add them, but based on how version_compare works, technically it would be the same test as alpha/beta so I'm not sure there is an added value.

dregad:
I just pushed a commit addressing various comments. 

> I wonder if we shouldn't have special handling for dev. For most plugins they don't need to worry about dev and if they do, then they should have the requirement as 2.0.0-dev, 2.0.0-beta, or 2.0.0-rc, 2.0.0, etc.

This is really a convenience feature (mostly for us) for the MantisCore dependency, and as I see it only affects people using the code by cloning the repository (we never actually formally *release* a dev version). Without this, we would have to change the built-in plugins' MantisCore minimum requirement until a formal release is cut.

> Not sure if we need to worry about dev vs. dev.1 / dev.2

I don't think we would ever use that, what would be the point ? dev is dev, and when it's finished we move to alpha/beta. I don't see the need for intermediate development versions.

Left to do:

- adding test cases for rc and possibly test cases that compare beta.1 vs. beta.2, or rc.1 vs. rc.2. 
- double-check what happens with alpha/beta/rc. (ref https://github.com/mantisbt/mantisbt/pull/995#discussion_r96340258)

docbook/Developers_Guide/en-US/Plugins_Building.xml
vboctor: I think we should recommend and advertise support to semver, since when plugins have dependencies on each other, we also want to use semver semantics to check that the required dependencies are available.

docbook/Developers_Guide/en-US/Plugins_Building.xml
vboctor: Are you supporting array of strings or array of associate array that enables multiple properties, or both?

manage_plugin_page.php
vboctor: That's fine.  I missed the array vs. string part.

tests/Mantis/PluginTest.php
vboctor: Often such unit tests are good documentation for expected behavior, so even if mechanically it doesn't add value, it may provide a better documentation for how versions relate to each other.  But you can argue semver.org provides that in this case.

vboctor:
> This is really a convenience feature (mostly for us) for the MantisCore dependency, and as I see it only affects people using the code by cloning the repository (we never actually formally release a dev version). Without this, we would have to change the built-in plugins' MantisCore minimum requirement until a formal release is cut.

If the core plugins requires MANTIS_VERSION instead of 2.0.0, then this problem won't exist.  I would rather do that than special casing dev.  Other option is to have them require 2.0.0-dev, which is not a big deal.

dregad:
> If the core plugins requires MANTIS_VERSION instead of 2.0.0

I don't like this approach. Even if some plugins are bundled with Mantis releases, they should still be considered as an independent piece of software, and therefore we should not rely on defining a dependency for Mantis Core based on a constant defined in said Core, as the result would always be true; in this case, we might as well not require a Core dependency at all...

> have them require 2.0.0-dev

OK. I'll remove the special handling code and change the requirement to 2.0.0-dev

docbook/Developers_Guide/en-US/Plugins_Building.xml
dregad: Already did that, see https://github.com/mantisbt/mantisbt/pull/995/files#diff-8fcd352b021ad2ba7feb99bc3be1d267R178

docbook/Developers_Guide/en-US/Plugins_Building.xml
dregad: Currently, the code only handles `array( 'Damien', 'Victor' );`. 

I was thinking that in the future we could allow 
``` php
array(
    array( 'name' => 'Damien', 'email' => 'damien@example.com' ),
    array( 'name' => 'Victor', 'email' => 'victor@example.com' ),
);
```

tests/Mantis/PluginTest.php
dregad: Indeed. And moreover I don't think it's necessary to build test cases that don't add semantic value over PHP's own unit tests for version_compare().

vboctor:
@dregad I don't see value in requiring MantisCore version for core plugins, since they are always distributed with core and it is not really a requirement for mix and match them with other versions.  So I suggest removing such requirement from these plugins.

Having said that, once you remove the special handling for 'dev', let's go ahead and get this in so that we include this in 2.1.0 release.  So +1 for me once you have done such change.

dregad:
> I don't see value in requiring MantisCore version for core plugins, since they are always distributed with core and it is not really a requirement for mix and match them with other versions

Consider this (real life, happened in my company once) upgrade scenario:
- install new mantis version in separate directory
- copy plugins directory's contents from old to new directory

Not having a MantisCore req, would potentially let an old plugin work with newer version of Core, which may be undesirable and could cause issues. Leaving the requirement in there does not cost us much (bump it once per major version or when a specific feature requires it), so I'd rather leave that in (belt and braces).

PR #996, opened 2017-01-16
Remove wrong database version information from installer

atrol:


atrol:
There are also other changes in 2.0.x, so I thought: Why not fix it in 2.0.x?

It's better to have 2.0.1 with some fixes quite soon instead of waiting for 2.1.0.
I expect some delay in getting out 2.1.0 because of the known markdown issues.

vboctor:
Markdown is experimental and off by default in 2.1.0.  There are no blocking changes for markdown.  2.1.0 is planned by end of the month.

atrol:
> and off by default in 2.1.0

It's active at www.mantisbt.org/bugs, so I didn't think of it when writing
> I expect some delay in getting out 2.1.0

In future I will push to master-x.x.x  just if it's a security or regression issue.

PR #997, opened 2017-01-16
Added filling tags and custom fields from POST/GET args when submitting bug

GunSmoker:
Bug report/submit page supports reading pre-fill values for fields from GET/POST arguments. This allows one to open bug submission form with prefilled values. However, "tags" column and custom fields do not use defaults from GET/POST arguments. This minor patch solves this.

### DIFF
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -124,6 +124,7 @@ if( $f_master_bug_id > 0 ) {
 	$f_additional_info		= $t_bug->additional_information;
 	$f_view_state			= (int)$t_bug->view_state;
 	$f_due_date				= $t_bug->due_date;
+	$f_tag_string			= '';
 
 	$t_project_id			= $t_bug->project_id;
 } else {
@@ -179,6 +180,7 @@ if( $f_master_bug_id > 0 ) {
 	$f_additional_info		= gpc_get_string( 'additional_info', config_get( 'default_bug_additional_info' ) );
 	$f_view_state			= gpc_get_int( 'view_state', (int)config_get( 'default_bug_view_status' ) );
 	$f_due_date				= gpc_get_string( 'due_date', date_strtotime( config_get( 'due_date_default' ) ) );
+	$f_tag_string			= gpc_get_string( 'tag_string', '' );
 
 	if( $f_due_date == '' ) {
 		$f_due_date = date_get_null();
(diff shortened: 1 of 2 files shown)


GunSmoker:
Here: http://www.mantisbt.org/bugs/view.php?id=22217

P.S.
BTW, I would like to make a pull request for suggestion 22203. However, it requires adding a new field to tickets table. How should I commit such changes? I mean, changes in DB.

atrol:
> How should I commit such changes? I mean, changes in DB.

It would be something like adding the following code to admin/schema.php
 ```
$g_upgrade[210] = array( 'AddColumnSQL', array( db_get_table( 'bug' ), "
	access_token			C(64)	NULL DEFAULT NULL ) );
 ```

GunSmoker:
Yeah, I found this myself, but I get impression that "200" stands for Mantis 2.0.0 or 2.00, so I thought that DB changes are not easily allowed.

dregad:
> I get impression that "200" stands for Mantis 2.0.0

It's only a sequence number. It's only happenstance that we are working on version 2.1.0 while the schema version is 210.

We do defer schema changes to the next major release though (i.e. 3.0).

vboctor:
> We do defer schema changes to the next major release though (i.e. 3.0).

That is incorrect.  We typically allow schema changes for minor releases like 2.1.0, 2.2.0, etc.  However, as we transition from 1.3.x to 2.x, we have made a decision to defer schema changes a bit.  We may decide to re-open schema changes for 2.2.0 for example.

bug_report_page.php
```diff
@@ -179,6 +180,7 @@
 	$f_additional_info		= gpc_get_string( 'additional_info', config_get( 'default_bug_additional_info' ) );
 	$f_view_state			= gpc_get_int( 'view_state', (int)config_get( 'default_bug_view_status' ) );
 	$f_due_date				= gpc_get_string( 'due_date', date_strtotime( config_get( 'due_date_default' ) ) );
+	$f_tag_string			= gpc_get_string( 'tag_string', '' );
```
vboctor: Should this be named `tags`?

vboctor:
This feature is not supported by REST API and hence closing this PR.

vboctor:
i meant this feature is NOW supported by REST API and hence closing this PR.

PR #999, opened 2017-01-19
Markdown: Fix Markdown converting '<' to &lt; inside backticks

jllano:
- markdown is unable to process '<' signs within the code block tag (backticks)
- markdown convert '<' signs into '&amp;amp;lt;' after the text proccessing process
- therefore we need restore the entity back into its original name '&amp;lt;'

Fixes #22175

### DIFF
--- a/plugins/MantisCoreFormatting/MantisCoreFormatting.php
+++ b/plugins/MantisCoreFormatting/MantisCoreFormatting.php
@@ -131,6 +131,11 @@ class MantisCoreFormattingPlugin extends MantisFormattingPlugin {
 
 		$t_string = MantisMarkdown::convert_text( $t_string );
 
+		# markdown is unable to process '<' signs within the code block tag (backticks)
+		# markdown convert '<' signs into '&amp;lt;' after the text proccessing process
+		# therefore we need restore the entity back into its original name '&lt;'
+		$t_string = str_replace( '&amp;lt;', '&lt;', $t_string );
+
 		return $t_string;
 	}
 
--- a/plugins/MantisCoreFormatting/core/MantisMarkdown.php
+++ b/plugins/MantisCoreFormatting/core/MantisMarkdown.php
@@ -36,7 +36,7 @@
  * @uses Parsedown Library
  */
 
-require_lib( 'parsedown' . DIRECTORY_SEPARATOR . 'Parsedown.php' );
+require_once 'library/parsedown/Parsedown.php';
 
 /**
  * A class that overrides default Markdown parsing for Mantis specific scenarios.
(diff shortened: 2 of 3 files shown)


jllano:
> to ensure the text processing (CoreFormatting) remain in possession in terms of security and processing

my apologies for the confusion, I will removed that line.

> preg_replace( '#&amp;lt;.+?.*?&gt;#i', '$0', $t_string )

Good catch, no need to add a preg_replace also, thank you

jllano:
Before:
screen shot 2017-01-19 at 8 54 55 pm

After:
screen shot 2017-01-19 at 8 55 11 pm

plugins/MantisCoreFormatting/MantisCoreFormatting.php
```diff
@@ -131,9 +131,10 @@ function ( ) {
 
 		$t_string = MantisMarkdown::convert_text( $t_string );
 
-		# restore '<' converting &lt; within backticks 
-		$t_string = str_replace( '&amp;lt;', '&lt;', preg_replace( '#&amp;lt;.+?.*?&gt;#i', '$0', $t_string ));
-			
+		# restore '<' signs converting &lt; within code block tag (backticks)
+		# markdown coudnt perform well once the text procces executed the text
```
dregad: 2nd line of comment does not make sense. 
- coundt -> doesn'nt ?
- define what _perform well_ means
- _the text procces_ (processing ?) _executed the text_ what does that mean ?

atrol:
There is 

> Fixes #22175,#22204

in the first commit message.
I don't see at the moment how this change fixes http://www.mantisbt.org/bugs/view.php?id=22204

jllano:
It does @atrol, please checkout screenshots above.

atrol:
Are we talking about the same thing?

I am talking about main_page.php with $g_news_enabled = ON;
@badfiles, is this what you mean in http://www.mantisbt.org/bugs/view.php?id=22204 ?

jllano:
I assumed @badfiles is referring the <h1> to <h6> tags are used to define HTML headings?

dregad:
@jllano see http://www.mantisbt.org/bugs/view.php?id=22204#c55241 - definitely **not** fixed by this PR (I just tested). Please update commit message to remove bug ID reference.

jllano:
@dregad I updated the commit, I think its much cleaner now.

Thanks

vboctor:
I've merged the fix for the issue, but will not merge the unit tests for now.  Will open a separate work item for revising how unit tests for markdown should be implemented so that:

- We are testing the production code, and not a copy of it.
- Making sure that markdown tests run when MantisBT tests run as part of Travis.

PR #1001, opened 2017-01-19
Feature 22203: public / external access to tickets by unique URL

GunSmoker:
Implementation for http://www.mantisbt.org/bugs/view.php?id=22203
Screenshots of working feature can be found as attaches here: http://www.mantisbt.org/bugs/view.php?id=22203#c55198

- The whole feature is disabled by default. It can be enabled by "$g_public_urls = ON;" in config_inc.php.
- New "token" field was added to bugs table. Field is initially null, it is filled by randomized value on request.
- Public Share URL can be copied from corresponding field in "Users monitoring issue" box when viewing bug; or it can be accessed when bug is created - via "View Public Share URL" button.
- SOAP API was adjusted to return "token" for bug.
- New bug_view_limited_inc.php (similar to bug_view_inc.php) was added, it implements a limited read-only view for access via Public Share URL.
- Limited view shows description and steps to reproduce for public bugs only. Description/steps to reproduce is not shown for private bugs (only summary, resolution, version, dates are shown).
- view.php switches between usual bug_view_inc.php (full view/edit) and bug_view_limited_inc.php (limited read-only view) - depending on URL (token is present and valid) and if user is already logged in.
- Access is controlled by new global $g_login_by_bug_token, which is false for all usual operations. It is set to (int)reporter.id by view.php when valid token is present and bug_view_limited_inc.php is invoked.
- authentication_api.php was adjusted to allow access when public share URL is used. Most changes are just extra safeguards.

### DIFF
--- a/admin/schema.php
+++ b/admin/schema.php
@@ -885,8 +885,10 @@ $g_upgrade[209] = array( 'AlterColumnSQL', array( db_get_table( 'api_token' ), "
 # Release marker: 1.3.0
 
 # Per-bug access token for "public share URL" feature, #22203
-$g_upgrade[210] = array( 'AddColumnSQL', array( db_get_table( 'bug' ), "
-	token					C(20)	NULL DEFAULT NULL" 
+$g_upgrade[210] = array('CreateTableSQL', array( db_get_table( 'bug_token' ), "
+	bug_id					I		UNSIGNED NOTNULL PRIMARY,
+	token					C(20)	NOTNULL",
+	$t_table_options
 	) );
 
 # ----------------------------------------------------------------------------
--- a/admin/schema.php
+++ b/admin/schema.php
@@ -884,6 +884,10 @@ $g_upgrade[209] = array( 'AlterColumnSQL', array( db_get_table( 'api_token' ), "
 
 # Release marker: 1.3.0
 
+# Per-bug access token for "public share URL" feature, #22203
+$g_upgrade[210] = array( 'AddColumnSQL', array( db_get_table( 'bug' ), "
+	token					C(20)	NULL DEFAULT NULL" 
+	) );
 
 # ----------------------------------------------------------------------------
 # End of schema definition, clear local variables
--- a/api/soap/mantisconnect.wsdl
+++ b/api/soap/mantisconnect.wsdl
@@ -163,6 +163,7 @@
    <xsd:element name="monitors" type="tns:AccountDataArray" minOccurs="0"/>
    <xsd:element name="sticky" type="xsd:boolean" minOccurs="0"/>
    <xsd:element name="tags" type="tns:ObjectRefArray" minOccurs="0"/>
+   <xsd:element name="token" type="xsd:string" minOccurs="0"/>
   </xsd:all>
  </xsd:complexType>
  <xsd:complexType name="IssueDataArray">
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -138,6 +138,11 @@ function mc_issue_get( $p_username, $p_password, $p_issue_id ) {
 	$t_issue_data['custom_fields'] = mci_issue_get_custom_fields( $p_issue_id );
 	$t_issue_data['monitors'] = mci_account_get_array_by_ids( bug_get_monitors( $p_issue_id ) );
 	$t_issue_data['tags'] = mci_issue_get_tags_for_bug_id( $p_issue_id, $t_user_id );
+	if( isset($g_public_urls) && (ON === $g_public_urls) ) {
+		$t_issue_data['token'] = mci_null_if_empty( $t_bug->token );
+	} else {
+		$t_issue_data['token'] = mci_null_if_empty( '' );
+	}  
 
 	return $t_issue_data;
 }
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -138,7 +138,7 @@ function mc_issue_get( $p_username, $p_password, $p_issue_id ) {
 	$t_issue_data['custom_fields'] = mci_issue_get_custom_fields( $p_issue_id );
 	$t_issue_data['monitors'] = mci_account_get_array_by_ids( bug_get_monitors( $p_issue_id ) );
 	$t_issue_data['tags'] = mci_issue_get_tags_for_bug_id( $p_issue_id, $t_user_id );
-	if( isset($g_public_urls) && (ON === $g_public_urls) ) {
+	if( isset($g_public_urls) && (ON === $g_public_urls) && isset($t_bug->token) ) {
 		$t_issue_data['token'] = mci_null_if_empty( $t_bug->token );
 	} else {
 		$t_issue_data['token'] = mci_null_if_empty( '' );
@@ -1603,6 +1603,11 @@ function mci_issue_data_as_array( BugData $p_issue_data, $p_user_id, $p_lang ) {
 		$t_issue['custom_fields'] = mci_issue_get_custom_fields( $p_issue_data->id );
 		$t_issue['tags'] = mci_issue_get_tags_for_bug_id( $p_issue_data->id, $p_user_id );
 		$t_issue['monitors'] = mci_account_get_array_by_ids( bug_get_monitors( $p_issue_data->id ) );
+		if( isset($g_public_urls) && (ON === $g_public_urls) && isset($p_issue_data->token) ) {
+			$t_issue['token'] = mci_null_if_empty( $p_issue_data->token );
+		} else {
+			$t_issue['token'] = mci_null_if_empty( '' );
+		}  
 
 		return $t_issue;
 }
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -771,6 +771,7 @@ function mc_issue_add( $p_username, $p_password, stdClass $p_issue ) {
 		 access_has_project_level( config_get( 'set_bug_sticky_threshold', null, null, $t_project_id ), $t_project_id ) ) {
 		$t_bug_data->sticky = $p_issue['sticky'];
 	}
+	$t_bug_data->token = crypto_generate_uri_safe_nonce( 20 );
 
 	if( isset( $p_issue['due_date'] ) && access_has_global_level( config_get( 'due_date_update_threshold' ) ) ) {
 		$t_bug_data->due_date = SoapObjectsFactory::parseDateTimeString( $p_issue['due_date'] );
@@ -1603,7 +1604,7 @@ function mci_issue_data_as_array( BugData $p_issue_data, $p_user_id, $p_lang ) {
 		$t_issue['custom_fields'] = mci_issue_get_custom_fields( $p_issue_data->id );
 		$t_issue['tags'] = mci_issue_get_tags_for_bug_id( $p_issue_data->id, $p_user_id );
 		$t_issue['monitors'] = mci_account_get_array_by_ids( bug_get_monitors( $p_issue_data->id ) );
-		if( (ON === config_get_global( 'public_urls' ) ) && isset( $p_issue_data->token ) ) {
+		if( ( ON === config_get_global( 'public_urls' ) ) && isset( $p_issue_data->token ) ) {
 			$t_issue['token'] = mci_null_if_empty( $p_issue_data->token );
 		} else {
 			$t_issue['token'] = mci_null_if_empty( '' );
@@ -1669,6 +1670,12 @@ function mci_issue_data_as_header_array( BugData $p_issue_data ) {
 		$t_issue['attachments_count'] = count( mci_issue_get_attachments( $p_issue_data->id ) );
 		$t_issue['notes_count'] = count( mci_issue_get_notes( $p_issue_data->id ) );
 
+		if( ( ON === config_get_global( 'public_urls' ) ) && isset( $p_issue_data->token ) ) {
+			$t_issue['token'] = mci_null_if_empty( $p_issue_data->token );
+		} else {
+			$t_issue['token'] = mci_null_if_empty( '' );
+		}  
+    
 		return $t_issue;
 }
 
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -138,10 +138,10 @@ function mc_issue_get( $p_username, $p_password, $p_issue_id ) {
 	$t_issue_data['custom_fields'] = mci_issue_get_custom_fields( $p_issue_id );
 	$t_issue_data['monitors'] = mci_account_get_array_by_ids( bug_get_monitors( $p_issue_id ) );
 	$t_issue_data['tags'] = mci_issue_get_tags_for_bug_id( $p_issue_id, $t_user_id );
-	if( ( ON === config_get_global( 'public_urls' ) ) && isset( $t_bug->token ) ) {
+	if( ( ON === config_get_global( 'public_urls_enabled' ) ) ) {
 		$t_issue_data['token'] = mci_null_if_empty( $t_bug->token );
 	} else {
-		$t_issue_data['token'] = mci_null_if_empty( '' );
+		$t_issue_data['token'] = null;
 	}  
 
 	return $t_issue_data;
@@ -771,7 +771,7 @@ function mc_issue_add( $p_username, $p_password, stdClass $p_issue ) {
 		 access_has_project_level( config_get( 'set_bug_sticky_threshold', null, null, $t_project_id ), $t_project_id ) ) {
 		$t_bug_data->sticky = $p_issue['sticky'];
 	}
-	$t_bug_data->token = crypto_generate_uri_safe_nonce( 20 );
+	$t_bug_data->token = crypto_generate_uri_safe_nonce( PUBLIC_URL_TOKEN_LENGTH );
 
 	if( isset( $p_issue['due_date'] ) && access_has_global_level( config_get( 'due_date_update_threshold' ) ) ) {
 		$t_bug_data->due_date = SoapObjectsFactory::parseDateTimeString( $p_issue['due_date'] );
@@ -1604,10 +1604,10 @@ function mci_issue_data_as_array( BugData $p_issue_data, $p_user_id, $p_lang ) {
 		$t_issue['custom_fields'] = mci_issue_get_custom_fields( $p_issue_data->id );
 		$t_issue['tags'] = mci_issue_get_tags_for_bug_id( $p_issue_data->id, $p_user_id );
 		$t_issue['monitors'] = mci_account_get_array_by_ids( bug_get_monitors( $p_issue_data->id ) );
-		if( ( ON === config_get_global( 'public_urls' ) ) && isset( $p_issue_data->token ) ) {
+		if( ( ON === config_get_global( 'public_urls_enabled' ) ) ) {
 			$t_issue['token'] = mci_null_if_empty( $p_issue_data->token );
 		} else {
-			$t_issue['token'] = mci_null_if_empty( '' );
+			$t_issue['token'] = null;
 		}  
 
 		return $t_issue;
@@ -1670,12 +1670,6 @@ function mci_issue_data_as_header_array( BugData $p_issue_data ) {
 		$t_issue['attachments_count'] = count( mci_issue_get_attachments( $p_issue_data->id ) );
 		$t_issue['notes_count'] = count( mci_issue_get_notes( $p_issue_data->id ) );
 
-		if( ( ON === config_get_global( 'public_urls' ) ) && isset( $p_issue_data->token ) ) {
-			$t_issue['token'] = mci_null_if_empty( $p_issue_data->token );
-		} else {
-			$t_issue['token'] = mci_null_if_empty( '' );
-		}  
-    
 		return $t_issue;
 }
 
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -138,7 +138,7 @@ function mc_issue_get( $p_username, $p_password, $p_issue_id ) {
 	$t_issue_data['custom_fields'] = mci_issue_get_custom_fields( $p_issue_id );
 	$t_issue_data['monitors'] = mci_account_get_array_by_ids( bug_get_monitors( $p_issue_id ) );
 	$t_issue_data['tags'] = mci_issue_get_tags_for_bug_id( $p_issue_id, $t_user_id );
-	if( isset($g_public_urls) && (ON === $g_public_urls) && isset($t_bug->token) ) {
+	if( (ON === config_get_global( 'public_urls')) && isset($t_bug->token) ) {
 		$t_issue_data['token'] = mci_null_if_empty( $t_bug->token );
 	} else {
 		$t_issue_data['token'] = mci_null_if_empty( '' );
@@ -1603,7 +1603,7 @@ function mci_issue_data_as_array( BugData $p_issue_data, $p_user_id, $p_lang ) {
 		$t_issue['custom_fields'] = mci_issue_get_custom_fields( $p_issue_data->id );
 		$t_issue['tags'] = mci_issue_get_tags_for_bug_id( $p_issue_data->id, $p_user_id );
 		$t_issue['monitors'] = mci_account_get_array_by_ids( bug_get_monitors( $p_issue_data->id ) );
-		if( isset($g_public_urls) && (ON === $g_public_urls) && isset($p_issue_data->token) ) {
+		if( (ON === config_get_global( 'public_urls')) && isset($p_issue_data->token) ) {
 			$t_issue['token'] = mci_null_if_empty( $p_issue_data->token );
 		} else {
 			$t_issue['token'] = mci_null_if_empty( '' );
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -138,7 +138,7 @@ function mc_issue_get( $p_username, $p_password, $p_issue_id ) {
 	$t_issue_data['custom_fields'] = mci_issue_get_custom_fields( $p_issue_id );
 	$t_issue_data['monitors'] = mci_account_get_array_by_ids( bug_get_monitors( $p_issue_id ) );
 	$t_issue_data['tags'] = mci_issue_get_tags_for_bug_id( $p_issue_id, $t_user_id );
-	if( (ON === config_get_global( 'public_urls')) && isset($t_bug->token) ) {
+	if( ( ON === config_get_global( 'public_urls' ) ) && isset( $t_bug->token ) ) {
 		$t_issue_data['token'] = mci_null_if_empty( $t_bug->token );
 	} else {
 		$t_issue_data['token'] = mci_null_if_empty( '' );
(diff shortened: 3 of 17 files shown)


dregad:
Thanks for your contribution.

I haven't had time to look into details or test your code, but I already see some issues which need to be addressed. 

`$g_public_urls` config must be defined and documented (PHPDoc) in config_defaults_inc.php. Based on that, it will always be set, so you can get rid of all isset() calls; furthermore, config options must in principle be accessed via Config API (i.e. config_get() or config_get_global() depending on usage), and not directly. Consider also whether it should be added to *$g_global_settings* (I think yes), and/or to *$g_public_config_names* (probably not).

core.php
```diff
@@ -247,6 +247,10 @@ function __autoload( $p_class ) {
 	$g_login_anonymous = true;
 }
 
+if( !isset( $g_login_by_bug_token ) ) {
+	$g_login_by_bug_token = false;
+}
+
```
dregad: Probably not necessary, the variable is initialized when including auth API a couple lines later

docbook/Admin_Guide/en-US/config/misc.xml
```diff
@@ -121,6 +121,19 @@
...
+				</para>
+				<para>Enabling this option will show additional "Public Share URL" read-only field when viewing the bug (in "Users monitoring this issue" section) as well as additional "token" field in SOAP API. The URL can be copied, used and shared. 
+				</para>
+				<para>Access via public share URL does not require loggin in (e.g. access is anonymous). This option does not require $g_allow_anonymous_login to be enabled.
+				</para>
+				<para>When user visits public share URL - he will see a limited read-only view of the bug with "Full View" button asking to log in. User may see current status of the bug and minor details (such as modification date). Public bugs will show summary, description and steps to reproduce in limited view. Private bugs will show only summary, but no description and steps to reproduce. No other data nor actions are available in the limited view for public URLs.
+				</para>
+			</listitem>
+		</varlistentry>
+		<varlistentry>
```
dregad: Break long lines around 80, no more than 100

GunSmoker:
Fixed code formatting, except for extra parenthesis. Your guidelines says it is allowed: "Add parentheses as appropriate to make things improve readability (even if not syntactically required by operators precedence)". Since I am not a PHP dev, I came from language with different operator priorities - I would like to leave this as it is.

badfiles:
Maybe you would like to check out [this branch ](https://github.com/badfiles/mantisbt/commits/direct_access_keys) to find something useful.

GunSmoker:
@badfiles What exactly I should be looking at?

badfiles:
Maybe you'll find something useful. Nothing specific comes to my mind.

GunSmoker:
@badfiles Are there cat pictures or something?

GunSmoker:
If someone could indicate what is wrong with passing tests - it would be greatly appreciated. I can not reproduce these locally.

dregad:
> I can not reproduce these locally.

Meaning what, you can't run the PHPUnit tests ? Or they run but give different results ?

FYI, you can always review the Travis execution logs. In [this case](https://travis-ci.org/mantisbt/mantisbt/jobs/193494330#L507); the error is 
```
1) IssueAddTest::testCreateIssue
SoapFault: Error Type: SYSTEM NOTICE,
Error Description: Trying to get property of non-object
```

The problem even be reproduced via the GUI, the error below occured when reporting a new issue:

```
SYSTEM NOTICE
'Trying to get property of non-object' in '/path/to/mantisbt/core/bug_api.php' line 856
```

When developing, you should use the [recommended values for *$g_display_errors*](https://github.com/mantisbt/mantisbt/blob/release-2.0.0/config_defaults_inc.php#L4093), and set `$g_show_detailed_errors = ON;` if necessary.

atrol:
I have no time to have a deeper look at the moment, so just a few general comments.

- changes in manage_user_page.php are not related to this PR
- bug_view_limited_inc.php introduces quite a lot of redundant code (compare with bug_view_inc.php)
- introduces some redundant code following  if ( ( ON === config_get_global( 'public_urls' ) )

bug_report.php
```diff
@@ -302,6 +303,29 @@
 
 layout_page_header_begin();
 
+if ( ( ON === config_get_global( 'public_urls' ) ) && isset( $t_bug_data->token ) && ( !is_null( $t_bug_data->token ) ) ) {
```
vboctor: Calculating the share URL should be an API rather than being embedded in script pages.  For example, bug_get_share_url( $p_bug_id ).  The API should handle the config checks, and return a valid URL or empty/null.

bug_report.php
```diff
@@ -302,6 +303,29 @@
 
 layout_page_header_begin();
 
+if ( ( ON === config_get_global( 'public_urls' ) ) && isset( $t_bug_data->token ) && ( !is_null( $t_bug_data->token ) ) ) {	
+	if ( isset( $_SERVER ) ) {
```
vboctor: Use config_get_global( 'path' ) instead of constructing the base path here manually.

bug_view_inc.php
```diff
@@ -118,6 +118,29 @@
 $t_show_target_version = $t_show_versions && in_array( 'target_version', $t_fields )
 	&& access_has_bug_level( config_get( 'roadmap_view_threshold' ), $f_bug_id );
 
+if ( ( ON === config_get_global( 'public_urls' ) ) && isset( $t_bug->token ) && ( !is_null( $t_bug->token ) ) ) {
```
vboctor: Re-use the API that calculates the share url.

config/config_inc.php.sample
```diff
@@ -81,3 +81,4 @@ $g_return_path_email    = 'admin@example.com';	# the return address for bounced
 
 # --- Others ---
 # $g_default_home_page = 'my_view_page.php';	# Set to name of page to go to after login
+# $g_public_urls		= OFF;	# enable usage of public share URLs. The default setting is OFF.
```
vboctor: There is no need to add this config to the sample.  The sample is designed to include the most basic and commonly used configs that are typically overridden.

config_defaults_inc.php
```diff
@@ -2939,6 +2939,23 @@
...
+ * This option does not require $g_allow_anonymous_login to be enabled.
+ * When user visits public share URL - he will see a limited read-only view of the 
+ * bug with "Full View" button asking to log in. User may see current status of the 
+ * bug and minor details (such as modification date). Public bugs will show summary, 
+ * description and steps to reproduce in limited view. Private bugs will show only 
+ * summary, but no description and steps to reproduce. No other data nor actions are 
+ * available in the limited view for public URLs.
+ * @global integer $g_public_urls
+ */
+$g_public_urls = OFF;
```
vboctor: I just using $g_feature_name_enabled.  This way, the intention of the config variable is clear, and as needed, we can add more related configs as $g_feature_name_xyz, etc.  I'm not sure "public_urls" is the best name.  Are they public URLs, private URLs, secret URLs, ??  Not sure.  Maybe we should check for naming in features like YouTube, Google Docs, and other tools that offer similar access patterns.

core/authentication_api.php
```diff
@@ -74,6 +74,9 @@
 # @global int $g_cache_current_user_id
 $g_cache_current_user_id = null;
 
+# @global int $g_login_by_bug_token
+$g_login_by_bug_token = false;
```
vboctor: We shouldn't introduce a different authentication model.  We can have a viewer user that is used by this feature, similar to the user used for anonymous access feature.  The name of such user can be configurable, the same way the anonymous access have a config option for username.  Then once the page that supports such tokens validates the token, it calls auth_attempt_script_login( username ), which will login such user for the duration of the request.  This way no other API or changes are necessary for code used by the page.

manage_user_page.php
```diff
@@ -59,7 +59,6 @@
 
 $t_cookie_name = config_get( 'manage_users_cookie' );
 $t_lock_image = '<i class="fa fa-lock fa-lg" alt="' . lang_get( 'protected' ) . '" />';
-$c_filter = '';
```
vboctor: Changes to this page are unrelated to this PR as @atrol pointed out.

config_defaults_inc.php
```diff
@@ -4221,7 +4238,8 @@
 	'class_path','library_path', 'language_path', 'absolute_path_default_upload_folder',
 	'ldap_simulation_file_path', 'plugin_path', 'bottom_include_page', 'top_include_page',
 	'default_home_page', 'logout_redirect_page', 'manual_url', 'logo_url', 'wiki_engine_url',
-	'cdn_enabled', 'public_config_names', 'email_login_enabled', 'email_ensure_unique'
+	'cdn_enabled', 'public_config_names', 'email_login_enabled', 'email_ensure_unique',
+	'public_urls'
```
vboctor: It is unclear to me why it wouldn't be possible to set this configuration option via the database/UI.  We shouldn't limit this to just config_inc.php and hence global.

core/bug_api.php
```diff
@@ -847,6 +853,24 @@ function bug_cache_row( $p_bug_id, $p_trigger_errors = true ) {
 		}
 	}
 
+	if ( is_null( $t_row->token ) ) {
```
vboctor: If we need all issues to have a token, then I would put this into one of two places:

1. When user clicks Share URL, we shouldn't have such URL visible, unless user explicitly requests it.
2. As a database upgrade step.

This doesn't seem to be the right place to update the data.

view.php
```diff
@@ -23,15 +23,42 @@
...
 
 $t_file = __FILE__;
 $t_mantis_dir = dirname( __FILE__ ) . DIRECTORY_SEPARATOR;
 $t_show_page_header = true;
 $t_force_readonly = false;
 $t_fields_config_option = 'bug_view_page_fields';
+$f_bug_id = gpc_get_int( 'id' );
+if( ( ON === config_get_global( 'public_urls' ) ) ) {
+	$f_bug_token = gpc_get_string( 'token', '' );
+	$f_full_view = false;
```
vboctor: $f_ is for form variables.  In this case, bug_token and full_view are really temp variables that should use $t_ prefix.  If you want the variables exposed to included scripts below, technically speaking I would use $g_, but since $t_ is used for others, you can go either way.

docbook/Admin_Guide/en-US/config/misc.xml
```diff
@@ -121,6 +121,33 @@
...
+				<para>Access via public share URL does not require loggin in
+					(e.g. access is anonymous). This option does not 
+					require $g_allow_anonymous_login to be enabled.
+				</para>
+				<para>When user visits public share URL - he will see a limited
+					read-only view of the bug with "Full View" button asking
+					to log in. User may see current status of the bug and
+					minor details (such as modification date). Public bugs
+					will show summary, description and steps to reproduce in
+					limited view. Private bugs will show only summary, but no
```
vboctor: The concept of what to display is interesting.  I would rather if we don't add new rules and exceptions.  For example, for public issues, we should show all what a user with VIEWER access can see.  For private issue it is a bit tricky, since these are typically only visible for users with high access level like DEVELOPER, which means all issue details + private notes, etc would be visible.  The advantage of using the standard model of showing all what a user can see, is that it fits well with the configuration and business logic associated with all components that show up on the view issue page.  You can even re-use the same view issue page.  The "private issues" become the tricky part, we could exclude them from this, or think about some idea to cover them.

admin/schema.php
```diff
@@ -884,6 +884,10 @@
 
 # Release marker: 1.3.0
 
+# Per-bug access token for "public share URL" feature, #22203
+$g_upgrade[210] = array( 'AddColumnSQL', array( db_get_table( 'bug' ), "
```
vboctor: There are two models to come up with such tokens:

1. Use SHAxx on data include (issue_id + submission timestamp + global salt from config).  The global salt would be in $p_public_urls_salt config option for example.
2. Generate a crypto-secure token and store it for each issue.
3. Generate a crypto-secure token per share and have ability to revoke one but not the other.

The first approach doesn't require a schema change or upgrading of the data in the database, which is appealing. It doesn't allow revoking access to an issue though.  We could support revoked but it will require schema change to stamp an explicit token that overrides the SHA based approach for this specific issue (if we determine the revoke feature is needed in the future).

With the second issue, there is work for schema change + db upgrade, but there is the potential to revoke, which will revoke to all users.  Seems reasonable if we are going to use the revoke feature.

Third approach is likely an overkill.

api/soap/mantisconnect.wsdl
```diff
@@ -163,6 +163,7 @@
    <xsd:element name="monitors" type="tns:AccountDataArray" minOccurs="0"/>
    <xsd:element name="sticky" type="xsd:boolean" minOccurs="0"/>
    <xsd:element name="tags" type="tns:ObjectRefArray" minOccurs="0"/>
+   <xsd:element name="token" type="xsd:string" minOccurs="0"/>
```
vboctor: Are you expecting just to return this when getting issues and ignore it when add/updating issues?  I don't know if "token" is a meaningful name.  Should it be just the random token or the full url? e.g. "share_url" or "anonymous_share_url", etc.

api/soap/mc_issue_api.php
```diff
@@ -138,6 +138,11 @@ function mc_issue_get( $p_username, $p_password, $p_issue_id ) {
 	$t_issue_data['custom_fields'] = mci_issue_get_custom_fields( $p_issue_id );
 	$t_issue_data['monitors'] = mci_account_get_array_by_ids( bug_get_monitors( $p_issue_id ) );
 	$t_issue_data['tags'] = mci_issue_get_tags_for_bug_id( $p_issue_id, $t_user_id );
+	if( ( ON === config_get_global( 'public_urls' ) ) && isset( $t_bug->token ) ) {
+		$t_issue_data['token'] = mci_null_if_empty( $t_bug->token );
+	} else {
+		$t_issue_data['token'] = mci_null_if_empty( '' );
```
vboctor: Should this just be set to null?

api/soap/mc_issue_api.php
```diff
@@ -1659,6 +1670,12 @@ function mci_issue_data_as_header_array( BugData $p_issue_data ) {
 		$t_issue['attachments_count'] = count( mci_issue_get_attachments( $p_issue_data->id ) );
 		$t_issue['notes_count'] = count( mci_issue_get_notes( $p_issue_data->id ) );
 
+		if( ( ON === config_get_global( 'public_urls' ) ) && isset( $p_issue_data->token ) ) {
```
vboctor: Is this attribute important enough to be in the header information for the list of issues?

bug_view_limited_inc.php
```diff
@@ -0,0 +1,676 @@
+<?php
```
vboctor: I have done a fair bit of work over time to reduce the number of pages we have.  For example:

- Combine simple/advanced pages.
- Combine view/print pages.

I really would rather avoid having yet another page to view the issue.  As I said in other comments, if we can drive what should appear via the access level (VIEWER), that would make a lot of sense.  If the developer team doesn't want something to show, then avoid sharing the url or use private notes for example.  There is likely gaps between what this approach can do and the requirements you are trying to achieve that may be worth a discussion.

I also expect what you think should be viewed in public/private issues when viewed via a share url may be slightly different from what others believe should appear.  Another thing that we need to think about.

GunSmoker:
@dregad 

> Meaning what, you can't run the PHPUnit tests ? Or they run but give different results ?

I tried to run phing. It gives: "Buildfile: build.xml does not exist!". I figure out phing is automation build tool, so I tried to run tests manually. Looking at travis_script.sh, I figure out I could run:
> C:\PHP\php.exe C:\PHP\phpunit.phar --bootstrap ./tests/bootstrap.php ./tests/AllTests.php > tests\results.txt

Well, it runs, but does not show any errors. If I explicitly add some syntax bug - it will fail.
Needless to say, reporting bugs manually via UI or submitting via SOAP API also works without any error.

I tested this on clean installation as well. Same result. PHP is configured for development, so display_errors is ON.

> FYI, you can always review the Travis execution logs

Yes, but it does not show exception line, and it does not show call stack. So, pretty useless.

> The problem even be reproduced via the GUI, the error below occured when reporting a new issue.

Thanks for that! It appears that your tests does not account DB changes? Otherwise, I do not understand why your bug table does not have new 'token' field. Or did I mess up with DB changes?

GunSmoker:
@atrol 

> changes in manage_user_page.php are not related to this PR

These are not mine. I think I have messed up when trying to keep my clone up to date with ongoing commits to official Mantis. Sorry, I am not familiar with Git.
 
> bug_view_limited_inc.php introduces quite a lot of redundant code (compare with bug_view_inc.php)

Do you mean file itself or its content? If it is about content - I thought that it is best to keep more code, even if it is not used, so one can simply enable it in the future, if one wants to show more info.

> introduces some redundant code following if ( ( ON === config_get_global( 'public_urls' ) )

I could not understand why your Travis tests fails (see my reply above to dregad), so I probably added more safe-guards checks.

GunSmoker:
@vboctor 

> Do we really need a db change vs. generate tokens from hashes.

Technically, I can use something like signature: token = SHA1(bugid + server-secret), but that means that we can't revoke access for specific URL / bug, if we will want to do that. DB field just offer more flexibility, even if it is not used right now. I occasionally used wiping out tokens in our practice, so I guess it is useful to have.

P.S. I already got some comments sent to me directly to e-mail, asking if it would be better to have per-user/per-bug tokens with ability to grant and revoke access.

> Can we avoid having another issue view page vs. just configure the list of fields that should show up on the standard view issue page.

There are few reasons for standalone page. 

First, is historic - I initially made isolated page, so I could upgrade Mantis more easily. But, of course, this is not the point for you. 

Another reason: if you make bug_view_inc to display limited read-only view - then you will end up with "ugly" page which has a lot of blank spots. bug_view_limited_inc rearranges fields, to keep limited fields grouped together. See screenshots: http://www.mantisbt.org/bugs/view.php?id=22203

Of course, I can add more conditions and rearrange field in bug_view_inc. But I think that would add a lot of unnecessary complication. 

Lastly, (in my opinion) a full featured editable work place and the limited read-only "screenshot" / "business card" has very little in common and could be significantly different visually. Even if it is not so right now.

> I would like not to change the authentication code, but set the active user to a user with the appropriate access level (VIEWER). Such users would see pretty much all of the issue details excluding private notes for example.

How one can select user_id for that? Basically, I tried to change current/active user to reporter of bug report, but perhaps I did it wrong and there is a simpler way? I checked that current user comes from cookies, but is there a way to avoid that? Basically, I do not want to override cookies. For example, if you are already logged in, but do not have access to this specific bug, it would be nice to let you view bug via Public URL, but do not log you off.

> Should we support share urls for private issues? Should it behave differently than public issues?

I see it that way: if one can see bug report, then nothing stops him from taking screenshot and share / post it. Public URLs is a similar thing, except it greatly reduces amount of exposed information. Currently I made just one the difference between private / public: private bugs does not show description and steps to reproduce (I assume it can contain sensitive info), public bugs do.

Here are two usage cases, from which this feature comes:
1. Your software encounters bug; 
2. Your software submits bug to your bug tracker;
3. Your software receives token;
4. Your software thanks user and offer him a Public Share URL, so he can check when bug will be fixed.

1. Your customer report issue with your software to your help desk;
2. You confirm that this is a bug in your software;
3. You add new bug to Mantis;
4. You copy Public Share URL and give it to user, so he can check when bug will be fixed.

Considering these usage cases, I do not see that there must be some extra limitations. After all, you share bug with the person who actually reported it, there is no leak of info.

Anyway, if one does not like default rules - fine, just don't enable the feature, you will not lose anything. If one wants other rules - nothing stops one from altering this feature in the future to his liking (e.g. adding settings, etc.). I think it is better to have something working, instead of ideal feature which is not implemented (like votes in Mantis).

bug_report.php
GunSmoker: @vboctor 

I was not sure which file I should use for that function. Do you have any suggestions?

config_defaults_inc.php
GunSmoker: @vboctor

I simply do not have deep knowledge of Mantis internals.

core/bug_api.php
GunSmoker: @vboctor 

Token can be requested by other means, not just via UI. For example, SOAP API. Or any future code which might refer this feature.

I thought it would be slow to upgrade large DB.

docbook/Admin_Guide/en-US/config/misc.xml
GunSmoker: @vboctor 

Check out screenshots: http://www.mantisbt.org/bugs/view.php?id=22203#c55235

It is just a short presentation to know status of bug, there are no notes or other sensitive info.

api/soap/mantisconnect.wsdl
GunSmoker: @vboctor 

Fogbugz returns just token (not full URL), so I just followed the same practice. I assume that is because one can access your server via other URL than saved in your settings? Not sure, but I think this is not a great deal.

It is read-only server-generated value - same as bug ID, so it is totally ignored for "add"/"update".

api/soap/mc_issue_api.php
GunSmoker: @vboctor

I really not sure. Should I remove it?

GunSmoker:
@dregad 

I think I have fixed the problem with tests. It was my mistake. Thanks for your help!

cproensa:
Some thoughts:
- The proposed implementation modifies authentication api, adding some special cases, that can be problematic with consistency, mantainability, etc
- Adding fields to bug table, BugData structure... at first, i would avoid adding field there and keep those structures only to basic bug data. If neccessary, create a dedicated table, or use current token functionality, or look for a way to extend current tokens to allow linking to bugs.

Regarding the use of a separated view page, or integrating in current one:
- We should avoid cluttering current view page with special cases to check for every element if it must be shown or not
- Using a logged user to reuse current view page can be problematic. Victor suggested to create a dedicated user, like it's done with anonymous access. This dedicated user should ideally have "viewer" access. This user makes sense to be the same anonymous user. It has to be worked to allow this special case login even if anonymous login is disabled in the application.
- A standalone page can make sense, where this page can act like a logged-out page. No navigation, no menus, and only show a the bug description is a very specific format.

I also wonder, if this could be implemented as a plugin. I think the only tricky part is the plugin page being accessed by unlogged users. If this is possible, the plugin can manage all the stuff for tokens, configuring what information to show, and how to retrieve it from the apis.

GunSmoker:
@cproensa @vboctor 

You do not like changes in auth API, which I can fully understand. However, you are suggesting to make analog of anonymous_account setting. But this setting is checked like everywhere, isn't it? I mean, even if I make this setting, I can not avoid altering auth API.

GunSmoker:
I have removed safe-guard checks from auth API. Seems to work OK, but someone with a deep knowledge of auth API should really get a look.

api/soap/mantisconnect.wsdl
vboctor: The URL is either detected from the URL used to access the API or is fixed as part of the configuration, so I'm not sure this is a reason to just return the token.  Returning the full URL allows the URL to be opaque to the client and enable changing how such URL works and what script it points to without changing the clients.

We can also consider whether this should be a separate API to get the share URL instead of this being part of the normal issue get method.  That also enables us to separate the code into a plugin if we decide to do so.

docbook/Admin_Guide/en-US/config/misc.xml
vboctor: The limited view page seems to be missing interesting information to get the full status of an issue including:
- Status
- Public notes

Do we need something similar to `$g_bug_view_page_fields` and flags to control aspects like showing public notes.

bug_monitor_list_view_inc.php
```diff
@@ -110,6 +110,16 @@
 		<?php } ?>
 	</td>
 </tr>
+<?php if ( ( ON === config_get_global( 'public_urls_enabled' ) ) && isset( $t_share_link ) && ( !is_null( $t_share_link ) ) && ( $t_share_link != '' ) ) { ?>
```
vboctor: We should not have the URL visible by default or printed when an issue is printed.  But have a page that discloses the URL when a Share button is clicked.  This also reduces changes to core pages.

bug_report.php
```diff
@@ -341,6 +344,11 @@
 	array( string_get_bug_view_url( $t_bug_id ), sprintf( lang_get( 'view_submitted_bug_link' ), $t_bug_id ) ),
 	array( 'view_all_bug_page.php', lang_get( 'view_bugs_link' ) ),
 );
+
+if ( $t_share_link != '' ) {
+	$t_buttons[] = array( $t_share_link, lang_get( 'share_this_bug' ) );
```
vboctor: I would drop this button and again limit it to a Share button on the View Issue page that shows a page that discloses the URL.

vboctor:
@GunSmoker I think re-factoring this as a plugin will drive the right behaviors.  My feedback was heading in that direction, but I think re-factoring as a plugin will force the right design behaviors:

- The plugin will own its own table for tokens.
- The plugin will own its own the public (limited) view page. 
- No changes in core pages or core APIs.
- No changes to core schema
- Should be able to add a "Share" button to the View Issue page that creates/provides a token for the issue.
- Can expose an API to be used by the app after creating an issue.
- Installing the plugin == enabling the feature.

Future changes post v1 of the plugin can include:

- You can enable UI for revoking a share URL, by creating a new token.
- Customizations to what appears on the public view page.

If there is a plugin limitation that doesn't make this possible, we can address that by enhancing MantisBT core.

GunSmoker:
@vboctor 

I can understand why you do not want to add extra functionality to the Mantis core, which is already overloaded with options. However, I also think that you are seeing this from a wrong perspective.

Mantis has "Bug Tracker" in its name, which means it is supposed to be a specialized tool to report bugs. Did you saw two usage cases in https://github.com/mantisbt/mantisbt/pull/1001#issuecomment-274263740 ? Do you think it is "unusual" or "rare" usage cases for a bug tracker, so it should be offloaded to a plugin?

For example, if you develop desktop software, you usually have hundreds and thousands of users who are willing to send you a bug report. The simpler the send process is - the better. Ideally - just click a button. Creating extra user account on your bug tracker site definitely is not an option for these people, because most people will only report one or few bugs. Creating accounts works well when you have few people reporting many issues. 

I think that the ability to send bug report and track its progress should be a core feature, on top of which other features may be build. 

For example, bug tracker can receive e-mails and creates bug report for each incoming e-mail. It can send: "thank you for your report, you can track status of your report here: {public-link}" e-mail back to sender to confirm acceptance of the report and provide a way to monitor issue. You do not deal with accounts if you are communicating via e-mail. This is an existing feature in JIRA, Fogbugz, Redmine, Bugzilla, but not in Mantis (AFAIK). Additionally, bug tracker can create a limited "e-mail"/"anonymous" account assigned with sender's e-mail, so this person can get extra benefits when he "logs in" by following public link from e-mail. Such as seeing his other reports (from the same e-mail). This is called "community user account" in Fogbugz.

Another example: developer can enter any e-mail into "Monitor" edit and bug tracker will send a e-mail to this address (assuming no bug tracker's account exists with this e-mail): "This bug report may be interesting for you: {public-link}". This is called "Sharing" feature in JIRA.

And what about SOAP API? Can it be modified with a plugin? What about WSDL? Ability to get public link after submitting the bug is the whole reason for this feature.

P.S. BTW, there are two other important pieces which Mantis misses, but other bug trackers have: a technical alias (bug ID / bug hash) and count / occurencies. Which I also planned to add, as I tired of workarounds.

bug_monitor_list_view_inc.php
GunSmoker: @vboctor Sorry, can you clarify? I do not see "monitor" section on printed issue, therefore there is no public link. Besides, exporting or printing also shows private notes, etc., so why would you want to hide URL?

bug_report.php
GunSmoker: Why? That is one more click when you submit bug from your help/support desk.

docbook/Admin_Guide/en-US/config/misc.xml
GunSmoker: @vboctor I agree on the status, but I think that notes should be excluded. Unless you want to make "e-mail" / "community" / "anonymous" account, so user to log in by public link (possibly from e-mail), and offer him a little bit more functionality.

PR #1002, opened 2017-01-21
Enable searchable projects dropdown menu

syncguru:
Use list.js library to enable filtering of the projects dropdown menu when number of projects/subprojects exceeds 10.

Fixes #5731

screenshot from 2017-01-20 16-03-49

screenshot from 2017-01-20 16-04-53

screenshot from 2017-01-20 16-15-35


js/common.js
```diff
@@ -48,6 +48,14 @@ $(document).ready( function() {
         ToggleDiv( t_div );
     });
 
+    var options = {
+    	valueNames: [ 'project-link' ]
+    };
+    var list = new List('projects-list', options);
+    if(list.items.length <= 10) {
```
vboctor: Is there a reason for hiding the search box in javascript vs. server side?

core/layout_api.php
```diff
@@ -614,10 +620,17 @@ function layout_navbar_projects_list( $p_project_id = null, $p_include_all_proje
 		if( $p_project_id !== null ) {
 			check_selected( $p_project_id, ALL_PROJECTS, false );
 		}
-		echo '> ' . lang_get( 'all_projects' ) . ' </a></li>' . "\n";
+		echo '> ' . lang_get( 'all_projects' ) . ' </a></li>' . " \n";
```
vboctor: This line change seems to be unnecessary.

js/common.js
syncguru: Couple of reasons:
1- It is a frontend convenience, backend logic should not worry about it (if possible)
2- Easier to calculate the number of child elements vs. recursively counting projects and subprojects

syncguru:
@vboctor In response to your comments:
1- Fixed 
2- Filtering occurs by name. No concept of project vs. subprojects on the client
3- Added to library/readme.md

css/ace-mantis.css
```diff
@@ -294,6 +294,35 @@ i.datetimepicker {
     color: #333333;
 }
 
+.dropdown-menu .divider {
+    margin: 2px 0;
+}
+
+.list > li {
```
vboctor: Should the styles here be specific to project list rather than a generic name "list"?  You have a id `projects-list` and class `projects-searchbox`  for this already.

syncguru: The style changes are generic and needed for any future use of this list.js library with Mantis Modern UI. The projects-list entry is very specific to the implemented search box in a dropdown thus I qualified it. There is no right or wrong here, it is just a judgement call.

vboctor: Sounds good.

dregad:
> There is no where in the checkin where the url for the list control is available. This is important to get updates, check license, read more about the library, etc. I think that our library/readme should include info about javascript libraries as well.

I don't mind, but if we document this one, then for consistency I think we should also do the same for all the other ones too: dropzone, ace template, and probably bootstrap and jQuery as well.

syncguru:
@dregad That's a fair request but I don't think it should part of this PR. We need a new issue to track your request.

dregad:
Sounds like unnecessary administrative overhead to me, for something that is purely internal documentation and could easily be achieved here with a simple commit. But If you really want the extra work, that's fine by me; I let you open another issue and related PR in that case.

vboctor:
Let's update the readme organically as we we are updating the libraries or as we start using package managers.  In the meantime, anyone can update it if they feel like it.

PR #1003, opened 2017-01-21
Markdown: Fix broken layout in news page (main_page.php)

jllano:
Ensure to unset style wherein markdown is enclosing `<p>` tag around news headline

Fixes #22204


vboctor:
I've applied the fix I suggested above via f1aaddfb50f80374d1b622a887b493d4aeb1bcde

PR #1004, opened 2017-01-23
Markdown: Add Markdown own styling

jllano:
- customize markdown styling
- this commit use classes as preferable over the current use of hardcoded inline styles
- this commit added custom style for blockquote markdown

Fixes #22190

### DIFF
--- a/css/default.css
+++ b/css/default.css
@@ -107,4 +107,5 @@ td.print-overdue    { font-weight: bold; }
 
 .markdown-body pre {
 	border: 0;
+	margin-top: 9.5px;
 }
--- a/css/default.css
+++ b/css/default.css
@@ -92,3 +92,10 @@ td.print-overdue    { font-weight: bold; }
 .table-nonfluid {
    width: auto !important;
 }
+
+/* Markdown */
+.markdown-body blockquote {
+	padding: 0.13em 1em;
+	color: #777;
+	border-left: 0.25em solid #C0C0C0; font-size: 13px;
+}
--- a/css/default.css
+++ b/css/default.css
@@ -100,7 +100,11 @@ td.print-overdue    { font-weight: bold; }
 	border-left: 0.25em solid #C0C0C0; font-size: 13px;
 }
 
-.markdown-body p code {
+.markdown-body code {
 	color: #777;
 	background-color: #f5f5f5;
+}
+
+.markdown-body pre {
+	border: 0;
 }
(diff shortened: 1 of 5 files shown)


css/default.css
```diff
@@ -92,3 +92,10 @@ td.print-overdue    { font-weight: bold; }
 .table-nonfluid {
    width: auto !important;
 }
+
+/* Markdown */
+.markdown-body blockquote {
```
vboctor: I don't see why we would require a .markdown-body.  There will be a lot of places where markdown will be rendered due to calls to strings_display function calls that are not just issue details.  For example, the news deprecated function, or other areas.

jllano: > There will be a lot of places where markdown will be rendered

Like how many places do you think? 10 would be sound reasonable.

That way would make changes in design easier without affecting if admins customize the CSS should they want later.

jllano: I assumed that we would add the style only to the places that requires text formatting wherein markdown is available as well.

vboctor: We should support markdown everywhere where the following APIs are called by MantisBT core or plugin:

- string_display
- string_display_line
- string_display_links
- string_display_line_links
- string_rss_links

If so, then it is pretty much anywhere within the MantisBT emitted content.

Let's see what others think of the approach as well.

jllano:
screen shot 2017-01-25 at 2 52 35 pm


Notes section:
screen shot 2017-01-25 at 2 52 46 pm

jllano:
What do you think @dregad @atrol ?

jllano:
ping @dregad @atrol ?

dregad:
Didn't have time to test yet, sorry. Could you please resolve merge conflicts ?

jllano:
Conflicts resolved

jllano:
This https://github.com/mantisbt/mantisbt/pull/1004/commits/3fd0260b1b0119aa9f580a8a8229f5ee934aa7f2 fix the https://www.mantisbt.org/bugs/view.php?id=22485 issue

Before:
screen shot 2017-03-08 at 11 34 12 pm

After:
screen shot 2017-03-08 at 11 34 29 pm

vboctor:
I was thinking about HTML emails, and one thing came to mind is that we will need to convert issue markdown to HTML to include in such emails.  Often use of non-inline styles does not work in email clients.  I want to hold off on changing the way we handle styles until we have a basic html model that enables us to test this.

dregad:
Closed in favor of #1841

PR #1005, opened 2017-01-23
Fix markdown eating apostrophe or single quote

jllano:
Removed condition that fix the img tag causing apostrophe or single quote not working, as we don't need to reference the image from the internet at the moment. Instead, we will just add it or override the custom method in plugins/MantisCoreFormatting/core/MantisMarkdown later.

Fixes #22179


jllano:
Before:
screen shot 2017-01-23 at 10 01 02 pm

After:
screen shot 2017-01-23 at 10 01 39 pm

PR #1007, opened 2017-01-26
Markdown: Fix markdown converting '&' signs to '&amp;' inside the code block and 'a' inline element

jllano:
- Moved logic removing the html entity `&amp;` from MantisCoreFormattingPlugin into MantisMarkdown Extention

Fixes #22242, #22246

### DIFF
--- a/plugins/MantisCoreFormatting/MantisCoreFormatting.php
+++ b/plugins/MantisCoreFormatting/MantisCoreFormatting.php
@@ -107,31 +107,6 @@ class MantisCoreFormattingPlugin extends MantisFormattingPlugin {
 	}
 
 	/**
-	 * Process Markdown
-	 * @param string  $p_string    Raw text to process.
-	 *
-	 * @return string Formatted text
-	 */
-	private function processMarkdown( $p_string ){
-
-		$t_string = $p_string;
-
-		# We need to enabled quote conversion
-		# "> quote or >quote" is part of an html tag
-		# Make sure to replaced the restored tags with ">"
-		$t_string = str_replace( "&gt;", ">", $t_string );
-
-		$t_string = MantisMarkdown::convert_text( $t_string );
-
-		# markdown is unable to process '<' signs within the code block tag (backticks)
-		# markdown convert '<' signs into '&amp;lt;' after the text proccessing process
-		# therefore we need restore the entity back into its original name '&lt;'
-		$t_string = str_replace( '&amp;lt;', '&lt;', $t_string );
-
-		return $t_string;
-	}
-
-	/**
 	 * Plain text processing.
 	 *
 	 * @param string  $p_event     Event name.
@@ -216,7 +191,7 @@ class MantisCoreFormattingPlugin extends MantisFormattingPlugin {
 
 		# Process Markdown
 		if( ON == $s_markdown && $p_multiline ) {
-			$t_string = $this->processMarkdown ( $t_string );
+			$t_string = MantisMarkdown::convert_text( $t_string );
 		}
 
 		return $t_string;
--- a/plugins/MantisCoreFormatting/core/MantisMarkdown.php
+++ b/plugins/MantisCoreFormatting/core/MantisMarkdown.php
@@ -77,6 +77,12 @@ class MantisMarkdown extends Parsedown
 	 */
 	public static function convert_text( $p_text ) {
 		self::init();
+
+		# Enabled quote conversion
+		# Text processing converts special character to entity name 
+		# Make sure to restore "&gt;" entity name to its characted result ">"
+		$p_text = str_replace( "&gt;", ">", $p_text );
+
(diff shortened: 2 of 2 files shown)


jllano:
your thoughts @dregad @vboctor @atrol

dregad:
@jllano wouldn't this cause problems if markdown is switched off ?

jllano:
@dregad Absolutely no, that changes works only if markdown is enabled, it has nothing to do with the text processing as well.

plugins/MantisCoreFormatting/MantisCoreFormatting.php
```diff
@@ -123,11 +123,6 @@ private function processMarkdown( $p_string ){
 
 		$t_string = MantisMarkdown::convert_text( $t_string );
```
vboctor: Should we completely remove processMarkdown() and call directly MantisMarkdown::convert_text() and have it do all the processing, e.g. the str_replace just above this line?  This makes the code easier to unit test and remove this unnecessary indirection step.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -225,6 +225,60 @@ protected function blockQuoteContinue( $line, array $block ){
...
+	 * @param array $block A block-level element
+	 * @access protected
+	 * @return string html representation generated from markdown.
+	 */
+	protected function inlineCode( $block ) {
+
+		$block = parent::inlineCode( $block );
+		
+		if( isset( $block['element']['text'] )) {
+			$block['element']['text'] = str_replace( '&amp;', '&', $block['element']['text'] );
```
vboctor: It is unclear to me why we need to add such logic.

jllano:
> I have tested in the demo for parse down the code block with quotation, the ">quote" and "> quote". So I don't know why we need this special handling. We shouldn't need special handling to handle standard markdown, so unless we are handling a special MantisBT concept, then we should use parsedown without customization.

Can you specify the markdown you tested from the demo "code block with quotation"?

plugins/MantisCoreFormatting/MantisCoreFormatting.php
jllano: That make sense, we can move this logic also  `$t_string = str_replace( "&gt;", ">", $t_string );` to the class Extension.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
jllano: The logic there simply replaced any entity `&amp;` to `&`:

screen shot 2017-02-25 at 9 01 09 am


1. if we enter `&` sign within the code block or backtick, the text processing will convert that into entity name, and that becomes `&amp;`.
2. Now, markdown will check if they found `&` sign, then it will converted to `&amp;`. So that the output string now becomes `<code>&amp;amp;</code>`

Let me know if that make sense.

ajtruckle:
I pasted xsl and HTML code from Visual Studio and in mantis all ' turned visually into &quot;  

Thus the user cant paste the code. I also think the code window should be limited in height with scroll bar and a select all code button.

ajtruckle:
" turned into & quot

dregad:
> I also think the code window should be limited in height with scroll bar and a select all code button.

@ajtruckle this is a separate topic, please [open a new issue](http://www.mantisbt.org/bugs/bug_report_page.php) to track your enhancement request

ajtruckle:
@dregad Done so here: https://github.com/ejyothi/MantisBT-MarkDownEditor/issues/3

dregad:
@ajtruckle fine, but if you expect this to become an improvement in MantisBT itself, I'm sure you'll agree that a 3rd-party plugin's issue tracker is not the best place for it.

ajtruckle:
@dregad I was told already to direct my issues about the markdown editor side of things to that particular tracker rather than to MantisBT issue tracker....

dregad:
> direct my issues about the markdown editor side of things to that particular tracker rather than to MantisBT issue tracker

@ajtruckle That is indeed what you should do. 
I just wanted to make sure that you really meant your comment as an enhancement to the text areas' edit mode (i.e. markdown editor plugin or native text area), vs the way we render code blocks in MantisBT (in which case it does belong in our tracker)

ajtruckle:
Hhm. Well I reckon both actually.

Sent from my iPhone

On 28 Feb 2017, at 09:42, Damien Regad <notifications@github.com<mailto:notifications@github.com>> wrote:


direct my issues about the markdown editor side of things to that particular tracker rather than to MantisBT issue tracker

@ajtruckle<https://github.com/ajtruckle> That is indeed what you should do.
I just wanted to make sure that you really meant your comment as an enhancement to the text areas' edit mode (i.e. markdown editor plugin or native text area), vs the way we render code blocks in MantisBT (in which case it does belong in our tracker)

PR #1008, opened 2017-01-26
Automated crash reporting in Mantis

GunSmoker:
Implementation of http://www.mantisbt.org/bugs/view.php?id=22263

1. Added "alias" and "count".
2. Added "report bug" API.

Screenshots of working feature: http://www.mantisbt.org/bugs/view.php?id=22263#c55347

Details:

Alias is implemented as standalone table. That is mostly because I used UNIQUE constraint to ensure there are no concurrency issues - because not all databases support multiple NULLs for UNIQUE constraint. Count is implemented as integer field in bug table. 

Added ability to create a restricted API token which can only submit bug report and nothing else. It can be used by mc_login, mc_issue_report (see below) and bug_report.php.

Added SOAP.mc_issue_report API, which is sort of a combination of mc_issue_add and mc_attachment_add. It accepts restricted token (or the usual credentials). Added ability to create bug by posting to bug_report.php supplying auth_username and auth_password with valid token.

mc_issue_add, mc_issue_report and bug_report.php supports merging crashes by alias into single bug report by increasing count. Added various options to control how attaches should be handled for duplicates.

Count column is shown by default. It can be sorted.

Added "Me too!" button (like "I have this issue too!") to add ability to manually "vote" for manually created bug.


GunSmoker:
Don't worry, this is the last suggestion that I submit to Mantis :)

rombert:
I'll hold off the comments on the SOAP API changes and will wait for the other devs to comment on whether this is something that should go in.

> Don't worry, this is the last suggestion that I submit to Mantis :)

Why :-) ?

GunSmoker:
Just in case I have bothered you enough :)

vboctor:
Thanks @GunSmoker for your contribution.  I really appreciate your focus on scenarios and researching equivalent implementations in other bug tracker.  Keep them coming!

Before I go into reviewing the code, I would like to provide some high level comments:

- I like the idea of having this feature building on the infrastructure that can power the Voting scenario.
- I don't get why we need a different authentication mechanism / different type of authentication token.  We can have a user for the app (global or per app) that can report exceptions using the standard username / api token.
- Long time ago I've implemented SOAP API mc_issue_get_id_from_summary() for similar purpose.  Where app can generate a unique summary: "Exception (hash)" and then before reporting call this API, and create a note or a new issue accordingly.  However, I'm OK with a new API that would do this on MantisBT side and increment the vote count.
- I would use "Vote" instead of "Me Too!".  We can even consider in the future voting for or against the feature.
- If we are to create a new table, then I wonder if we should capture the users voting and their vote (for vs. against).  In this case, you can have 7 votes for and 3 against, and net would be 4.  The vote total can even be negative.  Another approach is to have the vote details in the history, and just have a counter.  If we just have counter, then maybe having it on the bug table is more efficient.
- I would call the column title "Vote" if we are factoring +1 and -1.  "Votes" can work too, but I wonder if it gets mixed up with count rather than total.

These are some thoughts to get the discussion going, let's not change the code yet until we have the discussion and agree on an approach.

GunSmoker:
@vboctor 

> I don't get why we need a different authentication mechanism / different type of authentication token. We can have a user for the app (global or per app) that can report exceptions using the standard username / api token.

If you try to implement crash reporting from application, you would probably do the following: 
1. Call mc_filter_search_issue_ids with your custom field for BugHash = current crash (or mc_issue_get_id_from_summary in older versions);
2. If none found - mc_issue_add;
3. If found - mc_issue_get + mc_issue_update (to +1 to count / occurrences);
4. mc_attachment_add.
So, crash reporter account should be able to search, create, read, edit and attach. Of course, if we introduce a single "report" API (like mc_issue_report) - then 1-4 will be collapsed into a single API call. However, even then you have to have account which can access this API - thus have rights to view / modify / attach.

And Mantis credentials would be hard-coded into application. 

This does not matter if you have web app. However, for any sort of desktop app - credentials can be extracted from executable. Which means that rogue user can read any report, add any report, modify any report. Not to mention that once credentials are extracted, rogue user can simply log in to your bug tracker (if it have public URL). While this is an unlikely case, I would sleep better if crash reporting feature would allow only "post new report" and nothing more. No rogue reads, no rogue edits, and definitely - no logins!

That's why I think that having some sort of restricted API token would be a very nice feature. You can't login with it, you can't read reports, you can't modify reports - the only thing that you can is to submit the reports.

> Long time ago I've implemented SOAP API mc_issue_get_id_from_summary() for similar purpose.

Yep, that is what I was using - until mc_filter_search_issue_ids was introduced.

> I would use "Vote" instead of "Me Too!". We can even consider in the future voting for or against the feature.
> If we are to create a new table, then I wonder if we should capture the users voting and their vote (for vs. against). In this case, you can have 7 votes for and 3 against, and net would be 4. The vote total can even be negative. Another approach is to have the vote details in the history, and just have a counter. If we just have counter, then maybe having it on the bug table is more efficient.
> I would call the column title "Vote" if we are factoring +1 and -1. "Votes" can work too, but I wonder if it gets mixed up with count rather than total.

I see this like so: on one hand - there are crash reports / bugs. I do not see much sense in "down-vote a bug". What does it mean? However, it does have sense to have count / occurrences and "I have this bug too!", which contributes to total count. 

On the other hands, there are suggestions / feature requests. And for these - yes, you may want a real voting, like "thumbs up / thumbs down". If there will be voting implemented - then additional "count" does not make any sense to me (for suggestions). And probably ability to withdraw your vote makes sense for suggestions and does not make sense for bugs.

So, for starters - a single "+1"-like button and a single field may be added. It may change its caption depending on if report is bug report or suggestion (e.g. "Me too!" / "Count" for bugs, and "Vote" / "Votes" for suggestions). Internally it would be the same, just different text. 

And then other enchantments can be added, if there will be such wish. Currently I have used history to track if user has already attributed to count, so button should be hidden. The same could be used to track count of "thumbs up" and "thumbs down", where "occurrences" would represent sum of both (yes, it can be negative for suggestions).

> We should have a limit of max number of crash reports (configurable) after which, we shouldn't accept more reports, or maybe increment a counter without triggering email notifications.

Yes, I have added this setting. 50 is default (same as Fogbugz). Additionally, switching status to acknowledged / confirmed / assigned would block further reports, which makes sense and allow one to manually control reports collecting (for example, reports may be submitted as "acknowledged" by default, and if developer is unable to identify bug - issue may be switched to "feedback" to gather additional reports).

GunSmoker:
@vboctor 

BTW, what happenned to this?
https://www.mantisbt.org/wiki/doku.php/mantisbt:issue_voting_requirements
https://www.mantisbt.org/bugs/view.php?id=668

It started in 2001, got some implementation in 2008 - and now all links are dead. I was able to find only some non-used texts in english.txt.

I personally do not need +1/-1 voting feature, but I also do not want to repeat or interfere.

core/bug_api.php
```diff
@@ -851,6 +1069,45 @@ function bug_cache_row( $p_bug_id, $p_trigger_errors = true ) {
...
+		return false;
+	}	
+	
+	# Bug is read-only?
+	if( $t_status >= config_get( 'bug_readonly_status_threshold' ) ) {
+		return false;
+	}	
+	
+	# Bug is closed?
+	if( $t_status >= config_get( 'bug_closed_status_threshold' ) ) {
```
vboctor: This is likely covered by the check for bug_collect_reports_status_threshold.  If the admin doesn't want to collect for closed bugs, then they can set bug_collect_reports_status_threshold to CLOSED which may be the default.

core/bug_api.php
```diff
@@ -851,6 +1069,45 @@ function bug_cache_row( $p_bug_id, $p_trigger_errors = true ) {
...
+		return false;
+	}	
+	
+	# Bug is fixed?
+	if( $t_resolution >= config_get( 'bug_resolution_fixed_threshold' ) ) {
```
vboctor: I would rather keep collecting until the bug reaches the desired status / # of hits.

core/bug_api.php
```diff
@@ -851,6 +1069,45 @@ function bug_cache_row( $p_bug_id, $p_trigger_errors = true ) {
...
+	$t_status = bug_get_field( $p_bug_id, 'status' );
+	$t_resolution = bug_get_field( $p_bug_id, 'resolution' );
+	
+	# Bug is closed for submission?
+	if( $t_status >= config_get( 'bug_collect_reports_status_threshold' ) ) {
+		return false;
+	}	
+	
+	# Bug is read-only?
+	if( $t_status >= config_get( 'bug_readonly_status_threshold' ) ) {
```
vboctor: I would collect data until we hit the 'bug_collect_reports_status_threshold' status and not worry about read-only status.  This provides more flexibility to the admin.

lang/strings_english.txt
```diff
@@ -1184,6 +1187,7 @@ $s_resolved = 'Resolved';
 $s_total = 'Total';
 $s_balance = 'Balance';
 $s_most_active = 'Most Active';
+$s_most_votes = 'Hot';
```
vboctor: Hot is a vague term. Is it issues with most comments, most votes, etc.  Let's use a term that is consistent with the criteria we are using.

admin/schema.php
```diff
@@ -884,6 +884,27 @@
 
 # Release marker: 1.3.0
 
+# Count / occurrences / votes 
+$g_upgrade[210] = array( 'AddColumnSQL', array( db_get_table( 'bug' ), "
+	votes					I		NOTNULL DEFAULT '1'",
```
vboctor: I wonder if we should start with a vote of 0, so that votes represents someone other than the reporter indicating a +1.  Otherwise, we end up with a case where some votes are captured in history and others are not.

admin/schema.php
```diff
@@ -884,6 +884,27 @@
...
+	) );
+$g_upgrade[211] = array( 'CreateIndexSQL', array( 'idx_bug_count', db_get_table( 'bug' ), 'votes' ) );	
+
+# A standalone table is required for "alias", because not all DB allow multiple NULLs in UNIQUE index
+$g_upgrade[212] = array('CreateTableSQL', array( db_get_table( 'bug_alias' ), "
```
vboctor: I would have this as a field on the bug with default null/empty and drop the uniqueness.  I would leave it up to the app to generate a unique identifier based on the exception that is reproducible.  Having uniqueness in the db is useful but not absolutely required.

admin/schema.php
```diff
@@ -884,6 +884,27 @@
...
+	alias					C(128)	$t_notnull",
+	$t_table_options
+	) );
+$g_upgrade[213] = array( 'CreateIndexSQL', array( 'idx_bug_alias_project', db_get_table( 'bug_alias' ), 'project_id' ) );	
+$g_upgrade[214] = array( 'CreateIndexSQL', array( 'idx_bug_alias_alias', db_get_table( 'bug_alias' ), 'alias' ) );	
+$g_upgrade[215] = array( 'CreateIndexSQL', array( 'idx_bug_alias_uniq', db_get_table( 'bug_alias' ), 'project_id,alias', array( 'UNIQUE' ) ) );
+
+# Add API token kind, allowing to have different types of API tokens
+$g_upgrade[216] = array( 'AddColumnSQL', array( db_get_table( 'api_token' ), "
+	kind					I		NOTNULL DEFAULT '0'",
```
vboctor: Instead of having types, I would rather add the concept of scopes to tokens.  For example, default can be "['All']', and in the case of reporting {'Issue.Report'}.  SOAP API and web app would check the scope to determine if a user should be able to login, report an issue, etc.  Hence, the tokens created with "['Issue.Report']" would only be able to report issues, and no more.  The format of this would be a json array of strings.

api/soap/mantisconnect.wsdl
```diff
@@ -45,6 +45,7 @@
    <xsd:element name="account_data" type="tns:AccountData" minOccurs="0"/>
    <xsd:element name="access_level" type="xsd:integer" minOccurs="0"/>
    <xsd:element name="timezone" type="xsd:string" minOccurs="0"/>
+   <xsd:element name="api_token_type" type="xsd:integer" minOccurs="0"/>
```
vboctor: Do we need to return such information to enable bug reporting scenario?  I would rather minimize the scope of this change to what is required to light up the scenario.

api/soap/mantisconnect.wsdl
```diff
@@ -163,6 +164,8 @@
    <xsd:element name="monitors" type="tns:AccountDataArray" minOccurs="0"/>
    <xsd:element name="sticky" type="xsd:boolean" minOccurs="0"/>
    <xsd:element name="tags" type="tns:ObjectRefArray" minOccurs="0"/>
+   <xsd:element name="alias" type="xsd:string" minOccurs="0"/>
```
vboctor: Alias is really a concept that needs to only be understood by the report API and the app should be able to generate it on its own.  There is no benefit of returning this as part of the issue information as part of our public API.

api/soap/mantisconnect.wsdl
```diff
@@ -269,6 +272,7 @@
  <xsd:complexType name="FilterSearchData">
   <xsd:all>
    <xsd:element name="project_id" type="tns:IntegerArray" minOccurs="0"/>
+   <xsd:element name="alias" type="xsd:string" minOccurs="0"/>
```
vboctor: If the report issue API is going to handle all the internals of the bug reporting, then I don't know we would need to expose the alias in filter API.

api/soap/mantisconnect.wsdl
```diff
@@ -546,6 +550,16 @@
...
+<message name="mc_issue_reportRequest">
+  <part name="username" type="xsd:string" />
+  <part name="password" type="xsd:string" />
+  <part name="issue" type="tns:IssueData" />
+  <part name="attach_note" type="xsd:string" />
+  <part name="file_name" type="xsd:string" />
+  <part name="file_type" type="xsd:string" />
+  <part name="content" type="xsd:base64Binary" /></message>
+<message name="mc_issue_reportResponse">
+  <part name="return" type="tns:IssueData" /></message>
```
vboctor: That would be IssueReportData or something along these lines containing IssueData as an element.

api/soap/mantisconnect.wsdl
```diff
@@ -1000,6 +1014,11 @@
     <input message="tns:mc_issue_addRequest"/>
     <output message="tns:mc_issue_addResponse"/>
   </operation>
+  <operation name="mc_issue_report">
+    <documentation>Same as mc_issue_add + mc_issue_attachment_add, except returns full issue data instead of just id.</documentation>
```
vboctor: That is inaccurate, this API will do the 4 steps you indicated in our discussion.

- Search for the issue.
- If found, increment the vote / hit count.
- If not found, add it.
- Add the note + attachment.

api/soap/mc_api.php
```diff
@@ -139,11 +139,22 @@ function mc_version() {
  */
 function mc_login( $p_username, $p_password ) {
 	$t_user_id = mci_check_login( $p_username, $p_password );
-	if( $t_user_id === false ) {
-		return mci_soap_fault_login_failed();
+	if( false === $t_user_id ) {
+		$t_user_id = mci_check_login( $p_username, $p_password, API_TOKEN_REPORT_ONLY );
```
vboctor: Re-work this to work with scopes instead of token types.  Expect a proper user name that is associated with the token.

api/soap/mc_filter_api.php
```diff
@@ -63,6 +63,7 @@
     'end_year' => FILTER_PROPERTY_DATE_SUBMITTED_END_YEAR,
     'tag_string' => FILTER_PROPERTY_TAG_STRING,
     'tag_select' => FILTER_PROPERTY_TAG_SELECT,
+    'alias' => FILTER_PROPERTY_ALIAS,
```
vboctor: I wouldn't expose the alias here.  Again that would be an internal operation that can be done internally.

api/soap/mantisconnect.wsdl
```diff
@@ -546,6 +550,16 @@
   <part name="issue" type="tns:IssueData" /></message>
 <message name="mc_issue_addResponse">
   <part name="return" type="xsd:integer" /></message>
+<message name="mc_issue_reportRequest">
+  <part name="username" type="xsd:string" />
```
vboctor: If we keep the concept of having a user + limited token, then this field would be useful.  If not, then we don't need to get an empty user here just to ignore it internally.

bug_report.php
```diff
@@ -74,12 +75,29 @@
 require_api( 'string_api.php' );
 require_api( 'user_api.php' );
 require_api( 'utility_api.php' );
+require_api( 'api_token_api.php' );
 
-form_security_validate( 'bug_report' );
+# Check if this is a usual submission via UI or scripted autosubmission
```
vboctor: I don't understand why we would change bug_report.php or reference the auto report concept for here.  This should be only a concept that is understood by the new SOAP API.

bug_update.php
```diff
@@ -73,6 +73,19 @@
...
 
+if( gpc_get_int( 'vote', 0 ) ) {
+	if( access_user_can_vote_for_bug( $f_bug_id, $t_current_user_id ) ) {
+		$t_existing_bug->vote();
+		
+		if( access_has_bug_level( config_get( 'monitor_bug_threshold' ), $f_bug_id ) ) {
+			bug_monitor( $f_bug_id, $t_current_user_id );	
+		}
+	}
+	form_security_purge( 'bug_update' );
```
vboctor: Unclear to me why you would die here and not run the full script.  It feels that voting is just one of many actions that bug update should handle.

config_defaults_inc.php
```diff
@@ -977,12 +977,12 @@
...
  * summary, sponsorship_total, due_date, description, steps_to_reproduce,
  * additional_info, attachment_count, bugnotes_count, selection, edit,
- * overdue
+ * overdue, votes
  *
  * @global array $g_view_issues_page_columns
  */
 $g_view_issues_page_columns = array (
-	'selection', 'edit', 'priority', 'id', 'sponsorship_total',
+	'selection', 'edit', 'priority', 'votes', 'id', 'sponsorship_total',
```
vboctor: I would put the votes after the issue id.

config_defaults_inc.php
```diff
@@ -2085,6 +2085,28 @@
...
+ * if its status is >= $g_bug_collect_reports_threshold. The bug starts 
+ * accept more bug reports again if its status becomes less than this 
+ * threshold. The default is ASSIGNED. 
+ * Read-only, closed and fixed bugs are closed for bug report submission 
+ * regardless of this option.
+ * See also $g_bug_max_auto_reports option.
+ * See also SOAP.mc_issue_report function.
+ * @global integer $g_bug_collect_reports_status_threshold
+ */
+$g_bug_collect_reports_status_threshold = ACKNOWLEDGED;
```
vboctor: I would default this to RESOLVED.

config_defaults_inc.php
```diff
@@ -2518,6 +2542,15 @@
 $g_report_bug_threshold = REPORTER;
 
 /**
+ * Enables / disables ability to create new bug report by posting to 
+ * bug_report.php with 'auth_username' and 'auth_password' fields set.
+ * Only API tokens are allowed.
+ * This can be used for automatic collection of bug reports.
+ * @global integer $g_bug_auto_reports_submission_enabled
+ */
+$g_bug_auto_reports_submission_enabled = OFF;
```
vboctor: If we are using a user for this, then why do we need an OFF/ON?  Wouldn't creating / deleting the token do the trick?

core/bug_api.php
```diff
@@ -84,6 +84,16 @@ class BugData {
 	protected $id;
 
 	/**
+	 * Alias
+	 */
+	protected $alias = null;
```
vboctor: We should drop the alias and make it a calculated field from bug information.

core/bug_api.php
```diff
@@ -254,6 +264,11 @@ class BugData {
 	private $loading = false;
 
 	/**
+	 * Indicates if create() has used existed bug or created new bug
+	 */
+	private $aliased = false;
```
vboctor: This doesn't belong on the bug information.  This is specific the report issue API and should be part of its response only.

core/bug_api.php
```diff
@@ -783,6 +971,36 @@ function update( $p_update_extended = false, $p_bypass_mail = false ) {
...
+			
+		db_param_push();
+		$t_query = 'SELECT votes FROM {bug}
+						WHERE id=' . db_param();
+		$t_result = db_query( $t_query, array(
+						$c_bug_id ) );
+			
+		$this->votes = db_result( $t_result );	
+
+		history_log_event_special( $c_bug_id, BUG_VOTED );
```
vboctor: Should we capture +1 as a parameter to the BUG_VOTED history event.  So that later we may want to add the -1 case and they would be consistent?

core/database_api.php
```diff
@@ -309,9 +309,10 @@ function db_query_bound() {
  * @param integer $p_limit     Number of results to return.
  * @param integer $p_offset    Offset query results for paging.
  * @param boolean $p_pop_param Set to false to leave the parameters on the stack
+ * @param boolean $p_silent    Set to true to supress error messages if query fails
```
vboctor: We have survived so far without this, why do we need it now?

my_view_inc.php
```diff
@@ -342,7 +342,9 @@
 	<td class="nowrap width-13 my-buglist-id">
 		<?php
 			print_bug_link( $t_bug->id, false );
-
+			if( $t_bug->votes > 1 ) {
+				echo ' ^', $t_bug->votes;
```
vboctor: Maybe we should use '+' instead of '^', since we are using the +1 concept.  In the future, it can be +5 or -5 based on votes.

config_defaults_inc.php
```diff
@@ -2085,6 +2085,28 @@
...
+$g_bug_collect_reports_status_threshold = ACKNOWLEDGED;
+
+/**
+ * Maximum number of automatic reports to collect per bug when bug is open
+ * for collection of new reports (e.g. bug is not read-only, not closed, 
+ * not fixed, and not switched to $g_bug_collect_reports_status_threshold state). 
+ * See also SOAP.mc_issue_report function.
+ * @global integer $g_bug_max_auto_reports
+ */
+$g_bug_max_auto_reports = 50;
```
vboctor: It would be useful to have all crash reports tagged with a tag like 'crash', this way it is easy to filter on such issues.  This can be part of the issue information provided to issue report or we can add a configuration option to auto-add this to all issues added through such APIs whether or not the app provides the tag.

core/bug_api.php
GunSmoker: Why?

Almost every other implementation allow to specify it from client side. Calculation from server side will be too much complicated. 

For example, should it depend on version? E.g. does application wants to collect bugs for each its version individually? Does application want to consider the bug with same alias in a version after bug was fixed to be a new bug or the same bug? 

When alias is defined by application this can be easily customizable by including or excluding version field from calculating bug alias/hash.

However, if you want to make server calculation - then you have to add many options to include/exclude the following pieces:
- Exception address
- Exception type
- Exception message
- Exception module
- Project module
- Version for both of the above
- Call stack - either full or partial

Mantis simply have no access to this information, whereas clien application - does.

core/bug_api.php
GunSmoker: Would it be OK to make this as additional output param to bug's create()?

core/bug_api.php
GunSmoker: Probably not. 

+1/-1 could be implemented as "BUG_UP_VOTED"/"BUG_DOWN_VOTED". Whereas an integer arg would suggest that you can up/down vote by more than 1 (e.g. +100).

Anyway, this can be modified later when real voting will be added.

lang/strings_english.txt
GunSmoker: Do you have a suggestion? It just bothers me to name this "Most votes", because currently no other UI mentions anything about votes. It is "bug count" (occurencies) and "Me too". I thought "votes" term would be delayed until a proper +1/-1 votes system will be added.

admin/schema.php
GunSmoker: This counter is supposed to indicate how much this bug happenned. So, it does not really make much sense when automatic report would have this at 0.

Besides, even for manual reports "0 votes" seems a bit counter-intuitive to me. Does the feedback reporter himself not want this issue to be implemented? Does the bug reporter not want his bug to be fixed? 

And about history inconsistence - it HAS to be that way. Automatic reporting absolutely should not be reflected in the history. 

The problem here is that unlike manual reporting, auto reports happens at large scale (at least with non-web applications). There can be hundreds, thousands, sometimes million machines and devices with your app installed. Each device can report the same bug, which means you can have millions upvotes. 

Do you really want you bug view page to have a list of a million upvotes from the same reporter account? This would be just a noise, with no useful information.

admin/schema.php
GunSmoker: Yes, it is not required. However, it is here for another reason, as indicated in the comment. I simply don't know how to add atomic operation in another way. E.g. if two applications/machines would try to post the same report (with same alias) at about the same time, Mantis should create only one report and set count to 2 (+1 on creation, +1 on second report). If you do not make this an atomic operation, you would end up with two reports with same alias - which is a big NO-NO.

I checked that Bugzilla uses transactions for that. Other implementations may use either transactions or INSERT OR UPDATE statements. I found that Mantis does not use any of those. So, introduce an unique key, try to insert, fail if it exists - it is a best try that I can do. 

If you have other suggestion how to implement atomic insert/update - you are welcome to share.

admin/schema.php
GunSmoker: I don't like fixed "types" either, but do you have any idea on what other scopes could be? Would it be solving problem that does not exist?

api/soap/mantisconnect.wsdl
GunSmoker: It is not strictly necessary, but desirable. 

Here is an usage case: imagine if you develop application. You don't want to reinvent the bicycle, so instead of writing your own error reporting solution, you pick up some 3rd party software. So, you feed you login credentials and other settings into software. Now, your application crashes and this software has to send bug report to Mantis. Currently this software would make a search for "fake alias", made insert or update. If this pull request will be implemented, then this software could simply use new report-only API or it could stick to old method, if there are some additional logic involved. 

How does this software recognize if it could call other methods from API? Try to call it and fail when access denied is returned? It could. But it also could just determinate what type of token you (as developer of application which uses this software) have provided. Type would determinate which API calls software can make.

api/soap/mantisconnect.wsdl
GunSmoker: Yes, it is not strictly necessary, but wouldn't it be an additional flexibility?

api/soap/mantisconnect.wsdl
GunSmoker: There can be several reporting-only accounts set up (one per project/app), and currently Mantis uses username (which has to be non-empty even with tokens) to determinate user id, against which it then checks token. So I don't see how this can be removed.

api/soap/mantisconnect.wsdl
GunSmoker: Why? Is this because you want to separate alias field?

bug_report.php
GunSmoker: Here is idea behind - the auto reporting consists basically from one function: you post your report and get its status back. That's it. 

Do you want to search for SOAP support library for your language/IDE and include that complex code in your application just for that? 

If you could just make a single POST request (via OS functions) to report a bug and get its status back - wouldn't it be great?

So, it is not required, but would be a great option. For example, Fogbugz has this feature.

bug_update.php
GunSmoker: Currently bug up vote is called from a single place: "Me too" button. It does only one single function: up votes. There is no need to run other code. 

Also, no user is supposed to edit bug and set/edit votes via UI. Therefore, it is either editing bug, or upvoting, but not both.

config_defaults_inc.php
GunSmoker: I have no strong objection (after all, this is customizable), but from my point of view, "count" determinates priority of the bug from exception-driven development point. That is why it is grouped with priority.

config_defaults_inc.php
GunSmoker: Here is why it is ACKNOWLEDGED: one can set max report to a very large number, may be even almost unlimited. Then they start to use their app and Mantis. All will be more or less fine, but only up until they made a really bad mistake, which will appear on almost all of their clients. 

Now what? Devs will receive insane amount of reports and could do nothing to stop it - until administrator will change max reports to lower value. If you have this value to something other than resolved, then devs can simply acknowledge the bug and stop reports from being accepted.

Surely, this would be a mistake in Mantis configuration made by administrator. He should not increase limit without changing this option too. But why not offer additional safe net?

Besides, from purely logic view: if you acknowledge the bug, why would you want to receive more reports?

P.S. Just in case, if it is unclear: only files (attaches) are rejected. Count (votes) will still be added - up until issue is resolved or become read-only.

config_defaults_inc.php
GunSmoker: Yep, it is super-useful, but nothing stops reporting application from adding tags as it likes. After all, you supply a full new bug data on each report. Should we duplicate this function in Mantis too?

config_defaults_inc.php
GunSmoker: This option controls only secondary entry point in bug_report.php.

I don't know if this option is really needed (I do not see usage cases for it), I have added it just in case.

core/database_api.php
GunSmoker: As I said to changes in DB structure: I simply don't know how to make atomic operations in Mantis.

How you survived so far: you present UI to edit bug with secret unique token. If user posts an edit, but token mismatch last used token - you tell user that someone else has edited the bug, and then the process repeats.

I don't see how this logic could be used with automating reporting via API. 

One can use a similar technique to insert bug, then check if another bug (with same alias) already exists, then delete dup, update votes. But wouldn't it be a bit tooo complex? Instead you can just run a single INSERT and fail on uniqueness constraint broken, then update.

my_view_inc.php
GunSmoker: Currently, no UI use term "vote". Only internal code operates with "votes". Therefore, it just looks a bit weird to me to have something like +100. I would rather delay this until real voting will be implemented, but I'll be fine either way.

core/bug_api.php
GunSmoker: Why? What is usage case for this?

Do you think about the same bug occurs after it was fixed? If yes - then this usage case would probably use version field to detect this case. And when found - bug should be simply re-open.

GunSmoker:
@vboctor 

The biggest question seems to be DB-related changes. I would gladly change it (and some your other remarks) once we decide how it should be implemented.

core/bug_api.php
vboctor: So should the event be BUG_UP_VOTED?

admin/schema.php
vboctor: I would separate the concept of reporting a bug from voting for it.  For example, imagine a world where user has a limit of 10 votes to use across open bugs, they may have reported 20 bugs, but put their votes on a subset of these.  We may event decide in the future to enable users to put multiple points on same issue (e.g. user voice enable 1 to 3 votes on a single feature request).

As for the million requests, that is not going to happen because of the max limit (50).  Your point about a million attempts to add is true and we must make sure that doesn't become a denial of service attach on the server, and hence declining new attempts to add reports should be as cheap as possible.

One thing that may be complicating the story is attempting to combine the votes concept and the hit-count concept.  Is that the case?  Should we just focus on the hit-count and use the appropriate terminology and remove ability for a user to do it interactively?  If a user says I want this too or I have this too, then it is not automatic crash reporting anymore and we don't have any info for it, it is just a note or a vote.

admin/schema.php
vboctor: One option I can think of is to have the following db schema upgrade steps:

- Step 1: Add the field without unique - SQL
- Step 2: Populate the alias for all bugs as the issue id - PHP function
- Step 3: Add uniqueness constrain - SQL

vboctor: Actually, I was trying to avoid a join to render a issues.  If this is just used at crash reporting time, that's fine.  I assume the count for each issue will be cached on the bug table.

core/bug_api.php
vboctor: From memory, I was referring to having bug_is_resolved() in addition to looking at the resolution.  For version case, the app can use same alias and hence it won't report, or different alias and hence a different bug will be open, right?

lang/strings_english.txt
vboctor: If we just focus on hit-count rather than voting then Most Hits.

api/soap/mantisconnect.wsdl
vboctor: The scenario of crash reporting only deals with one kind of token, the one that only enables reporting.  I would rather not have a surface area for this until we have a good model to define token scopes/types that we feel comfortable exposing via the public API.

api/soap/mantisconnect.wsdl
vboctor: Not really.  Again I want to reduce surface API of the feature in the public API and also I kind of think of the alias/hash as something that doesn't need to be displayed or returned.

bug_report.php
vboctor: We can use the REST API (see #1037) which I just submitted for review.  My goal is to get this in and add methods gradually.  Your feedback is welcome there.  Using the bug_report.php is not a path I would take.

I would be totally happy to have the new API added by this PR only show up in the REST API.

bug_update.php
vboctor: I would suggest removing this functionality and stick with just automated crash reports for this PR.  Then we can add votes as a separate PR/feature.

config_defaults_inc.php
vboctor: Let's call this hit-count.

config_defaults_inc.php
```diff
@@ -2085,6 +2085,28 @@
 $g_bug_closed_status_threshold = CLOSED;
 
 /**
+ * Bug becomes closed for submission of automated bug reports (attaches)
+ * if its status is >= $g_bug_collect_reports_threshold. The bug starts 
+ * accept more bug reports again if its status becomes less than this 
+ * threshold. The default is ASSIGNED.
```
vboctor: Default is not consistent with value below.  Typically we don't mention the default in config_defaults_inc since it right under it.  But we do in the configuration section of the manual.

config_defaults_inc.php
```diff
@@ -2085,6 +2085,28 @@
 $g_bug_closed_status_threshold = CLOSED;
 
 /**
+ * Bug becomes closed for submission of automated bug reports (attaches)
+ * if its status is >= $g_bug_collect_reports_threshold. The bug starts 
+ * accept more bug reports again if its status becomes less than this 
+ * threshold. The default is ASSIGNED. 
+ * Read-only, closed and fixed bugs are closed for bug report submission 
+ * regardless of this option.
+ * See also $g_bug_max_auto_reports option.
```
vboctor: I believe there is a phpdoc for see also.

config_defaults_inc.php
vboctor: Your P.S. is not clear and should be clarified in the documentation of the config option.

You opposite of your argument is also true.  You can imagine assigning the issue for investigation, and then you no longer receiving reports, though receiving reports may be good because you gather more information about who is hitting the issue.

In addition to the max number of reports per issue.  I wonder if we should have a max # of reports by a crash reporting user per hour.  That is to protect the system against getting a ton of reports to quickly, specially if there is a bug with alias calculation (e.g. exception message is included in hash and it has a non-unique value in it for each user).

config_defaults_inc.php
vboctor: That's good enough for now, we can revisit later.

config_defaults_inc.php
vboctor: Let's remove that and remove usage of bug_report.php.

core/access_api.php
```diff
@@ -443,6 +443,39 @@ function access_has_any_project( $p_access_level, $p_user_id = null ) {
 }
 
 /**
+ * Check if the current user has already voted (returns false) or not (returns true).
+ * @param integer      $p_bug_id       Integer representing bug id to check access against.
+ * @param integer|null $p_user_id      Integer representing user id, defaults to null to use current user.
+ * @return boolean whether user can vote for the bug
+ * @access public
+ */
+function access_user_can_vote_for_bug( $p_bug_id, $p_user_id = null ) {
```
vboctor: We should remove the voting functionality.

core/api_token_api.php
```diff
@@ -117,11 +124,15 @@ function api_token_validate( $p_username, $p_token ) {
 		return false;
 	}
 
-	$t_encrypted_token = api_token_hash( $p_token );
-
+	if( API_TOKEN_GENERIC == $p_kind ) {
+		$t_encrypted_token = api_token_hash( $p_token );
+	} else {
+		$t_encrypted_token = api_token_hash( $p_kind . $p_token );
```
vboctor: Why do we need a different encryption type?  I would drop this.  Each token should be long and unique on its own.  You can imagine a token type being modified in the future.

core/database_api.php
vboctor: OK

core/filter_api.php
```diff
@@ -1563,6 +1564,12 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
 		array_push( $t_project_where_clauses, '({bug}.last_updated BETWEEN ' . db_param() . ' AND ' . db_param() . ' )' );
 	}
 
+	# alias
+	if( !is_blank( $t_filter[FILTER_PROPERTY_ALIAS] ) ) {
```
vboctor: Please undo changes to filter APIs, no need to filter by alias or display it.

core/history_api.php
```diff
@@ -503,6 +503,9 @@ function history_localize_item( $p_field_name, $p_type, $p_old_value, $p_new_val
 			$p_new_value = get_enum_element( 'priority', $p_new_value );
 			$t_field_localized = lang_get( 'priority' );
 			break;
+		case 'votes':
```
vboctor: Use hitcount for case and language script.

api/soap/mantisconnect.wsdl
GunSmoker: Yes, but crash reporting software exists already, and it supports Mantis with login/password and, probably, (full) login/API token. When crash reporting API will be added, this software has to be modified to support new possibilities. 

So, how could such software detect what type of Mantis credentials were entered by developer? Is it login/password? Is it login/full token? Is it login/restricted token?

I understand that it does not make much sense to use full token or even login/password pair to call a single crash API function. However, there are people out there who already use such software and they have their settings all set up. Not everyone will immediately make changes in their configuration (e.g. replace login/password and/or full token with restricted token) when upgrading to new Mantis and crash reporting software.

bug_report.php
GunSmoker: I'm totally fine to replace option of reporting via bug_report.php with option of reporting via REST. It is even better.

However, I would like to have crash API present in both SOAP and REST.

api/soap/mantisconnect.wsdl
```diff
@@ -546,6 +550,16 @@
...
+  <part name="password" type="xsd:string" />
+  <part name="issue" type="tns:IssueData" />
+  <part name="attach_note" type="xsd:string" />
```
GunSmoker: Why? The note is for attach, that is why it is "attach_note" - it is super-clear. "note" is less clear to me. Is it a note for report itself? For attached file? For submission process itself?

core/api_token_api.php
GunSmoker: That is to avoid modification to places which does not take new concept of token type into account. E.g. just a safe net. 

If some old code would run a token check on restricted token - it will fail, because of kind addition. Otherwise you had to be extra sure that no other code checks tokens without also checking its type.

Yes, all token-related checks are supposed to go via api_token_validate (which takes type into account), but I would be better safe than sorry.

GunSmoker:
@vboctor 

I have added most of your suggestions.

- mc_issue_report -> mc_crash_report
- Alias was removed from almost everywhere. Because of that, a new mc_crash_get was added (before it was possible to search issue with filters).
- Reporting via bug_report.php was removed.
- UI for increasing hit count was removed.
- Hit count is displayed only for bugs, hidden for feature requests.
- All history changes were removed.

No changes to DB and token kinds.

PR #1009, opened 2017-01-27
Unbundle JS libraries from Ace theme files

syncguru:
The ace-admin theme comes bundled with a lot of other js libraries that we don't use. The goal of this PR is to remove unused libraries and for the libraries we use to unbandle them so that we can upgrade to latest version:

+ Add typeahead.js as separate library
+ Organize js/css libraries used in readme.txt

Summary of changes in file sizes (slimmed by over 300k):

js/ace.min.js 56490 => 41178 
js/ace-extra.min.js  6947 => 0 (deleted)
js/ace-elements.min.js 41782 => 0 (deleted)
js/typeahead.jquery-1.1.1.min.js 0 => 31010 (new)

css/ace.min.css  397644 => 298300
css/ace-part2.min.css 129383 => 29694
css/ace-rtl.min.css 117512 => 74972
css/ace-skins.min.css 78658 => 14106

@dregad libraries readme file updated to include all css/js packages we use (following up on #1002)

Fixes #22256


dregad:
Thanks for that. To make the readme file easier to read when markdown is rendered, it would be nice if you could add level 2 titles before the tables to separate them clearly. Something like
- `## PHP libraries` 
- `## Javascript/CSS libraries`

syncguru:
@dregad done.
@vboctor we should get this in sooner to get enough test coverage over the next few weeks.

PR #988, opened 2017-01-07
Merge of #862 in master

cproensa:
This includes the merge of pull request #862: Manage and edit stored filters 
merged into master branch, with the modifications for v2 UI

PR #1011, opened 2017-01-29
Manage filter pages text and styles

cproensa:
Small adjustments to texts and styles for the manage filters pages.
This is part of the suggestions discussed with Victor after merge of #988


syncguru:
Before and after snapshots?

cproensa:
before:
seleccion_133

after:
seleccion_132

PR #1012, opened 2017-01-30
Enlarge status color boxes & remove black frame

syncguru:
This change make the status color more prominent than before & less eye soar as the black frame is gone ... see before and after images.

Before:
screenshot from 2017-01-29 18-08-38

After:
screenshot from 2017-01-29 18-13-16

Fixes #21724

### DIFF
--- a/bug_view_inc.php
+++ b/bug_view_inc.php
@@ -476,7 +476,7 @@ if( $t_show_status || $t_show_resolution ) {
 		$t_status_label = html_get_status_css_class( $t_bug->status );
 
 		echo '<td class="bug-status">';
-		echo '<i class="fa fa-square-o fa-xlg ' . $t_status_label . '"></i> ';
+		echo '<i class="fa fa-square fa-status-box ' . $t_status_label . '"></i> ';
 		echo $t_status, '</td>';
 	} else {
 		$t_spacer += 2;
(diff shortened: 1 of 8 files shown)


cproensa:
As reporter of #21724, my request is more oriented to ease the colors visibility. The background was meant as a proposal as reusing the old style in a friendly way within the new UI, but not the predermined solution.

Later the issue may have developer into specific requests for having the background color, make it configurable, etc.
I think both items van be separated, one is improving visibility, which this PR seems to accomplish. 
Other request specific for background colors, may be kept separated.
So we could create a new issue, for those specific request, and let #21724 be solved by this PR

Anyway, when i have time i will try these changes and see how it works out. I'm glad that this has been accounted. Thanks.

dregad:
@cproensa thanks for the clarification. My comment was made based on this statement in the issue's description:

> Can this be considered, at least, as an optional style?

Personally, I don't have a strong opinion on colored boxes vs cell background. I'm fine with resolving #21724 and opening a new issue to tracking this feature request separately, assuming it is still needed after merging this PR.

vboctor:
It would be interesting to see how much customization can someone do by just overriding the css styles to achieve some of the desires of different users.  For example:

- We can have a style for a status cell.
- We can have another style for a status color box.

The default can do nothing for status cell and draw the status color box.   If someone wants to make the color takes the whole cell, then they can have both styles have the same background color.

The style for status box can also be customized to have a thin black border or no border.

I would rather have this approach rather than a config option or doing something that is not in line with what we started in 2.0.0 release.

syncguru:
@dregad updated the commit message

cproensa:
Roadmap and changelog  show the sqares too close one to another

seleccion_131

syncguru:
@cproensa, @dregad, @vboctor - fixed

screen shot 2017-02-04 at 5 50 27 pm

dregad:
LGTM. Thanks for the fix

PR #1018, opened 2017-02-04
Use constants instead of hardcoded values for filter view types

dregad:
Fixes [#21619](http://www.mantisbt.org/bugs/view.php?id=21619)


dregad:
As I was testing this, I realized that there was (still) some code duplication to handle display of filter view type toggle (see [#21618](http://www.mantisbt.org/bugs/view.php?id=21618)), so this PR now includes a  fix for that.

PR #1019, opened 2017-02-04
Add tags to permalink url parse

cproensa:
Permalink url parse did not read parameter for tags.
Previously it worked as a side effect of reading gpc vars at
filter_ensure_valid, which is not done any more.

Fixes: #22302


search.php
```diff
@@ -93,6 +93,8 @@
 $t_my_filter[FILTER_PROPERTY_VIEW_STATE] = gpc_get_string_array( FILTER_PROPERTY_VIEW_STATE, $t_meta_filter_any_array );
 $t_my_filter[FILTER_PROPERTY_VERSION] = gpc_get_string_array( FILTER_PROPERTY_VERSION, $t_meta_filter_any_array );
 $t_my_filter[FILTER_PROPERTY_MATCH_TYPE] = gpc_get_int( FILTER_PROPERTY_MATCH_TYPE, FILTER_MATCH_ALL );
+$t_my_filter[FILTER_PROPERTY_TAG_STRING] = gpc_get_string( FILTER_PROPERTY_TAG_STRING, '' );
+$t_my_filter[FILTER_PROPERTY_TAG_SELECT] = gpc_get_int( FILTER_PROPERTY_TAG_SELECT, 0 );
```
vboctor: What does the 0 mean?  Is there a constant for it?

cproensa: probably it's meta value "any", these are copied from the gpc_get that were done previously.
imo, this is the minimum change to avoid the regression, without going into refactor and deeper cleaning.
i agree this is a simple change, and can be targeted to 2.1

PR #1022, opened 2017-02-08
Add EmailData Event

Fazril:
Add EmailData Event before adding it to queue
fixes #22225

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -1096,6 +1096,8 @@ function email_store( $p_recipient, $p_subject, $p_message, array $p_headers = n
 	}
 	$t_email_data->metadata['hostname'] = $t_hostname;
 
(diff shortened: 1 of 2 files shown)


Fazril:
> This approach will break plugins like Eventlog that attempt to help user diagnose why user X received or didn't receive an email.

True, maybe the solution in the ticket description is better ? http://mantisbt.org/bugs/view.php?id=22225 

>  New events should be documented in the developer manual with details about information included, what plugins can do with it, etc.

Done

>  I would like to send to plugins an associative array with all the values that are relevant, EmailData can be one of these if we continue to that. Another option is not to depend on EmailData and trigger the events from all places that construct the email.

I think we need to discuss this in http://mantisbt.org/bugs/view.php?id=22225 

> The plugin should know the type of notification it is dealing with.

Done

Fazril:
Do you need any additionnal information ?

PR #1023, opened 2017-02-08
	Add Bug title for bug "view.php"

comxd:
image

### DIFF
--- a/bug_view_inc.php
+++ b/bug_view_inc.php
@@ -237,10 +237,10 @@ echo '<div class="col-md-12 col-xs-12">';
 
 # Bug title
 if( $t_show_summary ) {
-	echo '<div class="page-header">';
+	echo '<div class="col-md-12 col-xs-12">';
 	echo '<h1>', $t_summary, '</h1>';
 	if( $t_show_resolution ) {
-		echo '<small class="bug-resolution" title="', lang_get( 'resolution' ) , '">', $t_resolution, '</small>';
+		echo ' <small class="bug-resolution" title="', lang_get( 'resolution' ), '">', $t_resolution, '</small>';
 	}
 	echo '</div>';
 }
--- a/bug_view_inc.php
+++ b/bug_view_inc.php
@@ -235,22 +235,19 @@ $t_links = event_signal( 'EVENT_MENU_ISSUE', $f_bug_id );
 
 echo '<div class="col-md-12 col-xs-12">';
 
-# Bug title
-if( $t_show_summary ) {
-	echo '<div class="col-md-12 col-xs-12">';
-	echo '<h1>', $t_summary, '</h1>';
-	if( $t_show_resolution ) {
-		echo ' <small class="bug-resolution" title="', lang_get( 'resolution' ), '">', $t_resolution, '</small>';
-	}
-	echo '</div>';
-}
-
 # Bug details
 echo '<div class="widget-box widget-color-blue2">';
 echo '<div class="widget-header widget-header-small">';
 echo '<h4 class="widget-title lighter">';
-echo '<i class="ace-icon fa fa-bars"></i>';
-echo $t_form_title;
+
+if( $t_show_summary ) {
+	# Display Bug Title
+	echo $t_summary;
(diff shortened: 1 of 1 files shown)


atrol:
I am wondering if we should display the summary where we have "View Issue Details" at the moment.
So we would not waste any vertical space.

BTW, we should also replace the icon left beside "View Issue Details" as it looks like a burger menu and you might expect some menu entries when clicking on it.

comxd:
Hi all,

Ok I fully agree with you and I have edited my proposal.

image

comxd:
<p class="lead"></p>:
image

Summary into the table header with <b> only for the summary (not ticket ID):
image

I prefer the second proposal with the summary into the header with <b>.

vboctor:
@comxd I'm OK with the latest proposal.  Not sure if we need to just have the title as bold, maybe we should have issue id + summary as bold.

@syncguru @dregad @atrol any objections on going forward with latest proposal?

atrol:
I checkout out latest code of comxd's master branch, which is what you see at https://github.com/mantisbt/mantisbt/pull/1023#issuecomment-278695588

I am not sure if bold is needed.
Don't look at the resized GitHub preview, at least click on it to get the original resolution, or even better try the code in your test installation. My rendering using latest Firefox on Windows is even a bit better readable  than the screenshot with the original resolution.

comxd:
Hi all,

The title is very important, and title with no bold is not sufficiently visible in my opinion.

We can reduce the amont of bold? like 300/400?

syncguru:
-1 

- I am on-board with the need to have a issue title be prominent at the top of the details page.
- With that said, I don't like the designs presented thus far. I am strongly object to using the table header for title. Not to mention how bad the experience is on small screens.

I'd like to see us come closer to how Jira represent issue title as shown in this example: https://jira.atlassian.com/browse/TRANS-2343?jql=

syncguru:
I took a stab at it. Feedback welcome:

screen shot 2017-03-17 at 1 12 51 pm

screen shot 2017-03-17 at 1 22 18 pm

syncguru:
@comxd what do you think of the design above?

syncguru:
Created issue to track this: https://www.mantisbt.org/bugs/view.php?id=22539

dregad:
While I like the addition of the color square reflecting the issue's status, @comxd's proposal had 2 things which were IMHO better to make the title stand out: 
1. stronger contrast for the title bar's background color
2. bold text for issue summary

syncguru:
@dregad 

1- Try to click on the image to see it at full scale. The title should appear much more prominent than having it inside a think dark blue strip.

2- The title is bolded in small screen version (the one shown was not bolded).

dregad:
> Try to click on the image to see it at full scale

I did. And I don't like the proposal's light gray background is very close to that of the fields names, and it makes the bug data section looks different than the other sections on the page.

The bold text is a detail, I can live with or without it.

syncguru:
The usage of blue header has always been for static titles. Adding a dynamic title that could potentially be multi-line breaks the consistency. Beside it is quite unusable on small screen.

comxd:
Hi all,
I like the last version of @syncguru but with the title in bold like my previous proposal?

dregad:
@syncguru 

> The usage of blue header has always been for static titles

It has always been for *titles* (period). Whether the title itself is static or dynamic is irrelevant IMO.

> Adding a dynamic title that could potentially be multi-line breaks the consistency. 

I see more of a consistency break by NOT having a blue title on top of that section.

> Beside it is quite unusable on small screen.

I don't see in what way it would be less usable than your proposal. If there's a long title, then we'll have a taller section on a narrow screen. That is independent from the colors.

If what you mean is that we shouldn't be using the standard "title" CSS used for other sections, I'm fine with that. I see no problem with using a specialized class for this purpose. But I maintain that we should use the same colors and layout as other sections titles.

vboctor:

comxd:
Hi all, currently we use a Tampermonkey script, we share rendering result with you:

image

image

PR #1024, opened 2017-02-08
Expand image attachments by default

dregad:
Fixes [#21796](https://www.mantisbt.org/bugs/view.php?id=21796)


core/print_api.php
```diff
@@ -1894,16 +1894,19 @@ function print_bug_attachments_list( $p_bug_id, $p_security_token ) {
...
+ * @param array $p_attachment An attachment array from within the array returned
+ *                            by the file_get_visible_attachments() function.
  * @param string $p_security_token The security token to use for deleting attachments.
  * @return void
  */
 function print_bug_attachment( array $p_attachment, $p_security_token ) {
-	$t_show_attachment_preview = $p_attachment['preview'] && $p_attachment['exists'] && ( $p_attachment['type'] == 'text' || $p_attachment['type'] == 'image' );
+	$t_show_attachment_preview = $p_attachment['preview']
+		&& $p_attachment['exists']
+		&& ( $p_attachment['type'] == 'text' || $p_attachment['type'] == 'image' );
```
vboctor: The check for exists, can download, size, text or image are all done as part of setting the 'preview' flag.

core/print_api.php
```diff
@@ -1894,16 +1894,19 @@ function print_bug_attachments_list( $p_bug_id, $p_security_token ) {
...
 	if( $t_show_attachment_preview ) {
 		$t_collapse_id = 'attachment_preview_' . $p_attachment['id'];
 		global $g_collapse_cache_token;
-		$g_collapse_cache_token[$t_collapse_id] = false;
+		$g_collapse_cache_token[$t_collapse_id] = $p_attachment['type'] == 'image';
```
vboctor: I wonder if we should set to true here, since the idea of 'preview' flag is to determine whether we should auto-preview or not.  If we want to change the logic, then we should change it in core/file_api.php file_get_visible_attachments().

core/print_api.php
dregad: Thanks, I did not realize that; just reformatted existing code. Will change

core/print_api.php
dregad: > the idea of 'preview' flag is to determine whether we should auto-preview 

Actually my understanding of the preview flag (I guess, based on how it's been implemented until now) was that it indicates the *ability* to view the attachment inline, as opposed to automatically displaying it.

While I like to view images directly, I would prefer not to expand text files inline by default, as it can unnecessarily eat up a lot of vertical space. I don't think this specific handling can be achieved if we move the logic to file API.

dregad:
@vboctor this commit addresses your first comment. 

I'm not sure about the 2nd though, see [my note](https://github.com/mantisbt/mantisbt/pull/1024#discussion_r102018599). Do you actually want to display all preview-able attachments by default ?

vboctor:
Looks good @dregad - I have merged your change.

PR #1026, opened 2017-02-09
Add HTML5 required attribute to form fields that must be filled out for submission

sagamusix:
Implements https://www.mantisbt.org/bugs/view.php?id=22310


core/cfdefs/cfdef_standard.php
```diff
@@ -392,9 +397,10 @@ function cfdef_input_textarea( array $p_field_def, $p_custom_field_value ) {
  *
  * @param string $p_field_def          The custom field definition.
  * @param string $p_custom_field_value The custom field value to print.
+ * @param string $p_required          The "required" attribute to add to the field
```
dregad: missing a space to align

core/cfdefs/cfdef_standard.php
```diff
@@ -392,9 +397,10 @@ function cfdef_input_textarea( array $p_field_def, $p_custom_field_value ) {
...
  * @return void
  */
-function cfdef_input_date( $p_field_def, $p_custom_field_value ) {
+function cfdef_input_date( $p_field_def, $p_custom_field_value, $p_required = '' ) {
```
dregad: Parameter is not used, shouldn't print_date_selection_set() be adapted to accept it ?
Considering the ongoing work on refactoring date fields, maybe this should be deferred, in that case adding a TODO here and in print_date_selection_set() as a reminder to eventually update this would be useful.

core/custom_field_api.php
```diff
@@ -1392,7 +1393,7 @@ function print_custom_field_input( array $p_field_def, $p_bug_id = null ) {
 
 	global $g_custom_field_type_definition;
 	if( isset( $g_custom_field_type_definition[$p_field_def['type']]['#function_print_input'] ) ) {
-		call_user_func( $g_custom_field_type_definition[$p_field_def['type']]['#function_print_input'], $p_field_def, $t_custom_field_value );
+		call_user_func( $g_custom_field_type_definition[$p_field_def['type']]['#function_print_input'], $p_field_def, $t_custom_field_value, $p_required ? ' required ' : '' );
```
dregad: Break long line

dregad:
Thanks @sagamusix for your contribution. I tested this, and I think it would be a useful addition.

The one part I'm not sure about (did not analyze in depth), and maybe requires further work, is how things are handled in bug_change_status_page.php. E.g. a field required on update but not on resolved, could be set to blank when resolving the issue. Although I guess you could argue that this is an error in the definition of the field, and the admin should ensure *required on resolve* is checked in this case.

sagamusix:
I addressed the whitepsace and long line issues, and also added the required attribute to print_date_selection_set.

vboctor:
I haven't reviewed, but have a high level comment.  Should we use `required="required"` vs. `required`?  I think for cases like a checked checkbox we were using `checked="checked"`.

sagamusix:
@vboctor `checked="checked"` and `required="required"` are both XHTML. Since MantisBT uses HTML5, the short form `checked` and `required` should be preferred.

syncguru:
@sagamusix You are right. We don't need it. I think being consistent maybe more desired here. I'd prefer to have HTML5 version everywhere though.

dregad:
@sagamusix I just realized that the branch has conflicts, so if you could kindly resolve and rebase, it would be appreciated.

core/cfdefs/cfdef_standard.php
```diff
@@ -280,9 +280,10 @@ function cfdef_prepare_date_value( $p_value ) {
  * print_custom_field_input
  * @param array $p_field_def          Custom field definition.
  * @param mixed $p_custom_field_value Custom field value.
+ * @param string $p_required          The "required" attribute to add to the field
  * @return void
  */
-function cfdef_input_list( array $p_field_def, $p_custom_field_value ) {
+function cfdef_input_list( array $p_field_def, $p_custom_field_value, $p_required = '' ) {
```
cproensa: Should the parameter $p_required be a true/false flag?, like for example, `print_custom_field_input`

bug_change_status_page.php
```diff
@@ -291,7 +291,7 @@
 		<td>
 <?php
 			if( $t_has_write_access ) {
-				print_custom_field_input( $t_def, $f_bug_id );
+				print_custom_field_input( $t_def, $f_bug_id, $t_require );
```
cproensa: $t_require = $t_def['require_' . $t_custom_status_label];
This would usually contain a string "0" or "1"
Should we cast here to a bool?
Or should the definition array contain proper booleans already?

core/cfdefs/cfdef_standard.php
sagamusix: I found it easier to directly pass the HTML attribute rather than a boolean flag, all in all that seemed less error-prone and the code is structured more clearly.

sagamusix:
Conflicts have been resolved.

sagamusix:
@dregad Can it be merged now, before new conflicts creep in? :)

dregad:
Thanks for the reminder... I'm under the water these days, sorry.
Merged in f4137abe0c6e28ffd5a82fbe455b40d598109a25

PR #1028, opened 2017-02-10
Improve filter box user experience

syncguru:
+ Reduce resizing when expanding input fields
+ Clearly indicate disabled vs. enabled date selection fields
+ Use modern UI style checkboxes and radio buttons

Fixes #22359

Before:
screenshot from 2017-02-10 11-49-05

After:
screenshot from 2017-02-10 11-57-52


syncguru:
@dregad I would not go back to 2.1. This should just be in the next release as incremental improvement. I think we will have many of those.

vboctor:
Merged via ff0f6bb82fb7badbbed6c0cc553d235447f0972c and f782e9d9009e4641db56320b085098c04ce36f8e.

PR #1030, opened 2017-02-12
Sanitize window title

vboctor:
The window title is not sanitized.  That is not an issue when CSP is enable (default),
but if disabled, it can execute javascript that is set by a user who has access
to set configuration via Manage - Manage Configuration - Configuration Report page.

Fixes #22266


atrol:
+1

PR #1031, opened 2017-02-13
Fix time tracking without notes

vboctor:
Fixes #22326


atrol:
-1

After this change you are no longer able to add an attachment without a note if $g_time_tracking_without_note == ON,

vboctor:
Good catch @atrol - this should be fixed now.

PR #1032, opened 2017-02-15
Use date picker widget in place of date selection dropdowns

syncguru:
Fixes #8957

With this PR, built-in fields (i.e. date_submitted, last_updated) & custom fields of type date will use datetime widget. Usage by time tracking feature was updated as well.

screenshot from 2017-02-14 20-48-47


js/common.js
```diff
@@ -482,6 +456,28 @@ function toggleDisplay(idTag)
 	setDisplay( idTag, (document.getElementById(idTag).style.display == 'none')?1:0 );
 }
 
+// Datetime picker handler
+function enableDatTimePicker( p_element ) {
```
vboctor: typo in function name.

config_defaults_inc.php
```diff
@@ -1179,6 +1179,13 @@
  */
 $g_datetime_picker_format = 'Y-MM-DD HH:mm';
 
+/**
+ * Datetime picker widget format string for date only.
+ * For detailed instructions on date formatting
+ * @see http://momentjs.com/docs/#/displaying/format/
+ * @global string $g_datetime_picker_format
+ */
+$g_date_picker_format = 'Y-MM-DD';
```
vboctor: Is it possible to use short date format or a transformation of it rather than adding this?  If we are going to add this, it should be added to the manual.

atrol:
Did some cursory tests. Looks good.
No time to have a deeper look at it ...

syncguru:
@vboctor 
1- Rebased PR
2- Removed the new config option and used a simple conversion function to translate php date format to datetime picker date format (as described here http://stackoverflow.com/questions/30186611/php-dateformat-to-moment-js-format)

vboctor:
@synguru A a couple of comments as I did some quick testing.

- Date time text boxes in filter box and time tracking (on issue page and on time track page) are too narrow for the date time text.  We need to make the text box wider.
- Rename `convert_php_to_date_format()` to `convert_date_format_to_momentjs()`.

syncguru:
@vboctor Fixed the date field width & the helper function name

screen shot 2017-03-16 at 3 48 13 pm

For the smart defaults, I opened an issue to track that since the fix is not very obvious: http://mantisbt.org/bugs/view.php?id=22531

parkvillegirl:
Is there an installation package for this code?

PR #1033, opened 2017-02-19
Markdown: Fix indentation for blockList markdown

jllano:
Fixes #22344

Before:
screen shot 2017-02-20 at 1 46 57 am

After:
screen shot 2017-02-20 at 1 46 26 am

### DIFF
--- a/plugins/MantisCoreFormatting/core/MantisMarkdown.php
+++ b/plugins/MantisCoreFormatting/core/MantisMarkdown.php
@@ -233,4 +233,29 @@ class MantisMarkdown extends Parsedown
 		}
 		return static::$mantis_markdown;
 	}
+
+	/**
+	 * Fix line indent
+	 * Make sure to indent at least one space
+	 * @param string $line The Markdown syntax to parse
+	 * @return string html representation generated from markdown.
+	 */
+	function blockList($line){
+		$line['text'] = $line['text'] . " ";
+		return parent::blockList( $line );
+	}
+
+	/**
+	 * Customize the blockListContinue method
+	 *
+	 * @param string $line The Markdown syntax to parse
+	 * @param array $block A block-level element
+	 * @return void
+	 */
+	function blockListContinue( $line, array $block ){
+		
+		if( ! isset($block['interrupted']) ){
+			return;
+		}
+	}
 }


plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -233,4 +233,29 @@ private static function init() {
...
+	/**
+	 * Customize the blockListContinue method
+	 *
+	 * @param string $line The Markdown syntax to parse
+	 * @param array $block A block-level element
+	 * @return void
+	 */
+	function blockListContinue( $line, array $block ){
+		
+		if( ! isset($block['interrupted']) ){
```
vboctor: Incorrect formatting.  Here is the correct one:

```
if( !isset( $block['interrupted'] ) ) {
```

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -233,4 +233,29 @@ private static function init() {
...
 	}
+
+	/**
+	 * Fix line indent
+	 * Make sure to indent at least one space
+	 * @param string $line The Markdown syntax to parse
+	 * @return string html representation generated from markdown.
+	 */
+	function blockList($line){
+		$line['text'] = $line['text'] . " ";
```
vboctor: Use single quotes unless double quotes are needed.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -233,4 +233,29 @@ private static function init() {
 		}
 		return static::$mantis_markdown;
 	}
+
+	/**
+	 * Fix line indent
```
vboctor: It would be useful to focus on why the code is doing what it is doing, rather than what it is doing which is readable via the code.  For example, "Add a space at the end of the text in order to fix blah blah blah" and possibly reference the bug that this workaround is fixing.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -233,4 +233,29 @@ private static function init() {
...
+
+	/**
+	 * Customize the blockListContinue method
+	 *
+	 * @param string $line The Markdown syntax to parse
+	 * @param array $block A block-level element
+	 * @return void
+	 */
+	function blockListContinue( $line, array $block ){
+
```
vboctor: Remove this extra blank line.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
jllano: It seems a bug from the source itself. http://parsedown.org/demo. But let me check in the other extentions. https://github.com/erusev/parsedown/wiki/Extensions-and-Related-Libraries

dregad:
Please edit commit message, it's spelt *indentation*

jllano:
Unfortunately, the changes I made is not the fix and this will make the Ordered List (without backtick/code block) to not parse correctly.

But here is the issue I submitted to them https://github.com/erusev/parsedown/issues/477

jllano:
Closing this since this is a bug on library itself. We already opened an issue for this in upstream,

PR #1034, opened 2017-02-19
Fix Recently Modified filter to include closed issues

cproensa:
Fix "Hide status" property for the filter shown in my-view-page for
"Recently Modified" box. It should display closed issues.

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -2981,7 +2981,7 @@ function filter_name_valid_length( $p_name ) {
 }
 
 /**
- * Create a filter for getting issues modified in the last N days
+ * Create a filter for getting issues modified in the last N days.
  * @param integer $p_days Number of days counting from today
  * @param array $p_filter Add the filter conditions over this filter array. Return a new one if null
  * @return array Filter array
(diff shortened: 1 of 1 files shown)


my_view_inc.php
```diff
@@ -75,7 +75,9 @@
 
 # @TODO cproensa: make this value configurable
 $t_recent_days = 30;
-$c_filter['recent_mod'] = filter_create_recently_modified( $t_recent_days );
+$t_recent_mod_filter = filter_create_recently_modified( $t_recent_days );
+$t_recent_mod_filter[FILTER_PROPERTY_HIDE_STATUS] = META_FILTER_NONE;
```
vboctor: Why wouldn't this be part of filter_create_recently_modified().  This is the only place that we use this API and should produce the proper filter that can be re-used elsewhere overtime, like in the SOAP API or other places.

vboctor:
@cproensa please also change the PR to target `master-2.1` since it is a regression and a simple fix.

my_view_inc.php
cproensa: filter_create_recently_modified(), when used to create a new filter, sets the date filter to the needed days values, and keeps the rest of properties as given by a new default filter.
Hide status is then returned as the default value as per configuration "hide_status_default".

We seem to support that this filter box must show all status, regardless of "hide_status_default" configuration, so this property makes sense to be modified in my-view-page for this specific usage.

cproensa:
Rewritten with victor suggestion

vboctor:
@cproensa please update the comment with reference to the issue in MantisBT bug tracker.

vboctor:
@cproensa also please target `master-2.1`.

cproensa:
>  update the comment with reference
> also please target master-2.1.

all sorted out now

PR #1035, opened 2017-02-22
Only record effective custom fields changes in history

dregad:
When updating an issue, an unmodified custom field was logged as changed
in history.

Add check comparing old vs new value both in bug_update.php and custom
field API.

Fixes [#22408](http://www.mantisbt.org/bugs/view.php?id=22408)


atrol:
Not sure if this is the right approach, see http://www.mantisbt.org/bugs/view.php?id=22408#c55744

bug_update.php
```diff
@@ -307,6 +307,11 @@
 	$t_new_custom_field_value = gpc_get_custom_field( 'custom_field_' . $t_cf_id, $t_cf_def['type'], '' );
 	$t_old_custom_field_value = custom_field_get_value( $t_cf_id, $f_bug_id );
 
+	# No need to update the field if its value has not changed
+	if( $t_old_custom_field_value == $t_new_custom_field_value ) {
```
syncguru: What happens when old value is say 1 and new value is '1'?

atrol: Maybe you want to say: What happens when old value is say `1` and new value is `1.0`?
The output of  `var_dump ( '1' == '1.0' );` is `bool(true)`, so we need `===`

dregad: > so we need ===

Currently the logic does not use strict comparison (see https://github.com/mantisbt/mantisbt/blob/release-2.2.0/core/history_api.php#L78). 

I'm not saying this is correct, but if we change it here, I think we need to be consistent and update the history API as well.

dregad: I'm having second thoughts about updating the history API... this function is used all over the place and I'm concerned about the risk of regression. Since the loose comparison has not caused any issues that I'm aware of until today, I think we should apply the "if it ain't broke, don't fix it" principle.

dregad:
These new 2 commits address 

- the inconsistency when passing old/new custom field values to history_log_event_direct() in custom_field_set_value()
- @syncguru and @atrol's comment about comparing `'1' == '1.0'`; the code now uses strict comparison (see new issue [#22464](https://www.mantisbt.org/bugs/view.php?id=22464))

atrol:
Does not work

- assign a custom field of type TEXT to a project
- set custom field value of an issue to `1`
- change custom field value to `1.0`

Result: No entry in history for field value change to `1.0`

bug_update.php
cproensa: Then add a parameter to history_log_event_direct() that forces the strict comparison?

PR #56, opened 2012-08-09
REST API prototype - available methods: mc_version, mc_issue_add

bieli:
Instead of old NuSoap implementation of mantisbt API, my proposition it is REST/JSON API

Usage examples:
curl http://HOST/api/rest/mc_version
{"data":"1.2.11"}

curl -X POST http://HOST/api/rest/mc_issue_add -d "summary=test issue" -d "description=Description of test issue" -d "project=1" -d "category=cat1" -d "username=restuser" -d "password=pass"
{"data":"6"}

Solution is based on Slim PHP microframework (https://github.com/codeguy/Slim).
It is nice and fast future for connection mantisbt with older languages like Perl or Erlang :-) when WSDL  in NuSoap not working fine sometimes.

PR #1037, opened 2017-02-25
Modern RESTful API

vboctor:
This is a proof of concept for Modern RESTful API.  It would be good to get early feedback before going too deep with it.

Some of the design aspects:

- **RESTful** API with **JSON** payloads for requests and responses.
- **Authentication** no support for username and password.  API tokens are the authentication mechanism.
- **Enable use of API from web UI** As we move more towards javascript calling into backend enable JS to call into REST API via cookie based auth.
- **Enable plugins to publish their own APIs/routes** - It should be easy for plugins to extend the REST API with their specific APIs.
- **Support for [Swagger](http://swagger.io) and Swagger UI** (see [example](http://petstore.swagger.io)) - use of swagger provides API meta data, ability to generate client code in all languages, and a web UI to play with the API.
- **Time for Breaking changes** - Since this is a new API, we can do breaking changes like removing deprecated functionality, fixing in-consistencies (e.g. sometimes returning a user object and sometimes a user id), dropping username/password support, etc.
- **Re-use of code between SOAP and REST** - avoid duplicate logic.  Eventually deprecate SOAP.
- **Based on proven framework (Slim Framework)** - get a bunch of powerful functionality and a proven framework.
- **Use Composer** - It's about time we start transitioning to Composer for new dependencies.

The focus of this PR is to agree on the approach and provide some basic methods.  It is not a goal to implement every single SOAP API functionality.  This can be done over time.


atrol:
No time to have a deeper look at the moment, I just would like to be sure that you are aware of https://github.com/mantisbt/mantisbt/pull/56

vboctor:
@atrol I looked at #56, thanks for pointing it out.  It was also utilizing slim framework, but this PR is more comprehensive in aspects like:

- Authentication model based on API tokens
- Support for Swagger and Swagger UI
- RESTful format for the API
- Re-use of code between SOAP and REST.
- Using Slim Framework via Composer.
- Using middleware for features like Auth and Version.

As of now, the PR has a good sample of how the API will look like.

api/rest/restcore/AuthMiddleware.php
```diff
@@ -0,0 +1,31 @@
+<?php
+require_api( 'authentication_api.php' );
+
+class AuthMiddleware {
+	public function __invoke( \Slim\Http\Request $request, \Slim\Http\Response $response, callable $next ) {
+		if( mci_is_mantis_offline() ) {
+			return $response->withStatus( 500, 'Mantis Offline' );
```
rombert: I think 503 Service Unavailable is more appropriate.

api/soap/mc_api.php
```diff
@@ -43,6 +43,11 @@
  */
 class SoapObjectsFactory {
 	/**
+	 * @var bool true: SOAP API, false: REST API
+	 */
+	static public $soap = true;
```
rombert: I don't think this belong where. We should either unwrap the SoapVar in the REST API, as long as it's a transitional measure, or move this code someplace else.

vboctor: I've renamed this class and some of its methods so it is not Soap specific, but ApiObjectsFactory.  I kept methods that are providing added value or may behave differently based on API type.  More APIs may be added to this later to handle distinction between REST and SOAP.

api/rest/restcore/AuthMiddleware.php
vboctor: Fixed.

vboctor:
I've addressed comments from @rombert + some more improvements

- REST API is an experimental feature that is disabled by default.
- Support anonymous access.
- Some cleanup to SoapObjectFactory and renames to not make it SOAP specific.
- Added headers + php docs

syncguru:
Not sure if you've seen this: https://github.com/zircote/swagger-php

api/soap/mc_api.php
```diff
@@ -36,19 +36,21 @@
...
+
 	/**
 	 * Generate a new Soap Fault
 	 * @param string $p_fault_code   SOAP fault code.
 	 * @param string $p_fault_string SOAP fault description.
 	 * @return SoapFault
 	 */
-	static function newSoapFault( $p_fault_code, $p_fault_string ) {
+	static function fault($p_fault_code, $p_fault_string ) {
 		return new SoapFault( $p_fault_code, $p_fault_string );
```
syncguru: I expected to see RestFault type as well, no?

vboctor: Will likely do this at one point.

vboctor:
@syncguru I would rather use standard swagger format and leverage tooling (e.g. editor.swagger.io) and documentation for such format, rather than use some other language in php comments that gets translated to swagger and require some build step.

rombert:
@vboctor  - have you given testing any thought? I am not sure whether we can reuse the SOAP tests in the same way we reuse the implementation, but having solid testing on the REST API would be very good IMO.

vboctor:
I've added support for plugins to register their own REST routes / APIs.  Also added a sample API with MantisGraph plugin.

vboctor:
Added support for cookie based auth.  This enables javascript code on the web UI to call into the REST API.

vboctor:
Added the following functionality:

- /users/me to get information about the user and accessible projects.  More information can be added here as necessary.  The goal is to provide info that would be typically needed by a client for the logged in user.
- /config to get information about a set of provided config options within a user id and project id scope.  This supports getting a batch and configs of type arrays, both are not supported by SOAP API.  It also provides special handling for enum config options.
- /lang to get information about a set of strings within the scope of a user id.

TODO: update swagger definition.

vboctor:
Few more updates

- Disabling cache via CacheMiddleware
- Updated swagger.json
- Updated copyright header

vboctor:
I've noticed that all configuration options of type array are marked as non-public.  This is likely because SOAP API didn't support getting configs of type arrays.  I suggest marking them as public as appropriate.  That will result in an error if a client attempts to retrieve such configs via SOAP API, but will work fine when retrieving them via the new REST API.  The REST APIs also offers getting configs in batch and in context of a user and a project.

vboctor:
The following improvements were added:

- Use RestFault instead of SoapFault for REST API
- Depending on the error, use the appropriate http status code.
- Added ability to delete issues.
- Improve json info returned for related issues.

dregad:
@vboctor I think it's truly great that Mantis will offer REST API, and I strongly support the idea.

That said, I regretfully do not have enough free time for a proper, in-depth review, actually test the API or to effectively contribute to the effort, there is just too much on my plate these days. Being aware of the benefits a modern webservice interface, and the fact that it's an often requested feature, I would not want to delay its implementation without good reason.

As an overall comment, I trust you have fully and globally analyzed the interface endpoints, as this is a promise to the webservice's consumers that can't easily (shouldn't) be changed down the line. I think it's very important that we have an overall consistency in the API, so we it would be useful to have a vision/roadmap to avoid adding mismatched bits and pieces as we go along. This should be documented somewhere (wiki ?)

More detailed review follows.

composer.json
```diff
@@ -0,0 +1,15 @@
...
+    "description": "Mantis Bug Tracker",
+    "type": "project",
+    "require": {
+        "slim/slim": "^3.0"
+    },
+    "license": "GPL v2",
+    "authors": [
+        {
+            "name": "Victor Boctor",
+            "email": "vboctor@gmail.com"
```
dregad: No offense, but you are not the only author of this project. IMO, this should be the same as we have in the headers, i.e. 

    "name": "MantisBT Team",
    "email": "mantisbt-dev@lists.sourceforge.net"

composer.json
```diff
@@ -0,0 +1,15 @@
...
+        }
+    ]
```
dregad: Depending on what the future plans for use of Composer are, additional information may be useful (e.g. keywords, homepage, support, PHP version and extensions requirements), etc.

.gitmodules
```diff
@@ -18,3 +18,6 @@
 	path = library/parsedown
 	url = https://github.com/mantisbt/parsedown.git
 	branch = mantisbt
+[submodule "api/rest/swagger"]
```
vboctor: There is two components to swagger:

- **Swagger.json** - It is the meta-data describing the APIs exposed by the RESTful service (equivalent of WSDL for SOAP).  This file is used to power Swagger UI or code generators that generate clients in multiple languages based on Swagger meta-data.
- **Swagger UI** - This is a JS web app that reads such swagger.json and provides the page that describes all APIs including documentation, authentication support, etc.  API consumers can use this UI to understand the API as well as try it out.  Think of it like a SoapUI or one of these tools.  Bundling this enables that for any MantisBT instance.

composer.json
vboctor: This was probably picked up from some machine configs.  I don't recall setting it.  Anyways, I'm already using these same MantisBT team details in headers for all files.

composer.json
vboctor: We can add more information as we consumer.  Composer has a bunch of schema to accommodate for different usages.  For example, customizing auto-load map, fetching published packages or packages from repositories, etc.  I don't want to pack stuff that is not needed for this specific task.  I've added a link to our website in the author information though.

api/rest/index.php
```diff
@@ -0,0 +1,57 @@
...
+$app->add( new VersionMiddleware() );
+
+event_signal( 'EVENT_REST_API_ROUTES', array( array( 'app' => $app ) ) );
+
+require_once( __DIR__ . '/restcore/config_rest.php' );
+require_once( __DIR__ . '/restcore/issues_rest.php' );
+require_once( __DIR__ . '/restcore/lang_rest.php' );
+require_once( __DIR__ . '/restcore/users_rest.php' );
+
+$app->run();
```
vboctor: Added variable for rest core path.

api/rest/index.php
```diff
@@ -0,0 +1,57 @@
...
+ * @package MantisBT
+ * @copyright Copyright 2004  Victor Boctor - vboctor@users.sourceforge.net
+ * @copyright Copyright 2005  MantisBT Team - mantisbt-dev@lists.sourceforge.net
+ * @link http://www.mantisbt.org
+ */
+
+# Bypass default Mantis headers
+$g_bypass_headers = true;
+
+require_once( __DIR__ . '/../../vendor/autoload.php' );
```
vboctor: We can do that once we use composer outside the scope of the REST API.

api/rest/index.php
```diff
@@ -0,0 +1,57 @@
...
+require_once( __DIR__ . '/../../core.php' );
+require_once( __DIR__ . '/../soap/mc_core.php' );
+require_once( __DIR__ . '/restcore/ApiEnabledMiddleware.php' );
+require_once( __DIR__ . '/restcore/AuthMiddleware.php' );
+require_once( __DIR__ . '/restcore/CacheMiddleware.php' );
+require_once( __DIR__ . '/restcore/VersionMiddleware.php' );
```
vboctor: These now use the rest core variable path.  Will explore autoloader as a follow up as part of broader composer usage beyond REST API.

api/rest/index.php
```diff
@@ -0,0 +1,57 @@
...
+
+# Hint to re-used mantisconnect code that it is being executed from REST rather than SOAP.
+# For example, this will disable logic like encoding dates with XSD meta-data.
+ApiObjectFactory::$soap = false;
+
+$app = new \Slim\App();
```
vboctor: Renamed $app to $g_app.

docbook/Admin_Guide/en-US/config/api.xml
```diff
@@ -2,11 +2,20 @@
...
-	<para>A WSDL file which describes the web service is available at <literal>/api/soap/mantisconnect.php?wsdl</literal> below your installation root.</para>
+	<para>MantisBT exposes a webservice API which allows remote clients to interact with MantisBT and perform many of the usual tasks, such as reporting issues, running filtered searches and retrieving attachments.</para>
+	<para>
+		The SOAP API is enabled by default and available at <literal>/api/soap/mantisconnect.php</literal>
+		below the MantisBT root.  A WSDL file which describes the web service is available at
+		<literal>/api/soap/mantisconnect.php?wsdl</literal> below the MantisBT root.
+	</para>
+	<para>
+		The REST API is experimental and is disabled by default.
+		A Swagger sandbox and documentation for REST API is available at <literal>/api/rest/swagger/</literal> below
```
vboctor: - `/api/rest/swagger` points to Swagger UI for API documentation and sandbox for making calls to the API from the browser, which helps with understanding the API.
- '/api/rest' is the root end point for the REST API.  The route group (e.g. issues) is added after that.

api/soap/mc_config_api.php
```diff
@@ -33,19 +33,19 @@
...
 	}
 
 	if( !mci_has_readonly_access( $t_user_id ) ) {
-		return mci_soap_fault_access_denied( $t_user_id );
+		return mci_fault_access_denied( $t_user_id );
 	}
 
 	if( config_is_private( $p_config_var ) ) {
-		return SoapObjectsFactory::newSoapFault( 'Client', 'Access to \'' . $p_config_var . '\' is denied' );
+		return ApiObjectFactory::fault( 'Client', 'Access to \'' . $p_config_var . '\' is denied', 403 );
```
vboctor: Good point.  I have done the following:
- Added constants for http status codes in core/constants_inc.php
- Create factory methods for each status code.  Hence raising an error now looks as follows:

```
return ApiObjectFactory::faultForbidden( 'some reason why' );
```

api/rest/restcore/ApiEnabledMiddleware.php
```diff
@@ -0,0 +1,36 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * A webservice interface to Mantis Bug Tracker
+ *
+ * @package MantisBT
```
vboctor: Existing SOAP files don't have a sub-package defined.  I don't think it is a big deal, but if we add it we should probably use `webservice` or `webapi` since there is a lot of shared code between SOAP and REST.

vboctor:
Answering @dregad 's questions about `composer`:

> Are you planning to publish MantisBT on packagist.org ?

No.  I think of it more like an app than a package that others would want to re-use via `composer`.

> Replacing submodules by composer dependencies ?

Probably, specially for components that are published as `composer` packages.

> Documenting (dev guide) the need to run composer install,

Will do.

> The release process in the wiki needs an update as well

I'm planning to do something there, but I'm hoping I would update the packaging script rather than adding an extra step.  Basically something along the lines of `if you find composer.json then run composer install`.  This way the packaging works with both old and new releases.

I would also like over time to have the packaging script handle high level steps, and rely on a scripts in the source tree for MantisBT to do some of the micro steps.  That has two benefits: 1. such steps are likely to overlap with MantisBT build, 2. the micro-steps can be different per branch, but the package process will still work.

docbook/Admin_Guide/en-US/config/api.xml
```diff
@@ -35,11 +44,17 @@
...
 	</variablelist>
 
-	<section id="admin.config.soap.disable">
-		<title>Disabling the SOAP API</title>
-		<para>If you wish to temporarily disable the SOAP API it is sufficient to set the specific access thresholds to NOBODY:</para>
+	<section id="admin.config.api.disable">
+		<title>Disabling the webservice API</title>
+		<para>If you wish to temporarily disable the webservice API it is sufficient to set the specific access thresholds to NOBODY:</para>
 		<para><literal>$g_webservice_readonly_access_level_threshold = $g_webservice_readwrite_access_level_threshold = $g_webservice_admin_access_level_threshold = NOBODY;</literal></para>
 		<para>While the SOAP API will still be accessible, it will not allow users to retrieve or modify data.</para>
```
vboctor: > Does that effectively disable both APIs or just SOAP ?

At the moment, this applies to both REST/SOAP calls.

> Do we need / want distinct access levels for REST and SOAP ?

If we keep the concept, then we should keep it consistent across REST and SOAP.  But I'm considering removing this concept in favor of an ON/OFF for API and per feature authorization rules.  Note that with the REST API, I added a config option for ON/OFF which applies only to REST API.  This is useful to keep REST API disabled while it is in preview/beta.  It may be useful in the future, if a user wants to disable MantisBT API, but it will be switched from OFF to ON as the API gets out of preview/beta.

docbook/Admin_Guide/en-US/config/api.xml
```diff
@@ -2,11 +2,20 @@
...
  <!ENTITY % BOOK_ENTITIES SYSTEM "Admin_Guide.ent">
 ]>
-<section id="admin.config.soap">
-	<title>SOAP API</title>
+<section id="admin.config.api">
+	<title>API</title>
 
-	<para>MantisBT exposes a SOAP API which allows remote clients to interact with MantisBT and perform many of the usual tasks, such as reporting issues, running filtered searches and retrieving attachments. The SOAP API is enabled by default and available at <literal>/api/soap/mantisconnect.php</literal> below your installation root.</para>
-	<para>A WSDL file which describes the web service is available at <literal>/api/soap/mantisconnect.php?wsdl</literal> below your installation root.</para>
+	<para>MantisBT exposes a webservice API which allows remote clients to interact with MantisBT and perform many of the usual tasks, such as reporting issues, running filtered searches and retrieving attachments.</para>
```
vboctor: Will skip this until we finalize the knobs that control exposure for REST API.

api/rest/.htaccess
```diff
@@ -0,0 +1,4 @@
+RewriteEngine On
+RewriteCond %{REQUEST_FILENAME} !-f
+RewriteRule ^ index.php [QSA,L]
```
vboctor: Done.  Added a comment that references http://docs.slimframework.com/routing/rewrite/

api/soap/mc_api.php
```diff
@@ -35,29 +35,67 @@
...
+	 * @param integer $p_status_code The http status code
+	 * @param string $p_faultstring The error description
+	 */
+	function __construct( $p_status_code, $p_fault_string = '' ) {
+		$this->status_code = $p_status_code;
+		$this->fault_string = $p_fault_string === null ? '' : $p_fault_string;
+	}
+}
+
+/**
```
vboctor: Will defer auto-loader work out this of review.  I would like to think it through along with wider composer usage.

docbook/Developers_Guide/en-US/Events_Reference.xml
```diff
@@ -114,6 +114,26 @@
...
 		</blockquote>
 
+		<blockquote id="dev.eventref.system.rest_api_routes">
+			<title>EVENT_REST_API_ROUTES (Execute)</title>
+
+			<blockquote>
+				<para>
+					This event is triggered by MantisBT to enable plugins to register their own routes
+					to be accessible via the REST API.  All APIs belonging to a plugin named 'Example',
+					should live under 'https://.../api/rest/Example/'.  The route registration is done
```
vboctor: Used `MUST`, added /plugin/Example, and updated MantisGraph plugin to comply.

docbook/Developers_Guide/en-US/Events_Reference.xml
```diff
@@ -114,6 +114,26 @@
...
+					using the Slim Framework app instance that is passed as a parameter.  See MantisGraph
+					core plugin for an example.
+				</para>
+
+				<para>
+					Before calling into the plugin routes, the user would be already authenticated and
+					authorization for API access in general.  However, it is the responsibility of the
```
vboctor: Fixed.

plugins/MantisGraph/MantisGraph.php
```diff
@@ -65,6 +68,26 @@ function hooks() {
...
+		$t_app->group( '/MantisGraph', function() use ( $t_app ) {
+			$t_app->get( '/summary', function( $req, $res, $args ) {
+				if( access_has_project_level( config_get( 'view_summary_threshold' ) ) ) {
+					return $res->withStatus( 200 )->withJson( create_reporter_summary() );
+				}
+
+				return $res->withStatus( 403 );
+			} );
+		} );
+	}
```
vboctor: I have added plugin_route_group() to get the route group name for the plugin and updated documentation to provide more details and reference Slim Framework Router documentation.

vboctor:
@dregad I think I addressed all your feedback.

> @vboctor I think it's truly great that Mantis will offer REST API, and I strongly support the idea.

Great.  I'm very excited as well.  I think it will make a big difference and it is definitely one of the main aspects of MantisBT that needed fresh paint 👍 

> That said, I regretfully do not have enough free time for a proper, in-depth review, actually test the API or to effectively contribute to the effort, there is just too much on my plate these days.  Being aware of the benefits a modern webservice interface, and the fact that it's an often requested feature, I would not want to delay its implementation without good reason.

No worries, I will drive this.

> As an overall comment, I trust you have fully and globally analyzed the interface endpoints, as this is a promise to the webservice's consumers that can't easily (shouldn't) be changed down the line. I think it's very important that we have an overall consistency in the API, so we it would be useful to have a vision/roadmap to avoid adding mismatched bits and pieces as we go along. This should be documented somewhere (wiki ?)

I'm starting the API disabled by default and we will advertise it as preview/beta/experimental with potential breaking changes.  I will review and test, but I think we will benefit from having more people play with it and provide feedback.  Also open it up for plugins to start building their APIs.

As for the guidelines, I started a [gist](https://gist.github.com/vboctor/c7c356be2059aa67fa939e4cec235afa) with the guidelines.  We can move it to the wiki once it is stable.

vboctor:
Few extra changes:

- Support cookie based auth using cookies rather than cookie value in authorization header.
- Created a route group `/internal` for APIs that are private to the web UI.
- Added API for auto-complete for platform, OS and OS build.  Use REST API to power auto-complete instead of `xmlhttprequest_api.php` and `xmlhttprequest.php`.
- Disabling REST API only impacts api-token auth and not cookie auth, to keep the UI working.
- Use regular arrays rather than associating arrays for cases where we are not returning objects with pre-defined keys
- Updated swagger.json to add new APIs, schemas for POST and response data objects, and fix some schema errors.

rombert:
@vboctor - I'm still curious to see how you plan to approach testing.

vboctor:
@rombert I started working on this, then figured out that it is hard to test without having more operations available via the REST API to enable full test cycles.  So will revisit later.  It shouldn't be hard to test a REST API.

api/rest/restcore/issues_rest.php
```diff
@@ -47,4 +49,23 @@ function rest_issue_get( \Slim\Http\Request $p_request, \Slim\Http\Response $p_r
...
+/**
+ * Create an issue from a POST to the issues url.
+ *
+ * @param \Slim\Http\Request $p_request   The request.
+ * @param \Slim\Http\Response $p_response The response.
+ * @param array $p_args Arguments
+ * @return \Slim\Http\Response The augmented response.
+ */
+function rest_issue_add( \Slim\Http\Request $p_request, \Slim\Http\Response $p_response, array $p_args ) {
+	$t_issue = $p_request->getParsedBody();
```
Ashwin-Kapes: What's the format for JSON body to post an Issue?

crouti:
Is there any maintained PHP client for this RESTful API ? Thanks

vboctor:
@crouti please use forums for such questions.

See Power Shell client for sample usage.
https://github.com/mantishub/mantis-powershell

Otherwise, REST should be simple enough to use from any language.  At the moment, no client SDK for it.  You can see REST API documentation and download Postman definitions from http://bit.ly/2mC1Hpy

crouti:
Thanks ! and sorry not being posting in the correct place.

PR #1038, opened 2017-03-01
Markdown: Fix handling of # characters

jllano:
- parsedown is not following exactly with # symbol treated as header
- removed incorrect logic on header custom extension

Fixes #22333

screen shot 2017-03-01 at 9 54 51 pm

### DIFF
--- a/plugins/MantisCoreFormatting/core/MantisMarkdown.php
+++ b/plugins/MantisCoreFormatting/core/MantisMarkdown.php
@@ -102,10 +102,10 @@ class MantisMarkdown extends Parsedown
 	protected function blockHeader( $line ) {
 		
 		# Text either indented 0-3 spaces
-		# The opening sequence of # characters must be followed by a space
 		# An opening sequence of 1–6 # characters
-		# The #'s can be followed by a newline also
-		if ( preg_match( '/^\s{1,3}|#{1,6}(?:[ \t]+|$)/', $line['text'] ) ) {
+		# The opening sequence of # characters must be followed by a space
+		# The #'s can be followed by a newline as well
+		if ( preg_match( '/^ {0,3}#{1,6}( |$)/', $line['text'] ) ) {
 			return parent::blockHeader($line);
 		}
 	}
--- a/plugins/MantisCoreFormatting/core/MantisMarkdown.php
+++ b/plugins/MantisCoreFormatting/core/MantisMarkdown.php
@@ -95,15 +95,17 @@ class MantisMarkdown extends Parsedown
 	 *
 	 * @param string $line The Markdown syntax to parse
 	 * @access protected
-	 * @return null|string HTML representation generated from markdown or null if
-                      markdown starts with # symbol immediately followed by 
-                      any character other than space.
+	 * @return null|string HTML representation generated from markdown or null 
+	 				for example text starts with # symbol and immediately followed by 
+					any character other than space.
 	 */
 	protected function blockHeader( $line ) {
 		
-		# Check if string starts with # symbol followed by space, 
-		# If pattern match, then let the parent blockHeader function handle it
-		if ( preg_match('/^#+ /', $line['text'] ) ) {
+		# Text either indented 0-3 spaces
+		# The opening sequence of # characters must be followed by a space
+		# An opening sequence of 1–6 # characters
+		# The #'s can be followed by a newline also
+		if ( preg_match( '/^\s{1,3}|#{1,6}(?:[ \t]+|$)/', $line['text'] ) ) {
 			return parent::blockHeader($line);
 		}
 	}
--- a/plugins/MantisCoreFormatting/core/MantisMarkdown.php
+++ b/plugins/MantisCoreFormatting/core/MantisMarkdown.php
@@ -95,9 +95,9 @@ class MantisMarkdown extends Parsedown
 	 *
 	 * @param string $line The Markdown syntax to parse
 	 * @access protected
-	 * @return null|string HTML representation generated from markdown or null 
-	 				for example text starts with # symbol and immediately followed by 
-					any character other than space.
+	 * @return string|null HTML representation generated from markdown or null
+	 *				for example text starts with # symbol and immediately followed by
+	 *				any character other than space.
 	 */
 	protected function blockHeader( $line ) {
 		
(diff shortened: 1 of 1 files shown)


plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -91,54 +91,20 @@ public static function convert_line( $p_text ) {
 	}
 
 	/**
-	 * Customize the logic on Header elements
+	 * Customize the blockHeader markdown
 	 *
 	 * @param string $line The Markdown syntax to parse
 	 * @access protected
-	 * @return void if markdown starts with # symbol | string html representation generated from markdown.
+	 * @return void if markdown starts with # symbol followed by any character | string html representation generated from markdown.
```
dregad: [This is not valid PHPDoc](https://phpdoc.org/docs/latest/references/phpdoc/tags/return.html). The return type(s) must be defined at the beginning, i.e. `@return string Description`.

Furthermore, it is not correct to use *void* as return type here; as per [PHP documentation](http://php.net/manual/en/function.return.php), when *return* is used without parameter, NULL is returned.

```
* @return string|null HTML representation generated from markdown, or null if
                      markdown starts with # symbol immediately followed by 
                      any character other than space.
```

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -91,54 +91,20 @@ public static function convert_line( $p_text ) {
...
-
-		# make sure config option bug_link_tag == '#' only
-		# check if string start with # symbol followed by numbers
-		# hash[numbers] should not be treated as header, then let the app handle it
-		if( '#' == config_get_global( 'bug_link_tag' ) && preg_match_all( '/^#\d+$/', $line['text'], $matches ) ) {
-			return;
+		
+		# Parsedown is not exactly follows with # symbol as stated in http://spec.commonmark.org/0.27/#atx-heading
+		# Header need a space between the # symbol and the text as required by the original ATX implementation
+		# This fix will search for a match of # symbol followed by space only
```
dregad: `Parsedown is not exactly follows with # symbol ` what do you mean by that exactly ? Please reformulate in proper English.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -91,54 +91,20 @@ public static function convert_line( $p_text ) {
...
+		if ( preg_match('/^#+ /', $line['text'] ) ) {
```
dregad: To be fully compliant with the spec (as per your link):

- the opening # may be indented 0-3 spaces `^ {0,3}`
- since Markdown headers are limited to 6 levels, I would suggest to use `{1,6}` instead of `+`
- the #'s can be followed by a newline as well, so it should be `( |$)`

Note that I didn't actually test the regex suggestions above.

jllano: @dregad, my apologies for the confusion here, the logic `preg_match('/^#+ /', $line['text']` is already correct, the regex will just search only if the pattern will match the string that start with # symbol followed by space, then Parsdown library has to do the rest.

I'll just change the documentation instead.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -95,13 +95,14 @@ public static function convert_line( $p_text ) {
 	 *
 	 * @param string $line The Markdown syntax to parse
 	 * @access protected
-	 * @return void if markdown starts with # symbol followed by any character | string html representation generated from markdown.
+	 * @return null|string HTML representation generated from markdown or null if
+                      markdown starts with # symbol immediately followed by 
+                      any character other than space.
```
dregad: These lines are missing the leading `*`

jllano: > I don't think it is correct, because if the regex doesn't match (e.g. if the raw text is # blah at beginning of the line, then the parsedown parent method won't be called.

I'm not sure what you meant for? The logic is just simple:

`if raw text is like # blah (#+space+text) then the parent::blockHeader() will be called`.

jllano:
I will be adding unit tests.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -91,54 +91,23 @@ public static function convert_line( $p_text ) {
...
-		# check if string start with # symbol followed by numbers
-		# hash[numbers] should not be treated as header, then let the app handle it
-		if( '#' == config_get_global( 'bug_link_tag' ) && preg_match_all( '/^#\d+$/', $line['text'], $matches ) ) {
-			return;
+		
+		# Text either indented 0-3 spaces
+		# The opening sequence of # characters must be followed by a space
+		# An opening sequence of 1–6 # characters
+		# The #'s can be followed by a newline also
+		if ( preg_match( '/^\s{1,3}|#{1,6}(?:[ \t]+|$)/', $line['text'] ) ) {
```
dregad: Your updated regex has several problems
- `\s` is wrong, see the spec [1] _The opening # character may be indented 0-3 **spaces**_
- `{1-3}` is wrong, should be `{0-3}` (same as above)
- remove the `|` before `#`
- `[ \t]` is wrong, see spec [1] _The opening sequence of # characters must be followed by a **space** or by the end of line_
- the `+` is useless, remove it

See http://spec.commonmark.org/0.27/#atx-heading

jllano: @dregad Have you tried to test it in your end? or not

jllano: this one? `preg_match( '/^\s{0,3}#{1,6}( |$)/', $line['text'] )`

jllano:
> Asking for the 3rd time, please fix the PHPdoc block #1035 (comment)

Sorry, What do you mean on this?

dregad:
Apologies, I copy/pasted the wrong link
You didn't fix https://github.com/mantisbt/mantisbt/pull/1038#discussion_r104141853

dregad:
@jllano please note that your PR conflicts with master branch (this is most likely a consequence of my resolving #1007)

jllano:
Ok, What do you think is best approach for this? rebase my changes on top of upstream? or create a new PR and apply the changes instead?

jllano:
Closing this PR in favor of #1043

PR #1039, opened 2017-03-01
Fix print page error when cookie not set

dregad:
When the VIEW_ALL_COOKIE is not set or blank, an attempt to display
print_all_bug_page.php will trigger a system error: 'Argument 1 passed
to filter_get_visible_sort_properties_array() must be an array, null
given'.

This is due to the filter ($t_filter_cookie_arr) not being initialized
in this case, leading to an error as it's defaulted to NULL while the
filter_get_visible_sort_properties_array() function expects an array.

Use a default filter when the cookie is not set or empty.

Fixes [#22442](http://mantisbt.org/bugs/view.php?id=22442)

### DIFF
--- a/docbook/Admin_Guide/en-US/Customizing.xml
+++ b/docbook/Admin_Guide/en-US/Customizing.xml
@@ -785,7 +785,7 @@ $g_notify_flags['owner'] = array(
                         about. Also send new issue notifications to managers so they can
                         screen them.
 <programlisting>
-$g_mail_receive_own = OFF;
+$g_email_receive_own = OFF;
 $g_default_notify_flags = array(
 	'reporter'      => ON,
 	'handler'       => ON,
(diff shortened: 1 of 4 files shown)


dregad:
I am not sure why Github is showing the commits from the 1.3.x branch merge, since master-2.2 branch's HEAD is at the merge commit (75ef07835bc3771c9632ed20aa6512ed9d6a17cc). I guess it's a consequence of changing the base branch and force-pushing the feature branch.

Anyway only 5bd51cc97edb92f3570b66ce8ea86b20e3d28405 needs to be reviewed.

dregad:
@vboctor responded to your comment, see https://github.com/mantisbt/mantisbt/commit/5bd51cc97edb92f3570b66ce8ea86b20e3d28405#commitcomment-21137326, but why didn't you comment directly in the PR ?

PR #1040, opened 2017-03-01
Fix manage user page filter '0'-'9' selection

atrol:
Fixes #22445

### DIFF
--- a/manage_user_page.php
+++ b/manage_user_page.php
@@ -170,7 +170,7 @@ foreach ( $t_prefix_array as $t_prefix => $t_caption ) {
 	} else {
 		$t_title = '';
 	}
-	$t_active = $t_prefix === $f_filter ? 'active' : '';
+	$t_active = (string)$t_prefix === $f_filter ? 'active' : '';
 		print_manage_user_sort_link( 'manage_user_page.php',
 			$t_caption,
 			$c_sort,


manage_user_page.php
```diff
@@ -170,7 +170,7 @@
 	} else {
 		$t_title = '';
 	}
-	$t_active = $t_prefix === $f_filter ? 'active' : '';
+	$t_active = (string)$t_prefix === $f_filter ? 'active' : '';
```
vboctor: It is strange that when the array key and value were set earlier they were already casted to a string.  Do you understand why this happens?  Is there a reason we have '===' specially if we are going to cast anyway?

atrol: > It is strange that when the array key and value were set earlier they were already casted to a string

Exactly my thoughts when looking at the problem

The surprising output of the following script is: `string(1) "A" string(1) "B" int(2) string(1) "3"`
At least it explains the behavior.

```php
<?php

$t_prefix_array['A'] = 'B';
$t_prefix_array[(string)2] = (string)3;

foreach ( $t_prefix_array as $t_prefix => $t_caption ) {
    var_dump( $t_prefix, $t_caption );
}
```

> Is there a reason we have '===' specially if we are going to cast anyway?

'===' has been introduced to fix http://www.mantisbt.org/bugs/view.php?id=21115 see https://github.com/mantisbt/mantisbt/commit/413e641fc6b30ad0491044add4e999d1a2909df2

atrol:
> documented behaviour

still surprising, especially for developers that are also using strongly typed languages.
I am quite sure that there are not that much PHP developers who are always aware of this  behaviour.

PR #1044, opened 2017-03-07
#0022473: Avatar size does not handle ratio

romaincabassot:
Removed HTML attributes width and height from avatar image to let the CSS handle the size and ratio.


romaincabassot:
Size & ratio is handled by the CSS ace-mantis.css line 229:
.avatar {
    border: 2px solid #c9d6e5;
    border-radius: 15%;
    box-shadow: none;
    margin: 0;
    **max-width: 80px;**
}

dregad:
Please open an issue in our tracker, describing what the problem is, and reference this pull request.

atrol:
@dregad it's there http://www.mantisbt.org/bugs/view.php?id=22473

This is a loss of functionality as you are no longer able to display avatars with different sizes.
You will need it to implement something like that:

avatar

romaincabassot:
@atrol Sorry I don't understand why it would no longer be able to display avatars with different sizes. 

Sure it would not be so uniform as if we state that all avatar will be 80x80 but with the CSS specifying that max-width will be 80px it should be enough homogeneous (and it will be possible to specify max-height: 80px to prevent from avatars very large or tall).

If Mantis wants to allow users to plug to other avatars sources I think it should let the admins of the avatar sources adapt them if they don't fit well in the Mantis theme but you cannot force them to have a fixed size because those avatars are likely used elsewhere.

For the second part of your message I'm not sure to understand it to... You ask me if I want to implement the display of avatars in the "View Issue Details"  for the "Reporter" and "Assigned to" fields?

Sorry if we are mistaken.

romaincabassot:
For example I'm writing a Ldap avatar plugin for Mantis (https://github.com/romaincabassot/MantisLdapAvatar) and I managed to retrieve the avatars from Ldap and I resize them to fit a max width/height of 80 but it's another thing to resize to 80x80 because you have to change the ratio or you have to crop the avatars.

atrol:
> you ask me if I want to implement the display of avatars in the "View Issue Details" for the "Reporter" and "Assigned to" fields?

No, this was just an example of a new feature where displaying avatars with different sizes is needed.
I implemented a quick and dirty solution for it to provide the screenshot by using `print_avatar ($t_bug->reporter_id, 30);`

I thought at first sight, that parameter $p_size is no longer usedy after your changes, so we would have lost some functionality.

I was wrong. The parameter is still used `Avatar::get( $p_user_id, $p_size );`

Sorry for the confusion.

romaincabassot:
No problem.
Yes the parameter is still used and the display of the avatar image is handled by the css.
As we pass by, showing the avatar in the reporter and assigned to section would be nice.

dregad:
> it will be possible to specify max-height: 80px to prevent from avatars very large or tall

Since we currently assume that the avatars have a 1:1 (square) aspect ratio, I believe that would indeed be necessary. And you should also check everywhere avatars are currently used, to ensure that a rectangular ratio is not introducing display issues messing with page alignment due to the avatar being too wide or tall.

vboctor:
Just confirming.  The space allocated for the avatar should always be a square.  Now if an avatar is a rectangle, it should be stretched to fit within the square, without changing the aspect ratio.  I believe CSS max height and max width would help.

So tests should be as follows:

- Avatar takes no more space than the allocated square (80x80, 30x30) hence doesn't impact layout where it is hosted.
- Avatar for square avatars looks perfect.
- Avatars with width great than specified size would have an empty space at top + bottom, but width fits within given pixels (e.g. 80 or 30).
- Avatars with height greater than specified  size would have an empty space at left + right, but height fits within given pixels (e.g. 80 or 30).
- If we style avatars as circles, then the resized avatar should be centered within the space allocated to it.

Hence, it doesn't matter where we use avatars now (or in the future), since the contract is that they are always centered within or fully utilizing the given square size.

romaincabassot:
@dregad @vboctor I'll have a look to this and update the PR.

romaincabassot:
It looks like that now: screenshot_20170308_163739

dregad:
Does it work with Timeline view as well ? (sorry I don't have time to check for myself)

vboctor:
If we are centering the picture within the allocated square, then it should work everywhere.

romaincabassot:
@dregad Yes it works. Here is another screenshot:
mantis_timeline

vboctor:
@syncguru can you have a look at this change?

@romaincabassot 

- Please override styles in `ace-mantis.css` and revert changes in `ace.css` and its minimized version.
- Update `print_avatar` everywhere instead of outputting html everywhere.  This will be useful for core and plugins as they print avatars.
- Update `print_avatar` to accept a CSS class prefix.  Then we can add to it `_(size)`, then `_container` and `_avatar`.  This way we have consistent naming pattern and we can provide default for the CSS prefix.
- If you need the html string for the avatar instead of printing it, you can add `prepare_avatar()` in `core/prepare_api.php` that returns the html string, then use it from `print_avatar()`.
- One of the places where we output an avatar we use an image without a class.  This should be fixed if we use `print/prepare_avatar()` api everywhere.
- Please collapse all commits into a single commit.

syncguru:
@romaincabassot

Can you post snapshots for small screens? We may need to add css classes for every size we support.

romaincabassot:
Here are the screenshots from my smartphone:
screenshot_2017-03-14-08-10-29
screenshot_2017-03-14-08-10-43
screenshot_2017-03-14-08-11-00

romaincabassot:
The commit 3903b6d contains all the modifications.

core/prepare_api.php
```diff
@@ -104,3 +104,52 @@ function prepare_version_string( $p_project_id, $p_version_id ) {
...
+ * Prepares avatar for raw outputting (only avatar image).
+ *
+ * @param Avatar $p_avatar 			An instance of class Avatar.
+ * @param string $p_class_prefix	CSS class prefix to add to the avatar's surrounding div and to the img.
+ *   The CSS classes to implement will be named [$p_class_prefix]_[$p_size]_avatar_container and 
+ *   [$p_class_prefix]_[$p_size]_avatar.
+ * @param integer $p_size    		Image maximum size.
+ * @return the HTML string of the avatar.
+ */
+function prepare_raw_avatar( $p_avatar, $p_class_prefix, $p_size) {
```
vboctor: Please revise spacing inside () across all changes.  For example, the function definition and calls just look as follows:

```
func1( $p_param1, $p_param2 )
```

Notice the spaces inside the ().

core/prepare_api.php
```diff
@@ -104,3 +104,52 @@ function prepare_version_string( $p_project_id, $p_version_id ) {
...
+	if( $p_avatar === null ) {
+		return '';
+	}
+
+	$t_image = htmlspecialchars( $p_avatar->image );
+	$t_text = htmlspecialchars( $p_avatar->text );
+
+	$avatar_class = $p_class_prefix . '_' . $p_size . '_avatar';
```
vboctor: Use `-` instead of `_` for CSS class names.

core/prepare_api.php
```diff
@@ -104,3 +104,52 @@ function prepare_version_string( $p_project_id, $p_version_id ) {
...
+	return '<img class="' . $avatar_class . '" src="' . $t_image . '" alt="' .
+			$t_text . '" />';
+}
+
+/**
+ * Prepares avatar for outputting.
+ *
+ * @param Avatar $p_avatar 			An instance of class Avatar.
```
vboctor: Use spaces to align argument descriptions or don't align them.

core/prepare_api.php
vboctor: Use `$t_avatar_class`

core/prepare_api.php
```diff
@@ -104,3 +104,52 @@ function prepare_version_string( $p_project_id, $p_version_id ) {
...
+	if( $p_avatar === null ) {
+		return '';
+	}
+
+	$t_link = htmlspecialchars( $p_avatar->link );
+
+	$container_class = $p_class_prefix . '_' . $p_size . '_avatar_container';
+	return '<div class="' . $container_class . '">' . 
+			'<a rel="nofollow" href="' . $t_link . '">' .
+			prepare_raw_avatar ($p_avatar, $p_class_prefix, $p_size) .
```
vboctor: Remove space before ( and add spaces inside ().

core/print_api.php
```diff
@@ -182,24 +182,15 @@ function print_successful_redirect( $p_redirect_to ) {
...
-
-	$t_image = htmlspecialchars( $t_avatar->image );
-	$t_link = htmlspecialchars( $t_avatar->link );
-	$t_text = htmlspecialchars( $t_avatar->text );
 
-	echo '<a rel="nofollow" href="' . $t_link . '">' .
-		'<img class="avatar" src="' . $t_image . '" alt="' .
-		$t_text . '" width="' . $p_size . '" height="' .
-		$p_size . '" /></a>';
+	echo prepare_avatar ($t_avatar, $p_class_prefix, $p_size);
```
vboctor: Fix spacing.

css/ace-mantis.css
```diff
@@ -214,25 +214,49 @@ pre {
...
 
-.ace-nav .nav-user-photo {
+.ace-nav .nav-user-photo_40_avatar_container {
+	width: 48px; 
+	height: 40px;
+	float: left;	
+}
+
+.ace-nav .nav-user-photo_40_avatar {
+	margin: -4px 8px 0 0;
```
vboctor: Fix indentation.

css/ace-mantis.css
```diff
@@ -214,25 +214,49 @@ pre {
...
+	text-align: center;
+}
+
 .table-responsive {
     margin-bottom: 0 !important;
 }
 
-.avatar
-{
+.bugnote_view_inc_80_avatar {
```
vboctor: How about calling this style `note_80_avatar` or `bugnote_80_avatar`?

atrol:
Changes in strings_english.txt and strings_french.txt should be removed from the PR.

romaincabassot:
@vboctor This commit should handle your comments.

romaincabassot:
@atrol Yes sorry

bugnote_view_inc.php
```diff
@@ -155,7 +155,7 @@
 ?>
 <tr class="bugnote" id="c<?php echo $t_activity['id'] ?>">
 		<td class="category">
-		<div class="pull-left padding-2"><?php print_avatar( $t_activity['user_id'] ); ?>
+		<div class="pull-left padding-2"><?php print_avatar( $t_activity['user_id'], 'bugnote_view_inc', 80 ); ?>
```
romaincabassot: @syncguru vboctor told me to do it this way:
>     Update print_avatar to accept a CSS class prefix. Then we can add to it _(size), then _container and _avatar. This way we have consistent naming pattern and we can provide default for the CSS prefix.
Can you state on this before I push a new commit? Doing it this way allows to use the function in different places (it's not the case now but it could be if you plan to add a little avatar in the bug detail page for the reporter and assignee).

lang/strings_english.txt
```diff
@@ -264,7 +264,7 @@ $s_view_state_enum_string = '10:public,50:private';
 $s_priority_enum_string = '10:none,20:low,30:normal,40:high,50:urgent,60:immediate';
 $s_severity_enum_string = '10:feature,20:trivial,30:text,40:tweak,50:minor,60:major,70:crash,80:block';
 $s_reproducibility_enum_string = '10:always,30:sometimes,50:random,70:have not tried,90:unable to reproduce,100:N/A';
-$s_status_enum_string = '10:new,20:feedback,30:acknowledged,40:confirmed,50:assigned,80:resolved,90:closed';
+$s_status_enum_string = '10:new,20:feedback,25:external wait,26:internal wait,30:acknowledged,40:confirmed,50:assigned,80:resolved,85:delivered,90:closed';
```
vboctor: Changes to language files seem to be unrelated to this change.  Btw, you should generally be changing custom_strings_inc.php for such customizations rather than changing core language files.

css/ace-mantis.css
```diff
@@ -214,25 +214,49 @@ pre {
     position: relative;
 }
 
-.ace-nav .nav-user-photo {
-    border-radius: 4px !important;
+.ace-nav .nav-user-photo-40-avatar-container {
```
vboctor: Should `nav-user-photo` be changed to `nav`, similar to `profile-activity`?

core/prepare_api.php
```diff
@@ -104,3 +104,52 @@ function prepare_version_string( $p_project_id, $p_version_id ) {
...
+ * Prepares avatar for outputting.
+ *
+ * @param Avatar $p_avatar          An instance of class Avatar.
+ * @param string $p_class_prefix    CSS class prefix to add to the avatar's surrounding div and to the img.
+ *   The CSS classes to implement will be named [$p_class_prefix]-[$p_size]-avatar-container and
+ *   [$p_class_prefix]-[$p_size]-avatar.
+ * @param integer $p_size           Image maximum size.
+ * @return the HTML string of the avatar.
+ */
+
```
vboctor: Remove blank line.

lang/strings_english.txt
romaincabassot: Sure...

css/ace-mantis.css
romaincabassot: nav-user-photo is defined in other css:
- ace-mantis.css (2 matches)
- ace-rtl.css (4 matches)
- ace-rtl.min.css (4 matches)
- ace.css (4 matches)
- ace.min.css (4 matches)
Maybe it should stay as is.

core/prepare_api.php
romaincabassot: Yep

vboctor:
One more thing, when constructing css class names, let's have the size at the end.  This will be consistent with other classes like `bigger-125`, `width-40`, etc.

css/ace-mantis.css
romaincabassot: Done

vboctor:
Thanks @romaincabassot - merged via f6229fbef8f797d3d48338e7e795d5f1d671ccbe

PR #1045, opened 2017-03-07
Add project_id param to project_*_name_unique()

dregad:
This optional parameter is used to exclude the given id from the
uniqueness check, allowing correct behavior (avoid error 701) when
updating an existing project.

This was a problem on MySQL, because with utf8_general_ci collation,
accented characters are treated as identical, i.e. 'The' == 'thé'.

Fixes [#22479](http://www.mantisbt.org/bugs/view.php?id=22479)

### DIFF
--- a/core/project_api.php
+++ b/core/project_api.php
@@ -242,18 +242,21 @@ function project_ensure_exists( $p_project_id ) {
 /**
  * check to see if project exists by name
  * @param string $p_name The project name.
+ * @param int    $p_id   Optional project id to exclude from the check, to
+ *                       allow uniqueness check when updating.
  * @return boolean
  */
-function project_is_name_unique( $p_name ) {
+function project_is_name_unique( $p_name, $p_project_id = null ) {
 	db_param_push();
 	$t_query = 'SELECT COUNT(*) FROM {project} WHERE name=' . db_param();
-	$t_result = db_query( $t_query, array( $p_name ) );
-
-	if( 0 == db_result( $t_result ) ) {
-		return true;
-	} else {
-		return false;
+	$t_param = array( $p_name );
+	if( $p_project_id ) {
+		$t_query .= ' AND id <> ' . db_param();
+		$t_param[] = (int)$p_project_id;
 	}
+	$t_result = db_query( $t_query, $t_param );
+
+	return 0 == db_result( $t_result );
 }
 
 /**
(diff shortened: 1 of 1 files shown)


core/project_api.php
```diff
@@ -242,29 +242,34 @@ function project_ensure_exists( $p_project_id ) {
 /**
  * check to see if project exists by name
  * @param string $p_name The project name.
+ * @param int    $p_id   Optional project id to exclude from the check, to
```
atrol: $p_project_id instead of $p_id

core/project_api.php
```diff
@@ -242,29 +242,34 @@ function project_ensure_exists( $p_project_id ) {
...
+
+	return 0 == db_result( $t_result );
 }
 
 /**
  * check to see if project exists by id
  * if it doesn't exist then error
  * otherwise let execution continue undisturbed
  * @param string $p_name The project name.
+ * @param int    $p_id   Optional project id to exclude from the check, to
```
atrol: $p_project_id instead of $p_id

core/project_api.php
```diff
@@ -442,7 +447,8 @@ function project_update( $p_project_id, $p_name, $p_description, $p_status, $p_v
 	}
 
 	if( strcasecmp( $p_name, $t_old_name ) != 0 ) {
-		project_ensure_name_unique( $p_name );
+var_dump(1);
```
atrol: Remove var_dump

core/project_api.php
```diff
@@ -241,30 +241,35 @@ function project_ensure_exists( $p_project_id ) {
 
 /**
  * check to see if project exists by name
- * @param string $p_name The project name.
+ * @param string $p_name       The project name.
```
vboctor: I would rather use `$p_exclude_project_id` instead of `$p_project_id`.

core/project_api.php
```diff
@@ -241,30 +241,35 @@ function project_ensure_exists( $p_project_id ) {
...
+ * @param int    $p_project_id Optional project id to exclude from the check,
+ *                             to allow uniqueness check when updating.
```
atrol: int is ok, but most of the time we use integer

core/project_api.php
```diff
@@ -241,30 +241,35 @@ function project_ensure_exists( $p_project_id ) {
...
+	return 0 == db_result( $t_result );
 }
 
 /**
  * check to see if project exists by id
  * if it doesn't exist then error
  * otherwise let execution continue undisturbed
- * @param string $p_name The project name.
+ * @param string $p_name       The project name.
+ * @param int    $p_project_id Optional project id to exclude from the check,
```
atrol: same as above

core/project_api.php
```diff
@@ -442,7 +447,7 @@ function project_update( $p_project_id, $p_name, $p_description, $p_status, $p_v
 	}
 
 	if( strcasecmp( $p_name, $t_old_name ) != 0 ) {
-		project_ensure_name_unique( $p_name );
+		project_ensure_name_unique( $p_name, $p_project_id );
```
atrol: $p_exclude_id

atrol: Forget it, got confused by the way it's been displayed in GitHub.

PR #1046, opened 2017-03-08
Markdown: Fix links within code block

jllano:
Within the code block or backtick, parsedown converts the predefined
characters '<' (less than) and '>' (greater than) to HTML entities, so
that become `&lt` and `&gt;`

This fixes reverts the HTML entities back to characters so that
it displays correctly.

Fixes #22320

BEFORE:
screen shot 2017-03-08 at 11 11 55 am

AFTER:
screen shot 2017-03-08 at 11 12 24 am


vboctor:
Just to be clear, I would expect the fix here to be that we don't replace # + 123 into a link, but leave it as is, since it is in a code block.  Everything else should work fine since parsedown tests handles anchor links fine.

jllano:
Firstly, MantisBT text processing transformed the text thru `string_process_bug_link()` as link, then Parsedown (parent) converts the predefined characters `<` (less than) and `>` (greater than) to HTML entities `&lt;` and `&gt;`.

The fix reverts any `&lt;` and `&gt;` entity back to its characters so that it displays correctly.

It doesn't affect with the header `H's` stuff with in the code block but leave as is.

jllano:
I attached an screenshots here https://github.com/mantisbt/mantisbt/pull/1046#issue-212630503

vboctor:
@jllano we shouldn't hyperlink what looks like issue references within code blocks.  Code blocks should appear exactly as they were typed by the user.  We need to change our core formatting plugin to only process issue references outside the code blogs rather than in the whole text being processed.

PR #1048, opened 2017-03-10
Use modern hash function when save user's password

fetus-hina:
Today, the MD5 hash algorithm is very weak and Mantis's current hash algorithm does not use salt and key stretching.
The hashed password that created by current algorithm is seriously weak. (The hashed password that stored in database maybe contained in a rainbow-table, or may calculates quickly with CPU and/or GPU)

This pull-request adds new password hashing algorithm and switch default to it.
As of today, new algorithm uses [bcrypt](https://en.wikipedia.org/wiki/Bcrypt) via PHP-built-in `password_hash()` function. (This function available from PHP 5.5)
If bcrypt becomes "weak", the PHP team (or the distros' maintainer) will upgrade algorithm to safer one. Then, stored hashed-password will upgrade automatically when user logged in after update PHP engine. This means "this new algorithm is maintenance-free."

Backward compatibility info: The hashed-password that created by MD5 or other older algorithms is also valid. Older password will upgrade automatically when first log in.

Note: The `password_hash()` function requires PHP 5.5+ and Mantis's current minimal requirement is [5.5.9+](https://github.com/mantisbt/mantisbt#requirements). But, the documentation that linked from [download page](https://www.mantisbt.org/download.php) says "5.3.2+" (it looks for Mantis v1.3.x, but Mantis website links to that version now). If we should support PHP 5.3.2-5.4.x, we should include the [ircmaxell/password_compat](https://github.com/ircmaxell/password_compat) library. (This PR does not so.)

As you may have noticed, I'm not good at English. I feel sorry for my difficult-to-read English sentences. I will saved when feedback is simple and easy English.

Thank you.


dregad:
Hello @fetus-hina many thanks for your contribution, this is greatly appreciated as I've had this on my todo-list for ages, but somehow never got around to coding it. I will review your pull request later.

See also related #50 (yes that's old)

Don't worry about your English, it is anyway much better than my Japanese ;)

> the documentation that linked from download page says "5.3.2+" (it looks for Mantis v1.3.x, but Mantis website links to that version now).

Thanks for pointing that out; I will fix it. (for the record, even though you won't have access to it [#22511](https://mantisbt.org/bugs/view.php?id=22511))

dregad:
> I will fix it. (for the record, even though you won't have access to it #22511)

It's done.

vboctor:
Is there a way to incorporate the hashing mechanism into the AuthPlugin?

dregad:
@fetus-hina I rebased your branch on top of latest master, made a few adjustments and updated documentation. Many thanks for your work, let me know if you have any objections with the changes I made.

I ran some basic tests, everything seems to work fine when "upgrading" the login method e.g. from MD5, but additional testing is required to ensure everything works as expected in various conversion scenarios as well.

> Is there a way to incorporate the hashing mechanism into the AuthPlugin? #1070

@vboctor, not sure what you mean by that, can you clarify ?

core/authentication_api.php
```diff
@@ -764,29 +783,29 @@ function auth_does_password_match( $p_user_id, $p_test_password ) {
...
-
 			# a null salt is the same as no salt, which causes a salt to be generated
 			# otherwise, use the salt given
 			$t_processed_password = crypt( $p_password, $p_salt );
 			break;
 		case MD5:
 			$t_processed_password = md5( $p_password );
 			break;
+		case SAFE_HASH:
+			$t_processed_password = password_hash( $p_password, PASSWORD_DEFAULT );
```
vboctor: We should use `PASSWORD_BCRYPT` to avoid the option of the algorithm being upgraded with PHP resulting in truncation.  We can aways manually update this in future releases as a new algorithm is released whose output can be stored into the available field size.  The documentation recommends 255 and we are currently 64.  Not sure if we can move to 191 vs. 255.  The bcrypt requires 60 characters, so we are covered there.

core/constant_inc.php
```diff
@@ -591,8 +592,11 @@
 define( 'DB_FIELD_SIZE_PASSWORD', 64 );
 define( 'DB_FIELD_SIZE_API_TOKEN_NAME', 128 );
 
-# Maximum size for the user's password when storing it as a hash
-define( 'PASSWORD_MAX_SIZE_BEFORE_HASH', 1024 );
+# Maximum size for the user's password when storing it as a hash.
+# The CRYPT_BLOWFISH algorithm used by password_hash() will silently truncate
+# passwords to 72 characters, so there is no point in allowing longer passwords.
+# @see http://php.net/function.crypt
+define( 'PASSWORD_MAX_SIZE_BEFORE_HASH', 72 );
```
vboctor: I would rather keep the limit as before.  For example, if an auth plugin that validates passwords supports a longer length (e.g. LDAP), then we shouldn't block it.  Let's not make decisions based on one specific algorithm.

config_defaults_inc.php
```diff
@@ -1837,14 +1837,25 @@
...
- * @global integer $g_login_method
- */
-$g_login_method = MD5;
+ * should be carefully chosen at install time. However, if possible, MantisBT
+ * will attempt to "fall back" to older methods, and to convert the DB-stored
+ * hashes to the current login method after a successful login (e.g. with
+ * default settings, a MD5 hash will be upgraded to SAFE_HASH).
+ *
+ * @global integer $g_login_method One of: SAFE_HASH (default), LDAP, BASIC_AUTH
+ *                                 or HTTP_AUTH. Other values (PLAIN, MD5, CRYPT
```
vboctor: Should we fail when extremely old algorithms are configured as current? e.g. PLAIN and CRYPT?  We should continue supporting converting them to newer SAFE_HASH as users login.

core/authentication_api.php
```diff
@@ -764,29 +783,29 @@ function auth_does_password_match( $p_user_id, $p_test_password ) {
...
+			break;
 		case BASIC_AUTH:
 		case PLAIN:
 		default:
 			$t_processed_password = $p_password;
 			break;
 	}
 
-	# cut this off to DB_FIELD_SIZE_PASSWORD characters which the largest possible string in the database
+	# cut this off to DB_FIELD_SIZE_PASSWORD characters which is the largest
```
vboctor: Though not new.  This truncation behavior is not great since making the password field wider later will cause users to no longer be able to login.  Users will have to reset password at this point.

core/authentication_api.php
```diff
@@ -717,6 +717,20 @@ function auth_does_password_match( $p_user_id, $p_test_password ) {
...
 	$t_password = user_get_field( $p_user_id, 'password' );
+
+	# Process SAFE_HASH separately from the legacy methods, to safeguard against
+	# timing attacks leveraging password_verify() function.
+	if(    SAFE_HASH == $t_configured_login_method
+		&& password_verify( $p_test_password, $t_password )
+	) {
+		# Update the password hash if it does not match the current algorithm
+		if( password_needs_rehash( $t_password, PASSWORD_DEFAULT ) ) {
+			user_set_password( $p_user_id, $p_test_password, true );
```
vboctor: This is not new.  But the `user_set_password` assumes that it is being called by a user and hence it does change the cookie which is not necessary since we are just no hashing.  So no reason to log user out of other browsers for example.  Technically speaking, we don't need to worry about verification token, but that should have no side effects.

core/authentication_api.php
```diff
@@ -717,6 +717,20 @@ function auth_does_password_match( $p_user_id, $p_test_password ) {
...
 
 	$t_password = user_get_field( $p_user_id, 'password' );
+
+	# Process SAFE_HASH separately from the legacy methods, to safeguard against
+	# timing attacks leveraging password_verify() function.
+	if(    SAFE_HASH == $t_configured_login_method
+		&& password_verify( $p_test_password, $t_password )
+	) {
+		# Update the password hash if it does not match the current algorithm
+		if( password_needs_rehash( $t_password, PASSWORD_DEFAULT ) ) {
```
vboctor: Use PASSWORD_BCRYPT algorithm.

core/authentication_api.php
```diff
@@ -717,6 +717,20 @@ function auth_does_password_match( $p_user_id, $p_test_password ) {
 	}
 
 	$t_password = user_get_field( $p_user_id, 'password' );
```
vboctor: For readability, we should probably call rename `$t_password` to `$t_password_hash`.

dregad: I thought about it, but decided to leave it as-is because it's not necessarily a hash (think PLAIN or BASIC_AUTH login methods), and the underlying table column is called _password_

core/authentication_api.php
dregad: Changing a password *should* cause all sessions to be terminated, in general.

However, in this specific case I agree that this is really a "technical" change (i.e. we have a new hash, but the password itself remains the same) so there is no need to actually log the user out.

Are you suggesting to call `user_set_field( $p_user_id, 'password', $p_test_password )` instead ? 
Note that this would require special handling for protected accounts (since *user_set_fields* will trigger an error in this case). Alternatively, adding a new parameter to user_set_field, or defining a new API function that doesn't reset the cookie / token could work as well.

fetus-hina:
@dregad 
I have no objections about your work. However, we seem to need discussion about length limitation.
Also I have no opinion about it, at this time.

core/authentication_api.php
dregad: Using the default allows the same code base to automatically use the most advanced available algorithm, which I think is nice. 

I thought about the database field size given the recommendation in the PHP manual, but given how much advance notice we'd get before a change of default algo would be effective, I think it would be a better approach to make a schema change (can be done now or later) to increase the password field's size.

>  The documentation recommends 255 and we are currently 64. Not sure if we can move to 191 vs. 255. The bcrypt requires 60 characters, so we are covered there.

I know, I had the BLOWFISH hash length in mind when I increased the field size back in 2015.

The 191 chars limit only applies to MySQL when the column is indexed, which is not the case of the _password_ field. So there should be no issue with using 255 here.

core/authentication_api.php
dregad: That's true, but I can't think of any alternative option. Do you have any suggestions ?

config_defaults_inc.php
dregad: The Admin Checks will issue a warning in this case (including for MD5 now), I think that's sufficient - let the administrator decide.

We could obsolete them completely if you want to but I think that's outside of this PR's scope.

core/constant_inc.php
dregad: Is it worth adding complexity to the code to deal with a theoretical scenario that is unlikely to be an issue ? 

I mean, a password longer than 72 chars, _really_ ? `This is a really long password that I don't want to type everyday` is 66 chars...

An alternative could be to let BCRYPT silently truncate the password, the likelihood of a collision is

core/authentication_api.php
dregad: See my reply below

core/constant_inc.php
dregad: I'm having second thoughts about this. Maybe it does make sense to continue allowing long passwords prior to hashing, and only capping the length when using blowfish algo.

dregad:
@vboctor do you have any further comments on my replies following your initial review ?

Also, it would be nice if you could provide feedback on https://github.com/mantisbt/mantisbt/commit/8e466bb3dc84bf30fa893ca206ddf89032dacf11#commitcomment-21980157

core/authentication_api.php
vboctor: ok, not a big deal.

core/authentication_api.php
vboctor: I'm not recommending a specific approach, you can make such decision.  I just think we should skip the steps that don't apply when we are changing the hash but the password remains the same.

core/authentication_api.php
vboctor: I'm not a fan of auto-rolling algorithm without a specific MantisBT upgrade.  But I'm OK with doing it later once the field size is updated to 255.  Even if we change the db schema in a new release, using PASSWORD_DEFAULT today would mean that an old MantisBT release deployed on a server with a new PHP version would break.

Ideally, I would rather if we can select a specific algorithm as they become available and as we know it is compatible with our schema.

core/authentication_api.php
vboctor: I would error instead of truncate when setting a password that exceeds db schema size.

core/constant_inc.php
vboctor: We really should only enforce limit on plain password if it doesn't fit into the db schema.  We should also retire plain password to get out of this business and only offer rolling it over to a secure algorithm.

config_defaults_inc.php
vboctor: We should also consider whether we want to call this SAFE_HASH vs. BCRYPT or some specific algorithm that we choose in the future.  That builds on using a specific algorithm and being more explicit about it.  But we may decide to do one without the other.

vboctor:
@dregad I responded to your replies.  I think these should be addressed or get to an agreement on.

core/authentication_api.php
dregad: OK, I'll implement that.

core/authentication_api.php
dregad: > using PASSWORD_DEFAULT today would mean that an old MantisBT release deployed on a server with a new PHP version would break.

Good point. I'll switch to PASSWORD_BCRYPT and add a TODO to mention that this should be changed once the the field size has been updated to 255.

core/authentication_api.php
dregad: OK

core/constant_inc.php
dregad: > We really should only enforce limit on plain password if it doesn't fit into the db schema

IIRC we already do that on the front-end (via input field _maxlength_).

> retire plain password 

Per our release management guidelines, removing deprecated feature should only happen in Major versions. 

We already have a warning in admin checks recommending to switch to SAFE_HASH when method != LDAP, but we can fine-tune that if you want to make it more explicit.

config_defaults_inc.php
dregad: I am not overly keen on the `SAFE_HASH` name but couldn't think of a better name, so I'm fine with changing it.

Considering the switch from PASSWORD_DEFAULT to PASSWORD_BCRYPT (https://github.com/mantisbt/mantisbt/pull/1048#discussion_r115010854), it makes sense to use a more specific name.

dregad:
@vboctor thanks. I'll make adjustments per my replies above, and let you know when it's ready for the next round of review. 

Note that you have not responded to 
> Also, it would be nice if you could provide feedback on https://github.com/mantisbt/mantisbt/commit/8e466bb3dc84bf30fa893ca206ddf89032dacf11#commitcomment-21980157

core/constant_inc.php
vboctor: > Per our release management guidelines, removing deprecated feature should only happen in Major versions.

I mean eventually we should retire plain password when it is the right time, till then a good admin check is a good option to use.

> IIRC we already do that on the front-end (via input field maxlength).

I recall that we had a fixed max length.  The max length in case of using a hash is not really useful.  But if we are using plain passwords, it is must to apply a 64 limit.  Hence, we should have an app for max length to enforce in the UI that should take into consideration the algorithm used.

config_defaults_inc.php
vboctor: > Considering the switch from PASSWORD_DEFAULT to PASSWORD_BCRYPT (#1048 (comment)), it makes sense to use a more specific name.

I would use BCRYPT as the name and internally use PASSWORD_BCRYPT to PHP API.  I don't not sure I'm for using PASSWORD_DEFAULT even in the future.  I would rather if we manually upgrade to better algorithms with specific names as they become available.

core/authentication_api.php
vboctor: I would rather have changes in hash algorithm be synchronized with new MantisBT releases rather than based on PHP version installed.  Hence, I would rather never use PASSWORD_DEFAULT.  See my comment below about naming of the hashing algorithm which is also related to this.

vboctor:
@dregad I replied to your comment about LDAP check positioning.

core/authentication_api.php
dregad: OK, that makes sense

core/authentication_api.php
dregad: Tracked in https://mantisbt.org/bugs/view.php?id=22840

core/authentication_api.php
dregad: Tracked in https://mantisbt.org/bugs/view.php?id=22841

dregad:
@vboctor I just pushed updated code which should address your comments. I have not fully tested it yet.

admin/check/check_crypto_inc.php
```diff
@@ -52,8 +52,8 @@
 
 if( config_get_global( 'login_method' ) != LDAP ) {
 	check_print_test_warn_row(
-		'login_method is set to MD5',
-		config_get_global( 'login_method' ) == MD5,
-		'MD5 password encryption is currently the strongest password storage method supported by MantisBT.'
+		'login_method is set to SAFE_HASH',
```
vboctor: Should we replace all `SAFE_HASH` to `HASH_BCRYPT`?  I suggest find across all files since there are other occurrences in other files.

core/authentication_api.php
```diff
@@ -764,30 +790,35 @@ function auth_does_password_match( $p_user_id, $p_test_password ) {
...
 		case CRYPT:
-
 			# a null salt is the same as no salt, which causes a salt to be generated
 			# otherwise, use the salt given
 			$t_processed_password = crypt( $p_password, $p_salt );
 			break;
 		case MD5:
 			$t_processed_password = md5( $p_password );
 			break;
+		case HASH_BCRYPT:
```
vboctor: How about either also adding `LOGIN_METHOD_HASH_MD5`, `LOGIN_METHOD_HASH_CRYPT`, `LOGIN_METHOD_HASH_CRYPT_FULL_SALT` and using them in the code, though keeping `MD5` and `CRYPT` and `CRYPT_FULL_SALT` for backward compatibility?  Or we can just rename this constant to BCRYPT.

core/user_api.php
```diff
@@ -1588,31 +1588,37 @@ function user_set_default_project( $p_user_id, $p_project_id ) {
...
-	# When the password is changed, invalidate the cookie to expire sessions that
-	# may be active on all browsers.
-	$c_cookie_string = auth_generate_unique_cookie_string();
-	# Delete token for password activation if there is any
-	token_delete( TOKEN_ACCOUNT_ACTIVATION, $p_user_id );
+	$t_query = 'UPDATE {user} SET password=' . db_param();
+	$t_param = array( auth_process_plain_password( $p_password ) );
 
-	$c_password = auth_process_plain_password( $p_password );
+	if( $p_hash_update ) {
```
vboctor: !$p_hash_update

admin/check/check_crypto_inc.php
dregad: Thanks for the heads up, I missed those as I only looked at the code and omitted the comments.

Note that I'm still working on updating the admin checks, that will be covered with an upcoming commit.

core/authentication_api.php
dregad: My initial idea was to distinguish "new" hashing methods vs legacy ones. Adding a prefix to all the constants makes sense though.

core/user_api.php
dregad: Fixed, thanks

core/authentication_api.php
dregad: Fixed

atrol:
Maybe we should change the password in the following schema step to prevent any issue we might get when trying to update the hash on first login.
```php
$g_upgrade[51] = array( 'InsertData', array( db_get_table( 'user' ), "(
		username, realname, email, password,
		date_created, last_visit, enabled, protected, access_level,
		login_count, lost_password_request_count, failed_login_count,
		cookie_string
	)
	VALUES (
		'administrator', '', 'root@localhost', '63a9f0ea7bb98050796b649e85481845',
		$t_timestamp, $t_timestamp, '1', '0', 90,
		3, 0, 0, '"
		. md5( mt_rand( 0, mt_getrandmax() ) + mt_rand( 0, mt_getrandmax() ) ) . md5( time() )
		. "'
	)" ) );
```

dregad:
> prevent any issue we might get when trying to update the hash on first login of a fresh install

I'm not really sure what issues you are thinking about here, but in any case I think it should be safe to  be consistent and update the default administrator password's hash from md5 to bcrypt.

atrol:
> I'm not really sure what issues you are thinking about here

Nothing special, just the fact that code which is not executed can't cause a problem.

dregad:
For the record, I am still working on this, I am not satisfied with the current handling of the password upgrade mechanism and would like to implement automated testing too.

Magissia:
Was these changes merged for release yet ?

Magissia:
Could we merge this as current hashing is clearly lacking security ?

dregad:
@Magissia it is not mergeable in its current form, and I have not gotten around to fixing the issues I have found in password upgrade logic. Too many things to do, and not enough time, unfortunately.

limesurvey-translations:
By todays standards a plain MD5 hashing would be considered to be a major vulnerability.
If a break-in happens, most passwords could be easily decrypted with rainbow tables.
When is this finally going to be fixed?

dododedodonl:
Any updates regarding the acceptance of this (or a similair) PR?

dregad:
Just to let everyone know that this is still on my todo list; unfortunately real life got in the way, and I am having a hard time investing time on open-source development these days so I'm no able to commit on a date for fixing this issue.

seanm:
Seems this bug got some press today:

https://it.slashdot.org/story/19/06/17/182208/a-quarter-of-major-cmss-use-outdated-md5-as-the-default-password-hashing-scheme

https://www.zdnet.com/article/a-quarter-of-major-cmss-use-outdated-md5-as-the-default-password-hashing-scheme/

kaizen1:
So is Mantis still using ancient MD5 hash?
Major security hole as everyone knows.

istiak101:
> So is Mantis still using ancient MD5 hash? Major security hole as everyone knows.

Yes.

l00ptr:
any news about that issue ? is it possible to sponsor the integration of it ?

PR #1050, opened 2017-03-12
Markdown: Moved processBugAndNoteLinks() method after the Markdown process the text`

jllano:
MantisBT formatting plugin converts #bugID into bug view links.
At that point the text turned into hyperlinks which should not be,
particularly within the code block.

Code blocks should appear exactly as it is as. Therefore wee need to change
the logic in string_process_bug_link() to only process issue references outside the code blocks.

Fixes #22320


jllano:
@dregad 

It seems like we need to alter the regex here in `core/string_api.php` string_process_bug_link()

screen shot 2017-03-16 at 9 19 22 am

The idea is to only process references if #bugID is outside the code blocks elements.

Let me know you approach first?

dregad:
> It seems like we need to alter the regex here in core/string_api.php string_process_bug_link()

Markdown being part of a plugin that may be uninstalled and a feature that can be turned off, I think it is conceptually wrong to move logic for markdown-specific text processing into core's string API

vboctor:
> Markdown being part of a plugin that may be uninstalled and a feature that can be turned off, I think it is conceptually wrong to move logic for markdown-specific text processing into core's string API

Note that `string_process_bug_link()` is a method that is not used within Mantis core code and is there because it wasn't moved into the CoreFormatting plugin when it was created.  So maybe we should move functions that are left-overs from when such processing was part of core into the CoreFormatting plugin.

vboctor:
The following needs to be revised in `source-integration` plugin in `Source.ViewAPI.php`

```
	# The commit message is manually transformed (adding href, bug and bugnote
	# links + nl2br) instead of calling string_display_links(), which avoids
	# unwanted html tags processing by the MantisCoreFormatting plugin.
	# Rationale: commit messages being plain text, any html they may contain
	# should not be considered as formatting and must be displayed as-is.
	echo string_nl2br(
			string_process_bugnote_link(
				string_process_bug_link(
					string_insert_hrefs(
						string_html_specialchars( $t_changeset->message )
		) ) ) );
```

dregad:
I'm not sure about the approach. What alternative would you propose to avoid regressions on the source integration plugin's requirements ? It needs to process bug and bugnote links and URLs but exclude other processing (string_restore_valid_html_tags)

vboctor:
Closing due to lack of activity.

PR #1053, opened 2017-03-13
Correct resolution comparison when changing issue status

atrol:
Fixes #22468

### DIFF
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -83,6 +83,7 @@ $f_change_type = gpc_get_string( 'change_type', BUG_UPDATE_TYPE_CHANGE_STATUS );
 $t_reopen = config_get( 'bug_reopen_status', null, null, $t_bug->project_id );
 $t_resolved = config_get( 'bug_resolved_status_threshold', null, null, $t_bug->project_id );
 $t_closed = config_get( 'bug_closed_status_threshold', null, null, $t_bug->project_id );
+$t_resolution_fixed = config_get( 'bug_resolution_fixed_threshold', null, null, $t_bug->project_id );
 $t_current_user_id = auth_get_current_user_id();
 
 # Ensure user has proper access level before proceeding
@@ -158,9 +159,9 @@ print_recently_visited();
 
 <?php
 	$t_current_resolution = $t_bug->resolution;
-	$t_bug_is_open = $t_current_resolution < $t_resolved;
+	$t_bug_resolution_is_fixed = $t_current_resolution >= $t_resolution_fixed;
 
-	if( $f_new_status >= $t_resolved && ( $f_new_status < $t_closed || $t_bug_is_open ) ) {
+	if( $f_new_status >= $t_resolved && ( $f_new_status < $t_closed || !$t_bug_resolution_is_fixed ) ) {
 ?>
 	<!-- Resolution -->
 	<div class="field-container">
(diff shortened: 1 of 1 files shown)


bug_change_status_page.php
```diff
@@ -158,7 +159,7 @@
 
 <?php
 	$t_current_resolution = $t_bug->resolution;
-	$t_bug_is_open = $t_current_resolution < $t_resolved;
+	$t_bug_is_open = $t_current_resolution < $t_resolution_fixed;
```
dregad: _$t_bug_is_open_ is somewhat confusing, as the notion of "open" usually refers to the issue's _status_ (i.e. open == status < bug_resolved_status_threshold)

I suggest renaming _$t_bug_is_open_ to **$t_bug_resolution_is_fixed** or something similar, to avoid confusion.

Furthermore, considering the default value for $g_resolution_enum_string, the `<` operator could be incorrect, since fixed = 20, but I would also consider at least 30:reopened (and 80:suspended as well) as "open", so my expectation would be that closing a _reopened_ issue would mark it as _fixed_. Although you could argue that this is a separate issue.

atrol: > I suggest renaming $t_bug_is_open to $t_bug_resolution_is_fixed

Done

> so my expectation would be that closing a reopened issue would mark it as fixed.

Right, this would be good, unfortunately we have
```
$g_resolution_enum_string = '10:open,20:fixed,30:reopened,40:unable to duplicate,50:not fixable,60:duplicate,70:not a bug,80:suspended,90:wont fix';
```
instead of
```
$g_resolution_enum_string = '10:open,20:reopened,30:suspended,40:fixed,50:unable to duplicate,60:not fixable,70:duplicate,80:not a bug,90:wont fix';
```

Furthermore we have
```
/**
 * Threshold resolution which denotes that a bug has been resolved and
 * successfully fixed by developers. Resolutions above this threshold
 * and below $g_bug_resolution_not_fixed_threshold are considered to be
 * resolved successfully.
 * @global integer $g_bug_resolution_fixed_threshold
 */
$g_bug_resolution_fixed_threshold = FIXED;

/**
 * Threshold resolution which denotes that a bug has been resolved without
 * being successfully fixed by developers. Resolutions above this
 * threshold are considered to be resolved in an unsuccessful way.
 * @global integer $g_bug_resolution_not_fixed_threshold
 */
$g_bug_resolution_not_fixed_threshold = UNABLE_TO_DUPLICATE;
```

I think that documentation of $g_bug_resolution_fixed_threshold is wrong.
```
Resolutions above this threshold
```
 
I think it should be
 ```
Resolutions above and including this threshold
 ```
Do you agree?
 
I don't see how we could have a clean solution without 
- reordering resolutions (which would mean data migration) or
- being able to configure individual resolutions as fixed/not fixed instead of the two threshold options.
 
I think this is beyond the scope of this PR, which is just to prevent strange resolution changes when closing issues, caused by comparing status values with resolution values.

dregad: > I think that documentation of $g_bug_resolution_fixed_threshold is wrong.
> ```
> Resolutions above this threshold
> ```
>  
> I think it should be
>  ```
> Resolutions above and including this threshold
>  ```
> Do you agree?

Yes.

> I don't see how we could have a clean solution without 
> - reordering resolutions (which would mean data migration) or
> - being able to configure individual resolutions as fixed/not fixed instead of the two threshold options.
>  
> I think this is beyond the scope of this PR, which is just to prevent strange resolution changes when closing issues, caused by comparing status values with resolution values.

Agreed. I vaguely remember discussing this before... Do you know if there is already an issue to track this (I didn't check) ?

atrol: >  I vaguely remember discussing this before

Found it: https://www.mantisbt.org/bugs/view.php?id=15653#c37182
Hard to believe, I thought we discussed it maybe one year ago, but it's been more than three and a half years ...

PR #1055, opened 2017-03-15
Refer #22423 added ID attribute to textarea in bugnote edit page

libregeek:
@syncguru I had an alignment issue when I enabled SimpleMDE in bugnote edit page. The editor toolbar and the contents in the note are center aligned instead of left.
selection_528

On analysis, I found that the **class="center"** in the following code is causing the problem. 
````
<tr>
    <td class="center" colspan="2">
        <textarea class="form-control <?php echo $t_bugnote_class; ?>" cols="80" rows="10" name="bugnote_text" 
        id="bugnote_text"><?php echo $t_bugnote_text ?></textarea>
    </td>
</tr>
````
I'm not an expert in UI and not sure about the impact. It would be really nice if you could provide your expert opinion. Here is the screenshot without the class. 
selection_529

### DIFF
--- a/bugnote_edit_page.php
+++ b/bugnote_edit_page.php
@@ -130,8 +130,8 @@ $t_bugnote_class = bugnote_get_field( $f_bugnote_id, 'view_state' ) == VS_PUBLIC
 <table class="table table-bordered table-condensed table-striped">
 <tr>
 	<td class="center" colspan="2">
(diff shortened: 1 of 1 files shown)


syncguru:
@libregeek your analysis is correct. Removing `class="center"` should fix it. Go for it.

PR #1057, opened 2017-03-20
Show images inline instead of downloading them

vboctor:
Fixes [#22543](http://www.mantisbt.org/bugs/view.php?id=22543)

### DIFF
--- a/file_download.php
+++ b/file_download.php
@@ -196,7 +196,7 @@ $t_show_inline = $f_show_inline;
 $t_mime_force_inline = array(
 	'image/jpeg', 'image/gif', 'image/tiff', 'image/bmp', 'image/svg+xml', 'image/png',
 	'application/pdf' );
-$t_mime_force_attachment = array( 'application/x-shockwave-flash' );
+$t_mime_force_attachment = array( 'application/x-shockwave-flash', 'text/html' );
 
 $t_mime_type = substr( $t_content_type, 0, strpos( $t_content_type, ';' ) );
 
--- a/file_download.php
+++ b/file_download.php
@@ -190,12 +190,23 @@ if( $t_content_type_override ) {
 	$t_content_type = $t_content_type_override;
 }
 
-# Don't allow inline flash
-if( false !== strpos( $t_content_type, 'application/x-shockwave-flash' ) ) {
-	http_content_disposition_header( $t_filename );
-} else {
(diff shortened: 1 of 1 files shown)


vboctor:
Just to clarify this PR handles the case when user clicks on the image by opening it in the browser rather than downloading it.

atrol:
Just hints, no time for a deeper look or tests.

Did you have a look at function http_content_disposition_header in core/http_api.php ?
It's called just a few lines above your changes.

- it might be the better place to implement the change
- it seems there was a need for browser dependant code (implemented in 2009) `if( is_browser_internet_explorer() || is_browser_chrome() )`. You might need similar code or we might be able to remove the code.

file_download.php
```diff
@@ -197,6 +197,17 @@
...
 }
 
+# Open image in the browser
+# https://www.thoughtco.com/mime-types-by-content-type-3469108
+if( stripos( $t_content_type, 'image/jpeg' ) !== false ||
+	stripos( $t_content_type, 'image/gif' ) != false ||
+	stripos( $t_content_type, 'image/tiff' ) != false ||
+	stripos( $t_content_type, 'image/bmp' ) != false ||
+	stripos( $t_content_type, 'image/svg+xml' ) != false ||
+	stripos( $t_content_type, 'image/png' ) != false ) {
```
dregad: Instead of potentially calling stripos 6 times, I would suggest to use a single preg_match call instead, something like

```
~image\/(?:jpeg|gif|tiff|bmp|svg\+xml|png)~
```

Note that I didn't actually test the regex.

DemiMarie:
Does this allow a CRLF injection attack by passing a file with a newline in the filename?

vboctor:
@atrol Changed to use the appropriate API.
@dregad I've used in_array() instead of a series of stripes.
@DemiMarie The latest code now re-uses existing code for generating the header.

vboctor:
Few other additions:

- @atrol The security issue you referred to in the issue doesn't apply here since I check the calculated mime type rather than the extension of the file.
- HTML mime content are forced not to show up inline.
- PDF files are also now opened inline.

file_download.php
vboctor: I use in_array() now.  I would rather only use regular expressions when I have to. For the sake of readability and less chance of messing them up as we update the later.

PR #1058, opened 2017-03-20
Fixed #22423 Added ID attribute to bugnote textarea in note edit page

libregeek:
Cleaned up version of #1055


libregeek:
@dregad I separated the commits. Please review. Thanks for sharing the git guidelines for Mantis.

dregad:
Thank you. 

For the future, I strongly recommend to always create a **dedicated feature branch** instead of using _master_ as the pull request's source. This way if you later update _master_ for any reason, it will not impact the pull request.

I have logged [#22548](http://www.mantisbt.org/bugs/view.php?id=22548) to track the center class thing.

I'll force update the branch now, to remove the original commit and revert.

libregeek:
Thanks for accepting and reviewing the PR. This is great step to have an editor for Mantis.

PR #1059, opened 2017-03-21
Fix filter_gpc_get custom fields filter properties

cproensa:
Fix issue where filter_gpc_get() parses custom fields properties and
overwrites existing ones in the input filter with empty when those
properties are not present as gpc parameters.
This function, when called with an existing filter, should overwrite
only those properties thare are present as request parameters.

Fixes: #22392

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -3212,10 +3212,16 @@ function filter_gpc_get( array $p_filter = null ) {
 
 	# custom field updates
 	$t_custom_fields 		= custom_field_get_ids(); # @todo (thraxisp) This should really be the linked ids, but we don't know the project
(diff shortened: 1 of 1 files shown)


vboctor:
Thanks @cproensa for the fix.  Does this also fix 22555?  I suggest we target the fix to `master-2.2` branch.

cproensa:
>  Does this also fix 22555

see my comment there: probably, yes
unless there is another issue related to name-casing

cproensa:
> Does this also fix 22555

no, after reviewing, 22555 is a different issue.

cproensa:
>  I suggest we target the fix to master-2.2 branch

changed

PR #1060, opened 2017-03-22
Add note_by to permalink url parse

cproensa:
Permalink url parse did not read parameter for note-by.
Previously it worked as a side effect of reading gpc vars at
filter_ensure_valid, which is not done any more.

Fixes: #22496


atrol:
👍

cproensa:
> I wonder if this logic should re-use filter_gpc_get().

permalink logic has not been rewritten so far.
The syntax is somewhat different than filter form parameters, so filter_gpc_get() is not directly reusable right now

vboctor:
@cproensa I suggest opening a code clean bug for the work to re-write permalink to re-use filter_gpc_get().

vboctor:
@cproensa This doesn't merge cleaning on `master` or `master-2.2`.  Can you target latest `master-2.2`.

vboctor:
Merged into `master-2.2` via 7951aa22e4ac1b839265abe8336bf070d5e4a8f8

PR #1061, opened 2017-03-22
Validate filter non-array values

cproensa:
Validate filter values that must not be arrays    
Add validation for values thus must be a single value.
Clean up type validation for both single and multiple values.

In view_all_set, make sure current filter is valid after load.    
After loading the filter array, make sure it's valid filter, to avoid
errors due to inconsistent data or outdated version.

Fixes: #22566

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -586,8 +586,45 @@ function filter_ensure_valid_filter( array $p_filter_arr ) {
 		$p_filter_arr[FILTER_PROPERTY_SORT_DIRECTION] = filter_get_default_property( FILTER_PROPERTY_SORT_DIRECTION, $t_view_type );
 	}
 
-	# validate or filter junk from other fields
-	$t_multi_select_list = array(
+	# Validate types for values.
+
+	# helper function to validate types
+	$t_function_validate_type = function( $p_value, $p_type ) {
+		$t_value = stripslashes( $p_value );
+		if( ( $t_value === 'any' ) || ( $t_value === '[any]' ) ) {
+			$t_value = META_FILTER_ANY;
+		}
+		if( ( $t_value === 'none' ) || ( $t_value === '[none]' ) ) {
+			$t_value = META_FILTER_NONE;
+		}
+		# Ensure the filter property has the right type - see #20087
+		switch( $p_type ) {
+			case 'string' :
+			case 'int' :
+				settype( $t_value, $p_type );
+				break;
+		}
+		return $t_value;
+	};
+
+	# Validate properties that must not be arrays
+	$t_single_value_list = array(
+		FILTER_PROPERTY_VIEW_STATE => 'int',
+	);
+	foreach( $t_single_value_list as $t_field_name => $t_field_type ) {
+		$t_value = $p_filter_arr[$t_field_name];
+		if( is_array( $t_value ) ) {
+			if( count( $t_value ) > 0 ) {
+				$p_filter_arr[$t_field_name] = reset( $t_value );
+			} else {
+				$p_filter_arr[$t_field_name] = filter_get_default_property( $t_field_name, $t_view_type );
+			}
+		}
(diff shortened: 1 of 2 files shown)


dregad:
In > Add validation for values thus must be a single value.

thus --> that ?
Change commit message if needed.

cproensa:
> Change commit message if needed.

thanks, changed

core/filter_api.php
```diff
@@ -586,8 +586,43 @@ function filter_ensure_valid_filter( array $p_filter_arr ) {
...
+		}
+		# Ensure the filter property has the right type - see #20087
+		switch( $p_type ) {
+			case 'string' :
+			case 'int' :
+				settype( $t_value, $p_type );
+				break;
+		}
+		return $t_value;
+	};
```
dregad: Is there any particular reason for using an anonymous function for this ?

core/filter_api.php
```diff
@@ -586,8 +586,43 @@ function filter_ensure_valid_filter( array $p_filter_arr ) {
...
+	$t_function_validate_type = function( $p_value, $p_type ) {
+		$t_value = stripslashes( $p_value );
+		if( ( $t_value === 'any' ) || ( $t_value === '[any]' ) ) {
+			$t_value = META_FILTER_ANY;
+		}
+		if( ( $t_value === 'none' ) || ( $t_value === '[none]' ) ) {
+			$t_value = META_FILTER_NONE;
+		}
+		# Ensure the filter property has the right type - see #20087
+		switch( $p_type ) {
```
vboctor: What should happen if passed type is not string/int?  Error out? Default to string?

core/filter_api.php
```diff
@@ -586,8 +586,43 @@ function filter_ensure_valid_filter( array $p_filter_arr ) {
...
+	# Validate properties that must not be arrays
+	$t_single_value_list = array(
+		FILTER_PROPERTY_VIEW_STATE => 'int',
+	);
+	foreach( $t_single_value_list as $t_field_name => $t_field_type ) {
+		$t_value = $p_filter_arr[$t_field_name];
+		if( is_array( $t_value ) && count( $t_value ) > 0 ) {
+			$p_filter_arr[$t_field_name] = reset( $t_value );
+		} else {
+			$p_filter_arr[$t_field_name] = filter_get_default_property( $t_field_name, $t_view_type );
```
vboctor: Do we want this line to run if the value is not an array?  Or is this just when the value is not set?

view_all_set.php
```diff
@@ -122,6 +122,7 @@
 			exit; # stop here
 		}
 	}
+	$t_setting_arr = filter_ensure_valid_filter( $t_setting_arr );
```
vboctor: Should we ensure value in the else for `if( false === $t_setting_arr )`?  In this place, if it is value, calling the ensure valid header doesn't make sense.

core/filter_api.php
cproensa: Only reason is to not cramp filter_api with more public functions, if this code is not needed elsewhere...

core/filter_api.php
cproensa: Current behaviour is that if passed type is not string/int, value is kept as is. I didnt change that, unless mistaken.

core/filter_api.php
cproensa: there's an error there. It should be:
- If value is array AND count>0, use first element
- If value is array AND count=0, use default
- If value is not array, it's ok, do nothing

view_all_set.php
cproensa: yes, an else branch there makes more sense.
as it is now, it would create an empty filter, that would not be used anyway.

cproensa:
updated with corrections to @vboctor findings. thanks

core/filter_api.php
vboctor: ok

vboctor:
Merged into `master-2.2` via 449f4d513a1190ae3ebab4629ab94675670aecb3

PR #1062, opened 2017-03-22
Upload(db): Don't encode content for db-mssqlnative.

obmsch:
Downloads are broken and inline preview doesn't work.

Move "case 'mssqlnative':" down next to 'oci8' in
db_prepare_binary_string (database_api) to effectively
return the string unchanged via 'default:'. Adjust comment.

Fixes #22208

### DIFF
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -743,7 +743,6 @@ function db_prepare_binary_string( $p_string ) {
 
 	switch( $t_db_type ) {
 		case 'mssql':
-		case 'mssqlnative':
 		case 'odbc_mssql':
 		case 'ado_mssql':
 			$t_content = unpack( 'H*hex', $p_string );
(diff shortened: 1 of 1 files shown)


core/database_api.php
```diff
@@ -743,7 +743,6 @@ function db_prepare_binary_string( $p_string ) {
 
 	switch( $t_db_type ) {
 		case 'mssql':
-		case 'mssqlnative':
```
dregad: To make things more obvious, I would move the _mssqlnative_ case before the _oci8_ one instead of removing it, and update the fall-through comment accordingly.

dregad:
Fixed in 43c4b6fb806dea867ed7a7c16f9dc85b0311d2da (1.3.x) and 4a06c6e803e3bc9da326b691db8d352a9862a60e (2.2)

PR #1063, opened 2017-03-22
Modify schema to get install/upgrade work with db-mssql.

obmsch:
Although the problem appears in step 209, the real cause is the combination of
steps 200 and 201. The Index ('user_id','name') created in step 201 (with a nullable
'user_id' as of step 200) forbids the alteration of 'user_id' to NOTNULL in step 209,
at least with mssql. So fix this in step 200 and set 'user_id' NOTNULL right from start.

While this is a 'post release' change, it is justifiable as it happens in the same
install/upgrade sequence and the final db (schema) is identical.

Fixes #22063

### DIFF
--- a/admin/schema.php
+++ b/admin/schema.php
@@ -846,7 +846,7 @@ $g_upgrade[199] = null;
 #
 $g_upgrade[200] = array('CreateTableSQL', array( db_get_table( 'api_token' ), "
 	id						I		UNSIGNED NOTNULL PRIMARY AUTOINCREMENT,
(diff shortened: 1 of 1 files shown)


dregad:
I think that should work, without introducing issues for already installed databases, but I haven't had time to fully analyze or test it yet.

Please shorten commit message's 1st line and wrap text at 72 chars (see http://www.mantisbt.org/wiki/doku.php/mantisbt:git_commit_messages)

PR #1066, opened 2017-03-24
Decouple widget colours css class names

cproensa:
Rename current widget-color-blue2 css class to a more generic class
name.
Move widget color definitions to ace-mantis.css for easier
customization.

Fixes: #22575


vboctor:
Note that changing style names will break plugins.  So if we consider this, we are better off doing it as one change rather than piece meal.

cproensa:
outdated & rejected

PR #1067, opened 2017-03-25
Move assigned projects outside of form

cproensa:
Move the list of assigned projects outside of the form in account_page.
This list is not editable, so it must not be part of the form.

Fixes: #21552

seleccion_151

### DIFF
--- a/account_page.php
+++ b/account_page.php
@@ -289,27 +289,28 @@ if( !empty( $t_projects ) ) {
 		<div class="widget-body">
 			<div class="widget-main no-padding">
 				<div class="table-responsive">
-					<table class="table table-bordered table-condensed table-striped">
+					<table class="table table-striped table-bordered table-condensed table-hover">
+						<thead>
+							<tr>
+								<th><?php echo lang_get( 'name' ) ?></th>
+								<th><?php echo lang_get( 'access_level' ) ?></th>
+								<th><?php echo lang_get( 'view_status' ) ?></th>
+								<th><?php echo lang_get( 'description' ) ?></th>
+							</tr>
+						</thead>
 						<?php
+						foreach( $t_projects as $t_project_id => $t_project ) {
+							$t_project_name = string_attribute( $t_project['name'] );
+							$t_access_level = get_enum_element( 'access_levels', $t_project['access_level'] );
+							$t_view_state = get_enum_element( 'project_view_state', $t_project['view_state'] );
+							$t_description = string_display_links( project_get_field( $t_project_id, 'description' ) );
 							echo '<tr>';
-							echo '<td class="category">' . lang_get( 'assigned_projects' ) . '</td>';
-							echo '<td>';
-							foreach( $t_projects AS $t_project_id => $t_project ) {
-								$t_project_name = string_attribute( $t_project['name'] );
-								$t_view_state = $t_project['view_state'];
-								$t_access_level = $t_project['access_level'];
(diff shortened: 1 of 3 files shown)


account_page.php
```diff
@@ -294,6 +273,48 @@
...
+<?php
+$t_projects = user_get_assigned_projects( auth_get_current_user_id() );
+if( count( $t_projects ) > 0 ) {
+?>
+	<div class="space-10"></div>
+
+	<div class="widget-box widget-color-blue2">
+		<div class="widget-header widget-header-small">
+			<h4 class="widget-title lighter">
+				<i class="ace-icon fa fa-user"></i>
```
syncguru: Should we use another icon for this section? Maybe the same one we use in manage projects page: http://fontawesome.io/icon/puzzle-piece/

cproensa:
> Should we use another icon for this section? Maybe the same one we use in manage projects

yes, makes sense.
I updated my-account page and manage-user page.

vboctor:
@cproensa 

- How is the assigned projects field different from other read-only fields on the form?
- Does this change also impact the manage user edit page?
- Maybe it would be good to align the table columns for the new table with the one above it for user information.

cproensa:
> How is the assigned projects field different from other read-only fields on the form?

The main issue is that this item is variable in size. The form button can be outside screen when a long enough list of project exists.
By moving this, the form remains in constant size.

> Does this change also impact the manage user edit page?

Only change is the widget icon as suggested by Rafik

> Maybe it would be good to align the table columns for the new table with the one above it for user information.

Afaik, we dont have any standards yet on managing columns sizes in this kind of table-forms. That would be a good point to be worked on, since I feel that on wide screens the widgets and spacings can get too wide-spread.

account_page.php
```diff
@@ -294,6 +273,48 @@
 	</div>
 </div>
 
+<?php
+$t_projects = user_get_assigned_projects( auth_get_current_user_id() );
+if( count( $t_projects ) > 0 ) {
```
vboctor: !empty() is more efficient

account_page.php
```diff
@@ -294,6 +273,48 @@
...
+		</div>
+		<div class="widget-body">
+			<div class="widget-main no-padding">
+				<div class="table-responsive">
+					<table class="table table-bordered table-condensed table-striped">
+						<?php
+							echo '<tr>';
+							echo '<td class="category">' . lang_get( 'assigned_projects' ) . '</td>';
+							echo '<td>';
+							foreach( $t_projects AS $t_project_id=>$t_project ) {
```
vboctor: Add spacing around `=>`

vboctor:
We can have re-usable code that prints out a section with a set of projects.  It would take the section title and array of projects/project ids as input.

cproensa:
seleccion_154

by the way, updated header styles for manage projects table

seleccion_155

> We can have re-usable code that prints out a section with a set of projects

Where would you reuse this section, as is?

vboctor:
> Where would you reuse this section, as is?

I was thinking we can potentially re-use for account page, manage projects page and manage user page.

cproensa:
>  re-use for account page, manage projects page and manage user page

in my opinion, won't fit...
- manage project page is more complex, as it features sorting. And doesnt show access level (does not make sense there)
- manage user page is close, but there the table is actually a form.

Anyway, the code for this table is very concise. Not a big problem to keep it ad-hoc.

PR #1069, opened 2017-03-25
Fix markdown with Indented-code-block

jllano:
Fixes #22545

BEFORE:
screen shot 2017-03-26 at 6 49 03 am

AFTER:
screen shot 2017-03-26 at 6 49 28 am


jllano:
I tried  but didn't fix it though.

PR #1070, opened 2017-03-26
Support for auth plugins

vboctor:
This PR adds support for authentication plugins.  The approach used here is:

- Use paginated logins like Microsoft and Google.  First page is username and second page is password.
- Remove all hard-coded assumptions in the code about authentication page and delegate to authentication_api.php.
- Enable plugins to take over the whole flow or replace the password page and login validation.

There can be multiple plugins and they can activate based on the user attempting to logins.  They get a chance to handle users whether or not they are registered in the MantisBT database, so plugins can validate credentials and auto-provision if necessary.

I expect that in the future, we can also add more auth plugin events that enable more specific behaviors that can fit into standard auth flow like:
- Password check (e.g. against a database) or LDAP while re-using the password page.
- Password strength validation.  This is more work, since it is currently not supported by core.

Here are the new screenshots:

**The login username page:**

native_login_form_for_username

**The credentials page:**

native_credentials_page

**Message that password can't be managed within MantisBT:**

sample_auth_no_password_change

See [SampleAuth](https://github.com/mantisbt-plugins/SampleAuth) plugin for an example and some extra documentation.

Fixes #4235


vboctor:
Here is a sample authentication plugin with further documentation:
https://github.com/mantisbt-plugins/SampleAuth

cproensa:
So, if a plugin is registered, does this replace authentication method for all users?
A feature that i'd really appreciate is:
- plugins registering an authentication method
- Each user is associated to one oh those authentication methods.

This way we could have, for example, a set of users that login with LDAP server, others throut some OAUTH service, and other through mantis native passwords...
Probably your refactoring here is a good start to achieve that feature.

vboctor:
@cproensa thanks for your feedback.

> if a plugin is registered, does this replace authentication method for all users?  ... This way we could have, for example, a set of users that login with LDAP server, others throut some OAUTH service, and other through mantis native passwords.

There are common models that I expect to happen:

- Team members using some kind of directory to authenticate (e.g. LDAP / SAML) where users/customers using native authentication.  Team members have higher access and it is required that they are not able to login via native method to block them once they leave the company.
- A single auth method against some directory for internal only MantisBT instance.

The work done so far enables building such plugins giving control to plugin to control native behaviors specially based on access levels rather than specific user ids.  However, we can imagine enabling based on user ids in the future.

> Probably your refactoring here is a good start to achieve that feature.

Yes, the refactoring removes a lot of hard-coded assumptions about authentication and enables new auth methods without having to change core code.  Before then it wasn't possible to do so, specially if it is required to block standard authentication at least for a set of users.  We can also add new authentication events that involves more targets changes to native authentication (e.g. lookup passwords elsewhere in another DB or LDAP.

>  a set of users that login with LDAP server, others throut some OAUTH service, and other through mantis native passwords

As a start there is support for a single auth plugin + native capabilities.  However, I'm thinking of adding support for multiple plugins to be active at the same moment replacing some of the native ones or adding new customer ones.  For example, LDAP/Anonymous can be easily moved out to plugins.  I expect to keep working on enriching this to enable more capabilities.  But this PR focuses so far on full control + falling back to native for a subset of the users.

atrol:
Cannot try the branch, I get
PHP Fatal error:  Uncaught Error: Call to a member function getAnonymousEnabled() on array in .../core/authentication_api.php:142

vboctor:
@atrol Issue fixed.  This was introduced with a later change where I moved to using `AuthFlags` class.

login_page.php
```diff
@@ -104,11 +104,18 @@
 	}
 }
 
+# Redirect to plugin login page if applicable.
+if( !auth_can_use_standard_login( NO_USER ) ) {
```
cproensa: auth_can_use_standard_login receives a user_id parameter. Is this call right?
And, regarding the "can_use_standard_login", which currently is defined as a threshold, how do you know the user before it has logged?

cproensa:
> But this PR focuses so far on full control + falling back to native for a subset of the users.

So far i'm thinking that some of the features that are delegated to the plugin should reside in core.
I'm looking with the point of view of having this extended to support a set of authentication plugins that are applied individually to users.
Additionally, we should consider having native password authentication to be implemented as any other authentication method, by hooking on this new abstraction.

In that scenario:
- We have "users", that are linked to "login" methods. So the passwords that are now in mantis_user_table would be managed by the PASSWORD authentication plugin.
- One "user" could have more than one "login" methods associated. With the following functional descriptions, having more than one method is feasible.

A plugin should provide (some may be optional)
- Capability to show a login page, like implemented in this PR
- Capability to show manage pages. For example, managing the passwords for the related user, or OUs for LDAP plugin, etc
- Capability to show a self registration page. If supported by the method.
- Flags (like implmented in this PR), that provide information to core about session behaviour: time to live, reauthenticate, permanent, etc

Some behaviours must be controlled centrally in core, like:
- AnonymousEnabled, AnonymousAccount
No need to delegate on plugins. This is a centralized setting. Also, note that AnonymousAccount would indicate a user_id, plugins however would manage "logins" not "users"
- SignupEnabled
 Also a centralized setting. Configure which of the plugins manages self-registration, then delegate on it.

Having thresholds to limit who can use which method, it's not practical. Many scenarios won't match a simple approach as that, and having that defined at user level is the simplest way.
- PasswordThreshold, PasswordManagedExternallyMessage
If considering that PASSWORD authentication is another kind of method, the ability to change a password is provided by the plugin itself. Core does not care (see following workflow)
- CreateApiTokensThreshold. If API-Tokens is a core functionality, this sould be managed in core only.
- UserStandardLoginThreshold. As said, this is not needed as there'd be a per user configuration.

A login workflow could work like this:
- First: a core login page, only asks for user name.
- After entering user name, the plugin specific login page is showed. If several plugins are associated for this user, a list can be shown to choose from.
This is what you experience when logging in to gmail, for example.

A manage workflow:
- Our current manage-user page only manages user name, email, etc, and a list of allowed auth methods for this user. Each method has its own page to manage its credentials, eg: password.

I know this is a bigger scope that this PR, and there's a lot of tricky isses, but it's worth discussing to know what we'd like to have in the future.


Regarding this PR:
I have tested lightly as is (without plugins), and seems to work fine.
I have looked at the sample plugin, works as expected (all the ideas i wrote, however, apply) 

I would expect that this could be merged, once validated, as a refactor for auth apis, so
- We benefit from the auth-apis cleanup
- The plugin functionality is agreed as not-ready for usage, since we can expect breaking changes as we develop the ideas.

dregad:
I don't have time to fully review or test, but I'm in agreement with @cproensa's points.

cproensa:
More comments

I don't like how is defined
`abstract class AuthPlugin extends MantisPlugin`

A plugin cant inherit multiple classes. If this pattern is followed, in the future this can be a conflict for a plugin that provides different activities (which may be related, and unable to break into separate plugins).

Technically, that fits better as interface declaration, where a plugin could then implement multiple "plugin interfaces"
But i think the approach should be different. Based on my experience with plugins, if the need here is to have a better coupling between core and plugin, with need for bidirectional/complex comunication . What i would do:

- `abstract? class AuthManager`: This has the logic/data that would be in abstract AuthPlugin
- the event ask for plugins to return an AuthManager instance, this is equivalent of what you have done with the flags object.
- core can implement it's own AuthManager's, like would happen with built-in password authentication.
- core communicates with each AuthManager in the same way, regardless of being a core or plugin provided.

In this scenario, the plugin itself could inherit AuthManager, or provide a new object as needed; core does not care, as long as it has the correct object interface.

Actually, what i have described is similar to the AuthFlags class, but instead of having only one Authclass created by core, and modified by plugins, we would have one object for each available method, with the relevant properties/functionality for each authentication scenario. (remember, how i said some of those AuthFlags settings should be moved back to core)

cproensa:
Minor suggesition: i would avoid using the "Auth" prefix, as we could mislead by mixing "authentication" with "authorization".

vboctor:
@cproensa thanks for the detailed review.  I have pretty much implemented all your core points.  More can come later and build on top of this PR as needed.

> So far i'm thinking that some of the features that are delegated to the plugin should reside in core.

Removed authentication configuration that doesn't happen within context of a user back to core.  That includes signup and anonymous access.  Removed plugin flags that use an access level threshold.  Now auth flags are retrieved in the context of a specific user that is registered already or that is unknown to the system.

> we should consider having native password authentication to be implemented as any other authentication method, by hooking on this new abstraction.

We should consider this, specially for authentication methods like LDAP and HTTP_AUTH.  However, that is outside the scope of this PR.  The focus here is to enable adding new authentication method or enabling the porting of some authentication methods to plugins.  I think a good goal is to move out any auth methods that doesn't store a hash of the password in our database.

> One "user" could have more than one "login" methods associated. With the following functional descriptions, having more than one method is feasible.

At the moment, it is up to the auth plugin to define what is possible for a specific user.  We can consider in the future providing a hook enable a secondary authentication model like github/google auth that matches the user based on verified email, but that is possible even before the work in this PR.  What is not possible is to disable the default authentication flows in favor of other models.

> Capability to show a login page, like implemented in this PR

This is still available, but I also implemented paginated login, where first page takes in username / password and second page takes in the password and associated options (remember me, secure session, forgot password).  The plugin can decide to use the first page, then take over the second or take over the whole flow.

> Capability to show manage pages. For example, managing the passwords for the related user, or OUs for LDAP plugin, etc

The plugin can expose its own page with such settings for now, or use existing hooks if any can enable that.  Otherwise, outside the scope of this PR.

> Capability to show a self registration page. If supported by the method.

The page that takes in username provides it to the auth plugin independent of whether it is in the db or not.  Hence, the plugin can decide to fail, auth then fail, auto-provisioning, or collect more info and then auto-provision.

> Flags (like implmented in this PR), that provide information to core about session behaviour: time to live, reauthenticate, permanent, etc

The flags are now in the context of a specific user.  So the plugin can decide which users should get which treatment, or leave some users to use native login.

> AnonymousEnabled, AnonymousAccount, SignupEnabled

These are now only in core and are removed from the AuthFlags class.

> CreateApiTokensThreshold should be in core only

This is now in core only and is only driven by protected users can't create new tokens.  But they can leverage existing ones, assuming it is published somewhere and the user knows it.

> PasswordThreshold, PasswordManagedExternallyMessage - If considering that PASSWORD authentication is another kind of method, the ability to change a password is provided by the plugin itself. Core does not care (see following workflow)

There is now a flag that is specific to a user indicating is they can use standard login or not.  If they can't, then the provided message is displayed to user when appropriate.

> UserStandardLoginThreshold. As said, this is not needed as there'd be a per user configuration.

This was changed to a boolean flag that applies to a specific user.  The can set password flag was also removed in favor of this flag.

> Paginated login

This is implemented and it is consistent with Google and Microsoft new login flows.  It is also better suited for extensible authentication.  I was hesitating to do this initially, but it is much better approach.  Also changed the re-authentication flow to go directly to the page that asks for the credentials as per native or plugin authentication.

> Each method has its own page to manage its credentials, eg: password.

This PR does the work to disable the default ability to manage the native password and it is up to the plugin to provide its own management page.  There message can hyperlink to their page.

> A plugin cant inherit multiple classes

Base class removed.  The plugin just needs now to hooked to EVENT_AUTH_USER_FLAGS and return the flags object or null if they don't want to handle the user. 

*TODO: I need to test how FIRST events handle when first plugin returns null.  Whether they return that or keep going until one returns a non-null value.*

> Actually, what i have described is similar to the AuthFlags class, but instead of having only one Authclass created by core, and modified by plugins, we would have one object for each available method, with the relevant properties/functionality for each authentication scenario. (remember, how i said some of those AuthFlags settings should be moved back to core)

This is done.  Now core only creates AuthFlags if the event signal doesn't return one.  i.e. there were no plugins or plugins decided not to handle auth for that specific user.

> i would avoid using the "Auth" prefix, as we could mislead by mixing "authentication" with "authorization".

Typically 'auth' used for authentication, and that is the case for APIs with `core/authentication_api.php`.  When referring to both authentication and authorization, AuthN and AuthZ are used for short.  Left as is.

vboctor:
@cproensa Also note that SampleAuth plugin has been updated to use standard login method for administrators and password-less login for normal users.

vboctor:
I have verified `event_type_first()` and it keeps going until it gets a non-null value.  So all should work as expected.

vboctor:
I have updated to PR description to reflect the latest approach and provide screenshots.

cproensa:
@vboctor thanks for the work
i'll need some days to get enough time to review the evolution of this.

cproensa:
I have been looking overall the latest changes. Looks good and we are getting closer in our views.
I guess that multiple plugins can be used, since the event is passing the username, and the FIRST to respond with an AuthFlags object wins the authentication responisbility.
This opens the question, how would user be managed? i guess in this scenario, each plugin should provide a way to configure which users are associated to the login method. Or, as you put in doc, be responsible of creating a user in core, and then enable authentication.

I'll explain my idea, which is another turn in this latest approach. Let's see if you find it useful. At least i think i would simplify the flow in some points.

First, i assume that user creation is part of core. By any mean, a user must exist as a registered user id+name+etc in mantis to be able to use the application. So i as an administrator, besides creating a new user, should be able to manage, centrally in user management, what login method a user is allowed (later we'll see self-registration)

I would add a plugin registration step. We dont have a methodology to register persistent external objects in core, so a event is fine. Something like EVENT_AUTH_PROVIDERS. Each auth plugin would reply with some data: METHOD technical name, METHOD verbose name, also an auth provider object (let's call it AuthProvider, may be AuthFlags, or a more generic interface implementation)
So in core initialization, it comples all AuthProviders that are available.
Note that the standard password login can be an AuthProvider itself, registered automatically by core, no need for it to be a standalone plugin.

For example. Upon initialization, core knows there are available providers: PASSWORD (native), LDAP (provided by a plugin)

In user manage page, each user has associated which method can use. For example:
user1: PASSWORD
user2: PASSWORD, LDAP
In the user configuration page, i could modify it based on the available AuthProviders.

We could support only one provider for each user for simplicity, or several, the difference is not much.

Let's go to login.
1) Login page, only asks for user id (like this pr)
- if entered user only has one provider, go to that plugin custom login page, where the plugin takes control of what to ask or do. For example, PASSWORD is the password login page.
- If the user have several providers, you could offer to choose which one to use.
- A shortcut can be made, where only one provider is available systemwide, go directly to that login page. This is the case when there are no plugins installed, then go directly to password login page.

Obviously, if a user has associated a provider that is not registerd in core (eg, unistalled the plugin), this user won't be able to login. In that case, i should update the user configuration.

Technically, one difference in this AuthProvider object to how it's AuthFlags now, is that AuthFlags is created based on a user, and AuthProvider is a general object, the user would be provided later to the object once the auth flow has started, or for each action call.

In this scenario, probably there is more simplicity with other stuff.
For example, core does not care about can-change-password, etc.
All you have is, for each provider associated to the user, a link to the plugin user configuration. For password this page would compile the password reset, etc.
simlarly, on user preferences, a link for a self-service page, like change password.

Regarding self-registration.
I would assume that if i want to enable self-registration, only one auth provider is designated to do so.
Then, the core configuration can be: self_registration_provider='PASSWORD', on none if disabled.
The link in main login page, would then point to the specific plugin registration page.


As i said, overall looks good. The above ideas i think are not very different, onlly to add a bit of high level order on where to put each thing
Anyway, those are not absolute. I havent reviewed all the low level details, so they can also be flawed at some point.
I will try to have more detailed incursions in this pr as time allows.

vboctor:
I had a discussion with @cproensa and we agreed that the latest comments build on top of what is done in this PR.  We agreed that this PR will be merged on Saturday night, and if any regressions are found by then we will address them.

core/authentication_api.php
```diff
@@ -75,6 +75,208 @@
...
+function auth_flags( $p_user_id = null, $p_username = '' ) {
+	if( is_null( $p_user_id ) ) {
+		$t_user_id = auth_get_current_user_id();
+	} else {
+		$t_user_id = (int)$p_user_id;
+	}
+
+	if( !$t_user_id && is_blank( $p_username ) ) {
+		# If user is not in db, must supply the name.
+		trigger_error( GENERIC_ERROR );
```
atrol: This line throws
SYSTEM NOTICE: 'Use of undefined constant GENERIC_ERROR ...

core/authentication_api.php
```diff
@@ -75,6 +75,208 @@
 $g_cache_current_user_id = null;
 
 /**
+ * Gets set of flags for authentication for the specified user.
+ * @param int|null|bool The user id or null for logged in user or NO_USER/false for user that doesn't exist
+ *                 in the system, that may be auto-provisioned.
```
atrol: Missing PHPDoc for parameter $p_username

core/classes/AuthFlags.class.php
```diff
@@ -0,0 +1,400 @@
...
+	/**
+	 * Gets the message to display to the user when they can't manage their password within MantisBT.
+	 *
+	 * @return string The message.
+	 * @see setPasswordManagedExternallyMessage()
+	 */
+	function getPasswordManagedExternallyMessage() {
+		if( empty( $this->password_managed_elsewhere_message ) ) {
+			return lang_get( 'password_managed_elsewhere_message' );
+		}
```
atrol: Missing else? 
return $this->password_managed_elsewhere_message ?

core/classes/AuthFlags.class.php
```diff
@@ -0,0 +1,400 @@
...
+	 * @see getReauthenticationEnabled()
+	 */
+	function setReauthenticationLifetime( $p_seconds ) {
+		$this->reauthentication_expiry = $p_seconds;
+	}
+
+	/**
+	 * Gets the number of seconds to re-authenticate the user after.
+	 *
+	 * @param int $p_seconds The number of seconds to prompt for re-authentication after.
```
atrol: There is no parameter $p_seconds

core/authentication_api.php
vboctor: Fixed.

core/authentication_api.php
vboctor: Fixed `$p_user_id` and `$p_username`.

core/classes/AuthFlags.class.php
vboctor: Fixed.

core/classes/AuthFlags.class.php
vboctor: Fixed

V1pr:
I'm an independent developer and plan to have my own MantisBT and use it to all my customers; making them able to log in, so I tried putting together SampleAuth - authing the users based on their e-mail addressess; used as login username.

However, I'm really not able to figure out, how this should work. If the user is still not authenticated, than how can the plugin now, if the user has administrator privileges?

I really have to build the auth logic INTO pages/login.php? Or I can implement the logic into the class itself, than how can I get an instance of the plugin class in the page (view)? A proper example will really do some good I think. (in place of 'TODO: use custom authentication method) I mean something like this:

`# TODO: use custom authentication method here.
$auth = getAuthInstance();
if ( $auth->login($f_username) ) {  // call the login logic implemented in the class
   $t_user_id = $mantisbt->provision_user($user_data[]);
} else {
  // redirect to login page or return null based on need - what will handle this one?
}`

IMHO, it would be much better to register the login/logout function as a flag or hook (and be able to use the built in login pages ):

`
$t_flags->setLoginHandler($this->login_logic);
`

Also I were not able to figure out, how to auto-provision the user (I've seen that this has been mention a few places though).

I'm open to share my solution once I'm able to put it together based on some guideance.

atrol:
@V1pr have a look at 
https://github.com/mantisbt-plugins/GoogleOauth
https://github.com/mantisbt-plugins/SimpleSAMLphpAuth/tree/initial
https://github.com/mantisbt-plugins/SampleAuth

V1pr:
@atrol 
Thanks, unfortunatelly that does not answer all my questions :(
GoogleOauth is not using the same hook az SampleAuth (I was referring to SampleAuth code in my previous comment) and it's not creating user/it's required to have a user in mantis.

I'll try to dig in a bit into SimpleSAMLphpAuth and SimpleSAML.

V1pr:
This is definetly not working (I'm using Mantis 2.13.1). I've dig into the code a bit more and found some really strange things.

If I open my Mantis login page in a new incognito window and echo the t_user_id in the SampleAuth plugin, it displays '1', which is my admin user's uid; and this make auto-provision impossible.

The SampleAuth plugin sets the CredentialsPage, but not the LoginPage. What was the idea behind this?

IMHO, if I set the CredentialsPage, than a custom "password input" page should be shown (but this is not referenced in the plugin's login.php); however, if I set the LoginPage in the SampleAuth plugin, then that should mean, that I want to use Mantis' login flow, but validate the user/password to a 3rd party backend at the end.
(if I got it right, than the standard flow looks like this: login_page.php (ask for username) -> login_password_page.php (if userid found, than redirect by flag) -> POST to login.php, which validates everything, but I cannot "hook" in it with the SampleAuth plugin)

While I was debuging, I've seen, that the user_id is set to '1' nearly every time, however, the username I gave in to the login screen is non-existent in the system... (I did not had the time to debug where is it set to 1, instead of null or 0, since the user is non-existent)

fmancardi:
@V1pr I'm working on something similar and had this finding: IMHO you always see user administrator because if config_get_global( 'admin_checks' ) == ON, MantisBT tries to get info about the user with login 'administrator'. 
If you remove this user or config_get_global( 'admin_checks' ) ==OFF, you are not going see th user_id = 1 anymore.
regards

V1pr:
@fmancardi  thanks for sharing, this is really interesting. @vboctor @dregad it this the expected working? A global setting can have this much influences on how an auth mech works?

V1pr:
@fmancardi take a look on my code. I've added autoprovision support to be able to redirect the auth in any cases (if you set credentials page, than you'll be able to have yourself redirected after the user inputs his username, if you set the authenticator page, than you can have a redirect after the password input)

PR #1071, opened 2017-03-26
Add user timeline to view user page

cproensa:
Display a timeline widget in "view user page", showing recent actions performed by the specific user.

The timeline is showed if access level is met by $t_timeline_view_threshold_access.
Issues that are shown are retrieved by joining with a filter, so that the only events displayed are those related to issues that the viewing user has access to.

Currently, the user timeline search is forced to All-projects. It's open to discussion if this information must be a global scope, or should be limited by current project selection.

These changes needed some refactoring of timeline_inc, timeline_api and history_api, to allow decoupling from my_view page, and user specific search.

seleccion_153

Fixes: #22585

### DIFF
--- a/core/history_api.php
+++ b/core/history_api.php
@@ -279,6 +279,7 @@ function history_query_result( array $p_query_options ) {
 		$t_query .= ' WHERE ' . implode( ' AND ', $t_where );
 	}
 
+	# Order history lines by date. Use the storing sequence as 2nd order field for lines with the same date.
 	$t_query .= ' ORDER BY {bug_history}.date_modified ' . $t_history_order . ', {bug_history}.id ' . $t_history_order;
 	$t_result = db_query( $t_query, $t_params );
 	return $t_result;
@@ -291,8 +292,12 @@ function history_query_result( array $p_query_options ) {
  * @param  integer $p_end_time       The end time to filter by, or null for all.
  * @param  string  $p_history_order  The sort order.
  * @return database result to pass into history_get_event_from_row().
+ * @deprecated		Use history_query_result() instead
  */
 function history_get_range_result_filter( $p_filter, $p_start_time = null, $p_end_time = null, $p_history_order = null ) {
+	error_parameters( __FUNCTION__ . '()', 'history_query_result()' );
+	trigger_error( ERROR_DEPRECATED_SUPERSEDED, DEPRECATED );
+
 	$t_query_options = array();
 	if ( $p_history_order !== null ) {
 		$t_query_options['order'] = $p_history_order;
@@ -316,8 +321,12 @@ function history_get_range_result_filter( $p_filter, $p_start_time = null, $p_en
  * @param  integer $p_end_time       The end time to filter by, or null for all.
  * @param  string  $p_history_order  The sort order.
  * @return IteratorAggregate|boolean database result to pass into history_get_event_from_row().
+ * @deprecated		Use history_query_result() instead
  */
 function history_get_range_result( $p_bug_id = null, $p_start_time = null, $p_end_time = null, $p_history_order = null ) {
+	error_parameters( __FUNCTION__ . '()', 'history_query_result()' );
+	trigger_error( ERROR_DEPRECATED_SUPERSEDED, DEPRECATED );
+
 	$t_query_options = array();
 	if ( $p_history_order !== null ) {
 		$t_query_options['order'] = $p_history_order;
@@ -475,14 +484,12 @@ function history_get_event_from_row( $p_result, $p_user_id = null, $p_check_acce
 function history_get_raw_events_array( $p_bug_id, $p_user_id = null, $p_start_time = null, $p_end_time = null ) {
 	$t_user_id = (( null === $p_user_id ) ? auth_get_current_user_id() : $p_user_id );
 
-	# grab history and display by date_modified then field_name
-	# @@@ by MASC I guess it's better by id then by field_name. When we have more history lines with the same
-	# date, it's better to respect the storing order otherwise we should risk to mix different information
-	# I give you an example. We create a child of a bug with different custom fields. In the history of the child
-	# bug we will find the line related to the relationship mixed with the custom fields (the history is creted
-	# for the new bug with the same timestamp...)
-
-	$t_result = history_get_range_result( $p_bug_id, $p_start_time, $p_end_time );
+	$t_query_options = array(
+		'bug_id' => $p_bug_id,
+		'start_time' => $p_start_time,
+		'end_time' => $p_end_time
+		);
+	$t_result = history_query_result( $t_query_options );
 
 	$t_raw_history = array();
 
--- a/core/history_api.php
+++ b/core/history_api.php
@@ -179,6 +179,112 @@ function history_count_user_recent_events( $p_duration_in_seconds, $p_user_id = 
 }
 
 /**
+ * Creates and executes a query for the history rows, returning a database result object.
+ * Query options is an array with parameters to build the query.
+ *
+ * Supported options are:
+ * - "bug_id" => integer | array    Limit search to these bug ids.
+ * - "start_time" => integer        Timestamp for start time to filter by (inclusive)
+ * - "end_time" => integer          Timestamp for end time to filter by (exclusive)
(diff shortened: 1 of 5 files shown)


core/history_api.php
```diff
@@ -238,6 +285,31 @@ function history_get_range_result_filter( $p_filter, $p_start_time = null, $p_en
...
 
 /**
+ * Creates and executes a query for the history rows related to bugs matched by the provided filter
+ * @param  array $p_filter           Filter array
+ * @param  integer $p_start_time     The start time to filter by, or null for all.
+ * @param  integer $p_end_time       The end time to filter by, or null for all.
+ * @param  string  $p_history_order  The sort order.
+ * @return database result to pass into history_get_event_from_row().
+ */
+function history_get_range_result_filter( $p_filter, $p_start_time = null, $p_end_time = null, $p_history_order = null ) {
```
cproensa: This function could be removed as it was introduced recently only to allow timeline on recursive subprojects, and has been replaced by history_query_result()

core/timeline_inc.php
```diff
@@ -19,17 +19,36 @@
...
 
+# Variables that are defined in parent script:
+#
+# $inc_timeline_filter	Filter array to be used to get timeline event
+#						If undefined, it's initialized as null.
+# $inc_timeline_user		User id to limit timeline scope.
+#						If undefined, it's initialized as null.
+#
+
+if( !isset( $inc_timeline_filter ) ) {
```
vboctor: Should $g_ prefix be used for these variables?

my_view_page.php
```diff
@@ -151,7 +151,13 @@
 
 <?php if( $t_timeline_view_threshold_access ) { ?>
 <div class="col-md-5 col-xs-12">
-	<?php include( $g_core_path . 'timeline_inc.php' ); ?>
+	<?php
+		# Build a simple filter that gets all bugs for current project
+		$inc_timeline_filter = array();
+		$inc_timeline_filter[FILTER_PROPERTY_HIDE_STATUS] = array( META_FILTER_NONE );
+		$inc_timeline_filter = filter_ensure_valid_filter( $inc_timeline_filter );
```
vboctor: Is it better to ensure valid here or in the API?  What pattern should be using and why?

core/helper_api.php
```diff
@@ -694,3 +694,20 @@ function helper_filter_by_prefix( array $p_set, $p_prefix ) {
...
 }
 
+/**
+ * Build the query part of an url, with an array of parameter name => value pairs
+ * This function does not prepends '?' to the resulting query string
+ * Parameter names and values must be valid for inclusion in a query string.
+ * @param array $p_params	Array of name => value pairs
+ * @return string	Resulting query string
+ */
+function helper_url_query_string( array $p_params ) {
```
cproensa: Thanks, i knew there was something already implemented, but couldn't find at that moment...

core/timeline_inc.php
cproensa: I have the same doubt.
We don't have a standard for this kind of variables that cross boundaries for included scripts.
I dont feel $g_ is a good prefix, since it can be mixed with general global variables.

Even if they are globals anyway, these are variables that the include script needs to be defined in parent script, an their usage is limited to this include pattern.
As opposed to general globals that are used throughout several api files...

Maybe the comment block at the start of the included script is enough, and keep them as $g_?

Note that we have, in other included scripts, variables with prefix $t_, which is a confusing situation. Having experience that, i tried to make it clear here.

my_view_page.php
cproensa: The approach i am using is to call filter_ensure_valid_filter:
- After creating a filter array, so it fills and formats all needed key fields.
- After modifying a filter from user input
- After loading a filter array from database, as it may fix minor format updates.

So, code that is modifiying a filter is responsible for validation.
The apis that use filters expect a valid structure, and shouldn't be needed to call  filter_ensure_valid_filter many times.

core/history_api.php
```diff
@@ -238,6 +285,31 @@ function history_get_range_result_filter( $p_filter, $p_start_time = null, $p_en
...
+	}
+	if ( $p_filter !== null ) {
+		$t_query_options['filter'] = $p_filter;
+	}
+	return history_query_result( $t_query_options );
+}
+
+/**
  * Creates and executes a query for the history rows matching the specified criteria.
  * @param  integer $p_bug_id         The bug id or null for matching any bug.
```
cproensa: I opt to not change expected parameters in this function.
If an array of bugs are requested, using the new history_query_result() should be better.

cproensa:
updated with some of victor commetns

my_view_page.php
vboctor: Makes sense.  Though "creating a filter array case" is a subset can of "modifying a filter".

core/timeline_inc.php
vboctor: At the moment I would suggest the following:

- Use `$g_` for all global variables.  If we want to limit this to just configuration, then we have a lot of existing code to revisit anyway.  We currently use that for API specific caches, etc.
- Use `$t_` for variables that are only used within a php function or a single php file.  Use of `$t_` for a variable that spans multiple files is not right and should be fixed over time.
- We could revisit the standards at one point, but for now, let's stick with what we have.

> Maybe the comment block at the start of the included script is enough, and keep them as $g_?

Yes.

> Note that we have, in other included scripts, variables with prefix $t_, which is a confusing situation. Having experience that, i tried to make it clear here.

We should fix this over time.

cproensa:
@vboctor 
Is the a reason for timeline_inc.php to be placed in core directory?
Shouldn't it be better place in root dir with the rest of web scripts?

my_view_page.php
cproensa: done

core/history_api.php
vboctor: Then let's remove it.

core/history_api.php
vboctor: Why wouldn't we remove this one too and just use the main one?

view_user_page.php
```diff
@@ -151,5 +153,22 @@
...
+
+<?php
+if( $t_timeline_view_threshold_access ) {
+	# Build a filter to show all bugs in current projects
+	$g_timeline_filter = array();
+	$g_timeline_filter[FILTER_PROPERTY_HIDE_STATUS] = array( META_FILTER_NONE );
+	$g_timeline_filter = filter_ensure_valid_filter( $g_timeline_filter );
+	$g_timeline_user = $f_user_id;
+	# Override current project, to let timeline show events for all-projects
+	$g_project_override = ALL_PROJECTS;
```
vboctor: Since the timeline on the My View Page based on selected project, I wonder if we should do the same here for consistency.

vboctor:
> Is the a reason for timeline_inc.php to be placed in core directory?
> Shouldn't it be better place in root dir with the rest of web scripts?

The goal should be to reduce the # of scripts that are accessible directly.  In such cases, similar to api files, inc files shouldn't be accessed directly by a user and hence shouldn't be in the main folder.  I know that we have several include files in main folder today, but these are older ones and I think we should move them into core folder too.

core/timeline_inc.php
```diff
@@ -82,8 +105,9 @@
 		timeline_print_events( $t_events );
 		echo '<div class="widget-toolbox">';
 		echo '<div class="btn-toolbar">';
-		echo '<a class="btn btn-primary btn-sm btn-white btn-round" href="my_view_page.php?days='
-			. $f_days . '&amp;all=1">' . lang_get( 'timeline_more' ) . '</a>';
+		$t_url_params['all'] = 1;
+		$t_href = $t_url_page . '?' . helper_url_query_string( $t_url_params );
```
dregad: This causes PHP error

```
[Tue Mar 28 06:44:47 2017] [error] [client 90.63.237.103] PHP Fatal error:  Call to undefined function helper_url_query_string() in /srv/www/bugs/core/timeline_inc.php on line 109, referer: https://www.mantisbt.org/demo.php
```

cproensa: tks, must be replaced with "http_build_query"
Do you have time to provide the change?, i cant until tomorrow at least...

PR #1072, opened 2017-03-27
FIX #22452 - Create new project button

romaincabassot:
https://www.mantisbt.org/bugs/view.php?id=22452

Moved 'Create New Project' button before the projects list.


vboctor:
I've added feedback on the MantisBT issue.

vboctor:
Please provide before/after screenshots.

romaincabassot:
Here is the screenshot:
mantis_projects

PR #1073, opened 2017-03-27
Fix dropzone alert if upload is not allowed

cproensa:
If uploading is not allowed, when viewing a bug, don't place dropzone
attributes in bugnote form. Otherwise, an error dialog is triggered.

Fixes: #22140


takenek:
Still when You try add bug you get same error.
This patch only resolve problem when You viewing a bug, but not when You add.
When You press add you get errors....

Best Regards
TaKeN

PR #1074, opened 2017-03-29
Leave xmlhttprequest.php for backward compatibility

vboctor:
Re-adding xmlhttprequest.php for plugins.  However, plugins should
be updated to use the new REST API approach.

This should fix the Snippets plugin that was broken by remote of this
page.

Fixes #22593


vboctor:
> If not already done, documenting this in the dev guide, preferably with instructions on how to migrate to the new API, would be useful.

This is document in the docs for `EVENT_REST_API_ROUTES` with reference to an example in the MantisGraph plugin.

> If feasible (might be tricky in an AJAX context), it would be a good idea to somehow display (or at least, log) an error message to inform about the deprecation. The comment at line 18 is not sufficient IMO.

We currently have Ajax logging.  I added to it that the method is deprecated.  If desired, I can do a follow up to add an error_log() just in case our LOG_AJAX is disabled.

For now, I will merge this and update our instance to get Snippets plugin working again.

PR #1075, opened 2017-03-29
Fix PHP error in PR #1071 (timeline)

cproensa:
Fix undefiend function error introduced by PR #1071 
e2d1de8a9f2fc5a7190efa252b2aa6df7e5c6310

EDIT (dregad): when adding the # reference in the title, Github does not generate the hyperlink in the target PR


dregad:
Since this is also fixing the very annoying issue on my view page (see https://www.mantisbt.org/bugs/view.php?id=22590), I'm going to merge it now and update the tracker.

PR #1076, opened 2017-04-01
Don’t show completely empty notes

vboctor:
For time tracking notes with no text, if user doesn’t have access
to time tracking information, then an empty note is displayed.
This fix hides the note in such case.

Fixes #22635


dregad:
I didn't actually test, but what happens if we have a time-tracking note without text, but with an attachment ?

vboctor:
Just tested this case and it works fine.  The attachments shows on its own.

PR #1077, opened 2017-04-04
Fix dropzone alert for bug report page

cproensa:
If dropzone tool is not used, when reporting a bug, don't place dropzone
attributes in form. Otherwise, an error dialog is triggered.

Fixes: #22140

### DIFF
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -237,9 +237,7 @@ if( $t_show_attachments ) {
 <div class="col-md-12 col-xs-12">
 <form id="report_bug_form"
 	method="post" <?php echo $t_form_encoding; ?>
-	action="bug_report.php?posted=1"
-	class="dropzone-form"
-	<?php print_dropzone_form_data() ?>>
+	action="bug_report.php?posted=1">
 <?php echo form_security_field( 'bug_report' ) ?>
 <input type="hidden" name="m_id" value="<?php echo $f_master_bug_id ?>" />
 <input type="hidden" name="project_id" value="<?php echo $t_project_id ?>" />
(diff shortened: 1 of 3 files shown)


bug_report_page.php
```diff
@@ -230,16 +230,21 @@
...
+	$t_form_class = 'class="dropzone-form"';
 }
 ?>
 <div class="col-md-12 col-xs-12">
 <form id="report_bug_form"
 	method="post" <?php echo $t_form_encoding; ?>
 	action="bug_report.php?posted=1"
-	class="dropzone-form"
-	<?php print_dropzone_form_data() ?>>
+	<?php echo $t_form_class ?>
```
atrol: Is there a special reason that you did not use the same approach you used in https://github.com/mantisbt/mantisbt/commit/5898638f65f08fdadce35864a3c4011dc0fac866 ?

No additional variable $t_form_class, but `echo ' class="dz dropzone-form" ';` just if upload is enabled.

bug_report_page.php
```diff
@@ -230,16 +230,21 @@
...
+	<?php if( $t_show_attachments ) {
```
vboctor: Why two php blocks next to each other?  Can they be combined?  Thought that may change with the previous suggestions.

bug_report_page.php
vboctor: Here are a couple of suggestions for approaching this:

1. Change `print_dropzone_form_data` to `prepare_dropzone_form_data()` and capture its output into a variable like the class in this case.  This way the code for outputting the form is easier to read without having IF statements in the middle of a `form` tag.

2. Change `print_dropzone_form_data()` to take `$t_show_attachments` as a parameter and based on that decide to output the class and the other fields.  I know that can cause two class attributes if the form has a class already, but in both cases where we use drop zone, that is not the case.

cproensa: 1. ok, that's better

I'd rather fix the javascript that checks the dropzone container, so the form can have "dropzone-form" class without raising an error.

PR #1078, opened 2017-04-05
Allow submission of blank custom datetime field when not require

syncguru:
Fixes #22663

### DIFF
--- a/core/cfdefs/cfdef_standard.php
+++ b/core/cfdefs/cfdef_standard.php
@@ -271,7 +271,7 @@ function cfdef_prepare_email_value( $p_value ) {
  * @return string
  */
 function cfdef_prepare_date_value( $p_value ) {
-	if( $p_value != null && $p_value != 0 ) {
+	if( $p_value != null ) {
 		return date( config_get( 'short_date_format' ), $p_value );
 	}
 }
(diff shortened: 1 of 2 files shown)


core/custom_field_api.php
```diff
@@ -1113,6 +1113,10 @@ function custom_field_validate( $p_field_id, $p_value ) {
 
 			break;
 		case CUSTOM_FIELD_TYPE_DATE:
+            # Date field can be left blank (when field is not required)
```
vboctor: Why is this check needed?  The following check is valid if value is empty, null or 0.
Also please use tabs for indentation instead of spaces.

syncguru: The following check results in $t_valid => false which then throw the error.

core/cfdefs/cfdef_standard.php
```diff
@@ -271,7 +271,7 @@ function cfdef_prepare_email_value( $p_value ) {
  * @return string
  */
 function cfdef_prepare_date_value( $p_value ) {
-	if( $p_value != null ) {
+	if( $p_value != null && $p_value != 0 ) {
```
dregad: The added condition is useless, as `null == 0`.

syncguru: @dregad you are right. Removed.

vboctor:

PR #1079, opened 2017-04-05
Fix regression of switching dates in filter to use date picker

syncguru:
The extra time annotation was not supposed to be part of the change. As it made it to master, it broke filters with date_submitted & last_updated values set.

Fixes #22653

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -497,7 +497,7 @@ function filter_ensure_fields( array $p_filter_arr ) {
 		}
 	}
 
-	# Veryfy custom fields
+	# verify custom fields
 	foreach( $t_filter_default['custom_fields'] as $t_cfid => $t_cf_data ) {
 		if( !isset( $p_filter_arr['custom_fields'][$t_cfid] ) ) {
 			$p_filter_arr['custom_fields'][$t_cfid] = $t_cf_data;
@@ -1314,9 +1314,6 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
 	log_event( LOG_FILTERING, 'START NEW FILTER QUERY' );
 
 	$t_limit_reporters = config_get( 'limit_reporters' );
-	$t_report_bug_threshold = config_get( 'report_bug_threshold' );
-	$t_where_param_count = 0;
-
 	$t_current_user_id = auth_get_current_user_id();
 
 	if( $p_user_id === null || $p_user_id === 0 ) {
@@ -1512,12 +1509,9 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
 	}
 
 	# creation date filter
-	if( ( 'on' == $t_filter[FILTER_PROPERTY_FILTER_BY_DATE_SUBMITTED] )
-			&& is_string( $t_filter[FILTER_PROPERTY_START_DATE_SUBMITTED] )
-			&& is_string( $t_filter[FILTER_PROPERTY_END_DATE_SUBMITTED] )
-			) {
-		$t_start_string = $t_filter[FILTER_PROPERTY_START_DATE_SUBMITTED];
-		$t_end_string = $t_filter[FILTER_PROPERTY_END_DATE_SUBMITTED];
+	if( ( 'on' == $t_filter[FILTER_PROPERTY_FILTER_BY_DATE_SUBMITTED] )	) {
+		$t_start_string = filter_read_last_updated_start_date( $t_filter );
+		$t_end_string = filter_read_last_updated_end_date( $t_filter );
 
 		$t_where_params[] = strtotime( $t_start_string );
 		$t_where_params[] = strtotime( $t_end_string );
@@ -1526,10 +1520,7 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
 
 
 	# last update date filter
-	if( ( 'on' == $t_filter[FILTER_PROPERTY_FILTER_BY_LAST_UPDATED_DATE] )
-			&& is_string( $t_filter[FILTER_PROPERTY_LAST_UPDATED_START_DATE] )
-			&& is_string( $t_filter[FILTER_PROPERTY_LAST_UPDATED_END_DATE] )
-			) {
+	if( ( 'on' == $t_filter[FILTER_PROPERTY_FILTER_BY_LAST_UPDATED_DATE] ) ) {
 		$t_start_string = $t_filter[FILTER_PROPERTY_LAST_UPDATED_START_DATE];
 		$t_end_string = $t_filter[FILTER_PROPERTY_LAST_UPDATED_END_DATE];
 
@@ -3133,12 +3124,13 @@ function filter_gpc_get( array $p_filter = null ) {
 	# date values
 	# creation date
 	$f_do_filter_by_date	        = gpc_get_bool( FILTER_PROPERTY_FILTER_BY_DATE_SUBMITTED, $t_filter[FILTER_PROPERTY_FILTER_BY_DATE_SUBMITTED] );
-	$f_submitted_start_date			= gpc_get_string( FILTER_PROPERTY_START_DATE_SUBMITTED, $t_filter[FILTER_PROPERTY_START_DATE_SUBMITTED] );
-	$f_submitted_end_date			= gpc_get_string( FILTER_PROPERTY_END_DATE_SUBMITTED, $t_filter[FILTER_PROPERTY_END_DATE_SUBMITTED] );
+	$f_submitted_start_date			= filter_read_submitted_start_date( $t_filter );
+	$f_submitted_end_date			= filter_read_submitted_end_date( $t_filter );
+
 	# last_updated date values
 	$f_do_filter_by_last_updated_date	= gpc_get_bool( FILTER_PROPERTY_FILTER_BY_LAST_UPDATED_DATE, $t_filter[FILTER_PROPERTY_FILTER_BY_LAST_UPDATED_DATE] );
-	$f_last_updated_start_date			= gpc_get_string( FILTER_PROPERTY_LAST_UPDATED_START_DATE, $t_filter[FILTER_PROPERTY_LAST_UPDATED_START_DATE] );
-	$f_last_updated_end_date			= gpc_get_string( FILTER_PROPERTY_LAST_UPDATED_END_DATE, $t_filter[FILTER_PROPERTY_LAST_UPDATED_END_DATE] );
+	$f_last_updated_start_date			= filter_read_last_updated_start_date( $t_filter );
+	$f_last_updated_end_date			= filter_read_last_updated_end_date( $t_filter );
 
 	$f_search				= gpc_get_string( FILTER_PROPERTY_SEARCH, $t_filter[FILTER_PROPERTY_SEARCH] );
 	$f_view_state			= gpc_get_int( FILTER_PROPERTY_VIEW_STATE, $t_filter[FILTER_PROPERTY_VIEW_STATE] );
@@ -3200,8 +3192,10 @@ function filter_gpc_get( array $p_filter = null ) {
 					$t_start_date = (int)$f_start_date;
 					$t_start = $t_start_date;
 				} else {
-					$t_date = gpc_get_string( 'custom_field_' . $t_cfid . '_start_date', null );
-					$t_start_date = strtotime( $t_date );
+					$t_year = gpc_get_int( 'custom_field_' . $t_cfid . '_start_year', null );
+					$t_month = gpc_get_int( 'custom_field_' . $t_cfid . '_start_month', null );
+					$t_day = gpc_get_int( 'custom_field_' . $t_cfid . '_start_day', null );
+					$t_start_date = mktime( 0, 0, 0, $t_month, $t_day, $t_year );
 					# calculate correct timestamps
 					$t_start = 1;
 					switch( $t_control ) {
@@ -3233,8 +3227,10 @@ function filter_gpc_get( array $p_filter = null ) {
 					$t_end_date = (int)$f_end_date;
 					$t_end = $t_end_date;
 				} else {
-					$t_date = gpc_get_string( 'custom_field_' . $t_cfid . '_end_date', null );
-					$t_end_date = strtotime( $t_date );
+					$t_year = gpc_get_int( 'custom_field_' . $t_cfid . '_end_year', null );
+					$t_month = gpc_get_int( 'custom_field_' . $t_cfid . '_end_month', null );
+					$t_day = gpc_get_int( 'custom_field_' . $t_cfid . '_end_day', null );
+					$t_end_date = mktime( 0, 0, 0, $t_month, $t_day, $t_year );
 					# calculate correct timestamps
 					$t_end = 1;
 					switch( $t_control ) {
@@ -3377,7 +3373,7 @@ function filter_is_named_filter( $p_filter_id ) {
 }
 
 /**
- * Returns true if the filter is accesible by the user, which happens when the user
+ * Returns true if the filter is accessible by the user, which happens when the user
  * is the owner of the filter, or the filter is public.
  * @param integer $p_filter_id	Filter id
  * @param integer $p_user_id	User id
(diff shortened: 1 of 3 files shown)


vboctor:
I'll let @cproensa review this first.

cproensa:
I have commented in issue #22653:

Still not addressed:
- Breaks backwards compatibility with stored filters and permalinks
- Changing $g_normal_date_format breaks filter functionality: string date in filter is stored based on user configuration. When building the query it's directly fed to strtotime(), which fails with unsupported formats like: $g_normal_date_format = 'd/m/Y H:i'
- Changing $g_normal_date_format show dates in widget incorrectly when loading stored dates with a different format. For example: "30/01/2015 00:00" is converted to 2030-01-20 00:00 when configuration changes.

However, my main concern is that filter structure has changed, which is not required for the new widget implementation, and that is the main reason for the problems.

syncguru:
> Beaks backward compatibility with stored filters and permalinks
That's a rare scenario if the user changed the format from the default. If the default was used then $g_short_date_format will work with $g_normal_date_format

> Changing $g_normal_date_format breaks filter functionality
That's the case before the change. So existing issue that I was not trying to address with the PR.

cproensa:
>> Beaks backward compatibility with stored filters and permalinks

> That's a rare scenario if the user changed the format from the default. If the default was used then $g_short_date_format will work with $g_normal_date_format

No. It happens without any modification, because the filter properties where the dates were stored are ignored

>> Changing $g_normal_date_format breaks filter functionality

> That's the case before the change. So existing issue that I was not trying to address with the PR.

No. Before this PR dates were stored  as separated date parts, and the construction of query did not depend on localization.

cproensa:
To make clear the situation:
This is the subset of json related to date as is stored as part of a filter:
```
"filter_by_date":true,"start_month":1,
"start_day":1,
"start_year":2015,
"end_month":1,
"end_day":30,
"end_year":2015,
```

This is how it's stored after the changes in filter structure:
```
"filter_by_date":true,
"filter_by_start_date":"01-01-2015 00:00",
"filter_by_end_date":"30-01-2015 00:00"
```

syncguru:
@cproensa I updated the fix for filter regression. It should now be backward compatible with old date fields (day, month, year) - take a look and let me know your thoughts.

cproensa:
I still see issues caused by storing a date string in a localized format.

My suggestion was that the way dates are stored in filter should not be changed, from a start.
Which means dates are to be stored in the filter array using the existing properties:
```
FILTER_PROPERTY_DATE_SUBMITTED_START_DAY
FILTER_PROPERTY_DATE_SUBMITTED_START_MONTH
FILTER_PROPERTY_DATE_SUBMITTED_START_YEAR
FILTER_PROPERTY_DATE_SUBMITTED_END_DAY
FILTER_PROPERTY_DATE_SUBMITTED_END_MONTH
FILTER_PROPERTY_DATE_SUBMITTED_END_YEAR

FILTER_PROPERTY_LAST_UPDATED_START_DAY
FILTER_PROPERTY_LAST_UPDATED_START_MONTH
FILTER_PROPERTY_LAST_UPDATED_START_YEAR
FILTER_PROPERTY_LAST_UPDATED_END_DAY
FILTER_PROPERTY_LAST_UPDATED_END_MONTH
FILTER_PROPERTY_LAST_UPDATED_END_YEAR
```

Then these properties would not exist as such:
```
FILTER_PROPERTY_START_DATE_SUBMITTED
FILTER_PROPERTY_END_DATE_SUBMITTED
FILTER_PROPERTY_LAST_UPDATED_START_DATE
FILTER_PROPERTY_LAST_UPDATED_END_DATE
```

Since the calendar widget sends and receives a date string: the "full date" string should be built when writting the UI elements, and converted back to the separated D/M/Y values when reading form sumbission.
Doing so avoids having the need to manage conversions when loading old filters, and also avoid impacting on UI-unrelated code, like building the sql query.
The old format uses a lot of variables, and as i stated, i'd really like to have a rewrite of date fields. However that would need more changes that just having a date string, and that's one reason i prefer keeping current format and wait for the full changes to make it happen.
As a bonus, the old format does not suffer from localization issues.

My suggested strategy is, then:
- filter_gpc_get(): reads the submitted string for each date (as given by the widget), convert it to D M Y parts and store them in the filter array
- filter_form_api.php: build the form UI stuff, getting the filter properties as date parts and converting here to a date string suitable for the widget.
- Same with date-typed custom fields. Convert to and from its existing format, only at the UI endpoints.
Anything else should then be unmodified.

Drawback: Existing date properties don't support hours:minutes. The calendar should then hide the time selection part. Not a big issue as we didn't support hours previously, and can be postponed until an actual rewrite of the date fields is made.

Unless there are more opinions in favor of keeping the new date format, my suggestion is what i have detailed.

cproensa:
After chatting with RAfik, it's clear we have opposing views on how to unblock this issue.
There's nothing wrong with that, though.

I will let the team now decide on next steps. I will agree on whatever is agreed to fix all the regressions, which are the priority right now, and offer my help on any work if is needed for this.

core/filter_api.php
```diff
@@ -1512,12 +1509,9 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
...
 
 	# creation date filter
-	if( ( 'on' == $t_filter[FILTER_PROPERTY_FILTER_BY_DATE_SUBMITTED] )
-			&& is_string( $t_filter[FILTER_PROPERTY_START_DATE_SUBMITTED] )
-			&& is_string( $t_filter[FILTER_PROPERTY_END_DATE_SUBMITTED] )
-			) {
-		$t_start_string = $t_filter[FILTER_PROPERTY_START_DATE_SUBMITTED] . ' 00:00:00';
-		$t_end_string = $t_filter[FILTER_PROPERTY_END_DATE_SUBMITTED] . ' 23:59:59';
+	if( ( 'on' == $t_filter[FILTER_PROPERTY_FILTER_BY_DATE_SUBMITTED] )	) {
+		$t_start_string = filter_read_last_updated_start_date( $t_filter );
```
vboctor: This function refers to last updated even though the if statement refers to date submitted.

vboctor:
I can't see the discussion where there are two opposing point of views.  I'm also not very familiar with the filtering code to provide the best code change, but I will share my thoughts:

- The format that is saved to the db MUST be culture neutral.  For example, date either split into 3 fields as per before or in some well known ISO format.
- Saved filters before regressions must be supported.
- Saved filters during the regressions can be broken (where they didn't work when created).

Custom fields of type date should follow same principles.  Locale neutral format and format shouldn't be changed.

What I don't have a strong opinion on is whether we change the format or not.  If it is a good change in  then I don't mind it.  From memory, we have a decent versioning story for filters and hence we don't have to do all the changes in one go.

I would love to see us support relative dates in filters (e.g. -30 days, +30 days).

I haven't gave time support much thought.  I don't like just current time showing up in the selector, but there may be a good default and a way to set the time when needed.

It would be useful if @syncguru and @cproensa clarify the exact point of disagreement.

core/filter_api.php
syncguru: Good catch

syncguru:
I totally agree that we should use culture neutral to store last_updated & date_submitted values in the filter structure. This is in line with @cproensa & @vboctor review comments should be addressed before this PR is merged.

As for the disagreement about how to approach this. I am of the mindset that this change is already in and we need to fix the regression. This PR fixes the regression and enable backward support of the 3 column date fields (M,D,Y) while switching to new date format for new or updated filter (the new format should be timestamp which is what we use with custom date fields in filter today).

@cproensa seems to object to any attempt to fixing this regression but prefer backing out the original change and leaving this work to him as part of the filter code changes that he is planning. @cproensa please comment further on your reasoning.

vboctor:
I would target the following:

- Fix the regression.
- Make sure the change in format we are doing is step in the right direction.

If both the above is true, and if the new format is consistent with custom field dates in the filter, then it seems to be a step towards what we may want to go in the future.  As I said I don't mind staging the desired changes into several version updates.

@cproensa can you explain your point of view?

cproensa:
Additionally to what i have explained previously in this PR and #22653:

> the new format should be timestamp which is what we use with custom date fields in filter today

- If we properly want to converge both native and custom fields date filtering, the filter structure is still different, so the intended goal is not accomplished by this.
- this change does not help in any way to support relative date formats
- having a modified filter format (the whole structure) is a overhead to support backwards compatibility (let's do it properly, only when we have a change that makes sense for the planned goal)
- The filter changes are part of a UI feature, which does not required them in the first place, and by doing them, it has created side effects that caused the regressions.

> @cproensa seems to object to any attempt to fixing this regression but prefer backing out the original change and leaving this work to him as part of the filter code changes that he is planning. @cproensa please comment further on your reasoning.

You are understanding this as if i have some interest on not letting other work to flow. I am sorry that you may have interpreted this way. I have tried to express my reasoning from the start.
Note that i offered to work on these changes to avoid him the repeated work of going back and forth with all these disagreements, NOT because i want to do the filter stuff by myself and no one else (thats a surprising assumption which i never implied)
I believe i am in the line of what was expected in this path, as an example this is a recent victor comment supporting to not break filter data for minor fixes:
https://www.mantisbt.org/bugs/view.php?id=22555#c56154


To summarize:
> Fix the regression.

This PR is still not correct. I opt to not give more feedback until we (the team, not Rafik or me individually) finally decide if the filter structure is going to change. For the sole reason of not having Rafik doing more work that necessary.

> Make sure the change in format we are doing is step in the right direction.

In my opinion, it's neither needed, nor positioning us in any different place that we are now.

If you prefer storing date values as timestamps, thats one thing, and we can go with it and work from there on.
But let's not justify it as a need for UI changes, or for features for which we haven't still defined the technical requeriments.

i'm getting tired of this, so please decide if we settle the change in filter data or not, and let's move on.

syncguru:
@cproensa I want to make sure that any change we do now aligns with how you see the filter structure evolving to. So far we have not touched on that. By communicating your line of thoughts, early designs ..etc we can make this conversation more productive.

cproensa:
There is no value in delaying this any longer. The work is done, several PRs have been merged already, and the less traumatic for your work is to keep fixing as is, and keep the new date format.
Even if i think that changing filter backend was not desirable for the purpose of the UI, we can cope with that and work the issues later, if any.

As a summary of what i found that needed fixing:
> Breaks backwards compatibility with stored filters and permalinks

BC with stored filters should be solved by converting old values to the new properties in filter_ensure_valid_filter().
For permalinks, reading "old" permalinks should be feasible with similar conversion. Can't say if generating permalinks with the new date syntax (your intention is timestamp) is recommended, as permalinks have the characteristic of being more human readable. i didn't refactor permalinks yet, so i don't know how much trouble can they be in the long run.

> Changing $g_normal_date_format breaks filter functionality: string date in filter is stored based on user configuration. When building the query it's directly fed to strtotime(), which fails with unsupported formats like: $g_normal_date_format = 'd/m/Y H:i'

By using timestamps, this localization issue wouldn't happen anymore

> Changing $g_normal_date_format show dates in widget incorrectly when loading stored dates with a different format. For example: "30/01/2015 00:00" is converted to 2030-01-20 00:00 when configuration changes.

Probably with the conversion to timestamp this gets solved too, but look carefully anyway.

I assume the final date value will be integer timestamps as you suggested. However, any fixed, non localized internal format will work too, as victor said.

Some coding comments:
- as said, conversion for old properties compatibility should be made in filter_ensure_valid_filter.
- reading gpc should only happen in filter_gpc_get. for example, there are calls to filter_read_date_value (which do gpc_get_string), from places like filter_get_bug_rows_query_clauses

cproensa:
As a general question.
Should we increase filter version v9->v10, to signal these structural changes, even if minor? (and manage in a more ordered way the conversions)
Or do the conversions silently?

cproensa:
target to 2.3

syncguru:
@dregad, @cproensa - I concede :)
Let me know what you need from me to revert the original change.

atrol:
Whatever you are doing, keep in mind that it would be quite good if we keep being able to run 1.3.x and 2.x.x in parallel accessing the same database.

cproensa:
> Let me know what you need from me to revert the original change.

if going that route, i suggest to
1) revert all the commits, so that 2.3.1 can be released without the calendar changes, and without any of the issues.
2) merge 2.3 branch at that point into master, to remove it also from master branch.
3) resubmit your PR1032 to master (total, or partially) and work from there, so we can review and modify as necessary without having issues pending in any released branch.

I have tried the step of reverting from master-2.3 and is (almost, 1 conlict, easy) a clean process. I have this branch ready if you will, so it can be PRed right away:
https://github.com/cproensa/mantisbt/tree/revert_PR1032_master-2.3
Also, merge into master is clean after that

The commits that i reverted that, unless mistaken, are part of these:
```
commit 757320b21db774b965f7efc447ef23e3907a0617
Author: José D. Grieco <jdgrieco@lwinformatica.com.br>
Date:   Mon Apr 10 11:14:11 2017 -0300

    Missing "echo" in date-picker-locale attrib in due date field

commit 61d790f8302290fb7d479b857d30f54aff0009a6
Author: Rafik Robeal <rafik@mantishub.net>
Date:   Thu Mar 16 20:00:36 2017 -0700

    Fix datetime text box max size

commit 12c1eebabb5e0766dcd889f8f46a41999f21e4d5
Author: Rafik Robeal <rafik@mantishub.net>
Date:   Thu Mar 16 15:37:58 2017 -0700

    Increase the width of datetime picker text box

commit 0193226341f726d071478fbd450417adaf156302
Author: Rafik Robeal <rafik@mantishub.net>
Date:   Mon Mar 6 16:04:13 2017 -0800

    Convert php date format string to moment.js format string

commit 008cc6c76c8b13be9efdf5645a7c92d91489f3ca
Author: Rafik Robeal <rafik@mantishub.net>
Date:   Thu Feb 16 13:44:04 2017 -0800

    Fix typo in js function name

commit 3b2a5e14f17dd4aba8c2b0dda85e7e0335792101
Author: Rafik Robeal <rafik@mantishub.net>
Date:   Tue Feb 14 20:43:30 2017 -0800

    Use datetime picker with time tracking

commit 6d2206657e33873d48c442acc440b8efcb862f6c
Author: Rafik Robeal <rafik@mantishub.net>
Date:   Tue Feb 14 19:21:46 2017 -0800

    Convert custom date field value to timestamp

commit eebd6a740a18cc8727dedbb6e09c5eed5b04df70
Author: Rafik Robeal <rafik@mantishub.net>
Date:   Mon Feb 13 11:49:40 2017 -0800

    Post merge fixes

commit 424ad6865697de49f2d3b0f4482050224ad33294
Author: Rafik Robeal <rafik@mantishub.net>
Date:   Mon Feb 13 11:31:03 2017 -0800

    Covert filter form to read and write date picker string value
    
    Fixes #8957

commit 1721c39a81e561553ee2bd999a81e71d6189d24e
Author: Rafik Robeal <rafik@mantishub.net>
Date:   Fri Feb 10 15:53:02 2017 -0800

    Replace date selection dropdown boxes with datetime picker
    
    Fixes #8957

commit 6511b27f1edde2ffe423466255c2be3fb043a8a1
Author: Rafik Robeal <rafik@mantishub.net>
Date:   Mon Feb 6 18:50:06 2017 -0800

    Use datetime picker for custom date fields
    
    Fixes #8957

```

syncguru:
@cproensa Sounds good, please submit as PR.

cproensa:
> @cproensa Sounds good, please submit as PR.

Done, #1093

> Bump filter version to v10 

Note that if changing filter version, i suggest that to be done on top of the changes proposed in #1091, because there i have made a big clean up of how filters are loaded and decoded, and has a solid base for introducing a proper version upgrade step.
Otherwise, i'm not confident that the upgrade can be made reliably.

dregad:
Following merge of #1093, should this be closed ?

cproensa:
> should this be closed ?

yes, @syncguru please submit a new PR to work with the datepicker changes again

elgutu:
Hi, I apply the patch on the three specified files and the filter of last updated doesn't work for me, but after reviewing the code, I figured out that there's was an error.
I comment the original code, and leave my own code in here so you see if it's neccesary a change (at least this work in my case with no problems, and now the filter works.)...
Sorry if this is already solved in a new version.
Cheers.

filter_api.php
function filter_read_last_updated_start_date( $p_filter )
{
        return filter_read_date_value( $p_filter,
        FILTER_PROPERTY_LAST_UPDATED_START_DATE,
                FILTER_PROPERTY_LAST_UPDATED_START_DAY,
                FILTER_PROPERTY_LAST_UPDATED_START_MONTH,
                FILTER_PROPERTY_LAST_UPDATED_START_YEAR);

/**
*        FILTER_PROPERTY_START_DATE_SUBMITTED,
*               FILTER_PROPERTY_DATE_SUBMITTED_START_DAY,
*               FILTER_PROPERTY_DATE_SUBMITTED_START_MONTH,
*               FILTER_PROPERTY_DATE_SUBMITTED_START_YEAR);
*/

}

/**
 * Read last_updated end date filter value
 * @param array $p_filter       Filter values
 * @return date string
 */
function filter_read_last_updated_end_date( $p_filter )
{
        return filter_read_date_value( $p_filter,
        FILTER_PROPERTY_LAST_UPDATED_END_DATE,
                FILTER_PROPERTY_LAST_UPDATED_END_DAY,
                FILTER_PROPERTY_LAST_UPDATED_END_MONTH,
                FILTER_PROPERTY_LAST_UPDATED_END_YEAR);


/**
*        FILTER_PROPERTY_END_DATE_SUBMITTED,
*               FILTER_PROPERTY_DATE_SUBMITTED_END_DAY,
*               FILTER_PROPERTY_DATE_SUBMITTED_END_MONTH,
*               FILTER_PROPERTY_DATE_SUBMITTED_END_YEAR );
*/
}

PR #1081, opened 2017-04-05
Implement modern tagging functionality

syncguru:
Fixes #21813

This is a much more streamlined ajax-based implementation of tagging functionality. 

Screenshots:

screen shot 2017-04-05 at 4 21 32 pm

screen shot 2017-04-05 at 4 21 52 pm

screen shot 2017-04-05 at 4 22 30 pm


core/tag_api.php
```diff
@@ -920,19 +920,12 @@ function tag_display_link( array $p_tag_row, $p_bug_id = 0 ) {
...
-	} else {
-		$i = 0;
-		foreach( $t_tag_rows as $t_tag ) {
-			echo( $i > 0 ? config_get( 'tag_separator' ) . ' ' : '' );
-			tag_display_link( $t_tag, $p_bug_id );
-			$i++;
-		}
+	$i = 0;
+	foreach( $t_tag_rows as $t_tag ) {
+		echo( $i > 0 ? config_get( 'tag_separator' ) . ' ' : '' );
```
vboctor: I know this is old code, but let's get the config for the separate before the loop into a temp variable.

vboctor:
I haven't tried it, but here are some high level comments:

- Don't use `xmlhttprequest_api.php`.  Please add new REST APIs for `/tags/` where a POST would attach a tag to an issue, one DELETE would delete it.  You can model the auto-complete as internal API similar to auto-complete for platform, build, os -- though you can argue clients could be interested in that as well.

- When CDN enabled, make sure we are loading from CDN (if available), and including the hash of the resource.

cproensa:
- The suggestions dropdown is hidden behind the div container limits
seleccion_156

- I miss a list of tags without having to write any character. Like: use a space as a special initial char to show all tags?

- Some tag names are not validated. for example "+xxx" "-xxxx". Those names were not valid previously.
Which makes sense because +/- are used in filters as inclusive/exclusive modificators for tag search

cproensa:
As i see, the tags are added to the UI regardless of the validation on server side.
- Auto-creation of tags: if creation failed, the tag is shown in UI anyway
- Matching capitalization: if i write "NAME", but the existent tag is "name", "name" should be added to the UI

Those should go away if the added tag is read from server response, or an error warning if it fails creation, etc.

vboctor:

PR #1082, opened 2017-04-06
Rewrite dropzone initialization

cproensa:
This replaces #1077 as a better approach.

Some changes has been done to Dropzone initialization:
- Container form now does not need any special class.
- Dropzone specific data attributes are now placed in the Dropzone div
  element, which is indicated as "dropzone" class.
- Auto upload feature is now enabled by adding the additional
  "auto-dropzone" class to current "dropzone" div.

Dropzone initialization searchs for elements with class "dropzone", then
searchs for the container form and uses it as parameter for the object,
and register event listeners.

Fixes: #22140, #22673

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -231,12 +231,11 @@ $(document).ready( function() {
         enableDateTimePicker(this);
 	});
 
-	if( $( ".dropzone-form" ).length ) {
-		enableDropzone( "dropzone", false );
-	}
-	if( $( ".auto-dropzone-form" ).length ) {
-		enableDropzone( "auto-dropzone", true );
-	}
+	$( 'form .dropzone' ).each(function(){
+		var classPrefix = 'dropzone';
+		var autoUpload = $(this).hasClass('auto-dropzone');
+		enableDropzone( classPrefix, autoUpload );
+	});
 
 	$('.bug-jump').find('[name=bug_id]').focus( function() {
 		var bug_label = $('.bug-jump-form').find('[name=bug_label]').val();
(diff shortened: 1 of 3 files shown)


js/common.js
```diff
@@ -231,12 +231,11 @@ $(document).ready( function() {
...
 	});
 
-	if( $( ".dropzone-form" ).length ) {
-		enableDropzone( "dropzone", false );
-	}
-	if( $( ".auto-dropzone-form" ).length ) {
-		enableDropzone( "auto-dropzone", true );
-	}
+	$( 'form .dropzone' ).each(function(){
+		var classPrefix = 'dropzone';
```
vboctor: Do we need this variable vs. just using the string directly?  Why not have the class hard-coded in enableDropzone()?

cproensa: I guess that `enableDropzone( classPrefix, autoUpload )` was written with the idea to allow different class names to match the dropzone elements. For example, allowing more than one DZ, for different forms, in the same page.
However, i think that the library is internally dependant on elements with "dropzone" class. At least, i couldn't make it work in other way.

So, the class name is still hardcoded, but i kept the class-prefix parameter in case we would want to improve that feature eventually.

cproensa:
First time dealing with this library, so i need help reviewing and testing this.

js/common.js
```diff
@@ -231,12 +231,11 @@ $(document).ready( function() {
...
+		var autoUpload = $(this).hasClass('auto-dropzone');
```
syncguru: I thought we removed auto-upload, right? It used to be in issue details page but it is gone now.

cproensa: auto-upload is not used currently.
Supporting the feature as is, is simple enough, in case is needed, or reused by plugins?

PR #1083, opened 2017-04-06
Enable scrolling for timeline widget using jquery.nicescroll

syncguru:
Fixes #22674

Here is the jquery.nicescroll library: https://github.com/inuyaksa/jquery.nicescroll
The timeline box max height is set for different screen sizes

Snapshots:

screen shot 2017-04-06 at 1 09 33 pm

screen shot 2017-04-06 at 1 10 10 pm


cproensa:
Can this scroll library also be used to fix horizontal scrolling of tables (#22132)?

If yes, then please fix that too now that the library is intoroduced.
If not, then i suggest to use another scroll library that solves both vertical and horizontal scroll usages

cproensa:
The fixed heights looks too small, as a first impression.
Note that timeline is presented as a vertical bar. I would expect that it takes advantage of the screen from top to bottom

Do you consider setting height based on viewport, to a reasonable value? for example:
max-height: 80vh;

resulting in:
seleccion_157


instead of this PR original:
seleccion_158

atrol:
I did not have a deeper look at the library, but we might start riding a dead horse

- No activity since about 8 months https://github.com/inuyaksa/jquery.nicescroll/graphs/contributors?from=2016-08-07&to=2017-04-07&type=c
- 245 open issue https://github.com/inuyaksa/jquery.nicescroll/issues
- Maybe not latest state of the art, containing unneeded legacy code, e.g. support is for IE6 - IE10, see http://nicescroll.areaaperta.com

syncguru:
> Can this scroll library also be used to fix horizontal scrolling of tables (#22132)?

@cproensa: Yes.
Not sure what you referring to here. Please open an issue for horizontal scrolling needs. It is not in scope for this PR.

@atrol, @dregad - Fair point. I did more research and found another library that is in active development.

syncguru:
Closing this PR in favor of https://github.com/mantisbt/mantisbt/pull/1088

PR #1084, opened 2017-04-07
Honor HTTP_X_FORWARDED_PROTO for Gravatar

vboctor:
When behind a proxy/load balancer and HTTP_X_FORWARDED_PROTO indicates
that MantisBT is accessed via https, make sure all resources are loaded via https.

Fixes #22689


vboctor:
I pushed a change to re-use the same logic in `config_defaults_inc.php`.

PR #1087, opened 2017-04-07
Allow db_minutes_to_hhmm to work from non-integers

epenet:
50 minutes gets stored in the database as (50/60 = 0.83)
When read back, this turns into (0.83*60 = 49.8)

Previous version of db_minutes_to_hhmm would trim down to 49.
New version of db_minutes_to_hhmm rounds it correctly before formatting.

### DIFF
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -815,7 +815,8 @@ function db_now() {
  * @return string representing formatted duration string in hh:mm format.
  */
 function db_minutes_to_hhmm( $p_min = 0 ) {
-	return sprintf( '%02d:%02d', $p_min / 60, $p_min % 60 );
+	$t_min = round($p_min);
+	return sprintf( '%02d:%02d', $t_min / 60, $t_min % 60 );
 }
 
 /**
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -811,11 +811,11 @@ function db_now() {
 
 /**
  * convert minutes to a time format [h]h:mm
- * @param integer $p_min Integer representing number of minutes.
+ * @param float $p_min float representing number of minutes.
  * @return string representing formatted duration string in hh:mm format.
  */
 function db_minutes_to_hhmm( $p_min = 0 ) {
-	$t_min = round($p_min);
+	$t_min = round( $p_min );
 	return sprintf( '%02d:%02d', $t_min / 60, $t_min % 60 );
 }
 


cproensa:
This functions seems to be used solely from time tracking functionality.
I don't have expereience with time tracking, but as i see it,  db_minutes_to_hhmm() receives a value of minutes.
Timetracking value in bugnote table also seems to be stored as minutes.

> 50 minutes gets stored in the database as (50/60 = 0.83

So, When does it happen what you describe?

Do you have a reference to a bugtracker issue that tracks this detected problem, and adds more context?

epenet:
The current function works great with integers, but not with floating point numbers.

The time tracking module provided by the core stores data as a "minutes" integer, however the timetracking (https://github.com/mantisbt-plugins/timetracking) plugin stores time as an "hours" floating point.

There are also some related bugs on the Mantis bug tracker.
http://www.mantisbt.org/bugs/view.php?id=20616
http://www.mantisbt.org/bugs/view.php?id=19557

I can easily tweak my own plugin (https://github.com/epenet/timetracking) to make sure rounding is done "before" calling the function, but I thought it was better to implement inside the core.

core/database_api.php
```diff
@@ -815,7 +815,8 @@ function db_now() {
  * @return string representing formatted duration string in hh:mm format.
  */
 function db_minutes_to_hhmm( $p_min = 0 ) {
-	return sprintf( '%02d:%02d', $p_min / 60, $p_min % 60 );
+	$t_min = round($p_min);
```
vboctor: Add spacing `round( $p_min )`.  Also note that as per your change the phpdoc should be updated since $p_min is no longer required to be an integer.

cproensa:
> I can easily tweak my own plugin (https://github.com/epenet/timetracking) to make sure rounding is done "before" calling the function, but I thought it was better to implement inside the core.

In the line of victor's comments:
I actually think using a helper function that performs the conversion in your plugin is best approach, and not use db_minutes_to_hhmm()
As is see it, that function should not have gone into db api in the first place, as its more suited in either helper_api, string_api, timetracking_api, etc

cproensa:
@epenet are you mantainer of https://github.com/mantisbt-plugins/timetracking?
is that plugin repo active?
you can probably submit the suggestion for rounding as a pull request to the main plugin

dregad:
@epenet following up on the above note, considering that @mykbaker doesn't seem to be actively maintaining his time tracking plugin anymore, if you're interested and unless he objects to it, I can grant you the rights to the repository.

mykbaker:
I think it is a great idea to add @epenet to the repo. I no longer use mantisbt and won't be maintaining the time tracking plugin moving forward.

dregad:
Thanks for the feedback @mykbaker. @epenet let us know if you're willing to take things over.

epenet:
@dregad sure, at least for a little while.
Not planning on making any changes to the v1 plugin though

epenet:
Implemented a new plugin_TimeTracking_hours_to_hhmm function, directly on https://github.com/mantisbt-plugins/timetracking

PR #1088, opened 2017-04-07
Enable scrolling for timeline widget box 

syncguru:
Fixes #22674
Here is the perfect-scrollbar library used in this PR: https://github.com/noraesae/perfect-scrollbar

Snapshots:

screen shot 2017-04-07 at 3 15 20 pm

screen shot 2017-04-07 at 3 16 02 pm


cproensa:
I'm not sure, but maybe this can be achieved with:
```
#timeline-container {
    position: relative;
    height: 74vh;
    overflow-y:auto;
}
```
without using additional library?


Regarding viewport units "vh",i am not expert, but:
Are these unit supported by all browsers that are supported by current mantis versions?
If not, should we define a css fallback?

cproensa:
My point about horizontal scroll is based on this usability issue:
https://www.mantisbt.org/bugs/view.php?id=22132

So, if we are going to include a new library related to scrolling bars, it's preferable to choose one that can solve both situations, instead of eventually adding two libraries
But, as i see it now: my proposal at #22132 uses a jquery plugin that shows a floating scroll for a container that overflows the browser screen area, so probably that is a different situation than what the library in this PR does.

syncguru:
@cproensa Yep - turns out that hv support is quite limited: http://caniuse.com/#feat=viewport-units
I will add standard css to handle the vertical height as in the previous PR.

cproensa:
having both:
```
height: 74vh;
max-height: XXXpx;
```
defeats the purpose of viewport units?

For it to be a fallback, they should override the same property. Should they both be "max-height"?

syncguru:
No. It does not defeat the purpose. It sets an upper limit. 
For old browsers, the upper limit will end up being the actual height. 
For new browsers, the height can be smaller than the max which is desired. It would exceed the max which is set differently for different media directives.

cproensa:
>  It sets an upper limit.
> For old browsers, the upper limit will end up being the actual height.

We are talking opposite then

With the latest css, i get this view:

seleccion_164

It has an effective _height_ of 600px. 74vh is greater than 600px, so the limit set by _max-height_ is used.
I expect that timeline would use the vertical space available.

syncguru:
I was intentionally leaving this scenario out. But I updated the css to handle your request.

cproensa:
> I was intentionally leaving this scenario out. But I updated the css to handle your request.

I don't know.. i pointed it out from the view of a user.
The motivation is that the timeline has a "vertical area" look and feel. I understood your feature as a mean to limit it's height when there are too many history entries, which is reasonable as in that case the vertical size can be too big, and forces the user to scroll the page down.

cproensa:
Did you have opportunity to review this?:
>I'm not sure, but maybe this can be achieved with:
    overflow-y:auto;

Seems like scrolling an overflowed container, like proposed here, is a basic feature covered by standard css (unless i'm bypassing any other consideration)

syncguru:
@cproensa I don't believe that's true. Have you verified your suggestion?

cproensa:
> @cproensa I don't believe that's true. Have you verified your suggestion?

i think yes, i had a quick test with this PR (at the initial state), by removing the library include, and adding that css bit to timeline style
Unless mistaken, seemed fine. 
Again, don't know if i missed something there...

syncguru:
@cproensa you are right. It is possible without the library with the new div/css org.
Will remove the lib shortly.

cproensa:
> @cproensa you are right. It is possible without the library with the new div/css org.

thanks, so im not crazy then :)

i think the library is aiming to provide a precisely positioned scroll, that doesn't affect layout misheaps, jumps, etc? That may be a use case, but not for our need though.

syncguru:
You are certainly not crazy. Ace theme used a plugin for scrolling but I guess it was mostly related to look and feel vs. functionality.

cproensa:
Now a cosmetic issue:
by using _height_, the timeline container always has a fixed size, even if no data exists
if using _max-height_ instead, the size is variable until the limit is reached

what do you think?
use of _max-height_ is closer to the description in 22674: _limit the length of the timeline box but at the same time support scrolling_
and, it keeps the look that current timeline has for a small set of rows.

I'm not strongly biased over one or another. 


Also, once finished:
- consider renaming the PR title to better reflect current changes
- the commits would benefit from squashing to remove noise.

syncguru:
@cproensa We need to have a visible box even if there are no timeline entries. I would just stick with the height.

cproensa:
> We need to have a visible box even if there are no timeline entries. I would just stick with the height.

ok

vboctor:

PR #1090, opened 2017-04-10
Fixes #22704 Due date non english format

jdgrieco:
Get timestamp for non english due date format (Fixes #22704)

### DIFF
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -242,7 +242,7 @@ layout_page_begin();
 		</th>
 		<td>
 			<input type="text" id="due_date" name="due_date" class="datetimepicker input-sm" size="16" maxlength="20"
-				data-picker-locale="<?php lang_get_current_datetime_locale() ?>"
+				data-picker-locale="<?php echo lang_get_current_datetime_locale() ?>"
 				data-picker-format="<?php echo convert_date_format_to_momentjs( config_get( 'normal_date_format' ) ) ?>"
 				<?php helper_get_tab_index() ?> value="<?php echo $t_date_to_display ?>" />
 			<i class="fa fa-calendar fa-xlg datetimepicker"></i>
--- a/bug_update.php
+++ b/bug_update.php
@@ -84,7 +84,8 @@ if( $t_due_date !== null ) {
 	if( is_blank( $t_due_date ) ) {
 		$t_updated_bug->due_date = 1;
 	} else {
-		$t_updated_bug->due_date = strtotime( $t_due_date );
+	    $t_due_date = DateTime::createFromFormat( config_get( 'normal_date_format' ), $t_due_date );
+		$t_updated_bug->due_date = $t_due_date->getTimestamp();
 	}
 }
 $t_updated_bug->duplicate_id = gpc_get_int( 'duplicate_id', 0 );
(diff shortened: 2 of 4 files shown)


bug_change_status_page.php
```diff
@@ -242,7 +242,7 @@
 		</th>
 		<td>
 			<input type="text" id="due_date" name="due_date" class="datetimepicker input-sm" size="16" maxlength="20"
-				data-picker-locale="<?php lang_get_current_datetime_locale() ?>"
+				data-picker-locale="<?php echo lang_get_current_datetime_locale() ?>"
```
vboctor: This change has already been merged and it shouldn't be in this PR.

bug_update.php
```diff
@@ -84,7 +84,8 @@
 	if( is_blank( $t_due_date ) ) {
 		$t_updated_bug->due_date = 1;
 	} else {
-		$t_updated_bug->due_date = strtotime( $t_due_date );
+	    $t_due_date = DateTime::createFromFormat( config_get( 'normal_date_format' ), $t_due_date );
```
vboctor: Fix indentation.

core/bug_api.php
```diff
@@ -314,7 +314,12 @@ public function __set( $p_name, $p_value ) {
 				break;
 			case 'due_date':
 				if( !is_numeric( $p_value ) ) {
-					$p_value = strtotime( $p_value );
+				    $p_value = DateTime::createFromFormat( config_get( 'normal_date_format' ), $p_value );
+				    if ( false === $p_value ) {
+				        $p_value = date_get_null();
+                    } else {
```
vboctor: Fix indentation

bug_update.php
vboctor: Should you handle false value here?

core/bug_api.php
```diff
@@ -314,7 +314,12 @@ public function __set( $p_name, $p_value ) {
 				break;
 			case 'due_date':
 				if( !is_numeric( $p_value ) ) {
-					$p_value = strtotime( $p_value );
+				    $p_value = DateTime::createFromFormat( config_get( 'normal_date_format' ), $p_value );
```
vboctor: A couple of questions to consider:
- Should be using db_null() or triggering an error?
- Should this method or another method return date_get_null() when date is not valid?

core/bug_api.php
jdgrieco: @vboctor  Sorry about incorret indentation

vboctor:
Commit messages should also include "Fixes #" followed by issue number (without space).

jdgrieco:
@vboctor In that case I should not have put the PR in the issue description, right? Is the acceptable way to add this PR in a note?

jdgrieco:
@syncguru As mentioned, I think this issue can exists in other date fields as well. Maybe in custom date fields! I not sure.

cproensa:
The fixes seem that will be included in #1079
Let's keep it there to not interfere with @syncguru work

cproensa:
This PR should be closed, as the original changes from #1032 have been reverted
The date picker will be reimplemented with the known issues in mind.

Anyway, thanks @jdgrieco for your submission

PR #1091, opened 2017-04-13
Filter load save refactor

cproensa:
This PR is cleaning and rewriting the filters functionality that deals with loading and storing filters.

Mainly:
- Clean up duplicated or unneeded code, rename some functions and variables to better match the current concepts
- Encapsulate the logic for loading and savig filters from database, which previously was duplicated in several places, doing ad-hoc un/serialization.
- Remove the usage of cookies to track filters
- Remove the usage of token table to track temporary filters
- Implement temporary filters stored in user session.
- Ensure that temporary filters are tracked through most of the actions in view_all_page

Directly fixes:
0008167: Filter settings saved when using Anonymous account
0008204: Filters not remembered when clicking through from "My View"

Some background on how filters worked so far:

There is a cookie "view_all_cookie". In the first implementation of filters, long time ago, this cookie would store the filter properties on client side.
Then (around 2004), stored filters were created. Also the filters table would store the user filter properties for the in-use filter settings. Now the cookie only stores a filter id from that table.
The rows in filter table that stores "in use" filters are stored with negative project ids, as a hack to not mix with proper stored filters. This way a user can have a different working filter "remembered" for each project.
*However, i dont't think that really was working now, because i see the cookie id always the same, even after changing projects.*

Additionally, there are temporary filters, which is a filter that is not to be stored in database (as the "working" filters for each project). The solution was to use the tokens table, where the TOKEN_FILTER entry for each user, stores the serialized data of a temporary filter. 
To use a temporary filter, a parameter "filter" was passed around with the token id (but not everywhere, so far this only worked for pagination).
Using tokens also have the disadvantage that only one token value can exist for each user. The token id actually was not used, as the token values is nos accessible by token_id, just by "fetching the value for said token type"

On top of that, said "working" filters and temporary filters as token, only support *one* filter for each scenario, per user. So we have the case of anonymous user, or any other concurrent login, that is interfering each other, as all of the real users are modifying the same filter.

What This PR changes:

- Remove use of the cookie. There's no need for it, since whenever we need the working "filter", go directly to filters table.
- Remove use of the filter token. Since this was not optimal, storing temporary filters has been implemented in session data.
- A real filter key is used so that multiple temporary filters can exists for each user session. The filter is now tracked through most of the actions in view_all_page, modifications to the temporary filter are kept to that temp filter, saving a temp filter now actually saves it, etc.

However, using a non-temporary filter still uses the filters table "unique per project" working filters.
For solving the anonymous user scenario, all the filters are forced to be temporary for this user.

It mainly affects to temporary data, so there should not be any significant issue with backwards compatibility
As always, some bugs and typos can happen. Testing is needed...

_Some potential changes, to be evaluated:_

All the filters used could be temporary. This has a better experience overall, when several open tabs track different filter properties, and does not interfere between them. The database (unnamed) filters could be redefined as "default filters" for each project, where the user, supported by some UI element, be able to set a filter to be the deafult for each project (currently the user cannot control how or when these last-used filters are loaded, and can be confusing)

When using the temporary filters, a new copy can be made after each modification. This means having a new array, with different id, stored for each filter change. This allows compatibility with browser navigation, where for example, going back in the browser after a filter change, recovers the state of the previous filter/view. On the bad side, this would increase session storage, and would be better once we have implemented the plan to store filters only as the modified properties, not the whole array.

### DIFF
--- a/core/current_user_api.php
+++ b/core/current_user_api.php
@@ -236,9 +236,9 @@ function current_user_ensure_unprotected() {
  * @access public
  */
 function current_user_get_bug_filter( $p_project_id = null ) {
-	$f_tmp_key = gpc_get_string( 'filter', 0 );
+	$f_tmp_key = gpc_get_string( 'filter', null );
 
-	if( 0 !== $f_tmp_key ) {
+	if( null !== $f_tmp_key ) {
 		$t_filter = filter_temporary_get( $f_tmp_key, filter_get_default() );
 	} else {
 		$t_user_id = auth_get_current_user_id();
--- a/core/current_user_api.php
+++ b/core/current_user_api.php
@@ -249,8 +249,6 @@ function current_user_get_bug_filter( $p_project_id = null ) {
 			$t_filter = json_decode( $t_token, true );
 		}
 		$t_filter = filter_ensure_valid_filter( $t_filter );
-	} else if( !filter_is_cookie_valid() ) {
-		$t_filter = filter_get_default();
 	} else {
 		$t_user_id = auth_get_current_user_id();
 		$t_filter = user_get_bug_filter( $t_user_id, $p_project_id );
--- a/core/current_user_api.php
+++ b/core/current_user_api.php
@@ -228,26 +228,25 @@ function current_user_ensure_unprotected() {
 }
 
 /**
- * Returns the issue filter parameters for the current user
+ * Returns the issue filter for the current user, which is retrieved by
+ * evaluating these steps:
+ * 1) Reads gpc vars for a token id, which means to load a temporary filter
+ * 2) Otherwise, get the filter saved as current, for the user, project 
  *
- * @param integer $p_project_id Project id. This argument is only used if a 'filter' string is not passed via the web request.
- *                              The default value is null meaning return the current filter for user's current project
-                                if a filter string is not supplied.
- * @return array User filter, if not set, then default filter.
+ * @param integer $p_project_id Project id to get the user's filter from, if needed.
+ * @return array	A filter array
  * @access public
  */
 function current_user_get_bug_filter( $p_project_id = null ) {
-	$f_filter_string = gpc_get_string( 'filter', '' );
-	$t_filter = array();
+	$f_filter_token = gpc_get( 'filter', null );
 
-	if( !is_blank( $f_filter_string ) ) {
-		if( is_numeric( $f_filter_string ) ) {
-			$t_token = token_get_value( TOKEN_FILTER );
-			if( null != $t_token ) {
-				$t_filter = json_decode( $t_token, true );
-			}
-		} else {
-			$t_filter = json_decode( $f_filter_string, true );
+	if( null !== $f_filter_token && token_exists( (int)$f_filter_token ) ) {
+		# If the token id exists, try to load the value
+		# At this point, only one value can exists for each token type and user
+		# so read the token based on type, regardless of the id that was provided
+		$t_token = token_get_value( TOKEN_FILTER );
+		if( null != $t_token ) {
+			$t_filter = json_decode( $t_token, true );
 		}
 		$t_filter = filter_ensure_valid_filter( $t_filter );
 	} else if( !filter_is_cookie_valid() ) {
--- a/core/current_user_api.php
+++ b/core/current_user_api.php
@@ -28,7 +28,6 @@
  * @uses filter_api.php
  * @uses gpc_api.php
  * @uses helper_api.php
- * @uses tokens_api.php
  * @uses user_api.php
  * @uses user_pref_api.php
  * @uses utility_api.php
@@ -39,7 +38,6 @@ require_api( 'constant_inc.php' );
 require_api( 'filter_api.php' );
 require_api( 'gpc_api.php' );
 require_api( 'helper_api.php' );
-require_api( 'tokens_api.php' );
 require_api( 'user_api.php' );
 require_api( 'user_pref_api.php' );
 require_api( 'utility_api.php' );
@@ -238,17 +236,10 @@ function current_user_ensure_unprotected() {
  * @access public
  */
 function current_user_get_bug_filter( $p_project_id = null ) {
-	$f_filter_token = gpc_get( 'filter', null );
+	$f_tmp_key = gpc_get_string( 'filter', 0 );
 
-	if( null !== $f_filter_token && token_exists( (int)$f_filter_token ) ) {
-		# If the token id exists, try to load the value
-		# At this point, only one value can exists for each token type and user
-		# so read the token based on type, regardless of the id that was provided
-		$t_token = token_get_value( TOKEN_FILTER );
-		if( null != $t_token ) {
-			$t_filter = json_decode( $t_token, true );
-		}
-		$t_filter = filter_ensure_valid_filter( $t_filter );
+	if( 0 !== $f_tmp_key ) {
+		$t_filter = filter_temporary_get( $f_tmp_key, filter_get_default() );
 	} else {
 		$t_user_id = auth_get_current_user_id();
 		$t_filter = user_get_bug_filter( $t_user_id, $p_project_id );
--- a/core/current_user_api.php
+++ b/core/current_user_api.php
@@ -239,7 +239,11 @@ function current_user_get_bug_filter( $p_project_id = null ) {
 	$f_tmp_key = gpc_get_string( 'filter', null );
 
 	if( null !== $f_tmp_key ) {
-		$t_filter = filter_temporary_get( $f_tmp_key, filter_get_default() );
+		$t_filter = filter_temporary_get( $f_tmp_key, null );
+		# if filter doesn't exist or can't be loaded, return a default filter (doesn't throw error)
+		if( null === $t_filter ) {
+			$t_filter = filter_get_default();
+		}
 	} else {
 		$t_user_id = auth_get_current_user_id();
 		$t_filter = user_get_bug_filter( $t_user_id, $p_project_id );
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -3756,3 +3756,56 @@ function filter_get_included_projects( array $p_filter, $p_project_id = null, $p
 
 	return $t_project_ids;
 }
+
+/**
+ * Returns a filter array structure for the given filter_id
+ * A default value can be provided to be used when the filter_id doesn't exists
+ * or is not accesible
+ *
+ *  You may pass in any array as a default (including null) but if
+ *  you pass in *no* default then an error will be triggered if the filter
+ *  cannot be found
+ *
+ * @param integer $p_filter_id Filter id
+ * @param array $p_default     A filter array to return when id is not found
+ * @return array	A filter array
+ */
+function filter_get( $p_filter_id, array $p_default = null ) {
+	# if no default was provided, we will trigger an error if not found
+	$t_trigger_error = func_num_args() == 1;
+
+	# This function checks for user access
+	$t_filter_string = filter_db_get_filter( $p_filter_id );
+	# If value is false, it either doesn't exists or is not accesible
+	if( !$t_filter_string ) {
+		if( $t_trigger_error ) {
+			error_parameters( $p_filter_id );
+			trigger_error( ERROR_FILTER_NOT_FOUND, ERROR );
+		} else {
+			return $p_default;
+		}
+	}
+	$t_filter = filter_deserialize( $t_filter_string );
+
+	# Check if the filter references a named filter
+	# This property only makes sense, and should be available on unnamed filters
+	# Note that an unnamed filter is used as a working copy for the user. This logic
+	# allows for automatic updating of a current filter when the user had selected
+	# a named filter and that filter has changed. Otherwise the updates in base filter
+	# wuould not reflect in the user's current filter
+	if( isset( $t_filter['_source_query_id'] ) && $p_filter_id != $t_filter['_source_query_id'] ) {
+		$t_source_query_id = $t_filter['_source_query_id'];
+		# check if filter id is a proper named filter, and is accesible
+		if( filter_is_named_filter( $t_source_query_id ) && filter_is_accessible( $t_source_query_id ) ){
+			# replace filter with the referenced one
+			$t_filter = filter_deserialize( filter_db_get_filter( $t_source_query_id ) );
+			# update the referenced stored filter id for the new loaded filter
+			$t_filter['_source_query_id'] = $t_source_query_id;
+		} else {
+			# If the filter id is not valid, clean the referenced filter id
+			unset( $t_filter['_source_query_id'] );
+		}
+	}
+
+	return $t_filter;
+}
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -1000,7 +1000,7 @@ function filter_deserialize( $p_serialized_filter ) {
  */
 function filter_serialize( $p_filter_array ) {
 	$t_cookie_version = FILTER_VERSION;
-	filter_clean_runtime_properties( $p_filter_array );
+	$p_filter_array = filter_clean_runtime_properties( $p_filter_array );
 	$t_settings_serialized = json_encode( $p_filter_array );
 	$t_settings_string = $t_cookie_version . '#' . $t_settings_serialized;
 	return $t_settings_string;
@@ -3740,30 +3740,41 @@ function filter_get( $p_filter_id, array $p_default = null ) {
 		}
 	}
 	$t_filter = filter_deserialize( $t_filter_string );
+	$t_filter = filter_clean_runtime_properties( $t_filter );
+	$t_filter['_filter_id'] = $p_filter_id;
 
+	$t_filter = filter_update_source_properties( $t_filter );
+
+	return $t_filter;
+}
+
+function filter_update_source_properties( array $p_filter ) {
 	# Check if the filter references a named filter
 	# This property only makes sense, and should be available on unnamed filters
 	# Note that an unnamed filter is used as a working copy for the user. This logic
 	# allows for automatic updating of a current filter when the user had selected
 	# a named filter and that filter has changed. Otherwise the updates in base filter
 	# wuould not reflect in the user's current filter
-	if( isset( $t_filter['_source_query_id'] ) && $p_filter_id != $t_filter['_source_query_id'] ) {
-		$t_source_query_id = $t_filter['_source_query_id'];
+	if( isset( $p_filter['_filter_id'] ) ) {
+		$t_filter_id = $p_filter['_filter_id'];
+	} else {
+		$t_filter_id = null;
+	}
+	if( isset( $p_filter['_source_query_id'] ) && $t_filter_id != $p_filter['_source_query_id'] ) {
+		$t_source_query_id = $p_filter['_source_query_id'];
 		# check if filter id is a proper named filter, and is accesible
 		if( filter_is_named_filter( $t_source_query_id ) && filter_is_accessible( $t_source_query_id ) ){
 			# replace filter with the referenced one
-			$t_filter = filter_deserialize( filter_db_get_filter_string( $t_source_query_id ) );
+			$t_new_filter = filter_deserialize( filter_db_get_filter_string( $t_source_query_id ) );
 			# update the referenced stored filter id for the new loaded filter
-			$t_filter['_source_query_id'] = $t_source_query_id;
+			$t_new_filter['_source_query_id'] = $t_source_query_id;
+			$p_filter = filter_copy_runtime_properties( $p_filter, $t_new_filter );
 		} else {
 			# If the filter id is not valid, clean the referenced filter id
-			unset( $t_filter['_source_query_id'] );
+			unset( $p_filter['_source_query_id'] );
 		}
 	}
-
-	$t_filter['_filter_id'] = $p_filter_id;
-
-	return $t_filter;
+	return $p_filter;
 }
 
 /**
@@ -3820,7 +3831,7 @@ function filter_temporary_set( array $p_filter, $p_filter_key = null ) {
 		$t_filter_key = $p_filter_key;
 	}
 
-	filter_clean_runtime_properties( $p_filter );
+	$p_filter = filter_clean_runtime_properties( $p_filter );
 	$t_session_filters = session_get( 'temporary_filters', array() );
 	$t_session_filters[$t_filter_key] = $p_filter;
 	session_set( 'temporary_filters', $t_session_filters );
@@ -3879,12 +3890,22 @@ function filter_get_temporary_key_param( $p_key_or_filter ) {
  * @param array $p_filter	Filter array (passed as reference, it gets modified)
  * @return array	Modified filter array
  */
-function filter_clean_runtime_properties( array &$p_filter ) {
-	if( isset( $p_filter['_temporary_id'] ) ) {
-		unset( $p_filter['_temporary_id'] );
+function filter_clean_runtime_properties( array $p_filter ) {
+	if( isset( $p_filter['_temporary_key'] ) ) {
+		unset( $p_filter['_temporary_key'] );
 	}
 	if( isset( $p_filter['_filter_id'] ) ) {
 		unset( $p_filter['_filter_id'] );
 	}
 	return $p_filter;
+}
+
+function filter_copy_runtime_properties( array $p_filter_from, array $p_filter_to ) {
+	if( isset( $p_filter_from['_temporary_key'] ) ) {
+		$p_filter_to['_temporary_key'] = $p_filter_from['_temporary_key'];
+	}
+	if( isset( $p_filter_from['_filter_id'] ) ) {
+		$p_filter_to['_filter_id'] = $p_filter_from['_filter_id'];
+	}
+	return $p_filter_to;
 }
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -951,6 +951,10 @@ function filter_get_default() {
 
 /**
  * Deserialize filter string
+ * Expected strings have this format: "<version>#<json string>" where:
+ * - <version> is the versio number of the filter structure used. See constant FILTER_VERSION
+ * - # is a separator
+ * - <json string> is the json encoded filter array.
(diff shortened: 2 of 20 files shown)


cproensa:
@atrol, this commit c5aabcb includes support for reading old stored filters
This may solve some of the issues regarding errors after mantis upgrades (you have reviewed those issues at some point)

I think that if install_stored_filter_migrate() failed at some upgrade step, the filters are still upgraded (if possible) or ignored on runtime.
Possibly, install_stored_filter_migrate() could also be modified to not fail if the conversion can not be made, as now the left over filters will be more gracefully handled.

dregad:
@cproensa are you still working on this, or is it ready to merge ?

cproensa:
>  are you still working on this, or is it ready to merge ?

It is completed as is.

However, feedback may be provided that may lead to changes.

core/filter_api.php
```diff
@@ -3434,3 +3418,202 @@ function filter_print_view_type_toggle( $p_url, $p_view_type ) {
...
+	$t_session_filters = session_get( 'temporary_filters', array() );
+	if( isset( $t_session_filters[$p_filter_key] ) ) {
+		# setting here the key in the filter array only if the key exists
+		# this validates against receiving garbage input as XSS attacks
+		$t_filter = $t_session_filters[$p_filter_key];
+		$t_filter['_temporary_key'] = $p_filter_key;
+		return filter_ensure_valid_filter( $t_filter );
+	} else {
+		if( $t_trigger_error ) {
+			error_parameters( $p_filter_id );
```
atrol: $p_filter_key

core/print_api.php
```diff
@@ -1524,19 +1531,19 @@ function print_small_button( $p_link, $p_url_text, $p_new_window = false ) {
  * @param string  $p_text           The displayed text for the link.
  * @param integer $p_page_no        The page number to link to.
  * @param integer $p_page_cur       The current page number.
- * @param integer $p_temp_filter_id Temporary filter id.
+ * @param integer $p_temp_filter_key Temporary filter id.
```
atrol: change also description from id to key

core/print_api.php
```diff
@@ -1552,12 +1559,17 @@ function print_page_link( $p_page_url, $p_text = '', $p_page_no = 0, $p_page_cur
  * @param integer $p_start          The first page number.
  * @param integer $p_end            The last page number.
  * @param integer $p_current        The current page number.
- * @param integer $p_temp_filter_id Temporary filter id.
+ * @param integer $p_temp_filter_key Temporary filter id.
```
atrol: change also description from id to key

core/filter_api.php
cproensa: i can't see where $p_filter_key is used in that code...

atrol: $p_filter_key is the first parameter of the function.
$p_filter_id is undefined.

core/current_user_api.php
```diff
@@ -228,30 +226,20 @@ function current_user_ensure_unprotected() {
...
 /**
- * Returns the issue filter parameters for the current user
+ * Returns the issue filter for the current user, which is retrieved by
+ * evaluating these steps:
+ * 1) Reads gpc vars for a token id, which means to load a temporary filter
+ * 2) Otherwise, get the filter saved as current, for the user, project 
  *
- * @param integer $p_project_id Project id. This argument is only used if a 'filter' string is not passed via the web request.
- *                              The default value is null meaning return the current filter for user's current project
-                                if a filter string is not supplied.
```
dregad: Missing ` *` at beginning of line

dregad:
Forgot about that PR... nice work @cproensa .
I can't remember if I tested it or just reviewed the code when I looked at it back in May. Will perform some tests later (as time allows)

core/filter_api.php
cproensa: OK, the review here is anchored at filter_get(), that why i was confused.
But you mean in filter_temporary_get(). I got it.

core/current_user_api.php
cproensa: That's on the removed code

cproensa:
updated with @atrol comments. Anyway, refresh your local branch, and this has been rebased and may be inconsistencies

dregad:
@cproensa looks like this needs another round of conflict resolution...

cproensa:
> looks like this needs another round of conflict resolution...

strange, a simple rebase to master has fixed it, no conflicts.

issues_rss.php
```diff
@@ -172,17 +172,18 @@
 }
 $t_show_sticky = null;
 
+# Override current user
+current_user_set( $t_user_id );
```
vboctor: I usually use `auth_attempt_script_login()` for impersonating a user for running some logic.  It has more logic than what `current_user_set()`.

print_all_bug_page.php
```diff
@@ -72,21 +72,11 @@
...
-	$t_filter_cookie_arr = filter_get_default();
-} else {
-	# check to see if new cookie is needed
-	if( !filter_is_cookie_valid() ) {
-		print_header_redirect( 'view_all_set.php?type=0&print=1' );
-	}
-	$t_setting_arr = explode( '#', $t_cookie_value, 2 );
-	$t_filter_cookie_arr = json_decode( $t_setting_arr[1], true );
-}
+$t_filter = filter_get( $t_cookie_value_id, filter_get_default() );
```
vboctor: Should we only create the default filter when needed, rather than constructing it and not using it in most cases?

view_all_set.php
```diff
@@ -106,31 +106,20 @@
...
-		$t_setting_arr = filter_ensure_valid_filter( $t_setting_arr );
-	}
-} else {
+$t_cookie_name = config_get( 'view_all_cookie' );
+$t_cookie_filter_id = gpc_get_cookie( $t_cookie_name, null );
+if( null === $t_cookie_filter_id ) {
 	# no cookie found, set it
 	$f_type = 1;
+} else {
+	$t_setting_arr = filter_get( $t_cookie_filter_id, filter_get_default() );
```
vboctor: Same comment about avoiding creating default filter when not needed.

view_all_set.php
```diff
@@ -106,31 +106,20 @@
...
 }
 
 # Clear the source query id.  Since we have entered new filter criteria.
-$t_setting_arr['_source_query_id'] = '';
+if( isset( $t_setting_arr['_source_query_id'] ) ) {
```
vboctor: Should we create a constant for strings like `_source_query_id`?

view_all_set.php
```diff
@@ -152,22 +141,16 @@
...
-		$t_setting_arr = filter_deserialize( $t_filter_string );
-		if( false === $t_setting_arr ) {
-			# couldn't deserialize, if we were trying to use the filter, clear it and reload
-			gpc_clear_cookie( 'view_all_cookie' );
+		$t_setting_arr = filter_get( $f_source_query_id, null );
+		if( null === $t_setting_arr ) {
+			# couldn't get the filter, if we were trying to use the filter, clear it and reload
+			gpc_clear_cookie( $t_cookie_name );
 			error_proceed_url( 'view_all_set.php?type=0' );
-			trigger_error( ERROR_FILTER_TOO_OLD, ERROR );
```
vboctor: Is `ERROR_FILTER_TOO_OLD` still being used?

query_store.php
```diff
@@ -97,10 +97,19 @@
...
-$t_new_row_id = filter_db_set_for_current_user( $t_project_id, $f_is_public,
-												$f_query_name, $t_filter_string );
+# Check that the user has permission to create stored filters
+if( !access_has_project_level( config_get( 'stored_query_create_threshold' ) ) ) {
+	access_denied();
+}
+
+# ensure that we're not making this filter public if we're not allowed
+if( !access_has_project_level( config_get( 'stored_query_create_shared_threshold' ) ) ) {
+	$f_is_public = false;
```
vboctor: Should this trigger access denied rather than changing the script parameters?  This case shouldn't typically happen anyways since form should do such validation.

core/filter_api.php
```diff
@@ -3661,3 +3663,91 @@ function filter_get( $p_filter_id, array $p_default = null ) {
...
+ * If no key is provided, it will search in the filter property that holds
+ * its key if it was loaded as a temporary filter.
+ * If neither key is found, a new one will be created
+ * @param array $p_filter     Filter array
+ * @param type $p_filter_key  Key to update, or null
+ * @return string	The key used for storing the filter.
+ */
+function filter_temporary_set( array $p_filter, $p_filter_key = null ) {
+	if( null === $p_filter_key ) {
+		if( isset( $p_filter['_temporary_key'] ) ) {
```
vboctor: Should this leverage `filter_get_temporary_key()`?

core/filter_api.php
```diff
@@ -3661,3 +3663,91 @@ function filter_get( $p_filter_id, array $p_default = null ) {
...
+ * If the filter was not originated from a temporary key, returns an empty string
+ * @param array $p_filter	Filter array
+ * @return string	Formatted parameter string, or empty
+ */
+function filter_get_temporary_key_param( array $p_filter ) {
+	$t_key = filter_get_temporary_key( $p_filter );
+	if( $t_key ) {
+		return 'filter=' . $t_key;
+	} else {
+		return '';
```
vboctor: Should this be consistent with `filter_get_temporary_key()` and return null when not set?

core/filter_api.php
```diff
@@ -3656,30 +3656,41 @@ function filter_get( $p_filter_id, array $p_default = null ) {
 		}
 	}
 	$t_filter = filter_deserialize( $t_filter_string );
+	$t_filter = filter_clean_runtime_properties( $t_filter );
+	$t_filter['_filter_id'] = $p_filter_id;
```
vboctor: Should a constant be created for `_filter_id`?

query_store.php
```diff
@@ -117,5 +117,15 @@
...
 } else {
-	print_header_redirect( 'view_all_bug_page.php' );
+	# Build a redirect to view_all_set to load the filter that was saved.
+	# This will make the filter name appear as selected in the filter selection box.
+	$t_params = array(
+		'type' => 3,
+		'source_query_id' => $t_new_row_id
+	);
+	if( filter_is_temporary( $t_filter ) ) {
+		$t_params['filter'] = filter_get_temporary_key( $t_filter );
```
vboctor: Make `filter` a constant.

core/filter_api.php
```diff
@@ -992,25 +992,29 @@ function filter_deserialize( $p_serialized_filter ) {
 	# in this case, the filter version field inside the array is to be used
 	# and if not present, set the current filter version
 
-	# check to see if new cookie is needed
+	# check fitler version mark
 	$t_setting_arr = explode( '#', $p_serialized_filter, 2 );
-	if( ( $t_setting_arr[0] == 'v1' ) || ( $t_setting_arr[0] == 'v2' ) || ( $t_setting_arr[0] == 'v3' ) || ( $t_setting_arr[0] == 'v4' ) ) {
+	$t_version_string = $t_setting_arr[0];
+	if( in_array( $t_version_string, array( 'v1', 'v2', 'v3', 'v4' ) ) ) {
```
vboctor: Should the upgrade logic here be part of the filter_upgrade method that you added?

core/filter_form_api.php
```diff
@@ -2391,8 +2391,10 @@ function filter_form_draw_inputs( $p_filter, $p_for_screen = true, $p_static = f
 		if( $p_static) {
 			return $p_label;
 		} else {
-			if( $t_source_query_id > 0 ) {
-				$t_data_filter_id = ' data-filter_id="' . $t_source_query_id . '"';
+			if( filter_is_temporary( $t_filter ) ) {
```
vboctor: Should we use a different format for temp vs. permanent filters and use one filter id query string to represent both?  We can even have filter_get() handle both?

issues_rss.php
cproensa: As i see: `rss_login()'  which is used previously in the code, is already calling `auth_attempt_script_login()`
I'm not sure what's is best here. Since this may imply that a user (named, or anonymous) has been logged in by those calls, may this `current_user_set()` be simply removed?

core/filter_api.php
cproensa: `filter_version_upgrade()` deals with the contents of the filter array.
That code in `filter_deserialize()` deals with how the data is physically stored in db. It's not actually upgrading anything, only trying to read it properly

query_store.php
cproensa: That code moved "as is" from `filter_db_set_for_current_user()`
you may be right

query_store.php
cproensa: We dont usually make constants for gpc parameter names. However, i see the benefit of having it as a constant.
There should be a future rewrite for how the filters are propagated through links, aiming at using filters in other contexts like summary pages, etc
I'd rather leave this point to that future work.

view_all_set.php
cproensa: Filter properties starting with "_" should be replaced with constant names. This have been discussed for some time.
The inconvenience is that this will touch a lot of code, and i have some parallel work in progress for more incremental changes to filters. That would cause some extensive conflicts for no functionality gains.
I have it in the to-do list, when there is some stabilization between PRs

print_all_bug_page.php
cproensa: I understand.
note that your comments are on outdated code. Now the offending call is placed in `current_user_get_bug_filter()`
affects `filter_get()` and `filter_temporary_get()`

I think of two options:
1- If the default value is passed as "true", a defaulted filter is built only if needed as return vale inside those functions
2- `filter_get_default()` has a static filter array, so it has to be built only once. Similar to what is done in `filter_get_default_array()`

The default filter concept needs to be reviewed in future works, aiming at: differentiate "empty" from "default" (as system default, or default per user) when implementing filters that only stores the modified values (simplify filter dialogs)

core/filter_form_api.php
cproensa: in this PR, and due to inheriting part of that logic, the usage of a database saved filter is differentiated by using "filter_id" parameter, versus a temporary with "filter" parameter, on page requests.
This could be joined into a single value, for example, inferring that numerical ids are db-filters, and string ids are temporary filters.
This would fit in a more extensive work where the usage of filters across pages is cleaned up and improved (i commented on it in other reply).

A better way of dealing this has deeper implications that i'd rather leave out of this PR, but is due to be redesigned in the future.

view_all_set.php
cproensa: it's not referenced any more

core/filter_api.php
cproensa: ok, updated

core/filter_api.php
cproensa: ok, updated

query_store.php
cproensa: ok, updated

view_all_set.php
cproensa: ok, updated

cproensa:
@vboctor 
Changed 3bf8403 by removing only the lang string.
If i understand, this would propagate to other lang files with next translation updates, and then the constant could be finally removed.

atrol:
> If i understand, this would propagate to other lang files with next translation updates, and then the constant could be finally removed.

Messages don't get automatically removed in all language files.

When running admin/test_langs.php you get some messages like

```
Testing language file 'strings_croatian.txt' (phase 1)...
WARNING: '$s_file_upload_msg' is not defined in the English language file
WARNING: '$s_bugnote_added_msg' is not defined in the English language file
```

So it seems that the generation of a certain language file gets just triggered if someone changed a message in it, but not if a message has been deleted in master (English) language.

@translatewiki could you explain how it works and maybe change your scripts to consider deleted messages?

print_all_bug_page.php
cproensa: @vboctor 
overlooked the simpler solution: it's fixed in e0c048c3bc620adee83400b5f3e70ad71136fb63

cproensa:
any feedback on this?

core/filter_api.php
vboctor: Makes sense.

query_store.php
vboctor: ok

view_all_set.php
vboctor: ok

vboctor:
@cproensa please update the PR to address the conflicts so we can merge.

core/filter_api.php
```diff
@@ -539,6 +556,30 @@ function filter_ensure_fields( array $p_filter_arr ) {
 	return $p_filter_arr;
 }
 
+/**
+ * A wrapper to compare filter version syntax
+ * Note: Currently, filter versions hve this syntax: "vN",  * where N is a integer number.
+ * @param string $p_version1    First version number
+ * @param string $p_version2    Second version number
+ * @param string $p_operator    Comparison test, if provided. As expected by version_compare()
+ */
```
atrol: missing @return

core/filter_api.php
```diff
@@ -539,6 +556,30 @@ function filter_ensure_fields( array $p_filter_arr ) {
...
+function filter_version_compare( $p_version1, $p_version2, $p_operator = null ) {
+	return version_compare( $p_version1, $p_version2, $p_operator );
+}
+
+/**
+ * Upgrade a filter array to the current filter structure, by converting properties
+ * that have changed from previous filter versions
+ * @param array $p_filter	Filter array to upgrade
+ */
```
atrol: missing @return

core/filter_api.php
```diff
@@ -3667,3 +3651,226 @@ function filter_get_included_projects( array $p_filter, $p_project_id = null, $p
...
+}
+
+/**
+ * Saves a filter as a temporary filter in session data.
+ * The filter will be updated or created, indexed by provided $p_filter_key,
+ * If no key is provided, it will search in the filter property that holds
+ * its key if it was loaded as a temporary filter.
+ * If neither key is found, a new one will be created
+ * @param array $p_filter     Filter array
+ * @param type $p_filter_key  Key to update, or null
```
atrol: @param string

lang/strings_english.txt
```diff
@@ -1654,7 +1654,6 @@ $MANTIS_ERROR[ERROR_GPC_ARRAY_UNEXPECTED] = 'A string was expected but an array
 $MANTIS_ERROR[ERROR_GPC_NOT_NUMBER] = 'A number was expected for %1$s.';
 $MANTIS_ERROR[ERROR_BUG_NOT_FOUND] = 'Issue %1$d not found.';
 $MANTIS_ERROR[ERROR_FILTER_NOT_FOUND] = 'Filter %1$s not found.';
-$MANTIS_ERROR[ERROR_FILTER_TOO_OLD] = 'The filter you are trying to use is too old to be upgraded. Please re-create it.';
```
atrol: Remove also constant ERROR_FILTER_TOO_OLD

cproensa:
> I'm not sure about the different between temporary filters and one filter per project concept that you referred to. I thought we would end up with active filter per project (i.e. temp filter) and saved filters (0 or more per project). I would appreciate if you can provide some clarification for that.


There are different things here:
1) Active filters. The filter currently in use for viewing issues is remembered across page views. As i understand, currently they are intended to be remembered one active filter for each project.
However, from my tests traces and db inspection, the active filter is wrongly stored for one project, due to (i guess) some misbehavior of the logic getting the cookie id etc.
I think that it was working fine in the old 1.2 branch, long time ago, probably at some point something has broken. And it's understandable to have happened since the current code is old, unorganized and misleading. This is one of the issues fixed with this PR.

2) Stored filters. These filters are saved with a name and can be retrieved and reused at any time. Technically, they are not used directly, because at the moment of applying them, they are copied over the current "active filter". This PR, and the further comments i made, does not affect this kind of filters.

3) Temporary filters. What we call now a temporary filter is one that is not saved as the "active filter". They are those that follow the links from my-view or summary pages. Ideally you could browse issues without altering the current "active filter" for that project.
The issue (partially) fixed in this PR is that these filter were not kept after following some actions in the issues view page: sorting columns, modifying the filter properties.

Temporary filters are unique per user, becasue they are stored as token entries. This means that only one temporary filter can be used concurrently by each user.
This PR moves the storage of temp filters into session data, which allows to have multiple concurrent ones.

Additionally, we have the issue that an active filter is unique per user, but our "anonymous" user is shared. The combined changes in this PR fixes it by forcing all filters to be temporary for the anonymous user



Then i've talked about how we could improve the filter system overall.
Current system has some problems:
- The active filter is always updated. This means that updating the filter in different tabs, affects to all of them. This is confusing.
- The fact of storing temporary filters is an unneeded overhead.

_Victor, after the explanations, this is the actual answer to your question:_
Ideally the filter parameters should be applied to the url request, no need to save server side state. This would redefine some concepts:
- Temporary filters doesn't need to exists. The url request is the state itself.
- The user have "default/remembered filter" per project, instead of the current "active" ones. All further usage of this filter is not saved, unless explicitly pinned by the user. A convenient UI should provide easily to eg: "remember this view"

Those are basically our current "permalinks", however those have issues that make them not applicable in this way.
A lot of work must be made, in all levels to get to that point, but i think it's a sane approach.

Also i have talked about elsewhere:
Decouple the sorting/columns/presentation from the actual filter.

vboctor:
@cproensa Thanks for the thorough explanation.  That helps.

- The explanation of different type of filters would be useful to have checked in, whether in the `core/filter_api.php` or the manual.  Not blocking this PR, but is a good follow up if not in this PR.
- I wonder if we should force temporary filters for anonymous and protected users.  Protected users often a good indication of a shared account.  For example, an admin may create shared accounts for customers, etc.  Not blocking this PR, but may be trivial.
- I'm not sure I have the same conviction about the way forward, so I suggest we discuss each semantic change before you implement the changes.  For example, is filter based on URL vs. cookie, etc.  The correct behavior is not super obvious!  Also changing semantics could confuse users, so we have to discuss it at some point.
- Let's rebase this PR and get it checked in.

cproensa:
PR is now rebased to master

Some testing is neededas there were some conflicts from rest api, which weren't trivial to fix

@vboctor i had to rewrite part of filter_db_get_available_queries() after your changes
I noticed that you introduced returning an array indexed by filter names. This may have issues since currently two filters may exists with the same name. This can affect when having two filters owned by different users, and being public one or both of them.

cproensa:
@vboctor  i have rewritten/remvoed  `filter_load()`, in favour of keeping the different types of filter retrieval differentiated for now...

Regarding your intention of having a common entry for the standard app filters, some things to account for:
- Most of them use a user id as parameter, but not always. For example: filter_create_recently_modified() uses a days parameter. So the parameters may need to be extended, or packed into an array...
- Be careful when executing these filters, as the user id passed in is not the user who will execute the filter (for whom the authorizations are evaluated for which bugs, projects,... to be shown)

core/filter_api.php
```diff
@@ -3778,6 +3742,41 @@ function filter_get( $p_filter_id, array $p_default = null ) {
 	return $t_filter;
 }
 
+/**
+ * Return a standard filter
+ * @param string $p_filter_name     The name of the filter
+ * @param integer|null $p_user_id   A user id to build this filter. Null for current user
+ * @return null|boolean|array       null filter not found, false invalid filter, otherwise the filter.
+ */
+function filter_standard_get( $p_filter_name, $p_user_id = null ) {
```
vboctor: How about having parameters as: filter name, project id, user id?  With the last two defaulted to null.

vboctor:
@cproensa  I suggest we merge this right after 2.11.0 is out and dogfood it on our bug tracker.  The branching should happen this coming weekend.

> I noticed that you introduced returning an array indexed by filter names. This may have issues since currently two filters may exists with the same name. This can affect when having two filters owned by different users, and being public one or both of them.

Feel free to refactor as needed.

> i have rewriten/removed filter_load(), in favour of keeping the different types of filter retrieval differentiated for now...

The reason I was trying to get a single entry point is to not bother the caller with knowing what kind of filter it is retrieving.  Should your changes require changes in callers, e.g. REST API?  Another alternative to a single entry point is to have a function that returns type from id, e.g. numeric positive -> db, numeric negative -> temporary, string -> standard.  Just a thought, will leave it up to you and will see the impact on a caller like REST API as a use case.

> Be careful when executing these filters, as the user id passed in is not the user who will execute the filter (for whom the authorizations are evaluated for which bugs, projects,... to be shown)

I thought we fixed the case where we had an issue with this.  Did you see other issues?  Or are you just make sure this is stated to continue to be aware of?

cproensa:
> The reason I was trying to get a single entry point is to not bother the caller with knowing what kind of filter it is retrieving. Should your changes require changes in callers, e.g. REST API? Another alternative to a single entry point is to have a function that returns type from id, e.g. numeric positive -> db, numeric negative -> temporary, string -> standard. Just a thought, will leave it up to you and will see the impact on a caller like REST API as a use case.

I agree that may be good. At this moment the use case is 1 place from the rest api. I'd rather leave it as is now. There are more changes needed for the global filter usage, so probably at that point it will be a better moment to implement your suggestion with more differentiated use cases. I take note anyway.


> >    I noticed that you introduced returning an array indexed by filter names. This may have issues since currently two filters may exists with the same name. This can affect when having two filters owned by different users, and being public one or both of them.

> Feel free to refactor as needed.

I don't think i can refactor, as this is a deeper issue. If the api is designed to work with filter names, how would you resolve two conflicting names?
Note that this also affects the web UI, but there some UI fix can be implemented, as we are not contracting to interface solely by name.

>> Be careful when executing these filters, as the user id passed in is not the user who will execute the filter

> Or are you just make sure this is stated to continue to be aware of?

That is. Let's be aware as it's easy to misunderstand.

core/filter_api.php
cproensa: That's ok. Currently project_id is not used from the caller. Can you extend this as needed after this PR?

vboctor: sounds good

core/filter_api.php
atrol: Ping @cproensa

cproensa: done

core/filter_api.php
cproensa: done

core/filter_api.php
cproensa: done

atrol:
Closed, I rebased and merged manually.

PR #1095, opened 2017-04-18
Fix missing line after revert in PR1093

cproensa:
A line in date_api was missing after applying the reverts contained in PR1093

Related Issue: #22663


atrol:
👍

PR #1096, opened 2017-04-19
Authentication Fixes

vboctor:
- Fix php error when user clicks signup.
- Revert change to `auth_can_change_password` custom function.
- Fix access denied loop when user not logged in

Fixes #22744
Fixes #22740

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -257,14 +257,7 @@ function auth_can_set_password( $p_user_id = null ) {
 		return false;
 	}
 
-	$t_can_change = array(
-		PLAIN,
-		CRYPT,
-		CRYPT_FULL_SALT,
-		MD5,
-	);
-
-	return in_array( config_get( 'login_method' ), $t_can_change );
+	return helper_call_custom_function( 'auth_can_change_password', array() );
 }
 
 /**
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -83,6 +83,12 @@ $g_cache_current_user_id = null;
  */
 function auth_flags( $p_user_id = null, $p_username = '' ) {
 	if( is_null( $p_user_id ) ) {
+		# If user id is not provided and user is not authenticated return default flags.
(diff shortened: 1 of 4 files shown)


atrol:
TBH I don't like function auth_flags() because of the quite high cyclomatic complexity and the "multi purpose" / optional parameters.

Small refactoring, IMO a bit better to read could be to change

```
	if( !$t_user_id && is_blank( $p_username ) ) {
		# If user is not in db, must supply the name.
		trigger_error( ERROR_GENERIC, ERROR );
	}

	if( $t_user_id ) {
		$t_username = user_get_name( $t_user_id );
		$t_email = user_get_email( $t_user_id );
	} else {
		$t_username = $p_username;

		# If the plugin cares about email, then it can check if user typed the email address
		# as the username.
		$t_email = '';
	}

```
to

```
	if( $t_user_id ) {
		$t_username = user_get_name( $t_user_id );
		$t_email = user_get_email( $t_user_id );
	} else {
		if( is_blank( $p_username ) ) {
			# If user is not in db, must supply the name.
			trigger_error( ERROR_GENERIC, ERROR );
		}

		$t_username = $p_username;

		# If the plugin cares about email, then it can check if user typed the email address
		# as the username.
		$t_email = '';
	}
```

Did some cursory tests, at least my issues mentioned at https://www.mantisbt.org/bugs/view.php?id=22740#c56603 are fixed.

It seems there are some more bugs around "Lost Password" functionality.
I tried the page without entering "E-Mail" address and it seems that I locked out myself after that.
No time for more tests and review at the moment, maybe again this evening, can't promise anything.

So +1 to merge, as it's better than before, preferable with my small refactoring included.

vboctor:
Fixed merged and deployed to our bug tracker.

@atrol I will look into the proposed refactoring.  I will also check the 'Lost Password' functionality.

atrol:
> @atrol I will look into the proposed refactoring

@vboctor did you look at it? Should I provide a PR for it?

PR #1097, opened 2017-04-19
Lost password email validation fix

vboctor:
This was caused when:
- anonymous authentication is OFF.
- email address is left empty.

This caused calling auth_get_current_user_id() when no user is authenticated
which causes user to get redirected to login page and then get directed to
lost password action page, which then complains that there is no valid form
security token.

The correct behvior is to prompt an error message that email address is invalid.

Fixes #22746


vboctor:
Merged via 9c9297e26ea1d91d7c94a1d640828965ecdf468b

PR #966, opened 2016-12-01
Move bug events into BugData

cproensa:
This is a continuation of #834 updated
After my comments in the previous PR, i've decide to create a new set of events to not break current behaviours and minimize risk.

The new events are triggered from the BugData create() and update() methods, and code has been modified to perform changes to bugs always through this object. This will allow to capture the events in any action from web, soap, etc.
Old events have not been modified and will keep working as before.

Fixes: #20577

PR #1098, opened 2017-04-21
Move bug events into BugData

cproensa:
This replaces PR #966 
rebased to current master


cproensa:
It adds on suggestions from #966
- Better documentation for deprecated events
- Events parameters are now a single array, which contains different elements, that previously were separated parameters. This i think is @vboctor suggestion for event parameters. Having named indexes helps to have the data better self-explained. It would also allow to add more parameters if needed without breaking the method specification. 
- add_create_callback, as a wrapper for add_update_callback, as suggested by vboctor

Other victor suggestions (https://github.com/mantisbt/mantisbt/pull/966#pullrequestreview-14807058)
> It would be great to start the new extensibility model for these events where we have an object per event.

my interpretation is the array parameter, for example:
seleccion_174


> For the event objects it is tempting to re-use BugData as a field on these events, but I would rather have the raw properties or event objects that we want to expose (normal or calculated)

If we ever have newer interfaces/classes, these events should be updated. Ideally, I think a scenario where BugData is a readonly class, and we have a class for Bug-Modification, which is oriented to being aware of the delta changes, providing more control for notifications, and work in coordination with other entities (bugnotes, attachments, etc). That's another story, though. At the moment, out only interface is BugData

> We could also consider having an event type where if a plugin fail, we would stop the execution
of the rest

ideally, execution could be stopped with an error. In a scenario where we supports raising error as exceptions, then that could be catched and managed.

> It would be great for the update method to know the before/after state. This enables better email notifications and other changes that depend on the delta rather than just the final state. Currently the code disables email notification from within update method and triggers it by the caller

in this PR, notifications are kept as they were, triggered in the callers. BugData already have some logic that triggers notifications, so that could be used. However that was not changed in order to keep this PR more simple.

core/bug_api.php
```diff
@@ -621,16 +642,18 @@ function process_mentions() {
...
-
-		$c_bug_id = $this->id;
 
 		if( is_blank( $this->due_date ) ) {
 			$this->due_date = date_get_null();
 		}
 
 		$t_old_data = bug_get( $this->id, true );
 
+		event_signal( 'EVENT_BUG_UPDATE_PRE', array( array( 'modified_bug' => $this, 'original_bug' => $t_old_data ) ) );
```
vboctor: I would rather use `old` and `new` and avoid the term bug.

core/bug_api.php
```diff
@@ -766,23 +794,62 @@ function update( $p_update_extended = false, $p_bypass_mail = false ) {
...
 
-			# @todo handle priority change if it requires special handling
-			email_bug_updated( $c_bug_id );
+		# Execute all registered update callbacks
+		while( !empty( $this->update_callbacks ) ) {
+			$t_callback = array_shift( $this->update_callbacks );
+			call_user_func_array( $t_callback['func'], $t_callback['params'] );
 		}
 
+		event_signal( 'EVENT_BUG_UPDATE_POST', array( array( 'modified_bug' => $this, 'original_bug' => $t_old_data ) ) );
```
vboctor: Same here `old` and `new`.

core/bug_api.php
```diff
@@ -1895,52 +1979,69 @@ function bug_resolve( $p_bug_id, $p_resolution, $p_fixed_in_version = '', $p_bug
 	$c_resolution = (int)$p_resolution;
 	$p_bugnote_text = trim( $p_bugnote_text );
 
-	# Add bugnote if supplied
-	# Moved bugnote_add before bug_set_field calls in case time_tracking_no_note is off.
-	# Error condition stopped execution but status had already been changed
-	$t_bugnote_id = bugnote_add( $p_bug_id, $p_bugnote_text, $p_time_tracking, $p_bugnote_private, 0, '', null, false );
-	bugnote_process_mentions( $p_bug_id, $t_bugnote_id, $p_bugnote_text );
+	# @TODO
+	# If time tracking is enabled, this code is executed outside of BugData validation.
```
vboctor: I don't recall the reason why we do this special behavior for time tracking notes?  Do you know why?

Btw, @ TODOs are picked up by some IDEs like PhpStorm, but they show the first line next to them, so it is generally good practice to try the have the todo on the same line and having enough info on first line to provide an idea of what needs to be done.  There are multiple occurrences of such usage of @ TODOs.

core/bug_api.php
```diff
@@ -1895,52 +1979,69 @@ function bug_resolve( $p_bug_id, $p_resolution, $p_fixed_in_version = '', $p_bug
...
+		# create callback for duplicate bug actions:
+		$t_bugdata->add_update_callback( 'relationship_upsert', array( $p_bug_id, $p_duplicate_id, BUG_DUPLICATE, /* email_for_source */ false ) );
 
 		# Copy list of users monitoring the duplicate bug to the original bug
-		$t_old_reporter_id = bug_get_field( $p_bug_id, 'reporter_id' );
-		$t_old_handler_id = bug_get_field( $p_bug_id, 'handler_id' );
-		if( user_exists( $t_old_reporter_id ) ) {
-			bug_monitor( $p_duplicate_id, $t_old_reporter_id );
+		if( user_exists( $t_bugdata->reporter_id ) ) {
+			$t_bugdata->add_update_callback( 'bug_monitor', array( $p_duplicate_id, $t_bugdata->reporter_id ) );
```
vboctor: This is a general comment about the approach for registering callbacks.  Since this is no longer a normal call to APIs like bug_monitor(), IDEs like PhpStorm won't pickup such usage as a call to the function.  Should we instead have a callback function that would call such APIs normally?

docbook/Developers_Guide/en-US/Events_Reference_Bug.xml
```diff
@@ -277,6 +317,126 @@
...
+					structure before the data is saved to the database. At this point,
+					the issue ID is not yet known, as the data has not yet been persisted.
+					Modifications on this structure can be made in the BugData object
+					received as parameter.
+					This event is triggered by the BugData API by any action that uses
+					the create() method.
+				</para>
+
+				<para>
+					Note: As the create action may be altered or stopped by other plugins,
```
vboctor: How should plugins trigger the stopping in a way that works well with web UI and APIs?  Is it via trigger_error?  It is worth testing this case and clarifying guidance in docs as needed.

cproensa:
> I'm not sure that changes by plugins will trigger the appropriate history changes. I suspect that we only track changes based on the original change.

BugData::update() always calls history_log_event_direct for its properties. Any change should be logged to history.

> This is a general comment about the approach for registering callbacks. Since this is no longer a normal call to APIs like bug_monitor(), IDEs like PhpStorm won't pickup such usage as a call to the function. Should we instead have a callback function that would call such APIs normally?

I see that would mean use anonymous functions, like they are used in some places in this PR by assigning a variable:
$callback = function () { ... };

> How should plugins trigger the stopping in a way that works well with web UI and APIs? Is it via trigger_error? It is worth testing this case and clarifying guidance in docs as needed.

yes, trigger_error is currently the only way to stop an invalid bug change. It's definitely not oriented to "works well with web UI and APIs", but we don't have any other way at the moment.
My suggestion for this case is using exceptions, but not in the scope of this PR.


> The lack of event specific objects makes event handling more complex and complete to detect the delta. The plugin has to do a diff to reverse engineer the change type. The plugins will also be blind to changes that happen only as part of the callbacks. 

That's the same behaviour we have now (for BugData properties). 
Ideally, we should have all bug related data objectified: notes, attachments, custom fields, etc. so we can  manage all the bug context as a whole, and manage proper deltas. That's a lot redesign which is out of scope of this PR. That would probably relate to a redesigned notification system too.

>  We should really try not to have bug modifying logic in the callbacks. For example: it is OK to send an email, update history, but we shouldn't set custom fields, add notes, add attachments, etc.

modifying in the callbacks was precisely the motivation to add them.
We want to modify custom fields, notes, attachments, only:
- After the pre-processing events are cleared. Otherwise we would have committed modifications even if the bug update failed due to an error, which may be intended to stop an invalid bug change.
- Before the post-update event is signalled. At that point the receivers of the event must have available the final state of the bug change and all the related entities.


Additionally.
I'm still not sure if the way of passing the data to the event as one single array:
`array( 'old' => BugData object, 'new' => BugData object )`
is in line of your suggestions instead of having both objects as raw parameters (like is done at the moment with events)

vboctor:
> BugData::update() always calls history_log_event_direct for its properties. Any change should be logged to history.

Good.

> I see that would mean use anonymous functions, like they are used in some places in this PR by assigning a variable: $callback = function () { ... };

Yes

> yes, trigger_error is currently the only way to stop an invalid bug change. It's definitely not oriented to "works well with web UI and APIs", but we don't have any other way at the moment. My suggestion for this case is using exceptions, but not in the scope of this PR.

Would it be too hard to support exceptions thrown from event handlers? i.e. for event X, plugins can throw exception Y and it will be caught by the `event_signal()` caller?  That will provide a way that is friendly to web ui/api.  Similar to our current APIs, the code used in such passes should avoid calls like ensure_condition1() which internally triggers error, but call is_condition1() and throw the exception if needed.  I'm OK with being part of a different PR, but don't think it needs to be associated with an exceptions everywhere scope of work.

> Ideally, we should have all bug related data objectified: notes, attachments, custom fields, etc. so we can manage all the bug context as a whole, and manage proper deltas. That's a lot redesign which is out of scope of this PR. That would probably relate to a redesigned notification system too.

To summarize my position here:
- I agree that we need to some ORM model with proper model objects and support for new/modified data that can be processed and then committed to db.  Active Record kind of stuff.  In fact I was going to open an issue for discussion this - "Modern Model".
- In the case where we have the "Modern Model" described above, I suspect we still need Event Objects.
- I agree that this is too much work for this PR.
- I dislike the fact that we are churning the code a lot, deprecating events and adding new ones, but we only solved part of the problem.

> I'm still not sure if the way of passing the data to the event as one single array:

I was just suggesting changing the naming of the old and new objects rather than changing how you send them to the events.

> modifying in the callbacks was precisely the motivation to add them.

I understand.  However, we still have the gap where plugins don't see a note change.  And have to manually stitch a status change + related note.  For example, there is no way for a plugin to stop changing a status without a note explaining the work or making sure that a custom field is populated with some value.  If we use event objects, then they can look at the whole change, approve it or modify it, then it goes through.

cproensa:
> Would it be too hard to support exceptions thrown from event handlers? i.e. for event X, plugins can throw exception Y and it will be caught by the event_signal() caller? That will provide a way that is friendly to web ui/api. Similar to our current APIs, the code used in such passes should avoid calls like ensure_condition1() which internally triggers error, but call is_condition1() and throw the exception if needed. I'm OK with being part of a different PR, but don't think it needs to be associated with an exceptions everywhere scope of work.

In some other place i suggested introducing an alternate error handler that packages the captured error in an exception, and then throws it. The context was to manage bug-action-group where one error would not stop the whole batch. This would be another application of the same concept.
The error->exception mapper should be used as a switchable mode, very like we now use a modifier to not populate error pages (ajax requests, etc.). One _OldStyleMantisErrorException_ is enough for this, and contains info for actual error code that were captured.
This allows to achieve that behaviour withouth refactoring all the code to use exceptions, which can also be introduced progressively.
Nonetheless, a exception can be created to manage the specific case of "invalid bugdata change".
Previous scenario means that both the dedicated exception, and any arbitrary error, can be managed equally.


> I dislike the fact that we are churning the code a lot, deprecating events and adding new ones, but we only solved part of the problem.

I dont know. For sure: we are lacking full coverage with the current events. Almost any action that doesn't come from bug-update-page or bug-create-page is not managed by those. 
The callback stuff is intended to add a little of consistency in the update workflow. We can leave it out and it won't be worse than we were before.

> I was just suggesting changing the naming of the old and new objects rather than changing how you send them to the events.

ok, but i'd rather be sure that this is a desired trait for future events too. This array thing makes more sense when an event would have a bigger number of parameters. So the plugin callee should not define:
`function event_handler( $p_event, $dontcare1, $dontcare2, $dontcare3, $relevant_param ) { ... }`
In this PR, two of the events only issues one parameter, so the array is overkill. However, I'm aiming for consistency here.

If you have other plans, of which there are many possibilities, there's the option to leave the array parameter out, to keep consistency with the old event convention.

vboctor:
> ok, but i'd rather be sure that this is a desired trait for future events too.

I think it is a common pattern for event handlers to get an event type + event args.  And the args is either a strongly typed class or a property bag with keys and values.  Events can start with single or two parameters, but then in the future, more may be added to provide more context or whatever, and we need a way to do them.  When referring to an array/property bag, keys are much better than array offsets.

Here is an example for triggering `AuthPlugin` event:

```
	$t_event_arguments = array(
		'user_id' => $t_user_id,
		'username' => $t_username,
		'email' => $t_email,
	);

	$t_flags = event_signal( 'EVENT_AUTH_USER_FLAGS', array( $t_event_arguments ) );
```

> removing callbacks

It is unclear to me if we are benefiting from having callbacks vs. running such follow up tasks after the call to ->update() or ->create()?  It seems that we update db, email, then call callbacks anyway.  So does it make a different if it is inside the update/create calls vs. called after from the calling code?  I get the value though in the case where the code needs to run after saving to validating via event, validating via core, saving to db and before email notifications.

So my feel is that they can be useful when we need such callbacks to be called in the middle of the create/update functions.  And maybe we overtime just reduce reliance on follow up bug updates to BugData in such functions.

If code can just run after update()/create() is done, then it shouldn't be a callback.  If code needs to run in the middle of these methods, then it should be but we should make it IDE friendly.

cproensa:
Updated with some of the requested changes
- Rewrite callbacks to call directly to api functions (IDE friendly)
- Rename event parameter indexex as "new', "old", etc
- Clean up the timetracking special cases for bugnotes

cproensa:
@vboctor reminder!

cproensa:
@vboctor 
What are the blocking issues in this PR, at this point?
This was started in Aug-2016...

vboctor:
@cproensa will have a look again.  Would be good to get others to have a look too.

vboctor:
@cproensa I'm working on more commands like `IssueAddCommand` and I think events will make a lot more sense in context of such commands.  The events will also use the REST API format for its args which will be more powerful and portable compared to what we had in the past.

cproensa:
> I'm working on more commands like IssueAddCommand and I think events will make a lot more sense in context of such commands

How is that? You can introduce new events, or trigger the existing ones, for each "command", but that won't solve the same issue that i try to address here.

cproensa:
outdated

PR #1099, opened 2017-04-22
Fix error message when user can’t handle issues

vboctor:
Fixes #22762


dregad:
It feels like reducing error message helpfulness to no longer return the user's ID with the error message. Maybe just I'm missing some context, but looking at 22762, I don't understand how the user id could be 0. Which one of the 3 changes is actually causing the display of _user0_, and in what usage scenario ?

Also, as a general note: It would be better to reword the commit message, so the reader can understand what the change is about (and why it's needed) without having to look at the code.

Finally, you're introducing a new error condition for disabled user; I have no objection to that, but again this should be described in the commit message, and it seems to be a distinct change from bug 22762.

vboctor:
@dregad 

- Commit message updated with more details and answers to your question.
- Unrelated check is removed, may add it in the future.

As per the commit message, the error messages will look as follows:

`Access denied for vboctor.  Reason: User does not have access right to handle issues.`

The reference to user id in the reason is redundant and was there from the previous error handling approach.

PR #1100, opened 2017-04-23
REST and SOAP API Improvements

vboctor:
It is recommended to review this PR one commit at a time to follow the chain of changes.

**REST API**
- Support getting multiple issues with pagination.  This can be based on filter or project.  Whether returning a single issue (by id) or a batch of issues, the same response format is now used which has `issues` array.
- Setup framework for testing REST API + tests for issue creation.
- When time tracking is not visible, return note type as `note` rather than `timelog` if it has a hidden time.
- Return versions as an object with id and name rather than just a string.
- Enum name should be the English (from config) value.
- Added enum `label` which has the localized version of the enum name.
- Fix some error messages that were not correctly reflecting the error case.
- Include status color as part of status info on issues.
- Use `created_at` rather than `date_submitted`.
- Use `updated_at` rather than `last_updated`.
- Make `sticky` a boolean rather than a string.
- Don't include `sponsorship_total`.

**SOAP and REST**

- Added some config option checks that were missing from SOAP/REST API (e.g. eta enabled, projection enabled, profile, build/platform/os, etc).
- Don't allow setting an undefined version when creating an issue.
- Do access checks based on bug/project rather than global access level for fields like due date.

### DIFF
--- a/api/rest/index.php
+++ b/api/rest/index.php
@@ -55,6 +55,7 @@ require_once( $t_restcore_dir . 'config_rest.php' );
 require_once( $t_restcore_dir . 'internal_rest.php' );
 require_once( $t_restcore_dir . 'issues_rest.php' );
 require_once( $t_restcore_dir . 'lang_rest.php' );
+require_once( $t_restcore_dir . 'projects_rest.php' );
 require_once( $t_restcore_dir . 'users_rest.php' );
 
 event_signal( 'EVENT_REST_API_ROUTES', array( array( 'app' => $g_app ) ) );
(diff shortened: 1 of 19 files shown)


rombert:
Test cases look good at a first glance to me ( did not test ).

Have you considered sharing the test cases between the SOAP API and the REST API, while leaving the test setup separated? That would quickly increase the coverage of the REST API and we would only maintain one test suite in the future.

In the areas in which the two APIs differ - e.g. SOAP API not getting new features - we can easily find a way to only run some checks for one of the APIs.

vboctor:
@rombert Thanks for the review.

> Have you considered sharing the test cases between the SOAP API and the REST API, while leaving the test setup separated? That would quickly increase the coverage of the REST API and we would only maintain one test suite in the future.

I haven't considered sharing test cases and I'm not in favor of it for the following reasons:

- As I'm churning the shared code, it would be good to at least have a stable non-shared test cases for SOAP to validate.  Churning both won't help.
- The API (capabilities), their requests, their responses, etc are likely to be different.  Hence, there will be a lot of if SOAP vs. REST even when validating the results.  There may be even gaps in functionality in one vs. the other.
- I expect less churn in SOAP and its associated test cases.  So hopefully not much maintenance there until it is fully retired.

PR #1101, opened 2017-04-25
Hide fields in the email notification when its not enabled in View Issue page

jllano:
This PR is about handling fields that can be practically hidden from View Issue page.

- There is no point in have a field in email notifications if the instance is configured to hide it from the UI.

Fixes #22813

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -1608,7 +1608,7 @@ function email_bug_info_to_one_user( array $p_visible_bug_data, $p_message_id, $
 		$t_message .= " \n";
 	}
 
-	$t_message .= email_format_bug_message( $p_visible_bug_data, $p_user_id );
+	$t_message .= email_format_bug_message( $p_visible_bug_data );
 
 	# build headers
 	$t_bug_id = $p_visible_bug_data['email_bug'];
@@ -1679,25 +1679,16 @@ function email_format_bugnote( $p_bugnote, $p_project_id, $p_show_time_tracking,
  * @param integer $p_user_id    A user identifier.
  * @return string
  */
-function email_format_bug_message( array $p_visible_bug_data, $p_user_id ) {
+function email_format_bug_message( array $p_visible_bug_data ) {
 	$t_normal_date_format = config_get( 'normal_date_format' );
 	$t_complete_date_format = config_get( 'complete_date_format' );
-
-	$t_bug_view_fields = config_get( 'bug_view_page_fields', null, $p_user_id, $p_visible_bug_data['email_project_id'] );
 
 	$t_email_separator1 = config_get( 'email_separator1' );
 	$t_email_separator2 = config_get( 'email_separator2' );
 	$t_email_padding_length = config_get( 'email_padding_length' );
 
-	$t_status = $p_visible_bug_data['email_status'];
-
 	$p_visible_bug_data['email_date_submitted'] = date( $t_complete_date_format, $p_visible_bug_data['email_date_submitted'] );
 	$p_visible_bug_data['email_last_modified'] = date( $t_complete_date_format, $p_visible_bug_data['email_last_modified'] );
-
-	$p_visible_bug_data['email_status'] = get_enum_element( 'status', $t_status );
-	$p_visible_bug_data['email_severity'] = get_enum_element( 'severity', $p_visible_bug_data['email_severity'] );
-	$p_visible_bug_data['email_priority'] = get_enum_element( 'priority', $p_visible_bug_data['email_priority'] );
-	$p_visible_bug_data['email_reproducibility'] = get_enum_element( 'reproducibility', $p_visible_bug_data['email_reproducibility'] );
 
 	$t_message = $t_email_separator1 . " \n";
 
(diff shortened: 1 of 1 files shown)


atrol:
Wouldn't it be better to hide the information in an earlier state (don't populate $p_visible_bug_data).
At first sight, email_build_visible_bug_data() could be the right place for it.
Why should we fetch information and decide to hide it afterwards?

jllano:
@dregad @atrol I already addressed your comments.

core/email_api.php
```diff
@@ -1676,6 +1676,7 @@ function email_format_bugnote( $p_bugnote, $p_project_id, $p_show_time_tracking,
 /**
  * Build the bug info part of the message
  * @param array $p_visible_bug_data Bug data array to format.
+ * @param integer $p_user_id    A user identifier.
```
vboctor: This phpdoc param doesn't correspond to a real one.

core/email_api.php
```diff
@@ -1729,7 +1742,7 @@ function email_format_bug_message( array $p_visible_bug_data ) {
 
 	# end foreach custom field
 
-	if( config_get( 'bug_resolved_status_threshold' ) <= $t_status ) {
+	if( isset( $t_status ) && config_get( 'bug_resolved_status_threshold' ) <= $t_status ) {
 		$p_visible_bug_data['email_resolution'] = get_enum_element( 'resolution', $p_visible_bug_data['email_resolution'] );
```
vboctor: We should add a check for `resolution` field.

atrol:
@jllano Is there a special reason that you don't treat some fields, e.g. date_submitted, last_updated, due_date, ... ?

jllano:
@atrol I just put the other fields remain as is, as there's no logic point for removing it (date_submitted, last_updated, due_date, etc...).

atrol:
> as there's no logic point for removing it

I am not sure if I understand what you mean with it.

Your PR is
> Hide fields in the email notification when its not enabled in View Issue page

When reading that, I would expect that I don't see any field in email notification if it's not in the list
of "View Issues Columns" on page "My Account" > "Manage Columns" (account_manage_columns_page.php).

While thinking about it I am wondering if using bug_view_page_fields is a good idea at all.
Maybe we should introduce bug_email_fields.

I would like to get also some feedback from other devs. @vboctor @dregad @cproensa

cproensa:
@atrol I agree with:
> While thinking about it I am wondering if using bug_view_page_fields is a good idea at all.
Maybe we should introduce bug_email_fields.

With the same reason that there are different sets of fields to: view, print, excel and csv
a mail fields configuration can be considered.

vboctor:
I have been trying to gradually work on email notifications.  As more of this work is done, we have less of a single email format, and more of a format that depends on the event.  For example, an @ mention or note will be different than submission and different than an issue update (which may depends on fields updated) and different than relationship added between two issues.  Hence, I don't think there is a single concept of email fields and even in a bug update notification, the fields may depend on what has changed.

In the meantime, I would go with a model where we hide fields/features that are disabled rather than getting a list of fields.  Kind of similar to what we do for fields like projection and ETA.  Bug view fields can be a good proxy in the meantime.

So I would revisit this as the modern email work is done.  I'm planning to work on this soon.

We could however use `bug_email_fields` whenever a verbose notification is generated that has the full issue details.  Though if such field is applicable in another notification, this model won't work well.

vboctor:
@dregad @cproensa @atrol please comment so we can clarify next steps for this PR.

cproensa:
> .. we have less of a single email format, and more of a format that depends on the event.

yes, this PR should be applicable to the verbose email format

> Bug view fields can be a good proxy in the meantime
> We could however use bug_email_fields whenever a verbose notification is generated that has the full issue details. 

I'm ok with either. However my opinions for not reusing bug-view-fields:
- To me it would be somewhat confusing receiving different email content if i change my viewing preferences, as i'm used to have a fixed email format, and not expecting that view page would affect email. I usually need to change view page more frequently that i'd need changing email fields.
- I see in the future, as part of cleaning up filter structure, the possibility of using "Views" which contains the pure presentation properties, and one of them is columns to show. The user could save and use custom views in the same way as stored filters (or one be a composition with the other), instead of the current columns configuration. In this scenario, These views would replace having a fixed set of configurations for each view/excel/print contexts.

vboctor:
@cproensa

> To me it would be somewhat confusing receiving different email content if i change my viewing preferences

I don't think it is confusing since if the intent of the administrator is not to use a field, then it should be hidden.  There are two ways to indicate that a field is not desired, an explicit way via an ON/OFF config option per field (similar to eta field) or an implicit way by assuming that if the field is not on the View Issue page, then it is not being used.

> I see in the future, as part of cleaning up filter structure, the possibility of using "Views" which contains the pure presentation properties

I think you are referring to columns in the View Issues page, rather than fields in the View Issue page.  The idea here is that for the time being based on current feature set, we need to provide a simple way to not include unused fields in verbose email notifications.  We can evolve that as we provide richer controls in the future.

atrol:
Forget parts of my comments.

After reading
> Hide fields in the email notification when its not enabled in View Issue page

I commented as if I had read
> Hide fields in the email notification when its not enabled in View Issues page

I am fine with using `bug_view_page_fields` to suppress fields in email notifications.

cproensa:
> I am fine with using bug_view_page_fields to suppress fields in email notifications

I was carried away with the same thinking.
I am ok with that too.

vboctor:
Merged via 66d035a046224ccbae12ada77e66215e658132ce - thanks @jllano

PR #1108, opened 2017-05-15
Fixes markdown formating for notes column

vboctor:
The 3 dashes marked the notes above it as a markdown header.  Fix is to just use a newline.

Fixes #22867

*Before:*

notes_column_with_markdown

*After:*

notes_column_with_markdown_fix


dregad:
Using a newline is acceptable, but I'm wondering if it would not be useful to use some other separator.

It's a bit of a corner case, but in my previous company, the `---` was actually used to parse the string and split the bugnotes. Can't do that with just a newline.

vboctor:
@dregad I have updated the separator to be `=-=` which doesn't confuse the markdown parser.

PR #1109, opened 2017-05-16
Changes for print_form_button()

cproensa:
Some proposals for buttons generated from print_form_button(), and buttons layout in general.

- Use `button` tag, instead of `input`, as the former can include icons. We use intensively these buttons in manage pages, one or several in each row of a table. Having the full text for all buttons add clutter to the page, so at some point they could be changed to icons
- Buttons generated from print_form_button() are not inlined by themselves, as they are placed as standalone forms, which have block display style. This does not correspond to the behavoiur of standard button or inputs, which are inline elements by default.
- Buttons generated from print_form_button() does not show spacing between them, whereas standard buttons in a form do have some margin.
- Experiment with a visibility toggle for some manage buttons, when these buttons are used for actions on every row in a table. By making them visible only when hovering on its relevant container, the page may look cleaner.

### DIFF
--- a/bugnote_view_inc.php
+++ b/bugnote_view_inc.php
@@ -153,7 +153,7 @@ $t_block_icon = $t_collapse_block ? 'fa-chevron-down' : 'fa-chevron-up';
 			$t_time_tracking_hhmm = '';
 		}
 ?>
-<tr class="bugnote" id="c<?php echo $t_activity['id'] ?>">
+<tr class="bugnote visible-on-hover-toggle" id="c<?php echo $t_activity['id'] ?>">
 		<td class="category">
 		<div class="pull-left padding-2"><?php print_avatar( $t_activity['user_id'], 'bugnote', 80 ); ?>
 		</div>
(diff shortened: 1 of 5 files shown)


cproensa:
Before: configuration report page

seleccion_191


After adding margin

seleccion_193


After adding visibility on hover (row):

seleccion_192

cproensa:
> You have opened several related issues on the tracker, but forgot to add the Fixes #xxx references in the commit messages.

yes.. as i see the changes as likely to be discussed and changed, dind't put the "fixes..." right away.
I'd wait for feedback on the topics, especially @syncguru

bugnote_view_inc.php
```diff
@@ -199,7 +199,7 @@
 		?>
 		<div class="clearfix"></div>
 		<div class="space-2"></div>
-		<div class="btn-group-sm">
+		<div class="btn-group visible-on-hover">
```
syncguru: It is not very clear to me if removing "-sm" is intentional. If so, why?

cproensa: two reasons:
- other buttons in the pages are within plain "btn-group"
- to make it apply the margins with the defined css selector: `.btn-group .single-button-form `.
Otherwise, the selector should be repeated for every .btn-group-xx

What do you think?

cproensa:
> You have opened several related issues on the tracker, but forgot to add the Fixes #xxx references in the commit messages.

Rebased, and added the bug references in the commits

PR #1111, opened 2017-05-20
Refactoring db_insert_id()

dregad:
- Refactor function to use $g_db_functional_type, avoid multiple calls to db_is_xxx.
- db_result() returns a string in some cases. Typecasting the return value to int ensures we comply with the PHPDoc.

Fixes [#22904](https://mantisbt.org/bugs/view.php?id=22904)


cproensa:
Ok, (didn't test though)

PR #1112, opened 2017-05-22
Use Composer instead of Submodules + update libraries

vboctor:
This PR includes the following changes:

- Replace all submodules with composer.
- Remove swagger-ui in favor of REST clients.
- Update PHPMailer
- Update Parsedown
- Update DisposableEmailChecker

Fixes #22913, #22939, #22940

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,6 @@ nbproject/
 
 # IIS7
 web.config
-!library/securimage/web.config
 
 # Libraries
 library/FirePHPCore
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,3 @@
-[submodule "library/phpmailer"]
-	path = library/phpmailer
-	url = https://github.com/mantisbt/PHPMailer
-	branch = mantis
 [submodule "api/rest/swagger"]
 	path = api/rest/swagger
 	url = https://github.com/mantisbt/swagger-ui.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,3 @@
-[submodule "library/adodb"]
-	path = library/adodb
-	url = https://github.com/mantisbt/ADOdb.git
-	branch = mantis-1.3
 [submodule "library/phpmailer"]
 	path = library/phpmailer
 	url = https://github.com/mantisbt/PHPMailer
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,10 +6,6 @@
 	path = library/phpmailer
 	url = https://github.com/mantisbt/PHPMailer
 	branch = mantis
-[submodule "library/disposable"]
-	path = library/disposable
-	url = https://github.com/mantisbt/disposable_email_checker
-	branch = mantisbt
 [submodule "library/securimage"]
 	path = library/securimage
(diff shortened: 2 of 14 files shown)


atrol:
I don't see that much benefit at the moment to introduce one more developer tool, maybe because I don't kwow that much about Composer.

Is it possible to maintain our own branches of 3rd party components with it?
Could we drop usage of Git submodules?

In future, do we want to use Git submodules for PHP and JavaScript, Composer for PHP and maybe npm for JavaScript?
I would prefer to use just a single tool as long as the tool is enough for what we need.

vboctor:
@atrol As per the plan proposed in #806, we are heading towards a modern build system.  Such build system will use two package managers: composer (php) and npm (javascript).  The build system will have some dev-only dependencies that introduce tools to be used for other tasks like compiling scss/less, javascript, phpunit, etc.  The goal is to be able to run one build command that would make your MantisBT ready to run in a development environment.  For users, all such tasks will be done at packaging time, so they won't have to worry about them.

Composer supports adding dependencies to packages published on Packagist or add dependencies on GitHub repositories directly.  So our goal is to use them to replace submodules.

Package managers have a lot of advantages:

- Self-documenting dependencies information (version, location, etc).
- Differentiating between dev/production dependencies.
- Pulling tools that can be used by the build system.
- Understanding dependency graph and resolving all necessary components and appropriate versions.
- Much easier to run an update command to get the latest versions for dependencies that are matching version constrains with understanding for [semver](http://semver.org).  Think about picking up latest dot releases for PHPMailer, SlimFramework, DisposableEmailChecker, etc by running one command at the beginning of a release cycle.  Major version upgrades require changing composer.json file.

We have already picked up Composer dependency as part of implementing the REST API and taking dependencies on SlimFramework and GuzzleHttp.  This change updates our build system to run composer.  This didn't break us before because we don't run REST API test cases as part of the build yet.

We should gradually move our submodules into composer dependencies as we are updating them to newer versions, similar to what I have done here with DisposableEmailChecker.

Here is a [very old post](http://blog.nelm.io/2011/12/composer-part-1-what-why/) about Composer from its author that he posted back in 2011.

atrol:
Thanks for the information @vboctor.

I really like the fact that I have to install just Git (nearly any version works) and enter two commands (git clone, git submodule update --init) to setup a developer environment on any of my test systems (Windows 7 and 10, macOS Sierra, Ubuntu 14.04 and 16.04)

I hope the new build system will not add too much additional effort to setup such a system.

dregad:
I'm fine with the switch from Submodules to Composer, as previously discussed elsewhere.

However, to reduce friction, I would prefer if we moved all submodules in one go as incremental changes can be quite painful from a maintenance perspective (e.g. when switching back and forth between branches).

syncguru:
I am 100% support of moving to package managers. It has been the standard practice in web development for quite some time now.

vboctor:
@atrol @dregad I have replaced all submodules with composer packages.  Hence, updating submodules is no longer necessary.

vboctor:
I'm planning to merge this after 2.5.0 is released.

PR #1113, opened 2017-05-23
Patch for setLogoutRedirectPage in authentication plugins

epenet:
It is necessary to cache the auth_logout_redirect_page() result, otherwise it gets cleared during auth_logout()


logout_page.php
```diff
@@ -32,6 +32,8 @@
 require_api( 'config_api.php' );
 require_api( 'print_api.php' );
 
+$t_logout_redirect = auth_logout_redirect_page();
```
vboctor: I would add a comment here explaining the reason for capturing the value in a temporary value, so it is clear and not optimized away later by mistake.

epenet:
I added the comment
I created the issue in the bug tracker http://mantisbt.org/bugs/view.php?id=22923
But I don't know how to squash the change

dregad:
> But I don't know how to squash the change

Don't worry, we can do it as we merge the pull request.

If you're interested in learning how to do this, have a look at how to use `git rebase --interactive` [[1]]. Note that since this will rewrite history, you will then have to do `git push --force` to update the remote. Feel free to ping me on gitter if you have questions.

[1]: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

PR #1115, opened 2017-05-26
Fix date custom field formatting

vboctor:
Fixes #22926

### DIFF
--- a/core/cfdefs/cfdef_standard.php
+++ b/core/cfdefs/cfdef_standard.php
@@ -262,8 +262,8 @@ function cfdef_prepare_list_value_for_email( $p_value ) {
 
 /**
  * Print the value of the email custom field.
- * @param $p_value The database value
- * @return string The html value.
+ * @param string $p_value The database value
+ * @return void
  */
 function cfdef_print_email_value( $p_value ) {
 	if( !is_blank( $p_value ) ) {
--- a/core/cfdefs/cfdef_standard.php
+++ b/core/cfdefs/cfdef_standard.php
@@ -107,7 +107,7 @@ $g_custom_field_type_definition[CUSTOM_FIELD_TYPE_EMAIL] = array (
 	'#function_value_to_database' => null,
 	'#function_database_to_value' => null,
 	'#function_print_input' => 'cfdef_input_textbox',
-	'#function_print_value' => null,
+	'#function_print_value' => 'cfdef_print_email_value',
 	'#function_string_value' => 'cfdef_prepare_email_value',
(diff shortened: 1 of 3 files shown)


atrol:
This fixes #22926, but the fix which introduced the issue (#22428) introduces even more issues, e.g.#22937

Maybe it's better in short term (2.4.2, 2.5.0) to revert the changes from #22428, and to provide a proper fix in 2.6.0 with some more tests.

vboctor:
@atrol I have fixed the email field formatting.  Will test all custom fields types tonight.

vboctor:
@atrol This PR now includes fix for date, email, and single line string custom fields.

core/cfdefs/cfdef_standard.php
```diff
@@ -261,12 +261,23 @@ function cfdef_prepare_list_value_for_email( $p_value ) {
 }
 
 /**
- * Format email address for email
- * @param string $p_value Value.
- * @return string
+ * Print the value of the email custom field.
+ * @param $p_value The database value
```
atrol: missing parameter type

core/cfdefs/cfdef_standard.php
```diff
@@ -261,12 +261,23 @@ function cfdef_prepare_list_value_for_email( $p_value ) {
...
+ * @return string The html value.
```
atrol: functions does not return any value

plugins/MantisCoreFormatting/MantisCoreFormatting.php
```diff
@@ -190,7 +190,7 @@ function formatted( $p_event, $p_string, $p_multiline = true ) {
 		$t_string = mention_format_text( $t_string, /* html */ true );
 
 		# Process Markdown
-		if( ON == $s_markdown && $p_multiline ) {
+		if( ON == $s_markdown ) {
```
atrol: This introduces side effects, e.g. check rendering of summary field on "My View" page. or "View Issue Details".
There are added new lines.

core/cfdefs/cfdef_standard.php
vboctor: Fixed

core/cfdefs/cfdef_standard.php
vboctor: Fixed

plugins/MantisCoreFormatting/MantisCoreFormatting.php
vboctor: Fixed side effects

PR #1118, opened 2017-05-30
Remove unused option unread_icon_arr

atrol:


atrol:
@vboctor I don't think that adding to the obsolete file is necessary as this option was not documented and has never been used in code (there was a function print_unread_icon which did access the option, but this function has never been called)

PR #1119, opened 2017-06-04
Supress display of "Access Denied" messages on view_user_page

atrol:
Fixes #22967

### DIFF
--- a/view_user_page.php
+++ b/view_user_page.php
@@ -108,8 +108,6 @@ $t_timeline_view_class = ( $t_timeline_view_threshold_access ) ? "col-md-7" : "c
 				    <?php
 						if( !is_blank( $u_email ) ) {
 							print_email_link( $u_email, $u_email );
-						} else {
-							echo " - ";
 						}
 					?>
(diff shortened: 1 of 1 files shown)


view_user_page.php
```diff
@@ -98,35 +98,34 @@
...
+			<tr>
+				<th class="category">
+					<?php echo lang_get( 'email' ) ?>
+				</th>
+				<td>
+				    <?php
+						if( !is_blank( $u_email ) ) {
+							print_email_link( $u_email, $u_email );
+						} else {
+							echo " - ";
```
vboctor: Should we just print `&nbsp;` or leave it empty instead of ` - `?

atrol:
I thought also about it when stumbling over this piece of code.
Another option could be to use something like lang_get( 'no_email_adress' );

Finally my conclusion is to leave it empty, as I see no reason to have a special case for emtpy e-mail addresses.

I opened https://www.mantisbt.org/bugs/view.php?id=22981 to track the issue and updated the PR.

PR #1120, opened 2017-06-05
Add config parameter to control format of email notifications

libregeek:
The markdown(MD) format within the plain text email notifications are causing confusions for our non-technical users who heavily depend on emails to know about issues. So we think that converting the MD to HTML in emails may help them greatly. We are working on a plugin which converts markdown text in email notifications to HTML format. Currently only the textarea fields (description, additional information, steps to reproduce, and notes) are converted.

The 'g_enable_html_email` parameter tells PHPMailer that it needs to use text/html as the Content-Type. The default value for this configuration parameter is OFF, so existing plain/text emails won't be affected.

Please review the PR and let us know your feedback.

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -705,6 +705,12 @@ $g_email_separator2 = str_pad('', 70, '-');
  */
 $g_email_padding_length	= 28;
 
+/**
+ * enable HTML format in emails
+ * @global integer $g_enable_html_email
+ */
+$g_enable_html_email = OFF;
+
 ###########################
 # MantisBT Version String #
 ###########################
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -1300,7 +1300,12 @@ function email_send( EmailData $p_email_data ) {
 			break;
 	}
 
-	$t_mail->IsHTML( false );              # set email format to plain text
+	if( ON == config_get( 'enable_html_email' ) ) {
+		$t_mail->IsHTML( true );              # set email format to HTML
+	} else {
+		$t_mail->IsHTML( false );              # set email format to plain text
+	}
+
 	$t_mail->WordWrap = 80;              # set word wrap to 80 characters
 	$t_mail->CharSet = $t_email_data->metadata['charset'];
 	$t_mail->Host = config_get( 'smtp_host' );


dregad:
I don't really like the approach of introducing a config option for this, especially considering the fact that MantisBT core doesn't do anything with it (i.e. we don't actually generate HTML content for e-mail notifications).

It would probably be preferable to add a new event instead, to allow plugins to alter this as needed.

vboctor:
Same here.  I don't think we should have an a config option for this.  I would like to see us move to HTML emails rather than text in core.

We could add an event that handles formatting and sending an email from the email queue.  But I wonder if we should consider having core do the HTML emails instead.

atrol:
> But I wonder if we should consider having core do the HTML emails instead.

👍 The core formatting plugin might be the right place for it.

libregeek:
I'm also not a big fan of adding another config option, but our users are complaining constantly and I thought of a quick solution and ended up with that.

Most of the webapps are moving towards HTML mail and majority of the email clients (except text based clients) supports HTML emails pretty well. So I think Mantis should seriously consider moving to HTML mail. If someone doesn't like HTML email, they could use EVENT_DISPLAY_EMAIL to remove HTML formatting. I would also like to see this as a user preference rather than a mandatory/system setting.

AFAIK, the bottleneck is the Content-Type header in email. This is hard-coded, and it would be really nice if Mantis could open up the API to modify the email content-type by the plugins. 

One another item in my wishlist is to have an event just before calling `email_format_bug_message`. This way plugins could change the format/fields in email if required.

dregad:
> AFAIK, the bottleneck is the Content-Type header in email. This is hard-coded, and it would be really nice if Mantis could open up the API to modify the email content-type by the plugins.

This is certainly feasible, I suggest you open an issue in our tracker, and if willing, submit a pull request to propose an implementation for review.

> One another item in my wishlist is to have an event just before calling email_format_bug_message. This way plugins could change the format/fields in email if required.

Same as above.

Since I think we all agree that the proposed solution is not something we'd implement, I'll close this PR.

atrol:
> Most of the webapps are moving towards HTML mail 

> I don't think we should have an a config option for this. I would like to see us move to HTML emails rather than text in core.

Being some kind of standard does not necessarily mean that HTML is good in all aspects.
e.g. HTML emails generate more network traffic, need more storage and are less secure than pure text emails.

We should not move from TEXT to HTML, but have the option to choose between them.
It should be an option, similar to what we have for markdown (plugin_config_get( 'process_markdown' ))

libregeek:
I agree, we can close this PR.

I don't have much experience in creating events, but I will definitely try and will submit an implementation for review.

PR #1121, opened 2017-06-09
Fixed initial setup.

7h3ju57:
Looks like a past commit broke the creation of the database during install, once you clicked install nothing would happen.

[#22985](https://mantisbt.org/bugs/view.php?id=22985)

### DIFF
--- a/admin/install.php
+++ b/admin/install.php
@@ -495,12 +495,12 @@ if( 2 == $t_install_state ) {
 </tr>
 
 <?php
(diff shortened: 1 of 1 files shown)


dregad:
Thanks for your contribution.

A note for future submissions: could you please set your git config's `user.email` correctly ? Judging by the commit in this PR, it is currently set as your name instead of your actual e-mail.

7h3ju57:
sure thing,

PR #1122, opened 2017-06-09
hard coded words in some pages

7h3ju57:
Some of the pages had hard coded words, replaced with plugin_get_lang

[#22987](https://mantisbt.org/bugs/view.php?id=22987)


dregad:
Next time, could you please create an issue in our tracker and reference it here ? Thanks in advance.

Also, same remark as in https://github.com/mantisbt/mantisbt/pull/1121#issuecomment-307451882 regarding author's e-mail.

PR #1123, opened 2017-06-09
Update bootstrap-datetimepicker to 4.17.47

dregad:
When opening the date/time picker widget on a field with an existing
date, the calendar showed the current date instead of the one in the
field. After clicking on the calendar icon a second time, the widget
displays the correct date.

Fixes [#22984](https://mantisbt.org/bugs/view.php?id=22984)


atrol:
@dregad is there a reason that you didn't merge?

dregad:
> is there a reason that you didn't merge?

Other than being busy and forgetting about it ? I don't think so / remember

atrol:
Thanks @dregad

PR #1124, opened 2017-06-19
Fix time tracking reports

miqmago:
Reports where not showing as the parameters where being fetched $t_bug and it was empty.

Right now they are obtained with gpc_get_string.

This also fixes the problem that posted dates where not remaining in fields once post request had been sent.

### DIFF
--- a/billing_inc.php
+++ b/billing_inc.php
@@ -54,7 +54,10 @@ require_api( 'utility_api.php' );
 <?php
 
 $t_today = date( 'd:m:Y' );
-$t_date_submitted = isset( $t_bug ) ? date( 'd:m:Y', $t_bug->date_submitted ) : $t_today;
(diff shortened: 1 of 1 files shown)


miqmago:
I'm not really understanding the travis error and how can it be fixed:

```
Creating project
ERROR 1146 (42S02) at line 1: Table 'bugtracker.mantis_project_table' doesn't exist
```

dregad:
The CI build error is due to an issue with Travis itself, and is not related to your PR. You can ignore it. In fact I just re-ran the failing job and it succeeded.

That being said, can you please relate your pull request with an issue on our bugtracker detailing what the problem is. If the issue does not exist, please create one. The bug ID also needs to be referenced in the commit message, e.g. `Fixes #xxxx`.

dregad:
@miqmago ping

PR #1125, opened 2017-06-30
Update file_download.php

rattkin:
https://www.mantisbt.org/bugs/view.php?id=12313

on my server, In my case $t_content_type is 'image/jpeg' there is no ';'

### DIFF
--- a/file_download.php
+++ b/file_download.php
@@ -198,7 +198,12 @@ $t_mime_force_inline = array(
 	'application/pdf' );
 $t_mime_force_attachment = array( 'application/x-shockwave-flash', 'text/html' );
 
-$t_mime_type = substr( $t_content_type, 0, strpos( $t_content_type, ';' ) );
+# extract mime type from content type
+if( strpos( $t_content_type, ';' ) === false ) {
+        $t_mime_type = $t_content_type;
+} else {
+	$t_mime_type = substr( $t_content_type, 0, strpos( $t_content_type, ';' ) );
+}
 
 if( in_array( $t_mime_type, $t_mime_force_inline ) ) {
 	$t_show_inline = true;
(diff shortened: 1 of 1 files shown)


file_download.php
```diff
@@ -198,7 +198,12 @@
 	'application/pdf' );
 $t_mime_force_attachment = array( 'application/x-shockwave-flash', 'text/html' );
 
-$t_mime_type = substr( $t_content_type, 0, strpos( $t_content_type, ';' ) );
+# extract mime type from content type
+if( strpos( $t_content_type, ';' ) === false ) {
+        $t_mime_type = $t_content_type;
+} else {
+	$t_mime_type = substr( $t_content_type, 0, strpos( $t_content_type, ';' ) );
+}
```
dregad: I would suggest 

``` php
# extract mime type from content type
$t_mime_type = explode( ';', $t_content_type, 2 );
$t_mime_type = $t_mime_type[0];
```

vboctor: I think it would be more readable to get the strpos return value in a $t_semicolon_position variable and re-use it when truncating if `;` was found.

file_download.php
```diff
@@ -198,7 +198,12 @@
 	'application/pdf' );
 $t_mime_force_attachment = array( 'application/x-shockwave-flash', 'text/html' );
 
-$t_mime_type = substr( $t_content_type, 0, strpos( $t_content_type, ';' ) );
+# extract mime type from content type
+if( strpos( $t_content_type, ';' ) === false ) {
+        $t_mime_type = $t_content_type;
```
vboctor: Use a tab for spacing and align with indentation in else.

file_download.php
dregad: Less efficient that way: 1 or 3 function calls + 1 if and 1 assignment, vs 1 function call and 2 assignments. It's readable enough with the comment.

vboctor:
@dregad I suggest you squash and merge with your desired code tweak.  Thanks @rattkin

dregad:
Fixed in 54929f3b4b5df695ce59d8d258fe9aa52482a7dc, many thanks @rattkin for your contribution !

PR #1126, opened 2017-07-02
Highlight the active plugin page in manage menu

dregad:
Before
selection_001

After
selection_003

Fixes [#23061](https://mantisbt.org/bugs/view.php?id=23061)

### DIFF
--- a/core/constant_inc.php
+++ b/core/constant_inc.php
@@ -689,6 +689,9 @@ define( 'AUTH_PAGE_CREDENTIAL', 'login_password_page.php' );
 # SLIM FRAMEWORK ATTRIBUTES
 define( 'ATTRIBUTE_FORCE_API_ENABLED', 'force_enable_api' );
 
+# Default parent pages for sub-menu items
+define( 'PAGE_CONFIG_DEFAULT', 'adm_permissions_report.php' );
+
 # Configuration management actions (adm_config_report.php)
 define( 'MANAGE_CONFIG_ACTION_CREATE', 'create' );
 define( 'MANAGE_CONFIG_ACTION_CLONE', 'clone' );
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -640,7 +640,7 @@ function print_manage_menu( $p_page = '' ) {
 		echo '</li>' . "\n";
 	}
 
-	# Plugins menu items - these are cooked links
+	# Plugins menu items - these are html hyperlinks (<a> tags)
 	foreach( $t_menu_options as $t_menu_item ) {
 		$t_active = $p_page && strpos( $t_menu_item, $p_page ) !== false
 			? ' class="active"'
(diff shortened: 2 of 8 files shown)


dregad:
Added a fix for [#22730](https://mantisbt.org/bugs/view.php?id=22730)

core/constant_inc.php
```diff
@@ -689,6 +689,9 @@
 # SLIM FRAMEWORK ATTRIBUTES
 define( 'ATTRIBUTE_FORCE_API_ENABLED', 'force_enable_api' );
 
+# Default parent page to use for Manage Configuration sub-menu items
+define( 'DEFAULT_CONFIG_PAGE', 'adm_permissions_report.php' );
```
vboctor: I would suggest `PAGE_CONFIG_DEFAULT`, so in the future, you can add similar constants with same prefix.  This is similar to what we do with other configs/constants.  For example, `MANAGE_CONFIG_ACTION_CREATE`.

core/html_api.php
```diff
@@ -642,7 +642,8 @@ function print_manage_menu( $p_page = '' ) {
 
 	# Plugins menu items - these are cooked links
 	foreach( $t_menu_options as $t_menu_item ) {
-		echo '<li>', $t_menu_item, '</li>';
+		$t_active = $p_page && strpos( $t_menu_item, $p_page ) !== false ? 'class="active"' : '';
+		echo "<li $t_active>", $t_menu_item, '</li>';
```
vboctor: I would have the space as part of the $t_active string, so it if it is not set, then you end up with no space.

core/html_api.php
```diff
@@ -642,7 +642,8 @@ function print_manage_menu( $p_page = '' ) {
 
 	# Plugins menu items - these are cooked links
 	foreach( $t_menu_options as $t_menu_item ) {
-		echo '<li>', $t_menu_item, '</li>';
+		$t_active = $p_page && strpos( $t_menu_item, $p_page ) !== false ? 'class="active"' : '';
```
vboctor: For the comment above, I don't know what is meant by cooked links.  Can you provide an example of a link to make it more readable?

core/constant_inc.php
dregad: Makes sense. Will do.

core/html_api.php
dregad: This term has been there [since grangeway added the comment back in 2009](https://github.com/mantisbt/mantisbt/commit/0d17ca9174941e550d3866f8a6c1b459c7cd6815#diff-b1c667913de013265f22c582987aa38cR879).

Looking at the code, *$t_menu_options* is populated by plugin event EVENT_MENU_MANAGE. The [documentation](http://mantisbt.org/docs/master/en-US/Developers_Guide/html-desktop/#dev.eventref.output.menu) 

> Return Value
> <Array>: List of HTML links for the management menu. 

So *cooked links* = HTML hyperlinks (`<a href="url">label</a>`).

core/html_api.php
dregad: OK

PR #1127, opened 2017-07-03
Add option to show summary of issues instead of ID in timeline

Zipher04:
It was hard for our team to know which issue was updated by just reading the issue IDs in timeline. We needed to mouse over each ID to see the summary and recognized the issue. I have added an option to change the IDs in timeline to summary of issues. Hope this can help someone has a similar need.

The changes are summarize as:

1. A new function *string_get_bug_view_link_with_summary()* is created. This function return a hyperlink to issue with summary as text.
2. And a new configuration *$g_timeline_show_issue_summary* is created. In order to change timeline display the summary in issue links.
3. Every derived TimelineEvent class reads the configuration and call the *string_get_bug_view_link_with_summary()* or *the original string_get_bug_view_link()*

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -3618,6 +3618,10 @@ $g_my_view_boxes = array(
  */
 $g_my_view_boxes_fixed_position = ON;
 
+/**
+ * Toggle whether 'Timeline' boxes show summary of issue after issue ID
+ */
+$g_timeline_show_issue_summary = OFF;
 
 #############
 # RSS Feeds #
--- a/core/classes/IssueAssignedTimelineEvent.class.php
+++ b/core/classes/IssueAssignedTimelineEvent.class.php
@@ -49,15 +49,21 @@ class IssueAssignedTimelineEvent extends TimelineEvent {
 	 * @return string
 	 */
 	public function html() {
+		$t_show_summary = config_get( 'timeline_show_issue_summary' );
+		if ( $t_show_summary ) {
+			$t_link = string_get_bug_view_link_with_summary( $this->issue_id );
+		} else {
+			$t_link = string_get_bug_view_link( $this->issue_id );
+		}
 		if( $this->user_id == $this->handler_id ) {
 			$t_html = $this->html_start( 'fa-flag-o' );
-			$t_string = sprintf( lang_get( 'timeline_issue_assigned_to_self' ), user_get_name( $this->user_id ), string_get_bug_view_link( $this->issue_id ) );
+			$t_string = sprintf( lang_get( 'timeline_issue_assigned_to_self' ), user_get_name( $this->user_id ), $t_link );
 		} else if( $this->handler_id != NO_USER ) {
 			$t_html = $this->html_start( 'fa-hand-o-right' );
-			$t_string = sprintf( lang_get( 'timeline_issue_assigned' ), user_get_name( $this->user_id ), string_get_bug_view_link( $this->issue_id ), user_get_name( $this->handler_id ) );
+			$t_string = sprintf( lang_get( 'timeline_issue_assigned' ), user_get_name( $this->user_id ), $t_link, user_get_name( $this->handler_id ) );
 		} else {
             $t_html = $this->html_start( 'fa-flag-o' );
-			$t_string = sprintf( lang_get( 'timeline_issue_unassigned' ), user_get_name( $this->user_id ), string_get_bug_view_link( $this->issue_id ) );
+			$t_string = sprintf( lang_get( 'timeline_issue_unassigned' ), user_get_name( $this->user_id ), $t_link );
 		}
 
 		$t_html .= '<div class="action">' . $t_string . '</div>';
--- a/core/classes/IssueMonitorTimelineEvent.class.php
+++ b/core/classes/IssueMonitorTimelineEvent.class.php
@@ -49,10 +49,16 @@ class IssueMonitorTimelineEvent extends TimelineEvent {
 	 * @return string
 	 */
 	public function html() {
+		$t_show_summary = config_get( 'timeline_show_issue_summary' );
+		if ( $t_show_summary ) {
+			$t_link = string_get_bug_view_link_with_summary( $this->issue_id );
+		} else {
+			$t_link = string_get_bug_view_link( $this->issue_id );
+		}
 		$t_string = $this->monitor ? lang_get( 'timeline_issue_monitor' ) : lang_get( 'timeline_issue_unmonitor' );
 
 		$t_html = $this->html_start( 'fa-eye' );
-		$t_html .= '<div class="action">' . sprintf( $t_string, user_get_name( $this->user_id ), string_get_bug_view_link( $this->issue_id ) ) . '</div>';
+		$t_html .= '<div class="action">' . sprintf( $t_string, user_get_name( $this->user_id ), $t_link ) . '</div>';
 		$t_html .= $this->html_end();
 
 		return $t_html;
--- a/core/classes/IssueNoteCreatedTimelineEvent.class.php
+++ b/core/classes/IssueNoteCreatedTimelineEvent.class.php
@@ -49,8 +49,14 @@ class IssueNoteCreatedTimelineEvent extends TimelineEvent {
 	 * @return string
 	 */
 	public function html() {
+		$t_show_summary = config_get( 'timeline_show_issue_summary' );
+		if ( $t_show_summary ) {
+			$t_link = string_get_bug_view_link_with_summary( $this->issue_id );
+		} else {
+			$t_link = string_get_bug_view_link( $this->issue_id );
+		}
 		$t_html = $this->html_start( 'fa-comment-o' );
-		$t_html .= '<div class="action">' . sprintf( lang_get( 'timeline_issue_note_created' ), user_get_name( $this->user_id ), string_get_bug_view_link( $this->issue_id ) ) . '</div>';
+		$t_html .= '<div class="action">' . sprintf( lang_get( 'timeline_issue_note_created' ), user_get_name( $this->user_id ), $t_link ) . '</div>';
 		$t_html .= $this->html_end();
 
 		return $t_html;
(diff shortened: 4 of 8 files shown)


vboctor:
Thanks @Zipher04 for your contributions.  Few points to discuss:

- Should we introduce a new function, or a parameter to the standard method to get a short link or a long link.  short is just the id, long is the one that contains the title.  The default will be short for backward compatibility, but the caller can provide based on a config value.
- In case of the mode that includes the summary, should we have a limit on the length of the summary we display?
- In case of the mode that includes the summary, should we use "id: summary" instead of "summary".
- For existing places where we use links like View Issue page, Timeline, etc.  Should we use the longer format?

Zipher04:
@vboctor 
Thank you for the suggestion. I will try

- Add a default parameter "$p_long_link = false" to the standard method. 
- Add a configuration "g_timeline_issue_summary_length".
- If "$p_long_link = true", return the link as "id: summary".
- I think the "Recently Visited" section in two-column format can use the long link too.

config_defaults_inc.php
```diff
@@ -3618,6 +3618,15 @@
...
 
+/**
+ * Toggle whether 'Timeline' boxes show summary of issue after issue ID
+ */
+$g_timeline_show_issue_summary = OFF;
+
+/**
+ * Control the maximum length of each summary shown in timeline when $g_timeline_show_issue_summary is ON.
+ */
+$g_timeline_issue_summary_length = 20;
```
atrol: The options should be added to array $g_public_config_names

core/string_api.php
```diff
@@ -666,7 +667,15 @@ function string_get_bug_view_link( $p_bug_id, $p_detail_info = true, $p_fqdn = f
 				$t_link .= ' class="resolved"';
 			}
 		}
-		$t_link .= '>' . bug_format_id( $p_bug_id ) . '</a>';
+		$t_link .= '>' . bug_format_id( $p_bug_id );
+		if ( $p_long_link ) {
+			$t_short_summary = $t_summary;
```
atrol: This line should be removed as the value of $t_short_summary is not used before the next assignment of $t_short_summary.

core/string_api.php
```diff
@@ -666,7 +667,15 @@ function string_get_bug_view_link( $p_bug_id, $p_detail_info = true, $p_fqdn = f
...
+			$t_summary_length = config_get( 'timeline_issue_summary_length' );
+			$t_short_summary = mb_strimwidth( $t_summary, 0, $t_summary_length, '...', 'utf-8' );
```
atrol: Remove 'utf-8' as we have set mb_internal_encoding.

BTW @dregad, I am note complete sure at the moment if we still have to check for `extension_loaded('mbstring')` or if we assume that the extension is always enabled.
At least during installation you made it mandatory since 1.3.5, see https://www.mantisbt.org/bugs/view.php?id=21959 but upgraded old installations might still run without mbstring enabled.

config_defaults_inc.php
atrol: Furthermore the options should be documented in Admin Guide.

core/string_api.php
```diff
@@ -644,9 +644,10 @@ function string_get_bug_page( $p_action ) {
  * @param integer $p_bug_id	     A bug identifier.
  * @param boolean $p_detail_info Whether to include more detailed information (e.g. title attribute / project) in the returned string.
  * @param boolean $p_fqdn        Whether to return an absolute or relative link.
+ * @param boolean $p_long_link 	 When set to false, the returned link is displayed as bug ID. When set to true, the return string is displayed as ID: summary.
```
atrol: > the return string is displayed as ID: summary

This is not what has been implemented, at the moment you display <ID><ShortenedSummary> without any space.
My proposal is to display <ID><Space><ShortenedSummary>

atrol:
> Add a configuration "g_timeline_issue_summary_length".

I think we should not have a hardcoded number for it, as there is no reason to shorten strings if there is enough space available.

Didn't try, maybe CSS `text-overflow: ellipsis;` can be used for it.
https://www.w3schools.com/cssref/css3_pr_text-overflow.asp

config_defaults_inc.php
Zipher04: I found that $g_public_config_names is for options available for SOAP API. But I am not sure how to update the SOAP API. If someone can help me on this, it will be much appreciated.

atrol: There is no need to change anything in SOAP API.
You have just to add the option to array $g_public_config_names in config_defaults_inc.php.
Keep in mind that the current entries are alphabetically sorted.

core/string_api.php
Zipher04: Thank you for the correction. I had corrected this to '<ID>:<space><ShortenedSummary>'

core/string_api.php
Zipher04: Thank you for the correction. I had removed this line.

config_defaults_inc.php
Zipher04: Thank you for the guidance. I have added this in 3953b7c. But I do not know how to check whether it works.

atrol: At least the part of the CI process that builds the documentation did not show any errors
https://travis-ci.org/mantisbt/mantisbt/jobs/254667677

You would have to setup Publican to see the final documentation on your system.
Should be no big deal if you are running Linux, for more details see https://github.com/mantisbt/mantisbt/blob/master/docbook/README.md

atrol:
One more general comment.

Please open an issue at https://www.mantisbt.org/bugs to track the change.

Furthermore, after having made all requested changes, squash your commits and reword the body of the commit message, so that it includes something like `Fixes #23150`, for example https://github.com/mantisbt/mantisbt/commit/66d035a046224ccbae12ada77e66215e658132ce

Zipher04:
Thank you for the reference of CSS, I will give it a try.

The things to do are:
1. Open a issue to track this PR. 
2. Document the new option in Admin Guide
3. Control the overflowed behavior by CSS.
4. Squash the commit.

Sorry if I had made some clutter on this PR. Because GIT and CSS are not so familiar to me. I am really appreciate your helps on this PR. Being a C/C++ programmer and use mantis for 5 years, I am glad I can do a little help on this project!

atrol:
> Sorry if I had made some clutter on this PR. Because GIT and CSS are not so familiar to me.

No worries.

> Being a C/C++ programmer

Similar to me, far away from being a HTML / CSS / PHP guru.

>  I am glad I can do a little help on this project!

Thanks for that! 

> Control the overflowed behavior by CSS.

Maybe we should define a CSS class for it, maybe use an existing one.
@syncguru, is adding something like
```
.issue_id {
	text-overflow: ellipsis;
	...
	...
}
```
to `default.css`  the right way to start?

Zipher04:
@atrol 
I have try the text-overflow method and modified the `default.css`. But when I resize the windows the link wrap around into multiple lines. In order to prevent the wrapping, I had to add the following CSS and limit the width of the div in px. But this is not what we want. We would like the width to be auto-sized.

```CSS
.issue_id {
	text-overflow: ellipsis;
	display: inline-block;
	overflow: hidden;
	white-space: nowrap;
	vertical-align: text-top;
	width: 240px;
}
```

I commit it to another branch. Please have a look at
https://github.com/Zipher04/mantisbt/tree/timelineSummaryCSS

atrol:
@Zipher04, as I told before, I am no CSS specialist and have not the time to learn enough to be able to help you.
I hope that @syncguru or @vboctor can push you in the right direction.

core/classes/IssueAssignedTimelineEvent.class.php
```diff
@@ -49,15 +49,18 @@ public function __construct( $p_timestamp, $p_user_id, $p_issue_id, $p_handler_i
 	 * @return string
 	 */
 	public function html() {
+		$t_show_summary = config_get( 'timeline_show_issue_summary' );
+		$t_link = string_get_bug_view_link( $this->issue_id, true, false, $t_show_summary );
```
dregad: These 2 lines of code are repeated in every html() method of all Timeline event classes.

I suggest instead to define a method that returns properly formatted HTML for the bug link in _TimelineEvent_ class, and call it everywhere you reference $t_link.

core/classes/IssueTagTimelineEvent.class.php
```diff
@@ -52,6 +52,9 @@ public function __construct( $p_timestamp, $p_user_id, $p_issue_id, $p_tag_name,
 	 * @return string
 	 */
 	public function html() {
+		$t_show_summary = config_get( 'timeline_show_issue_summary' );
+		$t_link = string_get_bug_view_link( $this->issue_id, true, false, $t_show_summary );
```
dregad: See above

core/classes/IssueStatusChangeTimelineEvent.class.php
```diff
@@ -91,29 +91,32 @@ public function skip() {
 	 * @return string
 	 */
 	public function html() {
+		$t_show_summary = config_get( 'timeline_show_issue_summary' );
+		$t_link = string_get_bug_view_link( $this->issue_id, true, false, $t_show_summary );
```
dregad: See above

core/classes/IssueNoteCreatedTimelineEvent.class.php
```diff
@@ -49,8 +49,11 @@ public function __construct( $p_timestamp, $p_user_id, $p_issue_id, $p_issue_not
 	 * @return string
 	 */
 	public function html() {
+		$t_show_summary = config_get( 'timeline_show_issue_summary' );
+		$t_link = string_get_bug_view_link( $this->issue_id, true, false, $t_show_summary );
```
dregad: See above

core/classes/IssueMonitorTimelineEvent.class.php
```diff
@@ -49,10 +49,13 @@ public function __construct( $p_timestamp, $p_user_id, $p_issue_id, $p_monitor )
 	 * @return string
 	 */
 	public function html() {
+		$t_show_summary = config_get( 'timeline_show_issue_summary' );
+		$t_link = string_get_bug_view_link( $this->issue_id, true, false, $t_show_summary );
```
dregad: See above

Zipher04:
Nice suggestion! I will do that this weekend.

PR #1128, opened 2017-07-14
Fix HTML syntax error

mamidenn:
470adca85fd8d004068f97b29ef94625041ed094 left a single `"` that was causing display errors on some browsers. I removed it.


vboctor:
Thanks @mamidenn for your contribution.  Can you please open a bug for this in our bug tracker so we can reference it from the commit.

Otherwise, change looks good.

mamidenn:
References 23116.

PR #1130, opened 2017-07-21
Fixed #23136 Added new event to change email content-type

libregeek:
Added a chain type event to modify the content-type of email notifications. This may help plugin developers to send HTML mails.

This is not a standalone event, and it may need another event to change the email body format. This is work in progress.

Please review and let me know your feedback.

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -1300,7 +1300,8 @@ function email_send( EmailData $p_email_data ) {
 			break;
 	}
 
-	$t_mail->IsHTML( false );              # set email format to plain text
+	# default email format is plain text
+	$t_mail->IsHTML( event_signal( 'EVENT_EMAIL_HTML_CONTENT' ) );	# default email format is plain text
 	$t_mail->WordWrap = 80;              # set word wrap to 80 characters
 	$t_mail->CharSet = $t_email_data->metadata['charset'];
 	$t_mail->Host = config_get( 'smtp_host' );
--- a/core/events_inc.php
+++ b/core/events_inc.php
@@ -149,4 +149,8 @@ event_declare_many( array(
 
 	# Authentication Events
 	'EVENT_AUTH_USER_FLAGS' => EVENT_TYPE_FIRST,
+
+	# Email Events
+	# Enable HTML Content-Type in email notifications
+	'EVENT_EMAIL_HTML_CONTENT' => EVENT_TYPE_CHAIN,
 ) );


libregeek:
I'm sorry, I'm slightly confused. As you know there was a discussion on this in #1120 and I thought Mantis doesn't want to move to HTML completely, but have an option to switch between HTML and Text in the core plugin. 

I'm under the impression that events are required even if we are using the core formatting plugin. Is this correct? If not I can modify this PR.

atrol:
Just brainstorming, not sure if it's the right approach.

There are two places where the format (HTML / PLAIN_TEXT / maybe even more in future) has to be considered.

1. when building the content befor it's stored in email queue (mantis_email_table)
2. later on when sending the email based on the content in the queue (via cron job or shutdown function)

So the first step is to ensure that the format information is stored in the queue.
This is needed if we want to allow that individual end users are able to configure if they want to get HTML or PLAIN_TEXT notifications.

The easiest way for this could be to add the format to the existing metadata in class EmailData

config_defaults_inc.php
`$g_email_notification_format = PLAIN_TEXT;`

function email_store
`$t_email_data->metadata['format'] = config_get ( 'email_notification_format' );`

function email_send
`$t_mail->IsHTML( $t_email_data->metadata['format'] == HTML );`

If we want the user specific setting, it should be implemented on "My Account" > "Preferences" similar to what has been implemented for "E-mail Full Issue Details" ($g_email_notifications_verbose) 
This means that the setting is stored in config_table but not in user_pref_table.
Advantages:
- no database schema change
- allows the administrator to globally change the format (sets $g_email_notification_format = HTML; in config_inc.php)
- allows single users to set back to PLAIN_TEXT if they want
- does not change existing installations
- allows us to get some epxerience with it and to change later on the default setting in config_defaults_inc.php to HTML in next major version

vboctor:
I would say that the goal is to have all email notifications as HTML, hence, long term, we don't need to have a way to choose between HTML and TEXT globally or per user.

However, it is fair to assume that there is a transition state where some notifications from core / plugins are still in TEXT and others are converted to HTML.  In such case, the code that creates the notification can choose the format as suggested by @atrol 

This will make the code much simpler, rather than every notification having to be implemented in the two formats.

We will also need to have string email formatting functions that target text vs. html.  Hence, the content of the email is escaped accordingly.  For example, in case of HTML, convert mark down to HTML, etc.

PR #1135, opened 2017-07-26
Fixed #23157: phpMailer SMTP Verify Peer

lucentminds:
This will add an additional config option for sending ssl/tls smtp e-mails in php 5.6.

### DIFF
--- a/config/config_inc.php.sample
+++ b/config/config_inc.php.sample
@@ -51,9 +51,15 @@ $g_anonymous_account		= '';
 
 # --- Email Configuration ---
 $g_phpMailer_method		= PHPMAILER_METHOD_MAIL; # or PHPMAILER_METHOD_SMTP, PHPMAILER_METHOD_SENDMAIL
-$g_smtp_host			= 'localhost';			# used with PHPMAILER_METHOD_SMTP
-$g_smtp_username		= '';					# used with PHPMAILER_METHOD_SMTP
-$g_smtp_password		= '';					# used with PHPMAILER_METHOD_SMTP
+
+# used with PHPMAILER_METHOD_SMTP
+$g_smtp_host			= 'localhost';
+$g_smtp_username		= '';
+$g_smtp_password		= '';
+# $g_smtp_connection_mode = ''; # Valid values are '' (no encryption), 'ssl' or 'tls'
+# $g_smtp_port = 25; # default unencrypted: 25  default tls: 587
+# $g_smtp_ssl_verify_peer = true; # Can be set to false to allow self-signed certs from mail server.
+
 $g_webmaster_email      = 'webmaster@example.com';
 $g_from_email           = 'noreply@example.com';	# the "From: " field in emails
 $g_return_path_email    = 'admin@example.com';	# the return address for bounced mail
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -680,6 +680,14 @@ $g_smtp_connection_mode = '';
  */
 $g_smtp_port = 25;
 
+/** 
+ * When sending e-mail via ssl or tls, the mail server you are sending to may be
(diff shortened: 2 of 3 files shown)


lucentminds:
According to php.net `verify_peer` is part of openssl context options. Since php 5.6.x peer verification is enabled by default. This change may add a small amount of complexity, but it should not limit what e-mail library is used. This setting is general for using ssl/tls with an smtp library. It is more specific to openssl than it is to an e-mail library.

It should allow mantis to use a trusted smtp server that uses a self-signed certificate.

atrol:
@vboctor, this means that users might get regressions after upgrading PHP from versions < 5.6.
Do we want to tell them to change source to fix it?

dregad:
I'm in favor of allowing admins to fine-tune SSL settings for SMTP.

However, I find the approach proposed here too specific and restrictive; what if someone needs to set another SSL context option ? Add yet another config to Mantis ? I feel it would be better to provide a more generic config like `$g_smtp_ssl_context_options`, defaulting to an empty array, and referring to http://php.net/manual/en/context.ssl.php for valid values.

The change in email_api.php then becomes (I assume PHPMailer will somehow choke if given invalid options, and that we would catch the error, but didn't test)

``` php
$t_mail->SMTPOptions = array ( 'ssl' => config_get( 'smtp_ssl_context_options' ) );
```

As for limiting what email library we could use in the future, I believe that switching away from PHPMailer is highly unlikely, so that should not be an argument against this PR.

lucentminds:
When I get some time, I'll test PHPMailer and to see what happens if you feed it invalid options.

atrol:
Seems there is a real need for it https://www.mantisbt.org/forums/viewtopic.php?f=2&t=25140

@vboctor do you still think
> I would rather not add a config for this.

dregad:
> When I get some time, I'll test PHPMailer and to see what happens if you feed it invalid options.

@lucentminds did you ever get around to testing PHPMailer's behavior ?

dregad:
Thanks for your feedback. In that case I would suggest that you revise your pull request, removing the new `$g_smtp_ssl_verify_peer` variable and replacing it with `$g_smtp_options`, with a descriptive PHPDoc block pointing to the relevant PHP documentation section; please also document the new config option in docbook/Admin_Guide/en-US/config/email.xml

atrol:
> I'm using phpmailer 6.0.3

I didn't check if it is relevant, but keep in mind that we are using 5.2.26

dregad:
> we are using 5.2.26

True. Based on my experience with other projects, upgrading to 6.x should be quite straightforward though, and I think we should do it.

dregad:
First attempt, not fully tested yet, but e-mails are being sent on my dev box https://github.com/dregad/mantisbt/commits/phpmailer6

lucentminds:
Ok. I'm running into warnings from php.
```
PHP Warning:  stream_context_create(): options should have the form [&quot;wrappername&quot;][&quot;optionname&quot;] = $value in c:\path\to\my\project\vendor\phpmailer\phpmailer\src\SMTP.php on line 317
```
Looks like [stream_context_create is looking for an array inside of an array](https://stackoverflow.com/questions/15749755/stream-notification-callback-doesnt-seem-to-work). So the first level string value's I put in there do not work without a warning. I've since changed the code like this...
```php
$mail->SMTPOptions = array(
   'ssl' => array(
      'Hey jude,' => "don't make it bad",
      'Take a sad song' => 'and make it better',
      'verify_peer' => 0,
      'Remember to let her into your heart' => 'Then you can start to make it better'
   )
);
```
This works with no errors, warnings, or notifications of any kind.

atrol:
> First attempt, not fully tested yet, but e-mails are being sent on my dev box

Same on my Ubuntu 16.04 test server

PR #1136, opened 2017-07-26
Show file added event in timeline

Zipher04:
Add an new event

I had added a new event for class TimelineEevent. In order to show events for file added in issue in the timeline.

Also I prevent events that have same issue id, user id, event type, and timestamp to show multiple times in timeline. Such that when multiple files are added to issue at a time. There will be only one timeline event for this.

A new item $s_timeline_issue_file_added is introduce in the language file. But I am only able to add the translation for English and Traditional Chinese.

### DIFF
--- a/core/classes/IssueFileAddedTimelineEvent.class.php
+++ b/core/classes/IssueFileAddedTimelineEvent.class.php
@@ -1,0 +1,58 @@
+<?php
+# MantisBT - A PHP based bugtracking system
+
+# MantisBT is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# MantisBT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Timeline event class for adding files to issues.
+ * @copyright Copyright 2014 MantisBT Team - mantisbt-dev@lists.sourceforge.net
+ * @link http://www.mantisbt.org
+ * @package MantisBT
+ */
+
+/**
+ * Timeline event class for creation of issues.
+ *
+ * @package MantisBT
+ * @subpackage classes
+ */
+class IssueFileAddedTimelineEvent extends TimelineEvent {
+	private $issue_id;
+
+	/**
+	 * @param integer $p_timestamp Timestamp representing the time the event occurred.
+	 * @param integer $p_user_id   An user identifier.
+	 * @param integer $p_issue_id  A issue identifier.
+	 */
+	public function __construct( $p_timestamp, $p_user_id, $p_issue_id ) {
+		parent::__construct( $p_timestamp, $p_user_id );
+
+		$this->issue_id = $p_issue_id;
+	}
+
+	/**
+	 * Returns html string to display
+	 * @return string
+	 */
+	public function html() {
+		$t_show_summary = config_get( 'timeline_show_issue_summary' );
+		$t_link = string_get_bug_view_link( $this->issue_id, true, false, $t_show_summary );
+		
+		$t_html = $this->html_start( 'fa-plus' );
+		$t_html .= '<div class="action">' . sprintf( lang_get( 'timeline_issue_file_added' ), user_get_name( $this->user_id ), $t_link ) . '</div>';
+		$t_html .= $this->html_end();
+
+		return $t_html;
+	}
+}
--- a/core/classes/IssueFileAddedTimelineEvent.class.php
+++ b/core/classes/IssueFileAddedTimelineEvent.class.php
@@ -46,8 +46,7 @@ class IssueFileAddedTimelineEvent extends TimelineEvent {
 	 * @return string
 	 */
 	public function html() {
-		$t_show_summary = config_get( 'timeline_show_issue_summary' );
-		$t_link = string_get_bug_view_link( $this->issue_id, true, false, $t_show_summary );
+		$t_link = string_get_bug_view_link( $this->issue_id );
 		
 		$t_html = $this->html_start( 'fa-plus' );
 		$t_html .= '<div class="action">' . sprintf( lang_get( 'timeline_issue_file_added' ), user_get_name( $this->user_id ), $t_link ) . '</div>';
--- a/core/timeline_api.php
+++ b/core/timeline_api.php
@@ -60,21 +60,22 @@ function timeline_events( $p_start_time, $p_end_time, $p_max_events, $p_filter =
 	$t_previous_history_event = null;
 	while ( $t_history_event = history_get_event_from_row( $t_result, /* $p_user_id */ auth_get_current_user_id(), /* $p_check_access_to_issue */ true ) ) {
 		$t_event = null;
-		$t_user_id = $t_history_event['userid'];
+		$t_user_id = (int)$t_history_event['userid'];
 		$t_timestamp = $t_history_event['date'];
 		$t_issue_id = $t_history_event['bug_id'];
+		$t_type = $t_history_event['type'];
 
 		if ( $t_previous_history_event != null ) {  //check repeated event
-			if (    $t_previous_history_event['userid'] == $t_history_event['userid']
-				 && $t_previous_history_event['date'] == $t_history_event['date']
-				 && $t_previous_history_event['bug_id'] == $t_history_event['bug_id']
-				 && $t_previous_history_event['type'] == $t_history_event['type']	
+			if (    $t_previous_history_event['userid'] == $t_user_id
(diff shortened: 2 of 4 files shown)


Zipher04:
An issue [#23161](https://www.mantisbt.org/bugs/view.php?id=23161) is opened for this.

atrol:
> But I am only able to add the translation for English and Traditional Chinese.

See my comment at https://www.mantisbt.org/bugs/view.php?id=23161#c57334

lang/strings_chinese_traditional.txt
```diff
@@ -1230,6 +1230,7 @@ $s_month_october = '10 月';
 $s_month_november = '11 月';
 $s_month_december = '12 月';
 $s_timeline_issue_created = '<span class="username">%1$s</span> 已建立問題 <span class="issue_id">%2$s</span>';
+$s_timeline_issue_file_added = '<span class="username">%1$s</span> 已新增檔案於議題 <span class="issue_id">%2$s</span>';
```
dregad: As mentioned by @atrol, translations should be made in TranslateWiki after the PR is merged. This should not be part of the change

lang/strings_english.txt
```diff
@@ -1592,6 +1592,7 @@ $s_month_december = 'December';
 
 # Timeline Feature Strings
 $s_timeline_issue_created = '<span class="username">%1$s</span> created issue <span class="issue_id">%2$s</span>';
+$s_timeline_issue_file_added = '<span class="username">%1$s</span> added attachment in issue <span class="issue_id">%2$s</span>';
```
dregad: Suggest rewording: `... added attachment(s) to issue ...`

core/classes/IssueFileAddedTimelineEvent.class.php
```diff
@@ -0,0 +1,58 @@
...
+
+		$this->issue_id = $p_issue_id;
+	}
+
+	/**
+	 * Returns html string to display
+	 * @return string
+	 */
+	public function html() {
+		$t_show_summary = config_get( 'timeline_show_issue_summary' );
```
dregad: This throws `APPLICATION WARNING #100: Configuration option "timeline_show_issue_summary" not found.` , which is a new feature added in #1127 (still under review/discussion); I would suggest to keep the 2 PRs independent from each other.

See also https://github.com/mantisbt/mantisbt/pull/1127#discussion_r130890947

core/timeline_api.php
```diff
@@ -57,12 +57,22 @@ function timeline_events( $p_start_time, $p_end_time, $p_max_events, $p_filter =
...
 		$t_event = null;
 		$t_user_id = $t_history_event['userid'];
 		$t_timestamp = $t_history_event['date'];
 		$t_issue_id = $t_history_event['bug_id'];
 
+		if ( $t_previous_history_event != null ) {  //check repeated event
+			if ( $t_previous_history_event['userid'] == $t_history_event['userid'] &&
+				 $t_previous_history_event['date'] == $t_history_event['date'] &&
+				 $t_previous_history_event['bug_id'] == $t_history_event['bug_id'] &&
+				 $t_previous_history_event['type'] == $t_history_event['type']	) {
```
dregad: Code formatting, for complex if statements, I find it more readable to have the `&&` on the left-hand side, and the closing `) {` on its own line, aligned with the `if(`, i.e.
``` php
if(    cond1
    && cond 2
    && ( cond 3 || cond4 )
) {
```

dregad:
Also, please amend the commit message to include the issue reference `Fixes #23161`

Zipher04:
Thank you for the review, I will update them this weekend.

core/timeline_api.php
```diff
@@ -101,11 +112,16 @@ function timeline_events( $p_start_time, $p_end_time, $p_max_events, $p_filter =
 						break;
 				}
 				break;
+			case FILE_ADDED:
+ 				$t_bugnote_id = $t_history_event['old_value'];
+				$t_event = new IssueFileAddedTimelineEvent( $t_timestamp, $t_user_id, $t_issue_id, $t_bugnote_id );
```
atrol: From class IssueFileAddedTimelineEvent
`public function __construct( $p_timestamp, $p_user_id, $p_issue_id ) `
`$t_bugnote_id` is not needed.

Zipher04: Thank you for the correction!

core/timeline_api.php
```diff
@@ -57,12 +57,23 @@ function timeline_events( $p_start_time, $p_end_time, $p_max_events, $p_filter =
...
 		$t_event = null;
 		$t_user_id = $t_history_event['userid'];
 		$t_timestamp = $t_history_event['date'];
 		$t_issue_id = $t_history_event['bug_id'];
 
+		if ( $t_previous_history_event != null ) {  //check repeated event
+			if (    $t_previous_history_event['userid'] == $t_history_event['userid']
+				 && $t_previous_history_event['date'] == $t_history_event['date']
+				 && $t_previous_history_event['bug_id'] == $t_history_event['bug_id']
+				 && $t_previous_history_event['type'] == $t_history_event['type']
```
atrol: Didn't try, but using $t_user_id, $t_timestamp and $t_issue_id should be better in terms of performance than $t_history_event['userid'], $t_history_event['date'] and $t_history_event['bug_id']

You could also introduce `$t_type = $t_history_event['type'];` and use it in your code and the following switch.

dregad: @atrol I implemented your suggestion

core/classes/IssueFileAddedTimelineEvent.class.php
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	/**
+	 * Returns html string to display
+	 * @return string
+	 */
+	public function html() {
+		$t_link = string_get_bug_view_link( $this->issue_id );
+		
+		$t_html = $this->html_start( 'fa-plus' );
```
vboctor: How about `fa-file-o` to get a file icon instead of a plus icon?  You can argue that plus applies to adding issues, notes, files, etc.

lang/strings_english.txt
```diff
@@ -1592,6 +1592,7 @@ $s_month_december = 'December';
 
 # Timeline Feature Strings
 $s_timeline_issue_created = '<span class="username">%1$s</span> created issue <span class="issue_id">%2$s</span>';
+$s_timeline_issue_file_added = '<span class="username">%1$s</span> added attachment(s) to issue <span class="issue_id">%2$s</span>';
```
vboctor: If we go with entry per file, then use `attachment`.

core/classes/IssueFileAddedTimelineEvent.class.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Timeline event class for adding files to issues.
+ * @copyright Copyright 2014 MantisBT Team - mantisbt-dev@lists.sourceforge.net
+ * @link http://www.mantisbt.org
+ * @package MantisBT
+ */
+
+/**
+ * Timeline event class for creation of issues.
```
vboctor: Stale comment.

core/timeline_api.php
```diff
@@ -57,13 +57,25 @@ function timeline_events( $p_start_time, $p_end_time, $p_max_events, $p_filter =
...
 	while ( $t_history_event = history_get_event_from_row( $t_result, /* $p_user_id */ auth_get_current_user_id(), /* $p_check_access_to_issue */ true ) ) {
 		$t_event = null;
-		$t_user_id = $t_history_event['userid'];
+		$t_user_id = (int)$t_history_event['userid'];
 		$t_timestamp = $t_history_event['date'];
 		$t_issue_id = $t_history_event['bug_id'];
+		$t_type = $t_history_event['type'];
 
-		switch( $t_history_event['type'] ) {
+		if ( $t_previous_history_event != null ) {  //check repeated event
```
vboctor: We don't combine similar actions, e.g. adding multiple tags.  I wonder if we should just keep it as an entry per file.

core/timeline_api.php
```diff
@@ -101,11 +113,15 @@ function timeline_events( $p_start_time, $p_end_time, $p_max_events, $p_filter =
 						break;
 				}
 				break;
+			case FILE_ADDED:
+				$t_event = new IssueFileAddedTimelineEvent( $t_timestamp, $t_user_id, $t_issue_id );
```
vboctor: One of the advantages of having a single event per file is that we can add the file name here, which would make the event more informative.  Similar to how the tag added event includes the tag name.

lang/strings_english.txt
dregad: Fixed

core/timeline_api.php
dregad: Added attachment filename to timeline event.

core/timeline_api.php
dregad: I agree. And in fact this is introducing a regression, e.g. when adding multiple tags in a single operations: only one timeline entry with the last processed tag is shown, instead of one per tag before this change.

core/classes/IssueFileAddedTimelineEvent.class.php
dregad: Fixed

core/classes/IssueFileAddedTimelineEvent.class.php
dregad: Fixed

PR #1138, opened 2017-08-04
21225 close parent with unresolved children

bgamrat:
Allow parent bug to close regardless of child status.  This defaults to OFF to preserve current functionality.

Courtesy of Game Creek Video.

### DIFF
--- a/bug_update.php
+++ b/bug_update.php
@@ -169,7 +169,13 @@ if ( !$t_reporter_reopening && !$t_reporter_closing ) {
 if( ( $t_resolve_issue || $t_close_issue ) &&
 	!relationship_can_resolve_bug( $f_bug_id )
 ) {
-	trigger_error( ERROR_BUG_RESOLVE_DEPENDANTS_BLOCKING, ERROR );
+
+	# this may be overridden with configuration to allow parents to close regardless of
+	# the status of any children
+	if ( OFF == config_get( 'allow_parent_of_unresolved_to_close' ) ) {
+		# retrieve all the relationships in which the bug is the source bug
+		trigger_error( ERROR_BUG_RESOLVE_DEPENDANTS_BLOCKING, ERROR );
+	}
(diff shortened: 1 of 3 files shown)


config_defaults_inc.php
```diff
@@ -1317,6 +1317,13 @@
 $g_default_bug_relationship_clone = BUG_REL_NONE;
 
 /**
+ * Allow parent bug to close regardless of child status.  This defaults to OFF
+ * to preserve current functionality.
+ * @global integer $g_allow_parent_of_unresolved_to_close
+ */
+$g_allow_parent_of_unresolved_to_close = OFF;
```
atrol: The options has to be added to array $g_public_config_names

atrol:
We should still keep the warning message if there are unresolved child issues, so function relationship_can_resolve_bug is not the right place to implement this option.

bug_update.php
```diff
@@ -169,7 +169,13 @@
 if( ( $t_resolve_issue || $t_close_issue ) &&
 	!relationship_can_resolve_bug( $f_bug_id )
 ) {
-	trigger_error( ERROR_BUG_RESOLVE_DEPENDANTS_BLOCKING, ERROR );
+
+	# this may be overridden with configuration to allow parents to close regardless of
+	# the status of any children
+	if ( OFF === config_get( 'allow_parent_of_unresolved_to_close' ) ) {
```
atrol: For the sake of consistency, we use loose comparison `OFF == config_get( ...` for such kind of comparison (at least 99% of the code)

atrol:
Please squash your commits and rebase to latest master.
Reword commit message to `Add option to allow closing of issues with unresolved children`
Add `Fixes #21225` to the commit description.

vboctor:
Merged via af213f1b58892657587900b7d58292376ba62bd7 with some tweaks.

PR #1139, opened 2017-08-04
AJAX: Send HTTP 500 error for invalid entry point

dregad:
Sending an HTTP error code back to the caller instead of simply some
outputting arbitrary text message ensures that the client call actually
fails, without relying on failure being caused by mismatch in output
type, or detecting the error by parsing the output.

Fixes [#23184](https://mantisbt.org/bugs/view.php?id=23184)

### DIFF
--- a/xmlhttprequest.php
+++ b/xmlhttprequest.php
@@ -43,6 +43,12 @@ if( function_exists( $t_function ) ) {
 	log_event( LOG_AJAX, 'DEPRECATED: Calling {' . $t_function . '}. Use REST API instead.' );
 	call_user_func( $t_function );
 } else {
-	log_event( LOG_AJAX, 'Unknown function for entry point = ' . $t_function );
-	echo 'unknown entry point';
+	$t_msg = 'Unknown function for entry point: ' . $t_function;
+	log_event( LOG_AJAX, $t_msg );
+	header(
+		'HTTP/1.1 ' . HTTP_STATUS_BAD_REQUEST . ' ' . $t_msg,
+		false,
+		HTTP_STATUS_BAD_REQUEST
+	);
+	echo string_attribute( $t_msg );
 }


rombert:
I would suggest 400 Bad Request for this ( see https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.1 ).

It does not look like the server is at fault, it's the client not using the correct location. We could also go with 404 Not Found if you have such a preference, but IMO 400 is better.

dregad:
Thanks for the feedback. 

I picked 500 because from my perspective it was the *server* not being able to serve the requested function (and as you may have noticed I used a custom error message string which provides the client further information and clarifies what the problem is).

Your argument for using a 4xx error makes sense though. I initially rejected 400 because the spec mentions _syntax_.

Definitely not 404 IMO, because it's a valid URL and the server is able to respond (even with an error condition).

xmlhttprequest.php
```diff
@@ -43,6 +43,8 @@
 	log_event( LOG_AJAX, 'DEPRECATED: Calling {' . $t_function . '}. Use REST API instead.' );
 	call_user_func( $t_function );
 } else {
-	log_event( LOG_AJAX, 'Unknown function for entry point = ' . $t_function );
-	echo 'unknown entry point';
+	$t_msg = 'Unknown function for entry point: ' . $t_function;
+	log_event( LOG_AJAX, $t_msg );
+	header( 'HTTP/1.1 500 ' . $t_msg, false, 500 );
+	echo $t_msg;
```
atrol: XSS possible?

xmlhttprequest.php
```diff
@@ -43,6 +43,8 @@
 	log_event( LOG_AJAX, 'DEPRECATED: Calling {' . $t_function . '}. Use REST API instead.' );
 	call_user_func( $t_function );
 } else {
-	log_event( LOG_AJAX, 'Unknown function for entry point = ' . $t_function );
-	echo 'unknown entry point';
+	$t_msg = 'Unknown function for entry point: ' . $t_function;
+	log_event( LOG_AJAX, $t_msg );
+	header( 'HTTP/1.1 500 ' . $t_msg, false, 500 );
```
vboctor: Don't hard code the HTTP error code.  Use the constants defined and used by REST API.

xmlhttprequest.php
dregad: I don't think XSS applies to HTTP headers (no HTML or Javascript executed there), but I'll check if header injection is possible (although offhand I don't think PHP allows this).

xmlhttprequest.php
dregad: right, forgot about those... Thanks

dregad:
@vboctor @atrol any thoughts on @rombert's suggestion to switch error code from 500 to 400 ?

xmlhttprequest.php
atrol: I though that `echo $t_msg;` could be used for XSS, not the header.

dregad: If I try to inject a CR/LF in the header, PHP returns _Warning: Header may not contain more than a single header, new line detected_ so I think it's safe. Confirmed by [PHP documentation](http://php.net/manual/en/function.header.php#refsect1-function.header-changelog).

dregad: > I though that echo $t_msg; could be used for XSS, not the header.

You're right. Fixed.

atrol:
`500 Internal Server Error` is defined as `unexpected condition`
Most of the time when I saw this, it was caused by something like a PHP crash.
I think the current case is not unexpected.

What about `501 Not Implemented` ?

rombert:
According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.5.2 , 501 means that 

>  The server does not support the functionality required to fulfill the request. This is the appropriate response when the server does not recognize the request method and is not capable of supporting it for any resource. 

One example is using an HTTP verb which is not supported by the endpoint, e.g. `PUT` when only `GET` and `POST` are supported.

All the 5xx status codes signal an error caused by the server server, which I believe is not our scenario. We are discussing an error which is due to the client, which is signalled with 4xx status codes.

atrol:
@rombert, you are right, it's the client who caused the error.

So we have to choose between 
- `400 Bad Request` where `due to malformed syntax` is questionable and
- `404 Not Found` where  `No indication is given of whether the condition is temporary or permanent` and `This status code is commonly used when the server does not wish to reveal exactly why the request has been refused` is questionable.

I tend to  `400 Bad Request`

rombert:
> I tend to 400 Bad Request

That would be my choice as well, yes

dregad:
Thanks for your feedback, will update the PR accordingly.

rombert:
LGTM, thank you.

PR #1140, opened 2017-08-05
#19302: Free text search should match against tags, custom-fields,...

hloehnert:
http://www.mantisbt.org/bugs/view.php?id=19302

Text search extended for tags, custom-fields, category.

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -2284,8 +2284,14 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
 				' OR ' . db_helper_like( '{bug_text}.description' ) .
 				' OR ' . db_helper_like( '{bug_text}.steps_to_reproduce' ) .
 				' OR ' . db_helper_like( '{bug_text}.additional_information' ) .
-				' OR ' . db_helper_like( '{bugnote_text}.note' );
+				' OR ' . db_helper_like( '{bugnote_text}.note' ) .
+				' OR ' . db_helper_like( '{tag}.name' ) .
+				' OR ' . db_helper_like( '{custom_field_string}.value' ) .
+				' OR ' . db_helper_like( '{category}.name' );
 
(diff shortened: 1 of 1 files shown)


hloehnert:
My database contains following entries:
bug: 67527
bugnote: 23533
bug_tag: 1832
custom_field_string: 63115

The view-issue-page-execution-time is ~0.15 sec with activated text-search and 4 hits.
So I can't see a measurable impact caused by the additional joins.

atrol:
> bug: 67527

Did you try with `Hide Status = None` ?

hloehnert:
- View all issues without any filters (hide status = none) on all projects returns me a list of 67482 entries. The difference of 45 entries to the 67527 entries in bug-table is due to a disabled project. Page execution time is ~0,5 sec.

- View all issues with an additional text-filter (hide status is still none) on all projects returns me a list of 163 entries with a page execution time of ~0,45 sec.

atrol:
> View all issues with an additional text-filter (hide status is still none) on all projects returns me a list of 163 entries with a page execution time of ~0,45 sec.


Strange result, as I got what I expected on one of my test systems.
It's been about 6 times slower after the changes when searching by text.

Independant from that, there is a serious conceptual issue.
After the changes, you are able to search based on data where you are not allowed to view the data.

E.g. set `Read Access` of a custom field to access level `manager`.
Log in as a developer and search by text.
You will see issues in the list where you are not allowed to see the custom field.

hloehnert:
@atrol you convinced me.
Nevertheless I will keep it on my master branch, because the shown implementation is sufficient for us.

atrol:
> Nevertheless I will keep it on my master branch, because the shown implementation is sufficient for us.

Nice example for one of the advantages of Open Source.
Maybe it's also sufficient for other users,

elmarculino:
Yes, same thing here. Thanks.

PR #1142, opened 2017-08-05
Add ticketcounter behind project title in project dropdown 

Elompenta:
Add configuration to display count of opened tickets to the project List

```
# Add Ticketcount behind project title in Project Dropdown
# $g_enable_ticketcounter_projectdropdown = ON; 
# $g_ticketcounter_projectdropdown_status = 90;
```


vboctor:
Thanks @Elompenta for your contribution.  I don't see the value of this feature justifing the extra db queries.

Elompenta:
@vboctor
Thanks for your Feedback

We use this improvement since many years, that supporters/programmers can find - with one view - customers/projects with open tickets.

In a standalone setup with one single Project this Feature make no sense - that is correct. ;)
But if you have more than 30 or 40 Projects (like our installation) it is a good worktime improvement and every supporter/programmer downt will miss this Feature.

If you dont like this Feature it is okay, but i think it is a good choice to ship the Feature and admins are able to activate it when needed.

best regards,
Sebastian

cproensa:
Some more thoughts:
- This feature, and similar others, are more suited to be "user preferences", rather than configuration options (enabled by admins). I don't see much value in this unless its is an easy configurable setting by the user itself.
- In the specific case of this feature: you have proposed showing "open" issues, but some scenarios may benefit from other information: count of "new" issues, "assigned to me", etc. So again, this would be ideally configurable by each user.

Regarding DB queries: we could add the counting in the projects cache, so query count would be minimal

More thoughts:
Maybe, instead of showing this information in the project selection dropdown, we could include it in the "project" main page (which is currently unused).
This project page would include information of current project issues is a similar way as proposed in this PR.
Viewing a single project would be no better that "my view", but if there are subprojects (or viewing from "all projects"), the information would be shown for each project, which is the idea of this PR, and a capability not present in "my view".

dregad:
@Elompenta are you planning to address the review comments ?

Elompenta:
@dregad 
Thanks for the follow up

I have shared the idea and the needed Code.
For a final Setup with Mantis conform Code i have - at this moment - not enough time.
Sorry

If anybody other like this Feature i think he would copy the Code, or replace the files with correct naming etc. like your post.

PR #1144, opened 2017-08-08
#10853 sorting

hloehnert:
http://www.mantisbt.org/bugs/view.php?id=10853

adopted changes from @dregad for actual master-branch
see https://github.com/dregad/mantisbt/tree/fix-10853-sort-category

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -1070,10 +1070,10 @@ function filter_get_query_sort_data( array &$p_filter, $p_show_sticky, array $p_
 	# until for loop below
 	if( !isset( $s_fk_map ) ) {
 		$s_fk_map = array(
-				'category_id' => array('table' => db_get_table( 'category' ), 'sort_field' => 'name' ),
-				'project_id'  => array('table' => db_get_table( 'project' ), 'sort_field' => 'name' ),
-				'handler_id'  => array('table' => db_get_table( 'user' ), ),
-				'reporter_id' => array('table' => db_get_table( 'user' ), ),
+				'category_id' => array( 'table' => '{category}', 'sort_field' => 'name' ),
+				'project_id'  => array( 'table' => '{project}' , 'sort_field' => 'name' ),
+				'handler_id'  => array( 'table' => '{user}' )  ,
+				'reporter_id' => array( 'table' => '{user}' )  ,
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


core/filter_api.php
```diff
@@ -1105,6 +1118,45 @@ function filter_get_query_sort_data( array &$p_filter, $p_show_sticky, array $p_
...
+			$t_fk_table_alias = $s_fk_map[$c_sort]['table'] . '_' . $c_sort;
+			
+			# Defining sort field for user fields
+			# The if statement below is a hack required to have (at least semi-)
+			# proper sorting of username fields (see #10853); the code somewhat
+			# replicates the logic from function user_get_name()
+			# use of CASE, COALESCE should be fine for all db (SQL92 standard)
+			# however string concatenation operator '||' has a different meaning
+			# in MySQL, so we use CONCAT function which exists in all DBs
+			if( !array_key_exists( 'sort_field', $s_fk_map[$c_sort] ) ) {
```
cproensa: it's assuming that a non existant sort_field in the array is always a user table. Maybe we need to extend this in the future, so checking for the actual sort criteria is good
.

core/filter_api.php
```diff
@@ -1063,6 +1065,17 @@ function filter_get_query_sort_data( array &$p_filter, $p_show_sticky, array $p_
...
 		$p_query_clauses['order'][] = '{bug}.sticky DESC';
 	}
+	
+	# For user fields, the complete definition of the sort_field is deferred
+	# until for loop below
+	if( !isset( $s_fk_map ) ) {
+		$s_fk_map = array(
+				'category_id' => array( 'table' => '{category}', 'sort_field' => 'name' ),
+				'project_id'  => array( 'table' => '{project}' , 'sort_field' => 'name' ),
+				'handler_id'  => array( 'table' => '{user}' )  ,
```
cproensa: The below code could be integrated in this array. Even if it's static, the configuration options won't change within the current script execution.
Probably a 3rd key is needed, as "select_field", which will be included in the select part. Usually will be the same as the sort_field, but in the user table case, it will contain the alias for the computed column.

cproensa:
More issues:
- 'sort_by_last_name' configuration can't be implemented sanely with sql
- Other fields (not included here) like resolution, reproducibility, projection... are enums and can be localized too

PR #1145, opened 2017-08-09
Fix the stopwatch from time_tracking feature.

EthraZa:
Before, it was lagging behind the real time and was not possible to manually edit the time field.

Fix [#22919](http://www.mantisbt.org/bugs/view.php?id=22919)

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -172,40 +172,43 @@ $(document).ready( function() {
 		startTime: null,
 		zeroTime: moment('0', 's'),
 		tick: function() {
-			$('input[type=text].stopwatch_time').val(this.zeroTime.clone().add(moment().diff(this.startTime)).format('HH:mm:ss'));
+			var elapsedDiff = moment().diff(this.startTime),
+				elapsedTime = this.zeroTime.clone().add(elapsedDiff);
+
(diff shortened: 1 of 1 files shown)


js/common.js
```diff
@@ -168,60 +168,68 @@ $(document).ready( function() {
...
-			}
-			if (minutes < 10) {
-				minutes = '0' + minutes;
-			}
-			if (hours < 10) {
-				hours = '0' + hours;
-			}
-			$('input[type=text].stopwatch_time').val(hours + ':' + minutes + ':' + seconds);
-			this.start();
+			$('input[type=text].stopwatch_time').val(this.zeroTime.clone().add(moment().diff(this.startTime)).format('HH:mm:ss'));
```
dregad: Please break the long line

EthraZa:
Requested changes have been made.
About the "self = this", it is neede inside the function, to access the outer scope. In JS functions have it's own scopes. But I have changed to use the self var only when it's really needed.

dregad:
Cheers. I'll test the PR again later tonight.

> About the "self = this", it is neede inside the function, to access the outer scope

Thanks for the explanation.

dregad:
@EthraZa Thanks for your contribution

PR #1147, opened 2017-08-09
Optimze function layout_print_sidebar

atrol:


dregad:
You have a typo in commit message (`Optimze` instead of _optimize_)

Suggest reword `Optimize and remove unused code in function layout_print_sidebar`

PR #1149, opened 2017-08-10
Fixes #19368 - If a float custom field is empty, display it as an emp…

bgamrat:
…ty string, rather than zero.  Courtesy of Game Creek Video.

### DIFF
--- a/core/cfdefs/cfdef_standard.php
+++ b/core/cfdefs/cfdef_standard.php
@@ -210,7 +210,7 @@ function cfdef_print_textarea( $p_value ) {
  * @param string $p_value The custom field value.
  */
 function cfdef_print_numeric( $p_value ) {
-	echo (int)$p_value;
+	echo is_numeric( $p_value ) ? (int)$p_value : '';
 }
 
 /**
@@ -218,7 +218,7 @@ function cfdef_print_numeric( $p_value ) {
  * @param string $p_value The custom field value.
  */
 function cfdef_print_float( $p_value ) {
-	echo (float)$p_value;
+	echo is_numeric( $p_value ) ? (float)$p_value : '';
 }
 
 /**


core/cfdefs/cfdef_standard.php
```diff
@@ -218,7 +218,10 @@ function cfdef_print_numeric( $p_value ) {
  * @param string $p_value The custom field value.
  */
 function cfdef_print_float( $p_value ) {
-	echo (float)$p_value;
+	if (is_numeric($p_value)) {
```
vboctor: I would change this to:
```
echo is_numeric( $p_value ) ? (float)$p_value : '';
```

vboctor:
Please also move the "Fixed #...." to the body of the comment rather than the title and keep the comment first line / title for what is being fixed.

```
Show empty string for empty float custom fields

Fixes #19368
```

atrol:
> Fixes #19368

#19368 is : `custom field (type: float) default value not work.` and has been fixed in version 1.3.0-beta.2.

Please enter a new issue for it.

core/cfdefs/cfdef_standard.php
```diff
@@ -218,7 +218,10 @@ function cfdef_print_numeric( $p_value ) {
...
+		echo (float)$p_value;
+	}
+	echo '';
```
atrol: Do we need to echo an empty string?

core/cfdefs/cfdef_standard.php
bgamrat: Changed as recommended

cproensa:
As far as i know, the same happens for "numeric" field type

vboctor:
Opened appropriate issues and merged via d24fa5e2c52850ea67935b37d53ec2a453b4596d and 84c9dc08ec43ebde3bbcd63528a5bcc32548e780.

Thanks @bgamrat for your contribution.

PR #1150, opened 2017-08-10
EVENT_LAYOUT_NAVBAR Customization of navbar

cproensa:
Allow a more detailed customization of navbar header content.
As previous methods like EVENT_LAYOUT_PAGE_HEADER event have disappeared  from version 2, and their implementation conflicts with current layout  with an always-visible navbar, seems logical to have a new event  EVENT_LAYOUT_NAVBAR that is used to insert content in the navbar header  area.
This allows for custom texts, company logo, etc

https://www.mantisbt.org/bugs/view.php?id=23210


core/events_inc.php
```diff
@@ -46,6 +46,7 @@
 	'EVENT_LAYOUT_CONTENT_END' => EVENT_TYPE_OUTPUT,
 	'EVENT_LAYOUT_PAGE_FOOTER' => EVENT_TYPE_OUTPUT,
 	'EVENT_LAYOUT_BODY_END' => EVENT_TYPE_OUTPUT,
+	'EVENT_LAYOUT_NAVBAR' => EVENT_TYPE_OUTPUT,
```
syncguru: This event name gives the impression that you can customize any part of the navbar, but the actual implementation seems to focus on the logo part - Is EVENT_LAYOUT_NAVBAR_LOGO a more appropriate name?
The idea here to allow developers to replace the default logo segment with any other custom menu.

core/layout_api.php
```diff
@@ -404,50 +404,63 @@ function layout_login_page_end() {
...
+	?>
+	<div id="navbar" class="navbar navbar-default navbar-collapse navbar-fixed-top noprint">
+		<div id="navbar-container" class="navbar-container">
+
+			<button id="menu-toggler" type="button" class="navbar-toggle menu-toggler pull-left hidden-lg" data-target="#sidebar">
+				<span class="sr-only">Toggle sidebar</span>
+				<span class="icon-bar"></span>
+				<span class="icon-bar"></span>
+				<span class="icon-bar"></span>
+			</button>
```
syncguru: Couple of test scenarios that I can think of:
1- Mobile screens
2- Right to left languages

cproensa:
> This event name gives the impression that you can customize any part of the navbar, but the actual implementation seems to focus on the logo part - Is EVENT_LAYOUT_NAVBAR_LOGO a more appropriate name?

Yes, that name is more suited for this event

There are open issues with the old header/bottom customizations. In my opinion, we should solve those as soon as possible, either reintroducing the lost functionality, or creating new tools to fill those scenarios.
If we are lacking that functionality, this at least introduces a minimal customization that goes in that way.

Some thought on this:
- should the event be of type event-first? like, do we want several plugins potentially writting in this area, or just one?
- is using a plugin event for this overkill?

> Documentation for this event carries more weight than ever since it is easy for poorly written customization to break the UI.

Yes. The usage of this area must be done with care. Here there is not any outer layout added for the inserted content, so keeping the layout is responsibility of the plugin (image size, content size, breaking divs, etc). Probably, using the bootstrap css classes to selectively render different content based on screen sizes is the right apporach. However, that would need to be provided as an example in documentation.

Other option, instead of handling raw content, is to manage a dedicated set of elements, eg: navbar_logo, navbar_title... that are rendered by mantis n a controlled way.

I'm concerned we still don't have a standard way to make this kind of customization, so this proposal, or any other, is greatly needed at this point of v2.

cproensa:
> Other option, instead of handling raw content, is to manage a dedicated set of elements, eg: navbar_logo, navbar_title... that are rendered by mantis n a controlled way.

So maybe it's as simple as:
- create a new option $g_navbar_logo_image
- if filled, put the logo image next to current "window_title"

Why not the existing $g_logo_image? because that is used for the login page, and the navbar logo probably would need to be smaller, to properly fit in the current design.
A warning can be made for this configuration, saying that the image must be of appropriate size to not distort navbar height.

Additionally. This logo image could be hidden by css for mobile layout?

atrol:
> is using a plugin event for this overkill?

I think so.

> create a new option $g_navbar_logo_image

👍 seems this is what most users are asking for. Installing an additional plugin for it is a bit too much.

cproensa:
>>   create a new option $g_navbar_logo_image

> :+1: seems this is what most users are asking for. Installing an additional plugin for it is a bit too much.

Let's see #1166 then

cproensa:
closed in favour of #1166

PR #1151, opened 2017-08-11
Check if admin dir has been disabled via .htaccess

dregad:
Verifies if the directory is accessible by retrieving the HTTP headers
via a HEAD request and checking the HTTP status code instead of checking
if it can be read at filesystem level.

Fixes [#23211](http://www.mantisbt.org/bugs/view.php?id=23211)

### DIFF
--- a/login_page.php
+++ b/login_page.php
@@ -141,11 +141,22 @@ $t_warnings = array();
 $t_upgrade_required = false;
 
 if( config_get_global( 'admin_checks' ) == ON ) {
-	# Check if the admin directory is accessible
-	$t_admin_dir = dirname( __FILE__ ) . '/admin';
-	$t_admin_dir_is_accessible = @file_exists( $t_admin_dir . '/.' );
+	# Check if admin directory is accessible as URL by retrieving HTTP headers.
+	# Note: using file_get_contents is faster than get_headers() which does a
+	# GET by default, and avoids changing the default stream context.
+	$t_admin_url = config_get_global('path') . 'admin/';
+	$context = stream_context_create( array(
+		'http' => array( 'method' => 'HEAD', 'follow_location' => 0 )
+	) );
+	@file_get_contents($t_admin_url, NULL, $context );
+	$t_admin_dir_is_accessible = strpos( $http_response_header[0], '200' ) !== false;
+
 	if( $t_admin_dir_is_accessible ) {
 		$t_warnings[] = lang_get( 'warning_admin_directory_present' );
+	} else {
+		# Check if admin directory is accessible at filesystem level
+		$t_admin_dir = dirname( __FILE__ ) . '/admin';
+		$t_admin_dir_is_accessible = @file_exists( $t_admin_dir . '/.' );
 	}
 
 	# Generate a warning if default user administrator/root is valid.


login_page.php
```diff
@@ -141,11 +141,22 @@
 $t_upgrade_required = false;
 
 if( config_get_global( 'admin_checks' ) == ON ) {
-	# Check if the admin directory is accessible
-	$t_admin_dir = dirname( __FILE__ ) . '/admin';
-	$t_admin_dir_is_accessible = @file_exists( $t_admin_dir . '/.' );
+	# Check if admin directory is accessible as URL by retrieving HTTP headers.
+	# Note: using file_get_contents is faster than get_headers() which does a
+	# GET by default, and avoids changing the default stream context.
+	$t_admin_url = config_get_global('path') . 'admin/';
```
vboctor: fix spacing within ().

login_page.php
```diff
@@ -141,11 +141,22 @@
...
+	$context = stream_context_create( array(
+		'http' => array( 'method' => 'HEAD', 'follow_location' => 0 )
+	) );
+	@file_get_contents($t_admin_url, NULL, $context );
```
vboctor: Fix spacing within ().

login_page.php
```diff
@@ -141,11 +141,22 @@
...
+	$t_admin_dir_is_accessible = strpos( $http_response_header[0], '200' ) !== false;
+
 	if( $t_admin_dir_is_accessible ) {
 		$t_warnings[] = lang_get( 'warning_admin_directory_present' );
+	} else {
+		# Check if admin directory is accessible at filesystem level
```
vboctor: What's the benefit of checking if it exists on the file system level?

dregad: This allows additional checks (requiring to include schema.php) to proceed (see further down, https://github.com/dregad/mantisbt/blob/b4589654f2e953db672d058b19e31eb03fc3a926/login_page.php#L208)

dregad:
> consider using GuzzleHttp for checking if the admin folder is accessible

I was actually considering that as an alternative, since a friend told me that using file_get_contents was not a good idea because many hosting providers actually restrict this function's usage.

hloehnert:
I'm using SSL with a self-made certificate. So I have to add 
`'verify' => false`
to the GuzzleHttp-client-options.

login_page.php
```diff
@@ -141,10 +141,29 @@
...
+	# Check if admin directory is accessible at filesystem level
 	$t_admin_dir = dirname( __FILE__ ) . '/admin';
 	$t_admin_dir_is_accessible = @file_exists( $t_admin_dir . '/.' );
-	if( $t_admin_dir_is_accessible ) {
+
+	# Check if admin directory is accessible as URL by retrieving HTTP headers.
+	try {
+		$t_guzzle = new GuzzleHttp\Client( array(
+			'base_uri' => config_get_global( 'path' ),
+			'timeout' => 1,
```
vboctor: Is there a reason for not using default timeout?

dregad: Default timeout is 0 (no timeout). I was thinking about avoiding unnecessary delay for what is basically a non-essential function. But considering it's basically Mantis accessing itself, on second thoughts it is probably not necessary.

dregad:
> the admin/ path not being accessible doesn't mean all scripts on the admin folder is not accessible (that would depend on htaccess rules)

That's true (although it would be somewhat silly on the admin's part to do something like that); do you think it would be better to check for an individual file instead of (or in addition to) the directory , let's say install.php ?

atrol:
I am not sure if using Guzzle for this check is a good idea at the moment.
Guzzle requires PHP >= 5.5 whereas we still have set `define( 'PHP_MIN_VERSION', '5.3.2' );`
Our Admin Guide tells that we require 5.5.x as the minimum version, but we don't enforce it during installation, upgrade or runtime.

> do you think it would be better to check for an individual file instead of (or in addition to) the directory

The additional check is certainly better in terms of security, but I am still asking myself why we should offer it.

In decreasing order in terms of security:
1. remove the folder
2. remove access permissions on operating system level
3. restrict the access by .htaccess

dregad:
> Guzzle requires PHP >= 5.5 whereas we still have set define( 'PHP_MIN_VERSION', '5.3.2' );

This is actually no longer the case since 2.7.0 (PR #1187 commit 380fde11d50c22ef94210822d60b5f576c9d4a34)

Any objections against merging this ?

atrol:
@dregad did you notice https://github.com/mantisbt/mantisbt/pull/1151#issuecomment-322240612 and [this note](https://mantisbt.org/bugs/view.php?id=23211#c57463)?

PR #1154, opened 2017-08-14
Make tag view threshold work at project level

cproensa:
The part about filtering requires more work, as the query clauses needed to be rewritten to account for proper visibility.

commit f0a92c0ae58574aae44ca7d971d48809bf82ea31

    Check bug permissions to show tags
    
    Check for bug access level, instead of global, to show tags in bug view
    page.
    
    Fixes: #23216

commit e5743f4532b1dd5dd7fc6e4c08e6349b6da96421

    Check project level to show tags in filter
    
    Check project level, instead of global level, to show tags in the filter
    form.
    
    Fixes: #23216

commit 1e6fe3929a9a5c641ef4c1f522f5108fc189b4cd

    Refactor tags filter query
    
    Modify the query clauses to use JOINs.

commit 11e5c025ba0e03b4d65190e305d7c37f37ff21e0

    Refactor calculate included projects for filter
    
    Move the logic to calculate all the individual projects included in the
    filter search, into a separate function, to allow for reuse.

commit f3210ccd4263770ee159d20a6ea871ef9736219c

    Refactor access_has_any_project_level()
    
    Refactor access_has_any_project_level() with a more detailed access
    check function, that can return an array of projects that match the
    requested access.

commit 3b6deb2ef27c5e2df4c75d8421ba6e3cddb16897

    Use tag view permission in filter query
    
    Account for permissions in each project in the filter scope,
    to match tags only on those with proper access to view tags.
    
    Fixes: #23216
commit f9cfce73fede39f52949a3dc22e3f58743176443

    Fix query for related tags
    
    Refactor the related tags function to use a filter search for the tag,
    leaving the rd work for access and visibility checks to the filter api.
    
    Previous query was not correct as it was comparing project access level
    with bug view state. Also, it didn't account for view tags permissions
    for each project.
    
    Fixes: #21913

commit 95f0f381becb7fced6a980f1d6dbfc21615d656c

    Fix duplicated caluses for tags filter query
    
    Fix duplicating the filter query clauses when the filter field "tag
    select" is provided.

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -2044,50 +2044,43 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
 					foreach( $t_tags_all as $t_tag_row ) {
 						$t_tag_alias = 'bug_tag_alias_' . ++$t_tag_counter;
 						array_push( $t_join_clauses,
-								'JOIN {bug_tag} ' . $t_tag_alias . ' ON ' . $t_tag_alias . '.bug_id = {bug}.id'
-								. ' AND ' . $t_tag_alias . '.tag_id=' . (int)$t_tag_row['id']
-								. $t_tag_projects_clause
+							'JOIN {bug_tag} ' . $t_tag_alias . ' ON ' . $t_tag_alias . '.bug_id = {bug}.id'
+							. ' AND ' . $t_tag_alias . '.tag_id=' . (int)$t_tag_row['id']
+							. $t_tag_projects_clause
 						);
 					}
 				}
 
 				if( count( $t_tags_any ) ) {
-					$t_check_not_null = array();
+					$t_tag_alias = 'bug_tag_alias_' . ++$t_tag_counter;
+					$t_tag_ids = array();
 					foreach( $t_tags_any as $t_tag_row ) {
-						$t_tag_alias = 'bug_tag_alias_' . ++$t_tag_counter;
-						array_push( $t_join_clauses,
-								'LEFT OUTER JOIN {bug_tag} ' . $t_tag_alias . ' ON ' . $t_tag_alias . '.bug_id = {bug}.id'
-								. ' AND ' . $t_tag_alias . '.tag_id=' . (int)$t_tag_row['id']
-								. $t_tag_projects_clause
-						);
-						$t_check_not_null[] = $t_tag_alias . '.tag_id';
+						$t_tag_ids[] = (int)$t_tag_row['id'];
 					}
-					# If the isn't a non-outer join, check that at least one of the tags has been matched by the outer joins
+					array_push( $t_join_clauses,
+						'LEFT OUTER JOIN {bug_tag} ' . $t_tag_alias . ' ON ' . $t_tag_alias . '.bug_id = {bug}.id'
+						. ' AND ' . $t_tag_alias . '.tag_id IN (' . implode( ',', $t_tag_ids ) . ')'
+						. $t_tag_projects_clause
+					);
+
+					# If the isn't a non-outer join, check that at least one of the tags has been matched by the outer join
 					if( !count( $t_tags_all ) ) {
-						if( count( $t_check_not_null ) > 1 ) {
-							array_push( $t_where_clauses, 'COALESCE(' . implode( ',', $t_check_not_null ) . ') IS NOT NULL' );
-						} else {
-							array_push( $t_where_clauses, $t_check_not_null[0] . ' IS NOT NULL' );
-						}
+						array_push( $t_where_clauses, $t_tag_alias . '.tag_id IS NOT NULL' );
 					}
 				}
 
 				if( count( $t_tags_none ) ) {
-					$t_check_null = array();
+					$t_tag_alias = 'bug_tag_alias_' . ++$t_tag_counter;
+					$t_tag_ids = array();
 					foreach( $t_tags_none as $t_tag_row ) {
-						$t_tag_alias = 'bug_tag_alias_' . ++$t_tag_counter;
-						array_push( $t_join_clauses,
-								'LEFT OUTER JOIN {bug_tag} ' . $t_tag_alias . ' ON ' . $t_tag_alias . '.bug_id = {bug}.id'
-								. ' AND ' . $t_tag_alias . '.tag_id=' . (int)$t_tag_row['id']
-								. $t_tag_projects_clause
-						);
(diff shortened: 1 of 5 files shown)


core/filter_api.php
```diff
@@ -3508,3 +3473,94 @@ function filter_print_view_type_toggle( $p_url, $p_view_type ) {
...
+ * @return array
+ */
+function filter_get_included_projects( array $p_filter, $p_project_id = null, $p_user_id = null ) {
+	if( null === $p_project_id ) {
+		$t_project_id = helper_get_current_project();
+	} else {
+		$t_project_id = $p_project_id;
+	}
+	if( $p_user_id === null || $p_user_id === 0 ) {
+		$t_user_id = uth_get_current_user_id();
```
atrol: typo auth_get_current_user_id
I would remove the null handling for $p_project_id and $p_user_id as the function is called just a single time where $p_project_id and $p_user_id are set.

cproensa:
@atrol 
> I would remove the null handling for $p_project_id and $p_user_id as the function is called just a single time where $p_project_id and $p_user_id are set.

That function is used once within this PR, but it will be used in many other places, for future works. There are some filter issues in my to-do list that needs that logic.

core/filter_api.php
```diff
@@ -3508,3 +3473,94 @@ function filter_print_view_type_toggle( $p_url, $p_view_type ) {
...
+ * @param integer $p_user_id      User id to use as current user when filtering
+ * @return array
+ */
+function filter_get_included_projects( array $p_filter, $p_project_id = null, $p_user_id = null ) {
+	if( null === $p_project_id ) {
+		$t_project_id = helper_get_current_project();
+	} else {
+		$t_project_id = $p_project_id;
+	}
+	if( $p_user_id === null || $p_user_id === 0 ) {
```
dregad: Might as well do a single loose-type comparison instead of 2 strict-type ones to handle both *null* and *0*, i.e. `if( !$p_user_id )`  or `if( $p_user_id != null)` if you prefer, so all situations are covered (e.g. *false*)

core/filter_api.php
```diff
@@ -2059,61 +1996,89 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
...
-				foreach( $t_tags_none as $t_tag_row ) {
-					array_push( $t_clauses, '{bug_tag}.tag_id = ' . $t_tag_row['id'] );
+				$t_tag_counter = 0;
+				if( count( $t_tags_all ) ) {
+					foreach( $t_tags_all as $t_tag_row ) {
+						$t_tag_alias = 'bug_tag_alias_' . ++$t_tag_counter;
+						array_push( $t_join_clauses,
+								'JOIN {bug_tag} ' . $t_tag_alias . ' ON ' . $t_tag_alias . '.bug_id = {bug}.id'
+								. ' AND ' . $t_tag_alias . '.tag_id=' . (int)$t_tag_row['id']
+								. $t_tag_projects_clause
```
dregad: Remove extra indent

core/filter_api.php
```diff
@@ -2059,61 +1996,89 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
...
+				if( count( $t_tags_any ) ) {
+					$t_tag_alias = 'bug_tag_alias_' . ++$t_tag_counter;
+					$t_tag_ids = array();
+					foreach( $t_tags_any as $t_tag_row ) {
+						$t_tag_ids[] = (int)$t_tag_row['id'];
+					}
+					array_push( $t_join_clauses,
+							'LEFT OUTER JOIN {bug_tag} ' . $t_tag_alias . ' ON ' . $t_tag_alias . '.bug_id = {bug}.id'
+							. ' AND ' . $t_tag_alias . '.tag_id IN (' . implode( ',', $t_tag_ids ) . ')'
+							. $t_tag_projects_clause
```
dregad: Remove extra indent

core/filter_api.php
```diff
@@ -2059,61 +1996,89 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
...
+				if( count( $t_tags_none ) ) {
+					$t_tag_alias = 'bug_tag_alias_' . ++$t_tag_counter;
+					$t_tag_ids = array();
+					foreach( $t_tags_none as $t_tag_row ) {
+						$t_tag_ids[] = (int)$t_tag_row['id'];
+					}
+					array_push( $t_join_clauses,
+							'LEFT OUTER JOIN {bug_tag} ' . $t_tag_alias . ' ON ' . $t_tag_alias . '.bug_id = {bug}.id'
+							. ' AND ' . $t_tag_alias . '.tag_id IN (' . implode( ',', $t_tag_ids ) . ')'
+							. $t_tag_projects_clause
```
dregad: Remove extra indent

core/filter_api.php
cproensa: I copied that code from `filter_get_bug_rows_query_clauses`
I don't think that $p_user_id === 0 will be an expected usage (this is, 0 == no user?)

cproensa:
> That function is used once within this PR, but it will be used in many other places, for future works. There are some filter issues in my to-do list that needs that logic.

And as long as it gets used, it may need to be modified with some different parameters or functionality... I am looking at custom fields issues that needs that logic, and i see possible variations needed.

core/filter_api.php
dregad: > I copied that code from filter_get_bug_rows_query_clauses

So my above comment applies to that as well :wink: 

> I don't think that $p_user_id === 0 will be an expected usage (this is, 0 == no user?)

True, it's not really expected, but it could happen and does not cost anything to make the API more robust by covering such a case.

core/filter_api.php
```diff
@@ -3508,3 +3473,94 @@ function filter_print_view_type_toggle( $p_url, $p_view_type ) {
...
+ * @param integer $p_user_id      User id to use as current user when filtering
+ * @return array
+ */
+function filter_get_included_projects( array $p_filter, $p_project_id = null, $p_user_id = null ) {
+	if( null === $p_project_id ) {
+		$t_project_id = helper_get_current_project();
+	} else {
+		$t_project_id = $p_project_id;
+	}
+	if( null === $p_user_id ) {
```
dregad: As mentioned in https://github.com/mantisbt/mantisbt/pull/1154#discussion_r133370618, should be loose-type `==` not `===` to catch $p_user_id = 0 or false in addition to null.

cproensa: done

PR #1156, opened 2017-08-17
Cache all needed projects for navbar layout

cproensa:
Reduce DB query count by caching all needed projects for the navbar layout.

Fixes: #23237

After:
seleccion_231

Before:
seleccion_230

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -1492,6 +1492,8 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
 		$t_private_and_public_project_ids = array();
 		$t_limited_projects = array();
 
+		# make sure the project rows are cached, as they will be used to check access levels.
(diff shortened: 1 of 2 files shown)


cproensa:
Added a second commit, where view_all_bug_page may read the project rows as part of the filter query, before the layout for navbar is done.

core/filter_api.php
```diff
@@ -1492,6 +1492,7 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
 		$t_private_and_public_project_ids = array();
 		$t_limited_projects = array();
 
+		project_cache_array_rows( $t_project_ids );
```
vboctor: It would be good to add the comments about why this is called, similar to the comment in the commit message.

cproensa: done

cproensa:
@atrol seems like this PR has been merged, but not closed.

PR #1157, opened 2017-08-18
Deprecate project_get_local_user_access_level()

cproensa:
This function is redundant, with access_get_local_level() having the
same functionality.

Fixes: #23242

### DIFF
--- a/core/project_api.php
+++ b/core/project_api.php
@@ -569,26 +569,13 @@ function project_get_name( $p_project_id, $p_trigger_errors = true ) {
  * @param integer $p_project_id A project identifier.
  * @param integer $p_user_id    A user identifier.
  * @return integer
+ * @deprecated     access_get_local_level() should be used in preference to this function
+ *                 This function has been deprecated in version 2.6
  */
 function project_get_local_user_access_level( $p_project_id, $p_user_id ) {
-	$p_project_id = (int)$p_project_id;
-
-	if( ALL_PROJECTS == $p_project_id ) {
-		return false;
-	}
-
-	db_param_push();
-	$t_query = 'SELECT access_level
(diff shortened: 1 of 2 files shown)


core/project_api.php
```diff
@@ -569,26 +569,10 @@ function project_get_name( $p_project_id, $p_trigger_errors = true ) {
...
-				  FROM {project_user_list}
-				  WHERE user_id=' . db_param() . ' AND project_id=' . db_param();
-	$t_result = db_query( $t_query, array( (int)$p_user_id, $p_project_id ) );
-
-	$t_level = db_result( $t_result );
-	if( $t_level ) {
-		return (int)$t_level;
-	} else {
-		return false;
-	}
```
dregad: Is it worth adding 

``` php
error_parameters( __FUNCTION__ . '()', 'access_get_local_level()' );
```

cproensa: done

cproensa:
> I assume you've replaced all occurrences of the deprecated function's usage (only 1 ?).

yes, only one usage

> Is it worth keeping the function as a stub for convenience, instead of deprecating it ?

i think so. Keep the function, as it may still be used by plugins or external code

Should we add "deprecated in version 2.6"? To know in some major release that it could be deleted

dregad:
Maybe I did not explain myself clearly. The question is, 
1. do we deprecate the function, keep it for a while then remove it, or
2. just keep project_get_local_user_access_level() as an alias for access_get_local_level() (no deprecation)

If 2, then https://github.com/mantisbt/mantisbt/pull/1157#discussion_r133952607 makes no sense.

> Should we add "deprecated in version 2.6"? 

Good idea.

PR #1160, opened 2017-08-19
Deprecate access_has_any_project()

cproensa:
This function may mislead into incorrect validations. Usually you want
to check that a user meets a threshold for any project, but that
threshold may be configured differently for each project, and the user
may also have different access levels in each project due to private
projects assignment.
In that scenario, $p_access_level can't be a static threshold, but a
"threshold identifier" instead, that must be evaluated for each project.
Function "access_has_any_project_level()" provides that functionality,
also covers the basic usage of this function.
For such reasons, this function has been deprecated.

Fixes: #21654

### DIFF
--- a/core/access_api.php
+++ b/core/access_api.php
@@ -417,12 +417,27 @@ function access_ensure_project_level( $p_access_level, $p_project_id = null, $p_
 
 /**
  * Check whether the user has the specified access level for any project project
+ *
+ * Warning: this function may mislead into incorrect validations. Usually you want to
+ * check that a user meets a threshold for any project, but that threshold may be configured
+ * differently for each project, and the user may also have different access levels in each
+ * project due to private projects assignment.
+ * In that scenario, $p_access_level can't be a static threshold, but a "threshold identifier"
+ * instead, that must be evaluated for each project.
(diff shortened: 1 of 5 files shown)


cproensa:
> Add deprecated warning in access_has_any_project()

done

vboctor:
Isn't it better to have the API access an access level value?  This has the following advantages:

- Load the config based on the right context (e.g. project, issue, user, etc), rather than hard-coded defaults.
- Ability to retrieve access level once and use it across multiple API calls.

If there a need to use an access level based on a context other than the default, wouldn't the API that accept an access level be useful?

cproensa:
> Isn't it better to have the API access an access level value? 
> If there a need to use an access level based on a context other than the default, wouldn't the API that accept an access level be useful?

The new function accepts an access level with the same behaviour as old function.
The access level parameter can "optionally" be a threshold name, for multiproject evaluation.

> Load the config based on the right context (e.g. project, issue, user, etc), rather than hard-coded defaults.
> Ability to retrieve access level once and use it across multiple API calls.

Thats what happened in previous usages of old function, and many of them were doing it wrong. Note the explanation in the function comments "warning: ...".

PR #1161, opened 2017-08-20
Add logging functionality for plugins

cproensa:
Add logging functionality for plugins
- Create a new loglevel LOG_PLUGIN to write logs specifically from plugins.
- Add function plugin_log_event() to wrap the log call for plugins.

Fixes: #22053

Print class name when logging backtrace caller info
If the caller function is a class method, print the class name alognside the function name in the log info.
Fixes: #23249

### DIFF
--- a/core/plugin_api.php
+++ b/core/plugin_api.php
@@ -35,6 +35,7 @@
  * @uses helper_api.php
  * @uses history_api.php
  * @uses lang_api.php
+ * @uses logging_api.php
  */
 
 require_api( 'access_api.php' );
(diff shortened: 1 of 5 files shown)


core/plugin_api.php
```diff
@@ -997,3 +999,20 @@ function plugin_init( $p_basename ) {
 		return false;
 	}
 }
+
+function plugin_log_event( $p_msg, $p_basename = null ) {
+	$t_current_plugin = plugin_get_current();
+	if( is_null( $p_basename ) ) {
+		$t_basename = plugin_get_current();
```
atrol: Use `$t_current_plugin` instead of `plugin_get_current()`

cproensa:
> I was going write something about refactoring logging to use PSR-3, then I realized I already did in https://mantisbt.org/bugs/view.php?id=22053#c54796

can you do a sketch on how that would be implemented, and open it as a separate feature issue?

luzpaz:
Would this PR have any impact on https://github.com/mantisbt-plugins/source-integration/issues/92 ?

PR #1162, opened 2017-08-20
Move notify user checkbox outside the form

cproensa:
On manage user page, the "notify user" checkbox is not part of the user
data, and is only used to trigger a notification if the form is saved.
Move the item outside the form, next to the update button to better
match the context of the action.

Fixes: #21695

Before:
seleccion_233

After:
seleccion_232


atrol:
Closed via https://github.com/mantisbt/mantisbt/commit/fa9d727fc4426da8bd8356349d3aa12155e56519

PR #1163, opened 2017-08-21
Timeline preserve parent script parameters

cproensa:
Correctly preserve parent script parameters for navigation.

Fixes: #23251

### DIFF
--- a/core/timeline_inc.php
+++ b/core/timeline_inc.php
@@ -47,6 +47,18 @@ $t_block_css = $t_collapse_block ? 'collapsed' : '';
 $t_block_icon = $t_collapse_block ? 'fa-chevron-down' : 'fa-chevron-up';
 
 $t_url_page = string_sanitize_url( basename( $_SERVER['SCRIPT_NAME'] ) );
+# Timeline shows shows next/prev buttons that reload the page with new timeline parameters
+# we must preserve parent script query parameters
+$t_url_params = array();
+if( !empty( $_GET ) ) {
+	# Sanitize request values to avoid xss
+	foreach( $_GET as $t_key => $t_value ) {
+		$t_url_params[$t_key] = htmlspecialchars( $t_value );
+	}
+}
+# clear timeline own parameters, which will be added later as needed
+unset( $t_url_params['days'] );
+unset( $t_url_params['all'] );
 ?>
 
 <div id="timeline" class="widget-box widget-color-blue2 <?php echo $t_block_css ?>">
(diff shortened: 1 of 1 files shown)


core/timeline_inc.php
```diff
@@ -47,6 +47,12 @@
 $t_block_icon = $t_collapse_block ? 'fa-chevron-down' : 'fa-chevron-up';
 
 $t_url_page = string_sanitize_url( basename( $_SERVER['SCRIPT_NAME'] ) );
+# Timeline shows buttons to reload the page with new timeline parameters
+# we must preserve parent script query parameters
```
dregad: Maybe more clear to say _shows next/prev buttons that reload_

core/timeline_inc.php
```diff
@@ -47,6 +47,12 @@
...
+$t_url_params = $_GET;
```
dregad: Maybe not a current, real-life scenario, but what if it's a POST request ?

cproensa: I thought the same...
But in that scenario, i think you could argue that the GET request is a viewing action and is safe to be repeated, while POST implies modification. The similar idea on why post usually is not cached.

core/timeline_inc.php
```diff
@@ -74,15 +80,15 @@
 				echo '&#160;&#160;';
 
 				echo '<div class="btn-group">';
-				$t_url_params = array( 'days' => $f_days + 7 );
-				$t_href = $t_url_page . '?' . http_build_query( $t_url_params );
+				$t_url_params['days'] = $f_days + 7;
+				$t_href = $t_url_page . '?' . htmlspecialchars( http_build_query( $t_url_params ) );
```
vboctor: We use `http_build_query` in several places without `htmlspecialchars` or any post processing.  If we need to post process, then we should do it in all locations ideally via a Mantis API that is used from all places.  If not, then it should be removed.

cproensa: It uses `htmlspecialchars` becasue it's printing all GET parameters that were present in the request. Those parameters are unverified content, so i think this avoid xss.
Other usages of `http_build_query`, where the parameters are trusted, don't need any postprocessing.

vboctor: @cproensa Makes sense.  But shouldn't the escaping happens on each untrusted value rather than the output of the `http_build_query`?  This would avoid escaping strings added by `http_build_query` itself like `&`s.

cproensa:
@vboctor 
> shouldn't the escaping happens on each untrusted value ...

done

PR #1164, opened 2017-08-21
Fix timetracking layout on view issue details

cproensa:
The layout for timetracking box was messed with misplaced tags,
resulting in a distorted layout for the following content in the page.

Fixes: #22291, #22469

### DIFF
--- a/bugnote_stats_inc.php
+++ b/bugnote_stats_inc.php
@@ -85,58 +85,58 @@ $t_block_icon = $t_collapse_block ? 'fa-chevron-down' : 'fa-chevron-up';
 # CSRF protection not required here - form does not result in modifications
 ?>
 <div class="col-md-12 col-xs-12 noprint">
-<a id="bugnotestats"></a>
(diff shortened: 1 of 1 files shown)


atrol:
Strange, I don't see that #22291 is fixed after this change, and I don't see that #22469 is still an issue in latest master.
Any idea what I am doing wrong? (Restarted web server, cleared browser cache)

cproensa:
Do you have
`$g_time_tracking_enabled = ON;`

cproensa:
@atrol you were right.
there was a misplaced bracket. It was good as long as the time table was expanded, otherwise the layour was still bad.

atrol:
Closed via https://github.com/mantisbt/mantisbt/commit/f6f845217b7df253d1d056ca27d3cd0cf55f3ad4 https://github.com/mantisbt/mantisbt/commit/a49974f45eba85af4ae7cf83378f93fd02ac368d

PR #1165, opened 2017-08-21
On group action page, show only allowed projects

cproensa:
On group action page, for "move" and "copy" actions, only allow the
selection of destination projects where the action can be performed.
Projects for which the user don't have report permission are now
disabled in the dropdown selection.

Fixes: #12444

### DIFF
--- a/bug_actiongroup_page.php
+++ b/bug_actiongroup_page.php
@@ -294,7 +294,8 @@ if( $t_multiple_projects ) {
 			switch( $f_action ) {
 				case 'COPY':
 				case 'MOVE':
-					print_project_option_list( null, false );
+					print_project_option_list( null /* $p_project_id */, false /* $p_include_all_projects */,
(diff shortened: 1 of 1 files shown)


bug_actiongroup_page.php
```diff
@@ -294,7 +294,7 @@
 			switch( $f_action ) {
 				case 'COPY':
 				case 'MOVE':
-					print_project_option_list( null, false );
+					print_project_option_list( null, false, null, false, true /* $p_can_report_only */ );
```
vboctor: Please add the parameter names next to each parameter for readability.

cproensa: ok, done

atrol:
Closed via https://github.com/mantisbt/mantisbt/commit/141020b4431557ab97f4caa4fb41c8bf383209c6

PR #1166, opened 2017-08-23
Config option for navbar logo image

cproensa:
As an alternative to #1150 

Create option $g_navbar_logo_image to print a logo in the navbar header area. This logo is empty by default.

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -902,10 +902,17 @@ $g_admin_checks = ON;
 $g_favicon_image = 'images/favicon.ico';
 
 /**
- * Logo
+ * Location for logo image
  * @global string $g_logo_image
  */
 $g_logo_image = 'images/mantis_logo.png';
+
+/**
+ * Location for logo image to be displayed in navbar header
+ * This image should be of small enough size to not distort the navbar layout.
+ * Note, navbar has: min-height: 45px
+ */
+$g_navbar_logo_image = '';
 
 /**
  * Logo URL link
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -404,50 +404,54 @@ function layout_login_page_end() {
  * @return void
  */
 function layout_navbar() {
-	$t_logo_url = config_get('logo_url');
+	$t_logo_url = config_get( 'logo_url' );
+	$t_toggle_class = ( OFF == config_get( 'show_avatar' ) ? 'navbar-toggle' : 'navbar-toggle-img' );
+	?>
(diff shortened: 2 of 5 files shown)


cproensa:
@syncguru 
Please have a look at some way to solve this issue. I don't really understand the navbar css that much.

When using different sized images:
This image is 30px. I'd want the image to be vertically centered in the bar space
seleccion_239


This image is 45 px, the exact size of navbar, so no issues here
seleccion_240


This image is 70px. I'd want the navbar space stretch to the height of the image
seleccion_241


This is the 70px image with wide screen. the bar height is extended to the image size
seleccion_242

vboctor:
I'm not sure it is really necessary to have the logo in the navigation bar.  It is likely that the logos won't look right given the variance in logo sizes and the fact that the background color is black.  I think the brand name should be sufficient in the nav bar and have the logo show up in the login screen as per the current implementation.

cproensa:
>  I think the brand name should be sufficient in the nav bar and have the logo show up in the login screen as per the current implementation.

On the contrary, i think it is not sufficient, **for those who need it**

cproensa:
Fixed resizing of navbar when the navbar-header is floated:

seleccion_249

It's pending the vertical align for the elements.

>  vboctor assigned cproensa

Actually, i am taking shots in the dark here, regarding to proper css handling of this.
It would be better if some of you who know how this ace styles work make a proper implementation

cproensa:
Align vertically logo and brand text, when smaller than bar height:

seleccion_250

http://caniuse.com/flexbox
https://www.mantisbt.org/docs/master/en-US/Admin_Guide/html/admin.install.requirements.client.html

sagamusix:
It might not even be necessary to add a new option for this (although it would be more consistent with Mantis 1.x so it would make transition easer).
Since there was no such option available when I switched to 2.x, I simply resolved this by using the already existing mechanism for adding custom CSS:
```
.navbar-brand
{
    background: url(../images/some_logo.png) no-repeat left center;
    padding-left: 36px !important; /* adjust for logo size */
}
```

cproensa:
> I simply resolved this by using the already existing mechanism for adding custom CSS:

Thanks. That's a bit dirty way of doing it, however, it may get the work done...

It has some disadvantages:
- image gets partially hidden if bigger than navbar-header div height. Note that div height is set by the inner text height+top/bottom padding, and may not match actual visible navbar height.
- adjusting for different @media widths need additional styles (my proposal was to hide logo on small screen layouts)

cproensa:
outdated & rejected

PR #1168, opened 2017-08-24
Add the custom fields values to the search results

elmarculino:
I always need to search and filter data from my custom fields. I'm using a version with this change for some time now without problems.

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -2269,6 +2269,10 @@ function filter_get_bug_rows_query_clauses( array $p_filter, $p_project_id = nul
 			$t_search_terms[trim( $t_match[1], "\'\"" )] = ( $t_match[0][0] == '-' );
 		}
 
+		# Subquery to get user access level 
+		$t_access_level_subquery = 'SELECT COALESCE((SELECT access_level FROM {project_user_list} WHERE user_id = ' . $t_user_id .
+		' AND project_id = ' . $t_project_id . '),(SELECT access_level FROM {user} WHERE id = ' . $t_user_id . '))';
+
 		# build a big where-clause and param list for all search terms, including negations
 		$t_first = true;
 		$t_textsearch_where_clause = '( ';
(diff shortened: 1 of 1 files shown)


atrol:
It's not that simple.
See my comment on a similar PR https://github.com/mantisbt/mantisbt/pull/1140#issuecomment-320712837

elmarculino:
> It's not that simple.
See my comment on a similar PR #1140 (comment)

Yes, same thing here, every one must see every issue. Thanks

elmarculino:
> Independant from that, there is a serious conceptual issue.
After the changes, you are able to search based on data where you are not allowed to view the data.

> E.g. set Read Access of a custom field to access level manager.
Log in as a developer and search by text.
You will see issues in the list where you are not allowed to see the custom field.

@atrol , I made some changes in the query to filter out this issues the user is not allowed to see . Is that right or the search must ignore the custom fields the user don´t have the mininum access level?

cproensa:
The access level check is wrong:
- access level for a user can be different on each project
- in your where clause, if the access level compares false, it will cause that issue to not appear in the filter result, which may not be correct.

elmarculino:
@cproensa  You are right. Changed the query to check the access level per custom field.

cproensa:
I am doing access level compliance when filtering custom field values, and it's not simple.
I'm thinking that for search, it would require one JOIN for each custom field, to properly account permissions at query level for each custom field id.

> Still need to look this project access level.

Your query is already several levels of subqueries, and getting the access levels right, is still more complexity.
Additionally to individual user-project membership, there's also "private_project_threshold"
`Threshold needed to be automatically included in private projects`

There are comments on this search topic in http://www.mantisbt.org/bugs/view.php?id=19302

elmarculino:
@cproensa @atrol  I changed the query to filter the custom fields by project/user. I didn't notice performance loss, but maybe my database isn´t large enought.  Can you check?

cproensa:
> I didn't notice performance loss, but maybe my database isn´t large enought. Can you check?

This is a search with current master, over a set of ~50k issues

```
DB-19	2.2630	filter_api.php:1184 filter_get_bug_count()
DB-20	2.2724	filter_api.php:1334 filter_get_bug_rows_result()

Page execution time: 4.8685 seconds
Memory usage: 3,346 KiB
Total queries executed: 80
Unique queries executed: 59
Total query execution time: 4.6009 seconds  
```

And this is the same search using this PR

```
DB-19	62.3848	filter_api.php:1184 filter_get_bug_count()
DB-20	146.3404	filter_api.php:1334 filter_get_bug_rows_result()

Page execution time: 209.0311 seconds
Memory usage: 3,411 KiB
Total queries executed: 80
Unique queries executed: 59
Total query execution time: 208.7688 
```

So, a 4500% increase in time?

elmarculino:
Yeah, not cool. Will continue using the first version. Thanks

PR #1169, opened 2017-08-27
Use CSS for spacing before checkbox/radio labels

dregad:
Use padding-6 class and remove hardcoded whitespace

Fixes [#23243](https://mantisbt.org/bugs/view.php?id=23243)

### DIFF
--- a/account_prefs_inc.php
+++ b/account_prefs_inc.php
@@ -158,11 +158,11 @@ function edit_account_prefs( $p_user_id = null, $p_error_if_protected = true, $p
 		<td>
 			<label for="bugnote-order-desc" class="inline padding-right-8">
 				<input type="radio" class="ace input-sm" id="bugnote-order-desc" name="bugnote_order" value="DESC" <?php check_checked( $t_pref->bugnote_order, 'DESC' ); ?> />
-				<span class="lbl"> <?php echo lang_get( 'bugnote_order_desc' ) ?> </span>
+				<span class="lbl padding-6"><?php echo lang_get( 'bugnote_order_desc' ) ?></span>
 			</label>
 			<label for="bugnote-order-asc" class="inline padding-right-8">
 				<input type="radio" class="ace input-sm" id="bugnote-order-asc" name="bugnote_order" value="ASC" <?php check_checked( $t_pref->bugnote_order, 'ASC' ); ?> />
-				<span class="lbl"> <?php echo lang_get( 'bugnote_order_asc' ) ?> </span>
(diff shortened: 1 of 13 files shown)


dregad:
@syncguru I picked _padding-6_ because it looks pretty much the same as the previous layout with a hardcoded space. I think 2 and 4 would be too small, but I guess that's a debatable matter of preferences. 

The [padding-6 class only adjusts margin-right](https://github.com/mantisbt/mantisbt/blob/master/css/ace.css#L9070) when applied to labels.

I can't use padding-2 and padding-4, because [you added global classes in ace-mantis.css](https://github.com/mantisbt/mantisbt/blame/master/css/ace-mantis.css#L89), not sure what for as they are don't seem to be referenced in the commits adding them, that indeed add a margin all around.

checkbox-no-space No space
checkbox-hardcoded-space Hardcoded space
checkbox-padding-6 padding-6

vboctor:
I'm OK with the new look and prefer it over the original based on snapshots above.  Will defer to @syncguru and @dregad for the approach.

dregad:
> I remain uneasy about padding in all directions. 

I [repeat](https://github.com/mantisbt/mantisbt/pull/1169#issuecomment-325299411), the proposed code does NOT add padding in all directions; the standard ace CSS only adds margin-right in this context, as demonstrated by my screenshot 
```css
input.ace + .lbl.padding-6::before {
    margin-right: 6px;
}
```
There is only padding in all directions for the `-2` and `-4`, because you added a default class doing this (but not for `-6`)

> Can you use .padding-right-4 (defined in ace-mantis.css)

I can certainly do, but I would like to understand why you added the global `.padding-4` class in the first place, considering that it is only ever used in [core/timeline_api.php](https://github.com/mantisbt/mantisbt/blob/master/core/timeline_api.php#L127). Same goes for `.padding-2` (used in bugnote_view_inc.php:158 and 160 and [core/columns_api.php](https://github.com/mantisbt/mantisbt/blob/master/core/columns_api.php#L1161)).

The standard ace behavior does exactly what we're looking for here, so I would like to challenge your suggestion. 

Also, woulnd't using `.padding-right-4` cause issues in RTL context ?

syncguru:
Oh, I see now that Ace overwrite that default padding-6 for lbl classes. That should fix the issue. I would double check the behaviour in RTL mode.

The default padding-6 shipped with Ace is in all directions. I needed the left and right versions so I added them. I also handle the RTL case which maybe another reason why I needed it.

PR #1170, opened 2017-08-27
Fix misplaced "Reset Prefs" button in user prefs with narrow screen

dregad:
Fixes [#23267](https://mantisbt.org/bugs/view.php?id=23267)


syncguru:
Before and after snapshots?

vboctor:
How does the `formaction` [browser minimum requirement](https://www.w3schools.com/tags/att_button_formaction.asp) compared to our browser minimum requirements?

dregad:
> Before and after snapshots?

@syncguru Before screenshot is in original issue; after change, the button is next to the other in the bottom bar

> How does the formaction browser minimum requirement compare to MantisBT browser minimum requirements?

@vboctor We're covered AFAICT
https://mantisbt.org/docs/master/en-US/Admin_Guide/html-desktop/#admin.install.requirements.client

PR #1171, opened 2017-08-28
Fix oracle sql syntax for table alias

cproensa:
Fix the oracle syntax to not remove those "AS" that are part of CAST
expressions.

Oracle doesn't suppor AS keyword as part of table aliases.
Previous code was modifying the query sql to remove all appearances of
"AS". However, "AS" can also appear as part of a CAST expression:
CAST( x AS y )

Fixes: #23268

### DIFF
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -1081,19 +1081,48 @@ function db_oracle_order_binds_sequentially( $p_query ) {
  */
 function db_oracle_adapt_query_syntax( $p_query, array &$p_arr_parms = null ) {
 	# Remove "AS" keyword, because not supported with table aliasing
-	$t_is_odd = true;
+	# - Do not remove text literal within "'" quotes
+	# - Will remove all "AS", except when it's part of a "CAST(x AS y)" expression
+	#   To do so, we will assume that the "AS" following a "CAST", is safe to be kept.
+	#   Using a counter for "CAST" appearances to allow nesting: CAST(CAST(x AS y) AS z)
+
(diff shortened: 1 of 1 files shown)


dregad:
I don't have time to look in details now, but I'm wondering if it would not be easier and simpler to limit processing to the query's FROM close (although the parser would have to take sub-queries into account, so implementing a recursive algorithm might be necessary)...

cproensa:
> but I'm wondering if it would not be easier and simpler to limit processing to the query's FROM

For simplicity, i would remove this and warn to not use table aliases with AS

Removing all "AS" was a hack, and this is still a bit more refined hack. However, any solution that requires a proper parser might be worse in performance...

dregad:
> Removing all "AS" was a hack, and this is still a bit more refined hack. 

Very true

> For simplicity, i would remove this and warn to not use table aliases with AS

That would work too I suppose, in which case we should ensure that we are not ourselves executing any queries with the offending syntax so as to not introduce regressions in Oracle support.

cproensa:
> in which case we should ensure that we are not ourselves executing any queries with the offending syntax so as to not introduce regressions in Oracle support.

Even if all queries were oracle compatible for this issue, a query by external code (plugin, customization, etc), could be non compatible. This is a problem because:
- it's defined as standard in the sql-99 syntax
- seems like usage of AS in table alias is relatively common
- there is no indicator that a query would be problematic, until it's run in an oracle instance, which is a small usage.
For those reasons i understand that this replacement was in place, and is reasonable

PR #1172, opened 2017-08-29
Add an option to reset current filters

cproensa:
Add an option, within user's account pages, to reset currently applied filters.
Having this option outside of main filtering pages is useful, in cases when the filter query throw errors, and such pages are inaccessible to reset the filters from the filter dialog.

This is related to:
https://www.mantisbt.org/bugs/view.php?id=21595
which i think can't be truly solved with current system. At least this publishes the option in a "safe" page, outside of the filter execution.

seleccion_248

This "filters" tab for account pages can be expanded with future functionality
eg: listing "my filters", "my view" boxes configuration, setting a per user default filter...


dregad:
I have not actually tested this, but I'm fine with the approach.

Considering the fact that this is a recurring problem, I believe it would be worth clearly documenting it as well as the workaround implemented here in the Admin Guide (troubleshooting section seems a good candidate)

vboctor:
The issue with this proposal is that there is a feature that user doesn't know its use and when the user hits a filtering error, it is not discoverable that they need to locate and use this feature.

I would rather if our error handler can support a registered recovering action that shows up if the page generates an error.  In such case, the filtering logic will register the recovery action before constructing and executing the queries.  If an error happens, then there is a button to reset filter as a recovery action.

vboctor:
Consider using php shutdown function registration for recovery actions.

PR #1174, opened 2017-08-31
Fix adding issues custom fields in REST API

elmarculino:
Fix for the [23264](https://www.mantisbt.org/bugs/view.php?id=23264) issue - Custom fields not been saved when adding issue through the Rest API.

The REST API pass a multi-dimensional array of custom field values and the mci_issue_set_custom_fields function expect an stdClass object.

### DIFF
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -159,17 +159,15 @@ function mci_issue_set_custom_fields( $p_issue_id, array &$p_custom_fields = nul
 	if( isset( $p_custom_fields ) && is_array( $p_custom_fields ) ) {
 		foreach( $p_custom_fields as $t_custom_field ) {
 
-			# convert the rest api multi-dimensional array into a stdClass object
-			if( is_array($t_custom_field) ) {
-				$t_custom_field = json_decode(json_encode($t_custom_field));
+			if( is_object( $t_custom_field ) ) {
+				$t_custom_field = ApiObjectFactory::objectToArray( $t_custom_field );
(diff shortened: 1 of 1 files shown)


api/soap/mc_issue_api.php
```diff
@@ -159,6 +159,11 @@ function mci_issue_set_custom_fields( $p_issue_id, array &$p_custom_fields = nul
 	if( isset( $p_custom_fields ) && is_array( $p_custom_fields ) ) {
 		foreach( $p_custom_fields as $t_custom_field ) {
 
+			# convert the rest api multi-dimensional array into a stdClass object
+			if( is_array($t_custom_field) ) {
+				$t_custom_field = json_decode(json_encode($t_custom_field));
```
dregad: I did not actually test to confirm the issue with HTTP 500 errors when adding custom field, but if all you need is conversion from *array* to *stdClass* then it's much more efficient to use a typecast:
```php
$t_custom_field = (object)$t_custom_field;
```

elmarculino: It's a multi-dimensional array. I can not cast all his levels.

vboctor:
Thanks @elmarculino for your contribution.

How does the API handle custom field names with spaces in their name?

elmarculino:
I didn't see any difference. It's working fine for me with this chance, but I always use the custom field id in the API request.

vboctor:
Merged via a1b8718b0c64b1f1292ecb8da15cbf1415a62b99

PR #1175, opened 2017-09-01
Fix missing enumeration break in REST API

elmarculino:
When requesting an issue in the REST API if an enumeration is missing the request simple fail. 

To avoid this silent break, the mci_enum_get_array_by_id function now log que error and return a not found message if enumeration element is not found.

### DIFF
--- a/api/soap/mc_enum_api.php
+++ b/api/soap/mc_enum_api.php
@@ -290,11 +290,25 @@ function mci_enum_get_array_by_id( $p_enum_id, $p_enum_type, $p_lang ) {
 		$t_result['name'] = MantisEnum::getLocalizedLabel( $t_enum_string_value, $t_enum_localized_value, $p_enum_id );
 	} else {
 		$t_enum_array = MantisEnum::getAssocArrayIndexedByValues( $t_enum_string_value );
+
+		if (!array_key_exists($p_enum_id, $t_enum_array)) {
+				log_event( LOG_WEBSERVICE, "Id '$p_enum_id' does not exist in the '$p_enum_type' enumeration"  );
+				$t_result['name'] = $t_result['label'] = 'NOT FOUND!';
+				return $t_result;
+		}
+
 		$t_result['name'] = $t_enum_array[$p_enum_id];
 		$t_result['label'] = MantisEnum::getLocalizedLabel( $t_enum_string_value, $t_enum_localized_value, $p_enum_id );
 
 		if( $p_enum_type == 'status' ) {
 			$t_status_colors = config_get( 'status_colors' );
+
+			if (array_key_exists($t_result['name'], $t_status_colors)) {
+					log_event( LOG_WEBSERVICE, "The color for status ". $t_result['name']  ." does not exist"  );
+					$t_result['color'] = 'COLOR NOT FOUND!';
+					return $t_result;
+			}
+
 			$t_result['color'] = $t_status_colors[$t_result['name']];
 		}
 	}


dregad:
Can you please reference the corresponding issue in our bugtracker (create one if needed)

elmarculino:
Related issue: 0023301

api/soap/mc_enum_api.php
```diff
@@ -290,11 +290,25 @@ function mci_enum_get_array_by_id( $p_enum_id, $p_enum_type, $p_lang ) {
 		$t_result['name'] = MantisEnum::getLocalizedLabel( $t_enum_string_value, $t_enum_localized_value, $p_enum_id );
 	} else {
 		$t_enum_array = MantisEnum::getAssocArrayIndexedByValues( $t_enum_string_value );
+
+		if (!array_key_exists($p_enum_id, $t_enum_array)) {
+				log_event( LOG_WEBSERVICE, "Id '$p_enum_id' does not exist in the '$p_enum_type' enumeration"  );
```
dregad: Not sure about logging here. While the error does occur in the webservice's context, this is really a system configuration problem, not an issue of the web service itself.

api/soap/mc_enum_api.php
```diff
@@ -290,11 +290,25 @@ function mci_enum_get_array_by_id( $p_enum_id, $p_enum_type, $p_lang ) {
...
+				$t_result['name'] = $t_result['label'] = 'NOT FOUND!';
```
dregad: I'm not overly keen on the hardcoded `NOT FOUND!`. For consistency with standard MantisBT behavior in such case, I believe the webservice should return `@<value>@` [just like Mantis does](https://github.com/mantisbt/mantisbt/blob/release-2.6.0/core/classes/MantisEnum.class.php#L223).

If a text message really has to be displayed, then it should be localized via *get_lang()* call (and in that case, again for consistency, the change should probably take place in *MantisEnum::getLabelForUnknownValue()* rather than here.

api/soap/mc_enum_api.php
```diff
@@ -290,11 +290,17 @@ function mci_enum_get_array_by_id( $p_enum_id, $p_enum_type, $p_lang ) {
...
-		$t_result['name'] = $t_enum_array[$p_enum_id];
+		$t_result['name'] = MantisEnum::getLabel( $t_enum_string_value, $p_enum_id );
 		$t_result['label'] = MantisEnum::getLocalizedLabel( $t_enum_string_value, $t_enum_localized_value, $p_enum_id );
 
 		if( $p_enum_type == 'status' ) {
 			$t_status_colors = config_get( 'status_colors' );
+
+			if( !array_key_exists( $t_result['name'], $t_status_colors ) ) {
+					$t_result['color'] = 'currentcolor';
+					return $t_result;
```
dregad: Remove extra indent.

api/soap/mc_enum_api.php
```diff
@@ -290,11 +290,17 @@ function mci_enum_get_array_by_id( $p_enum_id, $p_enum_type, $p_lang ) {
 		$t_result['name'] = MantisEnum::getLocalizedLabel( $t_enum_string_value, $t_enum_localized_value, $p_enum_id );
 	} else {
 		$t_enum_array = MantisEnum::getAssocArrayIndexedByValues( $t_enum_string_value );
```
vboctor: This line doesn't seem to be used anymore.
`$t_enum_array = MantisEnum::getAssocArrayIndexedByValues( $t_enum_string_value );`

dregad: Missed that. Fixed in ebac07b287bfc619dcc04333b7832dcf771d9eab, thanks !

PR #1176, opened 2017-09-03
Update strings_slovak.txt

KoapaSC:
I have added all missing string I found the newest verion.


atrol:
Translations are managed via TranslateWiki. Please refer our own Wiki [1] for the process to submit new translations or revise existing ones.

[1] http://www.mantisbt.org/wiki/doku.php/mantisbt:translationshowto

PR #1177, opened 2017-09-04
Fix REST API auth with show_realname = ON

dregad:
When accessing the webservice via browser as an authenticated user, and
$g_show_realname = ON, the request fails with HTTP 403 error.

This is due to using incorrect user_get_name() function to retrieve the
user's login name.

Calling user_get_field(<id>, 'username') instead.

Fixes [#23225](https://www.mantisbt.org/bugs/view.php?id=23225)

### DIFF
--- a/account_update.php
+++ b/account_update.php
@@ -68,7 +68,7 @@ if( !$t_account_verification ) {
 	# set a temporary cookie so the login information is passed between pages.
 	auth_set_cookies( $t_verify_user_id, false );
 	# fake login so the user can set their password
-	auth_attempt_script_login( user_get_field( $t_verify_user_id, 'username' ) );
+	auth_attempt_script_login( user_get_username( $t_verify_user_id ) );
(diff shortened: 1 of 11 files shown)


dregad:
@vboctor 
I've implemented the fix with a *user_get_field()* call, which is how we've historically been retrieving the username. 

The issue's reporter suggested to create a new function *user_get_username()*; which could make sense, as it would allow replacing 18 calls to `user_get_field( <id>, 'username' )` (19 if I count this one), so let me know if you prefer the new function approach.

dregad:
> Lets target this to the 2.6.x branch. 

Not sure why you want to target 2.6.x since this is an issue that was introduced in 2.3 it's not a regression so it should normally go in 2.7. But anyway I'm fine with that.

> We can add the new API as a separate code cleanup work item.

https://mantisbt.org/bugs/view.php?id=23331

PR #1182, opened 2017-09-06
Allow caching of generated js/css files

cproensa:
Allow client caching of dynamically generated css and js files.
Default headers disable explicitly caching of application pages, but
these generated files don't change usually between page loads.
By disabling default headers, the cacheability is defined by the server
configuration.

The included files are referenced by a URL built with
dummy parameters that allows cache busting on the browser cached copy,
based on the variables that may result in a different content, for
example, after changing logged user, language...

Fixes: #23324

### DIFF
--- a/core/helper_api.php
+++ b/core/helper_api.php
@@ -725,3 +725,38 @@ function helper_url_combine( $p_page, $p_query_string ) {
 
 	return $t_url;
 }
+
+/**
+ * Generate a hash to be used with dynamically generated content that is expected
+ * to be cached by the browser. This hash can be used to differentiate the generated
+ * content when it may be different based on some runtime attributes like: current user,
+ * project or language.
+ * An optional custom string can be provided to be added to the hash, for additional
+ * differentiating criteria, but this string must be already prepared by the caller.
+ *
+ * @param array $p_runtime_attrs    Array of attributes to be calculated from current session.
+ *                                  possible values: 'user', 'project', 'lang'
+ * @param string $p_custom_string   Additional string provided by the caller
+ * @return string                   A hashed md5 string
+ */
+function helper_generate_cache_key( array $p_runtime_attrs = [], $p_custom_string = '' ) {
+	# always add core version, to force reload of resources after an upgrade.
+	$t_key = $p_custom_string . '+V' . MANTIS_VERSION;
+	$t_user_auth = auth_is_user_authenticated();
+	foreach( $p_runtime_attrs as $t_attr ) {
+		switch( $t_attr ) {
+			case 'user':
+				$t_key .= '+U' . ( $t_user_auth ? auth_get_current_user_id() : META_FILTER_NONE );
+				break;
+			case 'project':
+				$t_key .= '+P' . ( $t_user_auth ? helper_get_current_project() : META_FILTER_NONE );
+				break;
+			case 'lang':
+				$t_key .= '+L' . lang_get_current();
+				break;
+			default:
+				trigger_error( ERROR_GENERIC, ERROR );
+		}
+	}
+	return md5( $t_key );
+}
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -214,6 +214,16 @@ function html_css() {
 		html_css_link( config_get( 'css_rtl_include_file' ) );
 	}
 	foreach( $g_stylesheets_included as $t_stylesheet_path ) {
+		# status_config.php is a special css file, dynamically generated.
+		# Add a hash to the query string to differentiate content based on its
+		# relevant properties. This allows a browser to cache them separately and force
+		# a reload when the content may differ.
+		if( $t_stylesheet_path == 'status_config.php' ) {
+			$t_stylesheet_path = helper_url_combine(
+				helper_mantis_url( 'css/status_config.php' ),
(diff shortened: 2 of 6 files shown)


cproensa:
@atrol I think this PR addresses your comments.
For example the generated url is `javascript_translations.php?lang=english`
Which will be different at the moment the language is switched, and the browser would treat it as a new resource to be downloaded.

css/status_config.php
```diff
@@ -28,9 +28,14 @@
 # Prevent output of HTML in the content if errors occur
 define( 'DISABLE_INLINE_ERROR_REPORTING', true );
 
+# Suppress default headers. This allows cachig as defined in server configuration
```
vboctor: typo `cachig`

javascript_config.php
```diff
@@ -27,9 +27,14 @@
 # Prevent output of HTML in the content if errors occur
 define( 'DISABLE_INLINE_ERROR_REPORTING', true );
 
+# Suppress default headers. This allows cachig as defined in server configuration
```
vboctor: typo `cachig`

javascript_config.php
```diff
@@ -27,9 +27,14 @@
...
+$g_bypass_headers = true;
```
vboctor: Should you skip headers that disable caching only when expected query parameter is supplied to handle the case where plugins include such files without query parameters?

cproensa:
@vboctor , regarding your comments:
Probably, to simplify, this can be changed into a single parameter, eg: "cache_key", which is a hash of a combination of variables (as needed in each case): user, language, project, core version...
The generation of this key can be then encapsulated for reuse.
Good idea, by including core version, an upgrade would force reload of these resources.

> Should you skip headers that disable caching only when expected query parameter is supplied

with the above approach, using a single parameter, your suggestion seems cleaner: only when the cache-key parameter is present, disable those headers in order to allow caching (if supported). It's a safer strategy overall.
But i don't think this is actually a need for plugins. I don't see an usual use case where a plugin may include those files without going through layout_api. Plugins may need to include it's own generated resources, but it's up to the plugin to implement a similar strategy on their own.

cproensa:
Added Victor's suggestions

core/html_api.php
```diff
@@ -297,8 +307,20 @@ function require_js( $p_script_path ) {
...
-	echo "\t" . '<script type="text/javascript" src="' . helper_mantis_url( 'javascript_config.php' ) . '"></script>' . "\n";
-	echo "\t" . '<script type="text/javascript" src="' . helper_mantis_url( 'javascript_translations.php' ) . '"></script>' . "\n";
+	# Add a hash to the query string to differentiate content based on its
+	# relevant properties. This allows a browser to cache them separately and force
+	# a reload when the content may differ.
+	$t_javascript_translations = helper_url_combine(
+		helper_mantis_url( 'javascript_translations.php' ),
+		'cache_key=' . helper_generate_cache_key( array( 'lang' ) )
+	);
+	$t_user_id = auth_is_user_authenticated()? auth_get_current_user_id() : null;
```
vboctor: Is the variable `$t_user_id` used anywhere?

core/helper_api.php
```diff
@@ -725,3 +725,36 @@ function helper_url_combine( $p_page, $p_query_string ) {
...
+ * Generate a hash to be used with dynamically generated content that is expected
+ * to be cached by the browser. This hash can be used to differentiate the generated
+ * content when it may be different based on some runtime attributes like: current user,
+ * project or language.
+ * An optional custom string can be provided to be added to the hash, for additional
+ * differentiating criteria, but this string must be already prepared by the caller.
+ *
+ * @param array $p_runtime_attrs    Array of attributes to be calculated from current session.
+ *                                  possible values: 'user', 'project', 'lang'
+ * @param type $p_custom_string     Additional string provided by the caller
```
atrol: *@param string

core/helper_api.php
```diff
@@ -725,3 +725,36 @@ function helper_url_combine( $p_page, $p_query_string ) {
...
+ * @return string                   A hashed md5 string
+ */
+function helper_generate_cache_key( array $p_runtime_attrs = [], $p_custom_string = '' ) {
+	$t_key = $p_custom_string;
```
atrol: Not needed if you change the following line to
`$t_key = $p_custom_string . '+V' . MANTIS_VERSION;`

cproensa:
updated, with including the same logic for plugin.php.
when plugin pages are run, at that point plugin.php has already set headers, started session, etc, so the same approach is needed.

core/helper_api.php
```diff
@@ -725,3 +725,35 @@ function helper_url_combine( $p_page, $p_query_string ) {
...
+		switch( $t_attr ) {
+			case 'user':
+				$t_key .= '+U' . ( $t_user_auth ? auth_get_current_user_id() : META_FILTER_NONE );
+				break;
+			case 'project':
+				$t_key .= '+P' . ( $t_user_auth ? helper_get_current_project() : META_FILTER_NONE );
+				break;
+			case 'lang':
+				$t_key .= '+L' . lang_get_current();
+		}
```
vboctor: Add a default that triggers a generic error.

cproensa: > Add a default that triggers a generic error.

added

PR #1184, opened 2017-09-17
Add no-op upgrade step in plugin_upgrade() 

dregad:
Implement the same functionality that was added to the MantisBT
installer in 061e66b5cfbda95e2ecbe87803c3388b2642cead.

This is a prerequisite to fix an issue with Source Integration plugin
https://github.com/mantisbt-plugins/source-integration/issues/36.

Fixes [#23367](https://mantisbt.org/bugs/view.php?id=23367)

### DIFF
--- a/core/plugin_api.php
+++ b/core/plugin_api.php
@@ -699,6 +699,12 @@ function plugin_upgrade( MantisPlugin $p_plugin ) {
 				}
 				break;
 
+			case null:
+				# No-op upgrade step
+				$t_sqlarray = false;
+				$t_status = 2;
+				break;
+
 			default:
 				$t_sqlarray = call_user_func_array(
 					array( $t_dict, $t_schema[$i][0] ),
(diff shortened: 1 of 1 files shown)


core/plugin_api.php
```diff
@@ -674,24 +676,40 @@ function plugin_upgrade( MantisPlugin $p_plugin ) {
...
+				}
+				break;
+
+			case null:
+				# No-op upgrade step
+				$t_sqlarray = false;
+				$t_status = 2;
+				break;
+
+			default:
```
vboctor: Ideally, we would error on default and have all the cases specifically handled.  That is not part of the scope of this consistency fix, but would be worth considering.

dregad: That's a good point, as unlike the MantisBT installer, we don't control the contents of the plugin schemas

dregad: Well, on second thoughts that would force us to list every possible ADOdb dictionary method... 

I think leaving _default_ is OK, especially considering that 
- we'd get a PHP warning from *call_user_func()*, which would then return a *null* status if given an invalid function name, 
- if given an existing function that does not return valid SQL statement code, ExecuteSQLArray() would fail

In both cases,  an ERROR_PLUGIN_UPGRADE_FAILED would be triggered.

PR #1185, opened 2017-09-18
Fix for oracle syntax after PR 1154

cproensa:
Oracle returns nulls as ''.
Change the check for an empty tags result.


vboctor:
@cproensa please create a bug and reference it in the commit message.

cproensa:
> please create a bug and reference it in the commit message.

Previous PR was merged after 2.6, so it has not been released yet
I can create a bug, but you must be careful to not set target version, etc.

vboctor:
@cproensa should we reference the same issue as the previous commit?

cproensa:
> should we reference the same issue as the previous commit?

Ok, with a "related" mark

PR #1187, opened 2017-09-20
Enforce usage of PHP >= 5.5.0

atrol:
Fixes #23378

### DIFF
--- a/core/php_api.php
+++ b/core/php_api.php
@@ -30,7 +30,7 @@
 /**
  * Constant for our minimum required PHP version
  */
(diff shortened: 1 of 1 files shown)


dregad:
If this was introduced in 2.6.0, maybe worth a 2.6.1 release.

atrol:
> If this was introduced in 2.6.0, maybe worth a 2.6.1 release.

I am not sure if this was introduced in 2.6.0, but backporting the change for a 2.6.1 release will not fix anything.

PR #1188, opened 2017-09-20
Remove code that is no longer needed when using PHP >= 5.5.0

atrol:
Fixes #23381

### DIFF
--- a/admin/install.php
+++ b/admin/install.php
@@ -246,7 +246,7 @@ if( $t_config_exists ) {
 
 		if( $f_db_type == 'mssql' ) {
 			print_test( 'Checking PHP support for Microsoft SQL Server driver',
-				version_compare( phpversion(), '5.3' ) < 0, true,
+				BAD, true,
 				'mssql driver is no longer supported in PHP >= 5.3, please use mssqlnative instead' );
 		}
 	}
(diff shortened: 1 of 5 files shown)


vboctor:
@atrol If an instance has database type as `mysql` do we still support that and map it to `mysqli`?

atrol:
> If an instance has database type as mysql do we still support that and map it to mysqli?

No, it's no longer supported, there is no mapping.
I added a check in install.php to tell users to use mysqli instead of mysql.

admin/install.php
```diff
@@ -246,9 +246,15 @@ function print_test( $p_test_description, $p_result, $p_hard_fail = true, $p_mes
...
 			print_test( 'Checking PHP support for Microsoft SQL Server driver',
-				version_compare( phpversion(), '5.3' ) < 0, true,
+				BAD, true,
 				'mssql driver is no longer supported in PHP >= 5.3, please use mssqlnative instead' );
 		}
+
+		if( $f_db_type == 'mysql' ) {
+			print_test( 'Checking PHP support for MySQL driver',
+				BAD, true,
+				'mysql driver is deprecated in PHP >= 5.5, please use mysqli instead' );
```
dregad: I would add that we no longer support it.

atrol: Changed the message.

PR #1189, opened 2017-09-21
Support issue update REST API

elmarculino:
Support updating an issue from a POST to the issues/update REST API url. 
Fixes #23396

### DIFF
--- a/api/rest/restcore/issues_rest.php
+++ b/api/rest/restcore/issues_rest.php
@@ -202,6 +202,11 @@ function rest_issue_note_delete( \Slim\Http\Request $p_request, \Slim\Http\Respo
  */
 function rest_issue_update( \Slim\Http\Request $p_request, \Slim\Http\Response $p_response, array $p_args ) {
 	$t_issue = (object)$p_request->getParsedBody();
+
+	if( !property_exists( $t_issue, 'id' ) ) {
+		$t_message = "Mandatory field 'id' is missing.";
+		return $p_response->withStatus( HTTP_STATUS_BAD_REQUEST, $t_message );
+	}
 	$t_issue_id = $t_issue->id;
 
 	$t_result = mc_issue_update( /* username */ '', /* password */ '', $t_issue_id, $t_issue );
--- a/api/rest/restcore/issues_rest.php
+++ b/api/rest/restcore/issues_rest.php
@@ -29,8 +29,8 @@ $g_app->group('/issues', function() use ( $g_app ) {
 	$g_app->get( '/{id}/', 'rest_issue_get' );
 	$g_app->post( '', 'rest_issue_add' );
 	$g_app->post( '/', 'rest_issue_add' );
-	$g_app->post( 'update', 'rest_issue_update' );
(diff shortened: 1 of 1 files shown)


dregad:
@elmarculino please reference an issue in our bugtracker, both in the PR description and the PR's commit (i.e. `Fixes #xxxx`)

elmarculino:
@vboctor in my tests when I get an issue, edit some fields and than update the issue via REST API, only the edited fields appear in the Issue Change History. I thought this was handled internally after calling the mc_issue_update() function. Isn't that an expected result?

api/rest/restcore/issues_rest.php
```diff
@@ -188,4 +192,32 @@ function rest_issue_note_delete( \Slim\Http\Request $p_request, \Slim\Http\Respo
...
+ */
+function rest_issue_update( \Slim\Http\Request $p_request, \Slim\Http\Response $p_response, array $p_args ) {
+	$t_issue = (object)$p_request->getParsedBody();
+
+	$t_issue_id = isset( $p_args['id'] ) ? $p_args['id'] : $p_request->getParam( 'id' );
+	if( is_blank( $t_issue_id ) ) {
+		$t_message = "Mandatory field 'id' is missing.";
+		return $p_response->withStatus( HTTP_STATUS_BAD_REQUEST, $t_message );
+	}
+
```
vboctor: You should call mc_issue_get() here, and merge the contents of the post body on top of the returned issue.  Hence, enabling the caller to just call PATCH with just the modified fields (e.g. handler and status).  This way the client doesn't need to get the full issue and then call patch with it.  This also behaves better if two clients are trying to to do update at the same time (e.g. one client updating handler + status and the other trying to change the priority.

api/rest/restcore/issues_rest.php
```diff
@@ -188,4 +192,32 @@ function rest_issue_note_delete( \Slim\Http\Request $p_request, \Slim\Http\Respo
...
+ *
+ * @param \Slim\Http\Request $p_request   The request.
+ * @param \Slim\Http\Response $p_response The response.
+ * @param array $p_args Arguments
+ * @return \Slim\Http\Response The augmented response.
+ */
+function rest_issue_update( \Slim\Http\Request $p_request, \Slim\Http\Response $p_response, array $p_args ) {
+	$t_issue = (object)$p_request->getParsedBody();
+
+	$t_issue_id = isset( $p_args['id'] ) ? $p_args['id'] : $p_request->getParam( 'id' );
```
vboctor: Add a check that if issue id is supplied in the body of the post, then it must match the one extracted from the URL.

api/rest/restcore/issues_rest.php
elmarculino: @vboctor when I get an issue, edit some fields and update the issue (RESP API), only the edited fields appear as modified in the Issue Change History. This merge is not executed in the mc_issue_update? And, what about the mandatory fields as summary, description or category?

vboctor:
Closing in favor of #1210

PR #1190, opened 2017-09-23
Add PHP and database information to manage_overview_page.php

atrol:
Fixes #23393


cproensa:
Do you think that some specific info on OS is useful?
From the phpversion() output i get `7.0.22-0ubuntu0.16.04.1`, but i don't know if that is consistent with other systems.

With $g_db->ServerInfo(), Why not show also "description"?
I think, at least, it would allow to differentiate Mysql from MariaDB.

atrol:
> Do you think that some specific info on OS is useful?

Could be useful

> And I agree with @cproensa comment about showing db description, it can be useful.

Done

PR #1192, opened 2017-09-24
Style the login logo with a css class

cproensa:
Fixes: #23382

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -1251,6 +1251,10 @@ function layout_scroll_up_button() {
 	echo '</a>' . "\n";
 }
 
+/**
+ * Render the div area with logo used in login pages
+ * @return void
+ */
 function layout_login_page_logo() {
 	?>
 	<div class="login-logo">
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -1250,3 +1250,13 @@ function layout_scroll_up_button() {
 	echo '<i class="ace-icon fa fa-angle-double-up icon-only bigger-110"></i>' . "\n";
 	echo '</a>' . "\n";
 }
+
+function layout_login_page_logo() {
+	?>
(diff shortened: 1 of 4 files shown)


core/layout_api.php
```diff
@@ -1250,3 +1250,11 @@ function layout_scroll_up_button() {
 	echo '<i class="ace-icon fa fa-angle-double-up icon-only bigger-110"></i>' . "\n";
 	echo '</a>' . "\n";
 }
+
+function layout_login_page_logo() {
+	?>
+	<div class="login-logo">
+		<img src="<?php echo helper_mantis_url( config_get( 'logo_image' ) ); ?>">
+	</div>
```
dregad: I'm not sure why an `<h1>` tag was used before, but is there any particular reason to change from that to a `<div>` ?

cproensa: > but is there any particular reason to change from that to a <div>?

div is a neutral tag that does not carry on any formatting, and is better suited to apply a css customization to change the positioning outside of the container flow, if needed.

dregad: Makes sense.

core/layout_api.php
```diff
@@ -1250,3 +1250,11 @@ function layout_scroll_up_button() {
 	echo '<i class="ace-icon fa fa-angle-double-up icon-only bigger-110"></i>' . "\n";
 	echo '</a>' . "\n";
 }
+
+function layout_login_page_logo() {
```
atrol: Missing PHPDoc block

cproensa:
note. It must be done also on signup, verify, lost-pwd, etc

atrol:
Closed via https://github.com/mantisbt/mantisbt/commit/883655a761cf9aa35fc82b311487d06a6cd9088e

PR #1193, opened 2017-09-25
Oracle, read config values as varchar

cproensa:
When using Oracle database, read config values as varchar type whenever
it's possible, to avoid performance issues of reading clob field types.

Even if the issue may also exist with other tables, config table is
especially sensitive, as it is fully fetched  for every page execution.

Fixes: #23395

Query execution time, before:
DB-5 **3.9610** config_api.php:136 config_get() 
SELECT config_id, user_id, project_id, type, value, access_reqd FROM m_config

After:
DB-5 **0.0315** config_api.php:727 config_cache_all() 
SELECT config_id, user_id, project_id, type, CAST(value AS VARCHAR(4000)) value, access_reqd FROM m_config WHERE dbms_lob.getlength(value)<=1000
DB-6 **0.0298** config_api.php:737 config_cache_all() 
SELECT config_id, user_id, project_id, type, value, access_reqd FROM m_config WHERE dbms_lob.getlength(value)>1000

Requires #1171

### DIFF
--- a/core/config_api.php
+++ b/core/config_api.php
@@ -132,15 +132,7 @@ function config_get( $p_option, $p_default = null, $p_user = null, $p_project = 
 			# @@ debug @@ echo 'pr= '; var_dump($t_projects);
 			# @@ debug @@ echo 'u= '; var_dump($t_users);
 			if( !$g_cache_filled ) {
-				$t_query = 'SELECT config_id, user_id, project_id, type, value, access_reqd FROM {config}';
-				$t_result = db_query( $t_query );
-				while( false <> ( $t_row = db_fetch_array( $t_result ) ) ) {
-					$t_config = $t_row['config_id'];
-					$t_user = $t_row['user_id'];
-					$t_project = $t_row['project_id'];
-					$g_cache_config[$t_config][$t_user][$t_project] = $t_row['type'] . ';' . $t_row['value'];
-					$g_cache_config_access[$t_config][$t_user][$t_project] = $t_row['access_reqd'];
-				}
+				config_cache_all();
 				$g_cache_filled = true;
 			}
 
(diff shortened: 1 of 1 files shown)


dregad:
@cproensa Nice work on the performance improvement !

I'm just wondering what would happen if the DBMS_LOB package is not installed (is it by default ?)... Is it feasible to test its presence and fallback to "normal" mechanism to retrieve config table ?

We should anyway document that the package is necessary (or at least strongly recommended) as part of system requirements.

core/config_api.php
```diff
@@ -727,3 +719,42 @@ function config_is_defined( $p_option, $p_value = null ) {
...
+	if( db_is_oracle() ) {
+		$t_query = 'SELECT config_id, user_id, project_id, type, CAST(value AS VARCHAR(4000)) AS value, access_reqd FROM {config} WHERE dbms_lob.getlength(value)<=1000';
+		$t_result = db_query( $t_query );
+		while( $t_row = db_fetch_array( $t_result ) ) {
+			$t_config = $t_row['config_id'];
+			$t_user = $t_row['user_id'];
+			$t_project = $t_row['project_id'];
+			$g_cache_config[$t_config][$t_user][$t_project] = $t_row['type'] . ';' . $t_row['value'];
+			$g_cache_config_access[$t_config][$t_user][$t_project] = $t_row['access_reqd'];
+		}
```
dregad: To reduce code duplication, I suggest to move this out of the if and use the same code as other RDBMS, i.e.
```php
if( db_is_oracle() ) {
    $t_query = ...
}
$t_result = ...
while() { ... }
if( db_is_oracle() ) {
    // CLOB query & loop
}
```
And/or maybe use a helper function to assign cache config and access values.

cproensa:
> I'm just wondering what would happen if the DBMS_LOB package is not installed (is it by default ?)... Is it feasible to test its presence and fallback to "normal" mechanism to retrieve config table ?

I think this is a built-in package, that is installed automatically.
Surely, it's already needed to use file upload, for blob fields.

core/config_api.php
cproensa: i see the code bloat, but:
- i'd rather not add another helper function, only for this usage
- i prefer showing a clear separation of the code paths

lets see how it can be cleaned

cproensa:
@dregad, code has been reorganised, if you like it more this way (commits should be squashed)

cproensa:
Other tables also suffer from this issue with clobs, eg: description fields for projects, tags, versions...
However, those are not as severe as config table.

Second offender is project table, which is also usually read for each page load. But a less number of rows are fetched, depending on project count and user assignment to projects.
If those cases are to be addressed, i think a better approach is to not cache descriptions by default (not part of this PR)

dregad:
> I think this is a built-in package, that is installed automatically.

That's what inferred from [Oracle doc](https://docs.oracle.com/database/121/ARPLS/intro.htm#ARPLS65020):

*Most Oracle supplied packages are automatically installed when the database is created. Certain packages are not installed automatically. Special installation instructions for these packages are documented in the individual chapters.*

but I just thought I'd get your confirmation.

> Surely, it's already needed to use file upload, for blob fields.

I have no idea how the oci8 extension internally handles blobs; ADOdb just makes use of oci_new_descriptor() and other functions. If it does use the package, then indeed we don't need to worry about it.

Thanks for refactoring the code, I think it's much clearer that way (at the cost of a bit higher memory usage, but that should be fine).

dregad:
> Second offender is project table, [...] If those cases are to be addressed, 

Ideally as a long-term solution we should get rid of some of these TEXT fields, they are really not necessary IMO.

Project table would only be an issue for instances with a large number of them; given the low usage of Oracle for Mantis, I'd be tempted to leave it like that for now, and wait for someone to complain...

core/config_api.php
```diff
@@ -727,3 +719,40 @@ function config_is_defined( $p_option, $p_value = null ) {
...
+	global $g_cache_config, $g_cache_config_access;
+
+	$t_config_rows = array();
+
+	# With oracle databse, ADOdb maps column type "L" to clob.
+	# Because reading clobs is significantly slower, cast them to varchar for faster query execution
+	# Standard max size for varchar is 4000 bytes, so a safe limit is used as 1000 charancters
+	# for multibyte strings (up to 4 bytes per char)
+	if( db_is_oracle() ) {
+		$t_query = 'SELECT config_id, user_id, project_id, type, CAST(value AS VARCHAR(4000)) AS value, access_reqd FROM {config} WHERE dbms_lob.getlength(value)<=1000';
```
dregad: If you don't mind, I'd suggest to split the long line
```php
$t_query = 'SELECT config_id, user_id, project_id, type, CAST(value AS VARCHAR(4000)) AS value, access_reqd
	FROM {config}
	WHERE dbms_lob.getlength(value)<=1000';
```

cproensa:
@dregad , updated with clean up suggestions

PR #1194, opened 2017-09-25
Remove non-clickable arrows from sidebar menu items

cproensa:
On the collapsed sidebar menu items, the arrow shown when they are
hovered, places an arrea that is not clickable.
Remove the arrows since the UI value is small compared to the annoyance
as the user expects the whole area to be clikable.

Fixes: #22245

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -940,7 +940,6 @@ function layout_sidebar_menu( $p_page, $p_title, $p_icon, $p_active_sidebar_page
 	echo '<i class="menu-icon fa ' . $p_icon . '"></i> ' . "\n";
 	echo '<span class="menu-text"> ' . lang_get_defaulted( $p_title ) . ' </span>' . "\n";
 	echo '</a>' . "\n";
-	echo '<b class="arrow"></b>' . "\n";
 	echo '</li>' . "\n";
 }
 


core/layout_api.php
```diff
@@ -940,7 +940,6 @@ function layout_sidebar_menu( $p_page, $p_title, $p_icon, $p_active_sidebar_page
 	echo '<i class="menu-icon fa ' . $p_icon . '"></i> ' . "\n";
 	echo '<span class="menu-text"> ' . lang_get_defaulted( $p_title ) . ' </span>' . "\n";
 	echo '</a>' . "\n";
-	echo '<b class="arrow"></b>' . "\n";
```
dregad: Since this is a workaround, I think it would be best to comment-out the line instead of removing it, and add a comment with a @TODO note explaining why with a reference to the issue.

cproensa: We usually always remove unused code instead of commenting

cproensa:
> I tried to implement that, but despite my efforts I couldn't achieve it... my CSS skills are not good enough to get through this mess of :before, :hover and various cascading styles...

I can't either. This may be a hard limitation from those pseudo-elements 

> This is a workaround
> I think it would be best to comment-out the line instead of removing it, and add a comment with a @todo note explaining why with a reference to the issue.

I don't think so defensive about it. If a better fix can be done about it, then go ahead.

dregad:
> If a better fix can be done about it, then go ahead.

I already admitted my inability to do so... :disappointed: 
@syncguru maybe you have an idea on how this could be fixed. If not, then I think we should go with Carlos' approach of this being a permanent design change, and not a temporary fix.

cproensa:
> this being a permanent design change, and not a temporary fix.

No need to be a permanent trait. 
Other refinements can be done instead, for example, paint a stronger border in the place where the arrows are, etc. Probably, avoiding the pseudo-element, other solutions will work fine.
I'm not good enough in the UI to make a proper replacement though.

dregad:
> No need to be a permanent trait.

Well permanent in the sense that neither you nor me will be able to do anything about it, therefore I think it's quite unlikely that it would change again in the foreseeable future :wink:

atrol:
👍 to get this workaround now instead of waiting even longer for another solution.

cproensa:
@vboctor 
can you put the arrow "on the right side" of the separator? like pointing to the expanded title, instead of towards the icon
The issue here is mainly that the inner area in the icon squares, where the arrow would appear, is usually the point where the mouse lands and expands the menu item.

atrol:
Another approach to fix it https://github.com/mantisbt/mantisbt/pull/1196

dregad:
I believe that @atrol's alternative approach is a much better solution, and suggest to close this PR in favor of #1196.

PR #1196, opened 2017-09-26
Make hover arrows clickable on collapsed sidebar menu

atrol:
An alternative approach to https://github.com/mantisbt/mantisbt/pull/1194

I tried latest Firefox, Chrome, IE and Edge.
Will have to check Safari this evening.

Looks quite good https://caniuse.com/#feat=pointer-events

Fixes #22245

### DIFF
--- a/css/ace-mantis.css
+++ b/css/ace-mantis.css
@@ -187,6 +187,11 @@ pre {
     cursor: pointer !important;
 }
 
+/* Clickable hover arrows on collapsed sidebar #22245 */
+.nav-list li > .arrow {
+    pointer-events: none;
+}
+
 .fa-xlg {
     font-size: 1.6em;
     line-height: 0.73em;


cproensa:
counter intuitive: no-click makes it clickable.
If compatibility is proved, this is ideal solution

atrol:
> counter intuitive: no-click makes it clickable.

Renamed to no-pointer-events

cproensa:
> Renamed to no-pointer-events

I mean, even the "no-pointer-events" clause is counter intuitive for what we are trying to achieve here.
I don't mind, it's just a thought

atrol:
> even the "no-pointer-events" clause is counter intuitive for what we are trying to achieve here

The clasee name is now similar to the CSS property name and expresses what is really happening: The arrow gets no longer any pointer events.
We don't make the arrow clickable (of course it seems to be the case for the end user), but we make the area behind the arrow clickable.

dregad:
This is a great solution, allowing to keep the UI as it was before. Nice one @atrol, thank you !

Instead of changing the markup *and* defining a new class, I would suggest an alternative approach leveraging the existing HTML, just adding in ace-mantis.css:
```css
/* Clickable hover arrows on collapsed sidebar #22245 */
.nav-list li > .arrow {
	pointer-events: none;
}
```

atrol:
> just adding in ace-mantis.css

Done

atrol:
> Will have to check Safari this evening.

Works also using lastest Safari Version 11.0 (12604.1.38.1.7)

cproensa:
What's the effect on a browser not compatible?
If it's the same as before the PR, then it's not a big deal

atrol:
> What's the effect on a browser not compatible?

I didn't test (e.g. would have to setup IE10).
According CSS3 specification, unknown properties must be ignored.
So I assume it will be the same as before and maybe there are some warnings in browser console.

From https://www.w3.org/TR/css-syntax-3/#style-rule
> Declaration for an unknown CSS property or whose value does not match the syntax defined by the property are invalid and must be ignored.

cproensa:
+1 anyway

PR #1197, opened 2017-09-26
Update mc_issue_api.php

dwings:
Fix to make this work with the REST API


dregad:
Please link this to an issue in our bugtracker, describing what the problem is

vboctor:
@dwings please also make sure that the change doesn't negatively impact SOAP API.

elmarculino:
There is already an issue in the bugtracker [23264](https://www.mantisbt.org/bugs/view.php?id=23264) and a PR #1174 for this bug.

vboctor:
Merged via a1b8718b0c64b1f1292ecb8da15cbf1415a62b99

PR #1199, opened 2017-09-27
Create event EVENT_BUG_ACTIONGROUP_FORM

cproensa:
Create event EVENT_BUG_ACTIONGROUP_FORM to add custom content to the
group action page form.
For example: custom inputs that can later be evaluated when processing
EVENT_BUG_ACTION.

Fixes: #13126

seleccion_261


vboctor:
@cproensa Don't we usually add these events in pairs? `_FORM` and `_DATA`?

cproensa:
> Don't we usually add these events in pairs? _FORM and _DATA?

@vboctor good pint, but not a simple one, imo

The current `_DATA` events, are those specific to a modification of an entity: eg bug creation, and update. An action from group-action page may consist of a bug update, which then would then trigger those _data events. 
Note that PR #1098 enables that consistent behaviour by routing all those bug modifications through the BugData object.

The same discussion that was held in that PR: other changes, not controlled by BugData, should be included in a similar model, or extend it as some kind of "transaction" where all context of a modification can be pre- and post-processed.

We could improve:
- EVENT_BUG_ACTION: currently only have "bug_id" and literal "action" as parameters. This could be enhanced to carry specific information of the action performed. Eg: note added, tags attached, etc. But this should be consistent also when those actions are performed outside from group-action page (eg attaching tags from issue page)
- EVENT_BUG_ACTION_DATA: Same comments that previous event, but offering the entered changes for validation. Same issues with consistency. Additionally, there is not a consistent way of packaging before and after data, as we do have with the BugData events...

So, this new event adds some control to group-actions, and some single actions that are routed through it (eg, move).
I'd consider the other proposals as separate work, it has some complexities on their own...

vboctor:
I was thinking of the simpler scenario where a plugin adds extra fields to the group action form and want to handle such fields on the action page.  This seems to be valuable even if the core changes are not packaged in an object.

cproensa:
>  the simpler scenario where a plugin adds extra fields to the group action form and want to handle such fields on the action page

That's what EVENT_BUG_ACTION can be used for, but this event is triggered after the action is done.

For example, i would add a checkbox for move action:
when an issue is moved to a different project, offer the option to remove the handler, if he's not a valid handler in destination project.
After the issue is moved, check the option with EVENT_BUG_ACTION and update handler if checked.

For the preprocess event:
I had the same thinking: #20576, for a event to preprocess the action. However, after further inspection, i moved my mind to properly hadle the BugData events.

A "preprocess" event for group-action, will have value for those operations currently not covered with it's own events. I'm not sure atm if there are many, or any,  of those. And, not having a system to manage the changeset, i guess the only way to alter the action may be to manipulate the $_GET/post values before they are read, or similar tricks.

In any case, that could be treated in a separate PR?

dregad:
@vboctor can this be merged (i.e. do you agree with @cproensa's proposal to treat the preprocess events as a separate issue) ?

PR #1201, opened 2017-09-30
Fix order of activities with same timestamp

cproensa:
When activities of the same type have the same timestamp, for example
when created from cloning issues, use the element id for sorting.

Fixes: #22441

### DIFF
--- a/core/bug_activity_api.php
+++ b/core/bug_activity_api.php
@@ -187,6 +187,11 @@ function bug_activity_sort( &$p_entries ) {
 			return $t_order == 'DESC' ? -1 : 1;
 		}
 
+		# same timestamp and same type, probably came from cloning an issue
+		if( $a['timestamp'] == $b['timestamp'] && $a['type'] == $b['type'] ) {
+			return (int)$a['id'] - (int)$b['id'];
+		}
+
 		if( $a['user_id'] < $b['user_id'] ) {
 			return -1;
 		}


cproensa:
@vboctor 
you wrote the activity sorting for showing inline attachments.

There is a related inconsistency:
cloning copy notes with current timestamps, but attachments are copied with original timestamps

If this were fixed to all have current timestamp:
In the scenario for clonign issues, i think correct positioning of attachment wouldn't be possible, if all notes and attachments had same timestamps, they cant be sorted, due to notes and attachments having different id ranges.

If keeping original timestamp, sorting issue would be solved, but there would be an inconsistency for activities having an older timstamp than issue creation

dregad:
Re: inconsistency of attached files timestamp when cloning issues - I would keep the original timestamp, but defer to @vboctor for the final decision. In any case, that would have to be treated in a separate issue/pull request, no ?

cproensa:
>  In any case, that would have to be treated in a separate issue/pull request, no ?

yes

with this PR, the result is:
- bugnotes in the new bug appear in the same order as in the original bug.
- attachments appear at the begining of activities, and still can't be interleaved with notes

which is a bit closer to the correct result than current.

> I would keep the original timestamp

note that it means: keep the original timestamps for both bugnotes and attachments.

in the long term, i think the correct solution is to have a master table for "activities" that stores references to bugnotes, attachmente, etc.
this way the sort key "timestamp, id" is always coherent.

vboctor:
@cproensa Let's track these issues with specific bugs.  The appropriate fix for bug notes and attachments is to have attachments associated with bug notes in the db.

PR #1202, opened 2017-10-01
Don't add users to monitoring list when resolving an issue as duplicate

atrol:
Fixes #23420


atrol:
> Assuming issue 1 is marked are a duplicate of issue 2.
> assuming reporter/handler of issue 1 are added to monitor list of issue 2.

Right, see also description https://www.mantisbt.org/bugs/view.php?id=23420

PR #1203, opened 2017-10-02
Fix PHP errors and warning in Issue Trends report

atrol:
Fixes #23425


plugins/MantisGraph/core/graph_api.php
```diff
@@ -497,6 +497,7 @@ function create_cumulative_bydate() {
 		$t_last_date = $t_date;
 	}
 
+	if ( $t_last_id == 0 ) return null;
```
vboctor: Reformat to have a {} block.

PR #1205, opened 2017-10-09
Don't deliver unneeded JavaScript if Graphs plugin is enabled

atrol:
Fixes #23446

We do request/deliver `chart.bundle-2.1.6.min.js`, `chart-2.1.6.min.js` and `MantisGraph.js` with *each* page request if Graphs plugin is enabled.

All in all about 322KB are transferred when accessing Mantis for the first time.
The files are cached, but they are still causing unneeded round trips to the server.

Most of the time the files are not needed as
- most of the users do not have access to the graphs ($g_view_summary_threshold = MANAGER)
- most of the pages don't need the scripts (might change in future, e.g. if we should have some graphs on `My View` page)

### DIFF
--- a/plugins/MantisGraph/MantisGraph.php
+++ b/plugins/MantisGraph/MantisGraph.php
@@ -99,7 +99,7 @@ class MantisGraphPlugin extends MantisPlugin  {
 	 * @return void
 	 */
 	function csp_headers() {
-		if ( config_get_global( 'cdn_enabled' ) == ON ) {
+		if( config_get_global( 'cdn_enabled' ) == ON ) {
 			http_csp_add( 'script-src', 'https://cdnjs.cloudflare.com' );
 		}
 	}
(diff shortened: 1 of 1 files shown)


plugins/MantisGraph/MantisGraph.php
```diff
@@ -122,14 +122,16 @@ function graph_filter_menu() {
 	 * @return void
 	 */
 	function resources() {
-		if ( config_get_global( 'cdn_enabled' ) == ON ) {
-			html_javascript_cdn_link( 'https://cdnjs.cloudflare.com/ajax/libs/Chart.js/' . CHARTJS_VERSION . '/Chart.min.js', CHARTJS_HASH );
-			html_javascript_cdn_link( 'https://cdnjs.cloudflare.com/ajax/libs/Chart.js/' . CHARTJS_VERSION . '/Chart.bundle.min.js', CHARTJSBUNDLE_HASH );
-		} else {
-			echo '<script src="' . plugin_file( 'chart-' . CHARTJS_VERSION . '.min.js' ) . '"></script>';
-			echo '<script src="' . plugin_file( 'chart.bundle-' . CHARTJS_VERSION . '.min.js' ) . '"></script>';
+		if( substr( gpc_get_string( 'page', '' ), 0, 12 ) === 'MantisGraph/' ) {
```
dregad: Suggest using `$this->basename` instead of hardcoding

cproensa:
there is a core function: `require_js()`
would this be suitable, to be used from a plugin page?
Otherwise, should be good to have an adaptation for plugins?

The same issue also appears with other plugins, and the fix presented here (which i sometimes use in my plugins) is not very clean.

atrol:
> Suggest using $this->basename instead of hardcoding

@dregad, I hardcoded `MantisGraph/` to be sure to exclude also any other plugin that might start with MantisGraph, e.g. something like MantisGraphAddOn.
$this->basename does not contain the `/`, so additional operations would be needed to cover this.

If we ever should change the plugin name, even the simplest test would show that something goes wrong.
So I preferred hardcoding because of better performance.

> there is a core function: require_js()
> would this be suitable, to be used from a plugin page?

@cproensa, require_js() adds the script file name to $g_scripts_included
$g_scripts_included is iterated in function html_head_javascript() to call html_javascript_link()
html_javascript_link() contains hardcoded path 'js/', so it can't be used at this place as the plugin js-files ore not stored in js folder.

> Otherwise, should be good to have an adaptation for plugins?

There are no rules where plugins store their js-files, so a path would be needed.
IMO should not be part of this PR.

> and the fix presented here (which i sometimes use in my plugins) is not very clean.

Agreed, but it's the easiest solution in short term and does not introduce any side effects.

plugins/MantisGraph/MantisGraph.php
```diff
@@ -122,14 +122,16 @@ function graph_filter_menu() {
...
+			if (config_get_global('cdn_enabled') == ON) {
```
cproensa: add spaces within ( )

cproensa:
> html_javascript_link() contains hardcoded path 'js/', so it can't be used at this place as the plugin js-files are not stored in js folder.

true, that's why i suggest a more general function

> There are no rules where plugins store their js-files, so a path would be needed.

a url should be provided as created by plugin_file or plugin_page

> IMO should not be part of this PR.

true, anyway i'm seeing it as useful feature, since i have encountered the same problem in my plugins

plugins/MantisGraph/MantisGraph.php
atrol: Changed, got messed by IDE.

dregad:
> If we ever should change the plugin name, even the simplest test would show that something goes wrong.
So I preferred hardcoding because of better performance.

@atrol point taken, although I don't think there would be so much overhead to make the code generic with

```php
$t_page_name = current( explode( '/', gpc_get_string( 'page', '' ) ) ); // or explode(...)[0]
if( $t_page_name === $this->basename ) {
```

That being said, I agree with @cproensa that the better approach would be to implement an API function to automatically deal with this case (we already have *plugin_require_api()*, so adding e.g. *plugin_require_js()* or something similar would make sense).

I leave it for you to decide whether to take the opportunity to revise the PR for a more generic solution now, or to just merge this (I'm fine with it as-is), and open an issue to track the rest for a later implementation.

PR #1207, opened 2017-10-10
Do not deliver code for outdated unsupported IE9 browser

atrol:
Fixes #23451

Our minimal client requirement for Internet Explorer is version 10 according Admin Guide.
https://www.mantisbt.org/docs/master/en-US/Admin_Guide/html-desktop/#admin.install.requirements.client

There is no need to go on delivering code that is just used for IE9 and older.

We should also think about supporting just IE >= 11 as IE10 is no longer supported since January 2016 and IE11 is available on all currently supported Windows versions (7, 8.1, 10).


atrol:
Ace comes with some more files than we have at the moment.
Some of them have been removed/changed some while ago , see https://github.com/mantisbt/mantisbt/commit/bd73abd5c8de7f2b5142b32327ff2038c775388c

So removing the files does not add any additional problem.

PR #1209, opened 2017-10-14
Numeric fields shouldn’t accept floats

vboctor:
Fixes #16133


atrol:
A value like `12345678901234567890` is no longer allowed after this change.

vboctor:
@atrol good catch.  Before this change, it was allowed, but displayed incorrectly, since the display method casted to int.  The core issue here is what to type cast numeric fields to (if any).  I don't recall right now why I added the casting earlier.  Will have to look that up.

PR #1211, opened 2017-10-16
DomainKeys Identified Mail (DKIM) Signatures

yaddran:
Implements #23493

This is a simple feature that exists in PHPMailer for years now. It helps a lot in preventing mails sent being classified as SPAM.


yaddran:
Done:
- opened a feature request number #23493
- referenced the PR from the issues and vice versa 
- added email_dkim prefix
- added the possibility to specify key as value

This feature is completely connected to email so however you treat smtp settings, for instance, should be the same way how you treat the DKIM settings. Since the email settings are using example.com as defaults and DKIM signature has to use the same values I think it is wise to have the same values example.com values for DKIM.

The validation is not necessary. The email sending will not stop. If one of the needed values are skipped or wrong, the PHPMailer will still send mail (silently skipping the problem) but it could happen that the signature is not added or that the signature later does not pass the DKIM check at the arriving mail server. I think that this is acceptable behavior. When you are trying to configure DKIM it is logical that you will check if it is working.

Todo:
- update documentation (a small pointer please :) )

core/email_api.php
```diff
@@ -1300,6 +1300,16 @@ function email_send( EmailData $p_email_data ) {
 			break;
 	}
 
+	#apply DKIM settings
+	if( config_get( 'dkim_enable' ) ) {
+		$t_mail->DKIM_domain = config_get( 'email_dkim_domain' );
+		$t_mail->DKIM_private = config_get( 'email_dkim_private_key' );
```
atrol: `private_key` is certainly better than `private`, but wouldn't something like `private_key_file_path` be even better?

atrol:
> update documentation (a small pointer please :) )

Options should be added to https://github.com/mantisbt/mantisbt/blob/master/docbook/Admin_Guide/en-US/config/email.xml

Not complete up to date, but should help
https://www.mantisbt.org/wiki/doku.php/mantisbt:docbook

core/email_api.php
```diff
@@ -1303,7 +1303,7 @@ function email_send( EmailData $p_email_data ) {
 	#apply DKIM settings
 	if( config_get( 'dkim_enable' ) ) {
 		$t_mail->DKIM_domain = config_get( 'email_dkim_domain' );
-		$t_mail->DKIM_private = config_get( 'email_dkim_private_key' );
+		$t_mail->DKIM_private = config_get( '$g_email_dkim_private_key_file_path' );
```
atrol: remove the `$g_`

atrol:
You added the documentation after `$g_smtp_port` but before `$g_smtp_connection_mode`, thus splitting the smtp settings.
The settings should be added after `$g_smtp_password`.

After applying this change and squashing the commits, the PR is fine from my side.
@vboctor might want some more changes.
Thanks for your contribution.

config_defaults_inc.php
```diff
@@ -680,6 +680,82 @@
...
+ * @global string $g_email_dkim_domain
+ */
+$g_email_dkim_domain = 'example.com';
+
+/**
+ * DomainKeys Identified Mail (DKIM) Signatures private key path
+ * Path to the private key. If $g_email_dkim_private_key_string is specified
+ * this setting will not be used.
+ * @see $g_email_dkim_private_key_string
+ * @see $g_dkim_enable
```
vboctor: Wrong config option name.

config_defaults_inc.php
```diff
@@ -680,6 +680,82 @@
...
+ */
+$g_email_dkim_private_key_file_path = '/path/to/readable/example.com/private_key.pem';
+
+
+/**
+ * DomainKeys Identified Mail (DKIM) Signatures private key value
+ * This string should contain private key for signing. Leave empty
+ * string if you wish to load the key from the file defined with
+ * $g_email_dkim_private_key_file_path.
+ * @see $g_dkim_enable
```
vboctor: Wrong config option name.

core/email_api.php
```diff
@@ -1300,6 +1300,16 @@ function email_send( EmailData $p_email_data ) {
 			break;
 	}
 
+	#apply DKIM settings
+	if( config_get( 'dkim_enable' ) ) {
+		$t_mail->DKIM_domain = config_get( 'email_dkim_domain' );
+		$t_mail->DKIM_private = config_get( 'email_dkim_private_key_file_path' );
```
vboctor: I suggest only setting the file path if the key string is not set.

config_defaults_inc.php
```diff
@@ -680,6 +680,82 @@
...
+
+/**
+ * DomainKeys Identified Mail (DKIM) Signatures private key path
+ * Path to the private key. If $g_email_dkim_private_key_string is specified
+ * this setting will not be used.
+ * @see $g_email_dkim_private_key_string
+ * @see $g_dkim_enable
+ * @global string $g_email_dkim_private_key_file_path
+ */
+$g_email_dkim_private_key_file_path = '/path/to/readable/example.com/private_key.pem';
```
vboctor: Specially now that we have a file path or a key string, let's default the file path to an empty string.

config_defaults_inc.php
atrol: It's a `@see` tag

core/email_api.php
yaddran: Why adding extra code when this is already handled by PHPMailer. We have specified in the documentation and comments the order the settings are applied in.

yaddran:
Hm... trouble with squash.... :(

atrol:
> Hm... trouble with squash.... :(
> How do I actually do this... ?

```
Use git rebase -i <after-this-commit> and replace "pick" on the second and subsequent commits with "squash" or "fixup
```
https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

yaddran:
hm... exactly what I did but then when I tried to push it complained that I need to pull first and them merge came up and then it came up as a new commit... :(
I should probably do git push --force after ?

atrol:
> I should probably do git push --force after ?

Right

atrol:
Closed in favor of https://github.com/mantisbt/mantisbt/pull/1213

PR #1213, opened 2017-10-18
Feature #23493 DomainKeys Identified Mail (DKIM) Signatures

yaddran:
Sorry but the previous PR gave me too much headache and I could not squash the commits to one. You can ignore PR 1211.

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -680,6 +680,82 @@ $g_smtp_connection_mode = '';
 $g_smtp_port = 25;
 
 /**
+ * Enable DomainKeys Identified Mail (DKIM) Signatures (rfc6376)
+ * To successfully sign mails you need to enable DKIM and provide at least:
+ * - DKIM domain
+ * - DKIM private key or key file path
+ * - DKIM selector
+ * - DKIM identity
+ * @see $g_email_dkim_domain
+ * @see $g_email_dkim_private_key_file_path
+ * @see $g_email_dkim_private_key_string
+ * @see $g_email_dkim_selector
+ * @see $g_email_dkim_identity
+ * @global integer $g_email_dkim_enable
+ */
+$g_email_dkim_enable = OFF;
+
+/**
+ * DomainKeys Identified Mail (DKIM) Signatures domain
+ * This is usually the same as the domain of your from email
+ * @see $g_from_email
+ * @see $g_email_dkim_enable
+ * @global string $g_email_dkim_domain
+ */
+$g_email_dkim_domain = 'example.com';
+
+/**
+ * DomainKeys Identified Mail (DKIM) Signatures private key path
+ * Path to the private key. If $g_email_dkim_private_key_string is specified
+ * this setting will not be used.
+ * @see $g_email_dkim_private_key_string
+ * @see $g_email_dkim_enable
+ * @global string $g_email_dkim_private_key_file_path
+ */
+$g_email_dkim_private_key_file_path = '';
+
+
+/**
+ * DomainKeys Identified Mail (DKIM) Signatures private key value
+ * This string should contain private key for signing. Leave empty
+ * string if you wish to load the key from the file defined with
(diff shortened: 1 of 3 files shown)


yaddran:
Hi, 2 days no reaction. Can this be accepted?

config_defaults_inc.php
```diff
@@ -680,6 +680,82 @@
...
+ 
+/**
+ * DomainKeys Identified Mail (DKIM) Signatures identity
+ * Identity you are signing the mails with (rfc6376)
+ * This is usually the same as your from email
+ * @see $g_from_email
+ * @see $g_email_dkim_enable
+ * @global string $g_email_dkim_identity
+ */
+$g_email_dkim_identity = 'noreply@example.com';
```
vboctor: Do we need this or should this `$g_from_email` be the one used to sign for the signing to match the from email address?

config_defaults_inc.php
```diff
@@ -680,6 +680,82 @@
...
+ * - DKIM selector
+ * - DKIM identity
+ * @see $g_email_dkim_domain
+ * @see $g_email_dkim_private_key_file_path
+ * @see $g_email_dkim_private_key_string
+ * @see $g_email_dkim_selector
+ * @see $g_email_dkim_identity
+ * @global integer $g_email_dkim_enable
+ */
+$g_email_dkim_enable = OFF;
```
vboctor: This config option should be added to `$g_public_config_names`.

config_defaults_inc.php
```diff
@@ -680,6 +680,82 @@
...
+
+/**
+ * DomainKeys Identified Mail (DKIM) Signatures domain
+ * This is usually the same as the domain of your from email
+ * @see $g_from_email
+ * @see $g_email_dkim_enable
+ * @global string $g_email_dkim_domain
+ */
+$g_email_dkim_domain = 'example.com';
```
vboctor: This config option should be added to `$g_public_config_names`.

config_defaults_inc.php
```diff
@@ -680,6 +680,82 @@
...
+/**
+ * DomainKeys Identified Mail (DKIM) Signatures selector
+ * DNS selector for the signature (rfc6376)
+ * DNS TXT field should have for instance:
+ *   host mail.example._domainkey
+ *   value v=DKIM1; t=s; n=core; k=rsa; p=[public key]
+ * @see $g_email_dkim_enable
+ * @global string $g_email_dkim_selector
+ */
+$g_email_dkim_selector = 'mail.example';
```
vboctor: This config option should be added to `$g_public_config_names`.

config_defaults_inc.php
vboctor: If we need this as a separate config, then it should be added to `$g_public_config_names`.

vboctor:
Please revise commit message to match the following.  If you have issues with squash and force push, I can apply this change, just address the other changes.

```
Support signing email via DKIM

Support DomainKeys Identified Mail (DKIM) to sign emails for improved
deliverability and trust.

Fixes #23493
```

config_defaults_inc.php
yaddran: Yes, it is needed. Here is the spec, http://dkim.org/specs/rfc4871-dkimbase.html#rfc.section.1.1 . It clearly specifies that the signer is not required to be one of the headers of the mail. Signer, in general, can be other identity and then it is up to the receiving party to decide about the action. However, to void those kind of decisions it is good practice not to use a different one.

yaddran:
Amended the last commit with the changes to $g_public_config_names and the commit message.

config_defaults_inc.php
vboctor: How about we allow this to be empty and default to from_email or set its default value to `'%from_email%'`?

yaddran: Maybe, but to be honest like this the dkim domain, selector and identity are all working together to show the needed setup... if we would remove this value we would loose this connection.

PR #1215, opened 2017-10-20
Allow password reset when it is blank

dregad:
Commit f4137abe0c6e28ffd5a82fbe455b40d598109a25 added HTML5 'required'
attribute to mandatory fields, including the Current Password on Edit
Account page.

Since the password can be blank (e.g. when $g_enable_email_notification
is OFF), this prevents the users from resetting their password.

This removes the mandatory attribute when the password is blank.

Fixes [#23507](https://mantisbt.org/bugs/view.php?id=23507)

### DIFF
--- a/account_page.php
+++ b/account_page.php
@@ -181,13 +181,16 @@ print_account_menu( 'account_page.php' );
 			if( $t_account_verification ) {
 				token_set( TOKEN_ACCOUNT_VERIFY, true, TOKEN_EXPIRY_AUTHENTICATED, $u_id );
 			} else {
+				# If the current password is blank, do not require the field
+				# so the user can reset the password (#23507)
+				$t_required = auth_does_password_match( $u_id, '' ) ? '' : 'required';
 			?>
 			<tr>
 				<td class="category">
-					<span class="required"><?php if( $t_force_pw_reset ) { ?> * <?php } ?></span> <?php echo lang_get( 'current_password' ) ?>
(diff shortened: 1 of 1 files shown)


account_page.php
```diff
@@ -181,13 +181,16 @@
...
 				token_set( TOKEN_ACCOUNT_VERIFY, true, TOKEN_EXPIRY_AUTHENTICATED, $u_id );
 			} else {
+				# If the current password is blank, do not require the field
+				# so the user can reset the password (#23507)
+				$t_required = auth_does_password_match( $u_id, '' ) ? '' : 'required';
 			?>
 			<tr>
 				<td class="category">
-					<span class="required"><?php if( $t_force_pw_reset ) { ?> * <?php } ?></span> <?php echo lang_get( 'current_password' ) ?>
+					<span class="<?php echo $t_required ?>"><?php if( $t_force_pw_reset ) { ?> * <?php } ?></span> <?php echo lang_get( 'current_password' ) ?>
```
vboctor: I would rather not generate `class=""` when field is not required.  Not sure if this would upset some validator.  Maybe also create `$t_required_class` and use it here.

dregad: `class=""` is [valid HTML](https://www.w3.org/TR/html5/dom.html#classes) so I really don't see the point of adding code complexity, or overhead of an additional variable.

dregad:
>  Should it be targeted to master-2.7

It is a regression indeed, I'll change the target.

PR #1216, opened 2017-10-21
Remove required attribute from check box custom fields

atrol:
Fixes #23512


atrol:
> Should this be targeted to master-2.7 since it is a regression?

Done, closed via https://github.com/mantisbt/mantisbt/commit/2263c7cf306465ba72211878bec77db30131e96c

PR #1220, opened 2017-10-23
New db query class, filter query rewrite and fixes

cproensa:
This PR has several layers of rewrite/refactor/improvements

Introduces a new object based query syntax, a filter query object for bug filtering, and fixes in the filters functionality based on these.

**DbQuery class:**
A self contained object for database queries. This class encloses the sql syntax, parameter substitution and utility functions to build, execute and fetch data.
- By managing internally the parameter syntax and value bindings, in a database independent way, solves the quirks and issues currently existing with db_param() approach.
- Wide options for parameter syntax, allowing named parameters ":label" within the query text, or anonymous parameters with a `param( value )` method
- Support for binding arrays, which will expand automatically to values lists. Eg: for `IN (...)` clauses.
- Support for binding another query object, as subquery composition.
- Supports rebinding of parameters for query reuse. It's not a native "prepare" functionality, but saves on user code.
- Revised utility constructs: `LIKE`, `ILIKE`, `IN` clauses
- Utility functions to fetch and loop on query results. Less boilerplate code for fetching rows.
- Managed IN clauses that transparently solves the oracle issue with >1000 elements
- It's backwards compatible with existing code, with an entry point compatible with current db_query() calls.

**BugFilterQuery class:**
A self contained, object based filter query building tool. This class inherits from DbQuery, so it's essentially a query object ready to be executed and fetched upon creation. It adds all the logic that build the final sql clauses to get the filtered bugs. The building logic was moved from filter_api, but revised with some fixes and improvements.
This approach for filter queries allows a cleaner integration of filter queries into some existing code, which previously was very windy, now is a seamless query composition. See for example the refactoring of `tag_stats_related()`, or `history_query_result()`. 
This is the foundation for extending filters usage as a key building tool for more functionality, eg: implement filtering in summary graphs, also delegating some complex permission checks by using a filter (viewable bugs for public/private projects, limit_reporter options, etc)

Some specific fixes:
- Correctly apply permissions for public/private notes visibility #23499 #23500
- Correctly apply permissions for "view handler" configurations #23502
- Correctly apply permissions for "view monitoring user" configurations #23504
- Allow filtering by "monitoring user = none" #23501
- Allow filtering by "category = none" #13177
- Sorting of some fields based on names, instead of ids #10853
- Improved relationship filtering #7264 #21867

Note that replacing the core db_query() may seem risky. It is in some way. We could keep the new class to be used only for bug filters and new code, but making full use of it would stress it faster and ensure stability asap.


dregad:
I haven't had time to look into details or test, but this seems to be promising, nice work @cproensa

vboctor:
Good work @cproensa - this is looks like it improves over what we have.

I was hoping that we adopt a library for building models easily, e.g. ActiveRecord pattern.  This way an Issue entity would be a model with its business logic and its advanced queries built in.  This gives us the benefit of a powerful library rather than trying to do the same on our own and defining our own syntax and semantics.  Thoughts?

cproensa:
> I was hoping that we adopt a library for building models easily, e.g. ActiveRecord pattern. This way an Issue entity would be a model with its business logic and its advanced queries built in. This gives us the benefit of a powerful library rather than trying to do the same on our own and defining our own syntax and semantics. 

Yes, that would be a long term goal for a major rewrite
However, the active record methodology would cover simple entities, and the need for arbitrary sql queries will remain. Filters are the main example.
At that point, having the tool to provide a common ground for all supported databases is still needed.

vboctor:
@cproensa I did some looking around and there seems to be three levels involved in data access layers:

1. PDO
2. DBAL (DB Access Layer)
3. Entity Mapper or ActiveRecord (depending on the library)

Currently we just have 2 via ADODB.  I was wondering if we should consider an option like [Doctrine](http://www.doctrine-project.org/).  We can start with leveraging its DBAL which builds on PDO.  It could be used side by side with ADODB until we gradually transition fully to it.

We can also gradually start adopting layer 3.

This was mainly in response to having more special parsing and workarounds in the query language.

cproensa:
> Currently we just have 2 via ADODB. I was wondering if we should consider an option like Doctrine. We can start with leveraging its DBAL which builds on PDO. It could be used side by side with ADODB until we gradually transition fully to it.

We use AdoDB for 2 things:
- DB queries, general DML
- Schema updates, general DDL

With this PR, you could plug in Doctrine DBAL to the query class and remove all adodb for DML usage in mantis.
DML is the easiest to replace. Schema DDL is more complex, so we should make sure that any candidate library can solve satisfactory our schema installation/upgrades.

vboctor:
> DML is the easiest to replace. Schema DDL is more complex, so we should make sure that any candidate library can solve satisfactory our schema installation/upgrades.

My priority is to use a better DBAL and models for better performance and programmability.  I'm OK with our current schema management and upgrades.  So I'm okay with keeping it as is or moving to the new library.  Do you have thoughts on a library to use?  Doctrine and Propel seemed like good candidates in my research, with Doctrine on top.  But I haven't tried coding with any of them.

I would like to use the query approach from the new library, rather than introducing a DbQuery.  You can still have a filtering logic that uses the query processor to apply a filter.

cproensa:
> My priority is to use a better DBAL and models for better performance and programmability

Is that an actual priority now? If not, let's not misconduct the conversation from this PR.

>  I'm OK with our current schema management and upgrades.

IIRC most of the current issues we have currently with AdoDb are related to schema & upgrades.
Plain query execution is the part that works better atm.

> I would like to use the query approach from the new library, rather than introducing a DbQuery

Is any other library, eg Doctrine, complete for providing compatibility at gereral SQL senteces, across multiple databases? I haven't looked it in detail, but looking at Doctrine, the general sql execution method is very much like current AdoDB "execute". We still have to sanitize and adapt syntax for the differences in Mysql, MSsql, postgres, and oracle. Then, there would be no practical usage of native library functionality directly, so if we are still in need to manage the syntax adaptation, out current db_api logic layer, the functionality in this PR, or any other approach is still needed. 

Or are you proposing using the sql composition syntax, eg in Doctrine, where the query is created with ->where() ->join()  etc methods, without using the raw sql?
If so, that implies a full rewrite of all queries AND breaking all compatibility.


IS ther any more specific feedback for this PR? @dregad, @atrol
I have more itereations of changes waiting for current PRs.

vboctor:
@cproensa 

> Is that an actual priority now? If not, let's not misconduct the conversation from this PR.

I was referring to accessing db as priority over managing schema.  Our current approach for both is not great, but the db access has more access on the majority of the code base than the way we create and update the schema.

If we are to change the way we access the database, I would rather than do a single step to move to a more modern model that leverages an established framework that reduces our boiler plate code rather than doing more code that we have to maintain ourselves.  Also have one transition for the MantisBT core and plugins rather than having multiple transitions.

> IIRC most of the current issues we have currently with AdoDb are related to schema & upgrades.
Plain query execution is the part that works better atm.

Then we should take this into consideration when choosing our new approach.

> Or are you proposing using the sql composition syntax, eg in Doctrine, where the query is created with ->where() ->join() etc methods, without using the raw sql?  If so, that implies a full rewrite of all queries AND breaking all compatibility.

Independent of the approach we use, I would like to transition to the new model gradually rather than a full transition in one version.  So we would start with new models, moving one entity at a time, etc.  But I would like to move to a model where the underlying framework handles the sanitization rather than using queries directly.

> IS ther any more specific feedback for this PR? @dregad, @atrol

My concern with this PR is high level rather than specific.  As I mentioned it disagree with the direction.  @dregad and @atrol may have feedback, but I think changing the way we interact with the DB is a major change that we should make sure we are heading the right direction.  This PR is one of those that would benefit from sharing an idea first before doing the full work.

cproensa:
>  I would like to transition to the new model gradually rather than a full transition in one version. So we would start with new models, moving one entity at a time, etc 

You are describing the ORM like approach, which is fine to use directly the functionality of the library of choice.
However, i doubt we can remove the need of doing raw queries (not object managed ORM stuff):
- complex queries: joins, counts, etc... AND more notably: filters.
- queries in plugins, which are outside of our control
- other abritrary queries wich may be needed, as probably not everything can be fitted in that entity-object model.

And more importantly, removing the compatibility with current syntax offers:
- Total break of BC for plugins
- Full rewrite of sql usage through the project


>  But I would like to move to a model where the underlying framework handles the sanitization rather than using queries directly.

I bet that those libraries does not handle full translation of plain sql code to all the different DB syntax variations. We do it within our current db api, and it's a very valuable way to ensure that our code, and external code from plugins, works across all supported databases.
Note: most of the issues we've suffered from db syntax are caused by the db_param implementation (more on this later).


>  I would rather than do a single step to move to a more modern model that leverages an established framework that reduces our boiler plate code rather than doing more code that we have to maintain ourselves

Let's suppose that the chosen framework offers the functionality to leverage DB syntax differences, etc. Then we could remove that layer of translation and rely on the framework instead. This is a fact whether we keep our current db api, or this PR proposal. Worse: probably our current db_param stuff will be equally flawed within the new framework, and would require either a full rewrite of all sql related code, of a translation layer for compatibility (again, it would become the same scenario that this PR solves)


> This PR is one of those that would benefit from sharing an idea first before doing the full work.

Actually, this is approximately inspired by all the discussions with Damien regarding our current db api, and ideas for a better approach.

These are the reasons for this PR:

1) Our db_param approach is inherently flawed. We have already made a lot of fixes to mend it, about: code building queries concurrently, inconsitent push/pull in adodb query execution, etc. It's been layers of tweaks, but still there are errors that can't be fixed due to its design.
The correct solution would have been to use a db_param object for each query, not a static, shared, object as it exists now. So, if going that route, why not have the query itself as a self contained object? The DbQuery object now contains the sql code, and it's related db_params.
Then, it makes sense to move all the db_api preprocessing logic within this class, even fix/improve the existing ones.
And, having the control of the class, improve the db_param functionality to allow flexible binding for: values, arrays, another dbquery object as subquery.

If you look at the methods, they loosely resemble the Doctrine sql query object.

Talking about other boilerplate code: This PR avoids writing a lot of code in the consumer side: no more db_param_push, db_fetch_row loop, etc


2) Another goal for this PR is to leverage filters as a better way to inspect arbitrary bug data. Currently we write complex queries that must account for user/project access, public/private, limit-reporters, and many other configurations related to access and visibility, and all that repeated across many places through the project. That is a risk for bugs and data leaks. We had, and still have, those kind of bugs.
Ideally, a query inspecting bugs should join with a filter to get the desired bug ids, and the filter code should manage all that logic.
The logic is written and mantained only in one place, the filter code.

The approach of this PR makes a the action of using a filter within any general query a straightforward task. No additional code other than defining a filter and binding it as a subquery. This alone is worth of the dbquery layer proposed in this PR.
You can see an example: 70e08a068c4d61d6d5b88d2f86e8868ca941966f
Another immediate application for example will be adding filter capabilities to the summary pages.


3) With the task of refactoring the filter query logic in the BugFilterQuery, a very much needed clean up, and some bug fixes, of that code is being made.


Could I have made (3) with current state of code? yes; (1) no; (2) maybe but not as cleanly.



Back to Victor vision, take this in mind:
_This DbQuery does not need to be the mandatory interface to access DB._
Assuming the benefits of this PR as explained are understood, this could exists as one way to build queries that provides our current level of DBRMS interoperabily, easier for complex queries, leveragin filters, etc
Other ways may be allowed, like those provided by the underlying framework, either with a managed query builder, object-entity mapping, etc.
Probably simple queries could be fed directly to the raw framework. 
However any of those native usages should accommodate to the specific usage, or ensure the DB interoperability of mantis.


With this PR I have replaced only 5 queries so far, which are using filter joins. Nothing else breaks in the application, no need to rewrite anyting else unless required.
Specifically, i would replace additional queries for currently tricky actions like:
- LIKE queries, which does not account correctly for upper/lower case, escape characters...
- queries with risk of surpassing oracle limit for 1000 "IN" items
- and just for cleanliness, some queries that add a lot of boilerplate code just to build/bind arrays, etc

dregad:
Hi all,

I'm a bit late to the party, and don't have time for a lengthy reply so I'll just say that I fully support @cproensa's approach here; it is a topic that I have discussed with him on several occasions.

I agree with @vboctor that in an ideal world all of that would be handled by the DB abstraction layer. I even considered implementing this query object concept in ADOdb, but gave up on the idea in the short term due to lack of time. 

Considering that this is an optional improvement that does not introduce backwards-incompatible changes that would force us to update all of the code base, and that the new layer could be alter adapted to a new DBAL if necessary, I don't see any reasons not to merge this.

atrol:
> any objections ?

I didn't have a look at the code until now and was just able to invest a few minutes for cursory testing.
After getting a issue when using an old stored filter, I stopped, but had no time for deeper analysis.

I don't want to block the merge, but having just about two weeks to fix regressions might be a bit too much of risk.
We should merge such kind of PRs not later than 1-2 days after we published a new release and run it on www.mantisbt.org/bugs.

While we are on it, we should also consider this PR to be merged at the same time https://github.com/mantisbt/mantisbt/pull/1091

In this special case with lot of filter related changes, we should be sure that @cproensa is available to fix regressions we might see.
@cproensa would be good to know when there will be a good moment for you.

cproensa:
> After getting a issue when using an old stored filter

let me know the details to have a look
(also, it may be more stable after  #1091?)

> @cproensa would be good to know when there will be a good moment for you

i give priority to these, so if any bug is found i'm ready to attend it

core/classes/BugFilterQuery.class.php
```diff
@@ -0,0 +1,1679 @@
...
+	}
+
+	/**
+	 * Utility function to process the values for a filter property that is related
+	 * to a database id number, used to represent users.
+	 * Manages special case meta-value-none, wich is translated to id "0", to be able
+	 * to match database values where "0" is the default for empty/none
+	 * Manages special case meta-value-myself, by replacing with the actual current user id.
+	 *
+	 * @param array $p_array	Input array with user ids
```
atrol: $p_users_array instead of $p_array

core/classes/DbQuery.class.php
```diff
@@ -0,0 +1,956 @@
...
+
+	/**
+	 * Compatibility method to support execution of legacy query syntax through db_query(...)
+	 * @param string $p_query		Query string
+	 * @param array $p_arr_parms	Values array for parameters
+	 * @param integer $p_limit		Query limit
+	 * @param integer $p_offset		Query offset
+	 * @return IteratorAggregate|boolean ADOdb result set or false if the query failed
+	 */
+	public static function compat_db_query( $p_query, array $p_arr_parms = null, $p_limit = -1, $p_offset = -1, $p_pop_param = true ) {
```
atrol: Missing PHPdoc for $p_pop_param

core/classes/DbQuery.class.php
```diff
@@ -0,0 +1,956 @@
...
+		if( !$this->db_result ) {
+			return false;
+		}
+		$this->current_row = db_fetch_array( $this->db_result );
+		return $this->current_row;
+	}
+
+	/**
+	 * Returns all rows as an array
+	 * @return array	Array with all rows from the result
```
atrol: array|boolean

core/classes/DbQuery.class.php
```diff
@@ -0,0 +1,956 @@
...
+				$t_value = false;
+			}
+		}
+		return $t_value;
+	}
+
+	/**
+	 * Alias for value()
+	 * @param integer|string $p_index_or_name	Column name or numeric index
+	 */
```
atrol: Missing return PHPdoc

core/filter_api.php
```diff
@@ -1048,8 +1059,12 @@ function filter_get_field( $p_filter_id, $p_field_name ) {
  * @param boolean $p_show_sticky   Whether to show sticky items.
  * @param array   $p_query_clauses Array of query clauses.
  * @return array $p_query_clauses
+ *
+ * @deprecated	Use BugFilterQuery class
```
atrol: We should not deprecate and generate trigger DEPRECATED errors as long as we use this function ourselves in function filter_get_bug_rows_query_clauses

core/filter_api.php
```diff
@@ -1205,8 +1220,13 @@ function filter_get_query_sort_data( array &$p_filter, $p_show_sticky, array $p_
  * array_unique here could cause problems with the query.
  * @param array $p_query_clauses Array of query clauses.
  * @return array
+ *
+ * @deprecated	Use BugFilterQuery class
```
atrol: Same as above, don't deprecate the function but go on using it.

core/filter_api.php
atrol: Hmm, filter_get_bug_rows_query_clauses is also deprecated and is not used anymore.

I am wondering if we could remove these functions instead of deprecation them.

core/filter_form_api.php
```diff
@@ -1530,10 +1542,37 @@ function print_filter_values_relationship_type( array $p_filter ) {
 	echo '<input type="hidden" name="', FILTER_PROPERTY_RELATIONSHIP_BUG, '" value="', string_attribute( $t_filter[FILTER_PROPERTY_RELATIONSHIP_BUG] ), '" />';
 	$c_rel_type = $t_filter[FILTER_PROPERTY_RELATIONSHIP_TYPE];
 	$c_rel_bug = $t_filter[FILTER_PROPERTY_RELATIONSHIP_BUG];
-	if( -1 == $c_rel_type || 0 == $c_rel_bug ) {
-		echo lang_get( 'any' );
+	if( BUG_REL_ANY == $c_rel_type ) {
+		switch ( $c_rel_bug ) {
+			case META_FILTER_NONE:
+				echo lang_get( 'none' );
```
atrol: missing break?
If not, add a comment `# Fall through`

core/filter_api.php
cproensa: > I am wondering if we could remove these functions instead of deprecating them.

I have left them in, as an example: my own use case, i have plugins that use them.... so it's possible that others do it too.

Rewritting them, as envelopes for the new functionality, is not factible either (i think). Note that they work around queries and db_params, which have side effects in caller context. The caller may also have specific code to manage those side effects (Eg, db_param stack and query execution).

One thing which may be cleaner, is to move the deprecated code outside of filter_api?

cproensa: > I am wondering if we could remove these functions instead of deprecating them.

I have left them in, as an example: my own use case, i have plugins that use them.... so it's possible that others do it too.

Rewritting them, as envelopes for the new functionality, is not factible either (i think). Note that they work around queries and db_params, which have side effects in caller context. The caller may also have specific code to manage those side effects (Eg, db_param stack and query execution).

One thing which may be cleaner, is to move the deprecated code outside of filter_api?

core/filter_form_api.php
cproensa: done (missing break)

core/classes/DbQuery.class.php
cproensa: done

atrol:
> > After getting a issue when using an old stored filter
> 
> let me know the details to have a look
> (also, it may be more stable after  #1091?)

Forget it. What I found was not a regression, but also a bug in current version, see https://www.mantisbt.org/bugs/view.php?id=23996 for what I found.

PR #1221, opened 2017-10-24
Allow display of detailed errors from REST API

dregad:
If MantisBT config `show_detailed_errors` is on, then the Slim app
object will be initialized with setting `displayErrorDetails` = true.

Fixes [#23537](https://mantisbt.org/bugs/view.php?id=23537)

### DIFF
--- a/api/rest/index.php
+++ b/api/rest/index.php
@@ -42,7 +42,15 @@ require_once( $t_restcore_dir . 'VersionMiddleware.php' );
 # For example, this will disable logic like encoding dates with XSD meta-data.
 ApiObjectFactory::$soap = false;
 
-$g_app = new \Slim\App();
+
+# Show SLIM detailed errors according to Mantis settings
+$t_config = array();
+if( ON == config_get_global( 'show_detailed_errors' ) ) {
+	$t_config['settings'] = array( 'displayErrorDetails' => true );
+}
+$t_container = new \Slim\Container( $t_config );
+
+$g_app = new \Slim\App( $t_container );
 
 # Add middleware - executed in reverse order of appearing here.
 $g_app->add( new ApiEnabledMiddleware() );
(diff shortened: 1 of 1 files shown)


api/rest/index.php
```diff
@@ -42,7 +42,15 @@
 # For example, this will disable logic like encoding dates with XSD meta-data.
 ApiObjectFactory::$soap = false;
 
-$g_app = new \Slim\App();
+
+# Show SLIM detailed errors according to Mantis settings
+$t_config = [];
```
vboctor: Do we use `[]` instead of `array()` now?  Should we keep using array() until we drop PHP 5.x support in the future?

api/rest/index.php
```diff
@@ -42,7 +42,15 @@
...
+if( ON == config_get_global( 'show_detailed_errors' ) ) {
+	$t_config['settings'] = ['displayErrorDetails' => true];
+}
+$t_container = new \Slim\Container( $t_config );
+
+$g_app = new \Slim\App($t_container);
```
vboctor: Fix spacing around `()`.

api/rest/index.php
atrol: Starting  with version 2.7.0 we enforce PHP >= 5.5, see https://github.com/mantisbt/mantisbt/commit/380fde11d50c22ef94210822d60b5f576c9d4a34

So using the [] array syntax should be no problem

> As of PHP 5.4 you can also use the short array syntax, which replaces array() with [].

http://php.net/manual/en/language.types.array.php

dregad: What @atrol said. Probably need to revise coding guidelines though.
That being said, if you prefer that we stick to array() syntax, that's fine with me.

api/rest/index.php
dregad: Will fix

api/rest/index.php
vboctor: I would stick with the old syntax at least until we update coding guidelines and upgrade mantisbt.org server.

dregad: OK, changed.

PR #1222, opened 2017-10-25
Force PHP compatibility for composer and update PHPMailer

vboctor:
Fixes #23542 and #23543

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,5 +27,10 @@
             "email": "mantisbt-dev@lists.sourceforge.net",
             "homepage": "https://www.mantisbt.org"
         }
-    ]
+    ],
+    "config": {
+        "platform": {
+            "php": "5.5.0"
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


atrol:
I am ok with this in general but would like to be sure that you are aware that this is a bit more than
> Force PHP compatibility for composer and update PHPMailer

There are some more upgrades and one downgrade
```
  - Updating phpmailer/phpmailer (v5.2.24 => v5.2.25): Loading from cache
  - Updating phpdocumentor/reflection-common (1.0 => 1.0.1): Downloading (100%)
  - Updating phpdocumentor/reflection-docblock (3.2.1 => 3.2.2): Loading from cache
  - Updating phpspec/prophecy (v1.7.0 => v1.7.2): Downloading (100%)
  - Updating symfony/yaml (v3.3.6 => v2.8.28): Downloading (100%)
```

vboctor:
@atrol These are updates due to running `composer update` based on the new constraints.  The PHPMailer is the only direct dependency, the rest are dependencies of dependencies.

PR #1223, opened 2017-10-25
Fix unattended upgrade

vboctor:
Fixes #23544


dregad:
"Fix upgraded unattended" is not very descriptive as commit message - does not say what is being fixed or why. Suggest amending the commit message along the lines of issue 23544 description

PR #1224, opened 2017-10-27
Bugnote text area not styled correctly when private by default

vboctor:
When bugnotes are private by default, they are not styled with bugnote-private class, hence, not yellow.

Fixes #23555

There may be an opportunity for refactoring here.


atrol:
> There may be an opportunity for refactoring here.

What about using a small helper function for it?

```php
function html_bugnote_class ( $p_bugnote_view_status ) {
	if ( $p_bugnote_view_status == VS_PRIVATE ) {
		echo 'bugnote-private';
	}
}
```

dregad:
I like the idea of the helper function

vboctor:
I will consider the refactoring separately.

PR #1226, opened 2017-11-01
Remove unused code of unsupported database types

atrol:
Fixes #23572

### DIFF
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -221,8 +221,6 @@ function db_get_type( $p_driver_type ) {
 	switch( $p_driver_type ) {
 		case 'mysqli':
 			return DB_TYPE_MYSQL;
-		case 'postgres':
-		case 'postgres7':
 		case 'pgsql':
 			return DB_TYPE_PGSQL;
 		case 'mssqlnative':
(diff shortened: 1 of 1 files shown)


vboctor:
What are your thoughts on the option below vs. the one you implemented?  It would be useful to share here.

```
function db_type() {
    $t_db_type = config_get_global( 'db_type' );
    # TODO: map mysql to mysqli, postgres variants to pgsql, etc.
    return $t_mapped_db_type;
}
```

Then update all other methods to handle standard types by calling db_type() rather than getting the config option and attempting to handle all the name variants.

This would not cause a failure when a MantisBT instance is using any of the alias names, but it will continue to work and provide the right behavior.  The mapping is encapsulated in one function.

cproensa:
>    # TODO: map mysql to mysqli, postgres variants to pgsql, etc.
>    return $t_mapped_db_type;

I think that is currently done with db_get_type()
The idea is that the mapped constants DB_TYPE_MSSQL, etc, are to be used when dealing with the generic dialects/functionality of those databases.
We sometimes use directly the global `$g_db_functional_type`, to avoid function calls in fast fetch loops.

The exact driver mssqlnative, odbc_mssql, etc should be used when the connection driver has specific behaviour. For example, some fix has to be done when dealing with odbc_mssql, but not with mssqlnative.

atrol:
> This would not cause a failure when a MantisBT instance is using any of the alias names, but it will continue to work and provide the right behavior

I see no need to introduce a mapping.
We didn't support all alias names in function db_check_database_support in 1.2.x, 1.3.x and 2.x.x, so there should not be that much installations be affected, as they failed some time ago.
E.g. there was just `pgsql` in 1.2 but no `postgres`, ` postgres64`or `postgres7`
https://github.com/mantisbt/mantisbt/blob/45df5c0e9aa579b54b7ca390a74c9495ea4df311/core/database_api.php#L140

atrol:
@vboctor after latest comments, do you agree to merge as is?

vboctor:
@atrol I don't have a strong opinion here, so go ahead.

PR #1228, opened 2017-11-05
REST API Fixes

vboctor:
Following are the list of fixes, some apply to non-rest flows as well:

- [23576](https://mantisbt.org/bugs/view.php?id=23576) Fix email error when reporting issues with date field
- [23594](https://mantisbt.org/bugs/view.php?id=23594) Fix handling custom date fields with value `{now}`
- [23577](https://mantisbt.org/bugs/view.php?id=23577) Ensure required custom fields specified on create issue
- [23579](https://mantisbt.org/bugs/view.php?id=23579) Return 400 when category doesn't exist
- [23575](https://mantisbt.org/bugs/view.php?id=23575) Make REST API category lookup case insensitive


api/soap/mc_api.php
```diff
@@ -797,12 +797,14 @@ function mci_get_category( $p_category_id ) {
 }
 
 /**
- * Convert a category name to a category id for a given project
+ * Convert a category name, or category object reference (array w/ id, name,
+ * or id + name) to a category id for a given project.
+ *
```
dregad: Suggest to clarify in the PHPDoc block(s) the difference between *mci_get_category_id*  and the "internal" variant, when to use one or the other.

vboctor: Maybe I should change the `_internal` variant to an anonymous method with the regular one since no one else is expected to call the internal function.

dregad: That would work too, I guess. Your call.

dregad:
> 23594 Fix handling custom date fields with value {now}

See my note in https://mantisbt.org/bugs/view.php?id=23594#c58154 - amend commit message if necessary.

PR #1232, opened 2017-11-10
Support issue history in REST API

vboctor:
Getting an issue via REST API should return issue history information.

See [Sample](https://gist.github.com/vboctor/1f5f39c28539f752ad811d53a4131ff4) responses.

The issue history is pretty comprehensive, however, the revisions for text area fields like description, additional information, steps to reproduce and notes are not included.  However, the act of changing such fields is included in the returned history.

Fixes #23616

### DIFF
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -90,7 +90,243 @@ function mc_issue_get( $p_username, $p_password, $p_issue_id ) {
 }
 
 /**
-* Get history details about an issue.
+ * Get history details for an issue that is visible to specified user.
+ * This is used by REST APIs and assumes that user access to issue is
+ * already done.
+ *
+ * @param integer $p_issue_id The issue id.
+ * @param integer $p_user_id The user id.
+ * @param string $p_lang The user language.
+ * @return array history entries or empty if user has no access to history.
+ */
+function mci_issue_get_history( $p_issue_id, $p_user_id, $p_lang ) {
+	$t_project_id = bug_get_field( $p_issue_id, 'project_id' );
+
+	$t_view_history_threshold = config_get( 'view_history_threshold', null, null, $t_project_id );
+	if( !access_has_bug_level( $t_view_history_threshold, $p_issue_id, $p_user_id ) ) {
+		return array();
+	}
+
+	$t_history_rows = history_get_raw_events_array( $p_issue_id, $p_user_id );
+
+	$t_history = array();
+
+	foreach( $t_history_rows as $t_history_row ) {
+		$t_type = (int)$t_history_row['type'];
+
+		$t_skip = false;
+
+		switch( $t_type ) {
+			case BUG_ADD_SPONSORSHIP:         # Deprecated, not exposed in REST API
+			case BUG_UPDATE_SPONSORSHIP:      # Deprecated, not exposed in REST API
+			case BUG_DELETE_SPONSORSHIP:      # Deprecated, not exposed in REST API
+			case BUG_REVISION_DROPPED:        # Not Supported
+			case BUGNOTE_REVISION_DROPPED:    # Not Supported
+				$t_skip = true;
+				break;
+		}
+
+		$t_event = array();
+
+		$t_event['created_at'] = ApiObjectFactory::datetime( $t_history_row['date'] );
+		$t_event['user'] = mci_account_get_array_by_id( $t_history_row['userid'] );
+
+		$t_field = $t_history_row['field'];
+		if( !is_blank( $t_field ) ) {
+			# map field names to external names
+			switch( $t_field )  {
+				case 'reporter_id':
+					$t_field = 'reporter';
+					break;
+				case 'handler_id':
+					$t_field = 'handler';
+					break;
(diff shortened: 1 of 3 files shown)


api/soap/mc_issue_api.php
```diff
@@ -90,7 +90,243 @@ function mc_issue_get( $p_username, $p_password, $p_issue_id ) {
...
+
+		$t_skip = false;
+
+		switch( $t_type ) {
+			case BUG_ADD_SPONSORSHIP:         # Deprecated, not exposed in REST API
+			case BUG_UPDATE_SPONSORSHIP:      # Deprecated, not exposed in REST API
+			case BUG_DELETE_SPONSORSHIP:      # Deprecated, not exposed in REST API
+			case BUG_REVISION_DROPPED:        # Not Supported
+			case BUGNOTE_REVISION_DROPPED:    # Not Supported
+				$t_skip = true;
```
atrol: not needed

api/soap/mc_issue_api.php
```diff
@@ -90,7 +90,243 @@ function mc_issue_get( $p_username, $p_password, $p_issue_id ) {
...
+					break;
+				case 'view_state':
+					$t_value = mci_enum_get_array_by_id( (int)$p_value, 'view_state', $p_lang );
+					break;
+				case 'resolution':
+					$t_value = mci_enum_get_array_by_id( (int)$p_value, 'resolution', $p_lang );
+					break;
+				case 'reproducibility':
+					$t_value = mci_enum_get_array_by_id( (int)$p_value, 'reproducibility', $p_lang );
+					break;
```
atrol: What about
```php
				case 'status':
				case 'priority':
				case 'severity':
				case 'view_state':
				case 'resolution':
				case 'reproducibility':
					$t_value = mci_enum_get_array_by_id( (int)$p_value, $p_field, $p_lang );
					break;
```

core/history_api.php
```diff
@@ -510,6 +510,203 @@ function history_get_raw_events_array( $p_bug_id, $p_user_id = null, $p_start_ti
...
+		case 'duplicate_id':
+			$t_field_localized = lang_get( 'duplicate_id' );
+			break;
+		case 'sponsorship_total':
+			$t_field_localized = lang_get( 'sponsorship_total' );
+			break;
+		case 'due_date':
+			$t_field_localized = lang_get( 'due_date' );
+			break;
+		default:
```
atrol: What about
```php
		case 'category':
		case 'status':
		case 'severity':
		case 'reproducibility':
		case 'resolution':
		case 'priority':
		case 'eta':
		case 'projection':
		case 'fixed_in_version':
		case 'target_version':
		case 'date_submitted':
		case 'os':
		case 'build':
		case 'platform':
		case 'summary':
		case 'duplicate_id':
		case 'sponsorship_total':
		case 'due_date':
			$t_field_localized = lang_get( $p_field_name );
			break;
		case 'view_state':
			$t_field_localized = lang_get( 'view_status' );
			break;
		case 'sticky':
			$t_field_localized = lang_get( 'sticky_issue' );
			break;
		case 'project_id':
			$t_field_localized = lang_get( 'email_project' );
			break;
		case 'handler_id':
			$t_field_localized = lang_get( 'assigned_to' );
			break;
		case 'reporter_id':
			$t_field_localized = lang_get( 'reporter' );
			break;
		case 'version':
			$t_field_localized = lang_get( 'product_version' );
			break;
		case 'last_updated':
			$t_field_localized = lang_get( 'last_update' );
			break;
		case 'os_build':
			$t_field_localized = lang_get( 'os_version' );
			break;
```

api/soap/mc_issue_api.php
vboctor: I would rather keep it as is.  The idea is that we should be moving to constants for DB fields and constants for enums and in such case having such shortcuts will have to be re-worked then.

core/history_api.php
vboctor: Same comment.  This current code is more friendly to moving to constants.

atrol: I thought that having the field names as some kind of unique index at various places is quite good, e.g.

- no need for additional constants
- adding new fields is easy, no need to add code at various places
- short and simple code
- better code coverage, even when doing just a few tests

After some refactoring by eliminating historic inconsistencies (e.g. `view_state` vs `view_status`) there would be some more advantages

- complete removal of the switch cases
- better performance, especially as PHP is quite bad in terms of how `switch` statements are executed internally

api/soap/mc_issue_api.php
vboctor: For some reason `continue;` doesn't work as expected, to make it clear I replaced `continue` with `break` and kept the setting of skip.

core/history_api.php
vboctor: I'm not sure I understand the proposal here, but I agree we need some cleanup and we should create a code cleanup issue to discuss.  It may be that you are proposing having a structure like cfdefs for native fields that maps a field to all associated constants and behaviors.  In fact, I think it would be useful to see if we can have a common structure for all fields that is common cross for native and custom fields.

PR #1233, opened 2017-11-12
Simplify require lib

atrol:


core/database_api.php
```diff
@@ -37,8 +37,6 @@
 require_api( 'logging_api.php' );
 require_api( 'utility_api.php' );
 
-require_lib( 'adodb/adodb-php/adodb.inc.php' );
```
vboctor: Can you check why this was added from git history?  It may have been added for some upgrade or install path.

core.php
```diff
@@ -124,15 +124,8 @@ function require_lib( $p_library_name ) {
...
-			global $g_vendor_path;
-			$t_library_file_path = $g_vendor_path . $p_library_name;
-
-			if( file_exists( $t_library_file_path ) ) {
-				require_once( $t_library_file_path );
-			} else {
-				echo 'External library \'' . $t_library_file_path . '\' not found.';
-				exit;
-			}
+			echo 'External library \'' . $t_library_file_path . '\' not found.';
```
vboctor: Is there a reason for not triggering an error here?

dregad: It's been a while since I looked at that code, but IIRC this is because it is possible that not all of the API are loaded at that point.

core/database_api.php
dregad: @vboctor That's just how you implemented the migration from bundled libs to Composer afea93e8c0019f99814b295ff5d23726dac7aa5d

core.php
atrol: @vboctor you introduced it the way it is, see https://github.com/mantisbt/mantisbt/commit/b30a3124f26b4937932071add3c21e6396859485
I don't know if there was a reason to use echo+exit.
Any change concerning the error handling shouldn't be part of this PR.

core/database_api.php
atrol: >  I will try a new install and uprade with it.

Tried without getting any problem.
@vboctor do you agree to merge?

PR #1235, opened 2017-11-13
Clickable summaries in view issues page

scfido:
In view all bug page, add a link to the summary column

### DIFF
--- a/core/columns_api.php
+++ b/core/columns_api.php
@@ -1505,8 +1505,9 @@ function print_column_summary( BugData $p_bug, $p_columns_target = COLUMNS_TARGE
 	} else {
 		$t_summary = string_display_line_links( $p_bug->summary );
 	}
-
-	echo '<td class="column-summary">' . $t_summary . '</td>';
+	
+	$t_bug_url = string_get_bug_view_url( $p_bug->id );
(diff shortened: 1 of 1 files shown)


vboctor:
Thanks @scfido for your contribution.  This implemented the following [feature request](https://www.mantisbt.org/bugs/view.php?id=8066) in our bug tracker.

Please change the commit message to:

```
Clickable summaries in view issues page

Fixes #8066
```

syncguru:
I wonder if we should make the entire row clickable vs. just summary field. Since the issues table customizable, users may end up with a different configuration that would benefit from similar functionality but on a different field.

atrol:
>  wonder if we should make the entire row clickable vs. just summary field

What will happen with the columns that are providing links to other pages than view.php, e.g. `edit`, `reporter_id` and `handler_id`.

syncguru:
Good question, @atrol 
I don't know for sure. It is better to try it out and evaluate the experience.

dregad:
> I wonder if we should make the entire row clickable vs. just summary field

@syncguru Maybe, but I suggest you take this on as a separate work item.

dregad:
Just stumbled upon this, and realized that I approved it over 3 years ago and never actually merged it... sorry !

PR #1236, opened 2017-11-13
fixing #23630

cybernet:
"Allow Reporter to close Issue" tick box is not centered


cybernet:
this needs any more than 10 minutes to review ?

PR #1239, opened 2017-11-16
Update email_api.php

akoscomp:
Add option to set in config.php the e-mail format, by set $g_email_is_html to true, the sent e-mail fromat will be HTML.


atrol:
@akoscomp please have a look at https://github.com/mantisbt/mantisbt/pull/1120 for the reasons why we reject your PR.

PR #1242, opened 2017-11-18
Use `ETag` and `If-None-Match` for issue get, update, and delete

vboctor:
Fixes related to `ETag` and `If-None-Match`:

- [23648](https://www.mantisbt.org/bugs/view.php?id=23648): Leverage ETag headers when getting issues
- [23650](https://www.mantisbt.org/bugs/view.php?id=23650): Leverage If-None-Match when deleting issues
- [23653](https://www.mantisbt.org/bugs/view.php?id=23653): Leverage If-None-Match when updating issues

Unrelated fixes:

- [23654](https://www.mantisbt.org/bugs/view.php?id=23654): Don't validate handler when updating issues without updating handler


atrol:
Please explain why the hash depends on user id.

vboctor:
@atrol you can imagine a service or a client that cache such issues and reuses the cache when it has an entry with the same ETag in the cache. Now, if such service or client is used by multiple users then they may see data that was cached from a higher privedge user. The current implementation returns different hash based on possibility of issue details being different across users.

atrol:
Maybe you have to add also a session id, as the same user can use different clients.

vboctor:
> Maybe you have to add also a session id, as the same user can use different clients.

@atrol The ETags round trip between server and client/browser, and hence another client won't have it.  Also if a user is connected from two clients, they will see the same issue contents and hence they should have the same `ETag` value.  Think of the etag value as the digest of an issue as they user see it and that should be consistent across clients.

Also if we include a session id (as defined by server) then a client will have to re download all issues when a session expired on a server, because according to ETags, all issues must be fully re-downloaded.

cproensa:
What happens with changes that dont update last_updated timestamp?
Or custom fields with dynamic content?

vboctor:
@cproensa We will have to make sure all our changes update `last_updated` or some `sync_key` (to be added) associated with the issue. I would say most of issue changes today update `last_updated`, and we should fix those that don't.

cproensa:
We have some discussion in the bugtracker about updates that causes modifications on last update, which causes the issues rise on the view issue page, according to users the behaviour is not desirable. So this is related in some way.

About the issue of dynamic content, i implememt colums to show realtime SLA time data. Are column contents also cached?

vboctor:
If there is a custom field that servers dynamic content, then it is a new type of custom field that clients should be able to render based on static data on the issue.

If the updating of `last_updated` is not desirable, then it is an issue for two reasons:
1. The issue was in fact updated.
2. We would need a sync_key that doesn't affect order in our db schema which is updated whenever issues are updated.

We also need someway to capture user access changes, since change to user global / project level can change how they see an issue.  But we don't have this right now, so we handle access to new issues, but not changes on how they view an issue.  This can be resolved by an `access_key` that is changed whenever the user access changes.

We need to work towards a reliable way for our javascript code (built on REST API) or a client that caches issues to stay up-to-date with cached issues.

vboctor:
One safe approach is to always calculate the response for the API, then hash it and use that as the ETag.  This approach is more conservative because the server does the work anyways, but it saves the bytes over the wire.

Pros:
- Any change in output due to code, access change, entity change is taken into consideration.
- Doesn't have requirements on the rest of the code.

Cons:
- Server does the work to construct the full REST API result even if not needed.

I'm thinking we should switch to this model, and then we can revisit in the future if we have better capabilities to optimize further.

vboctor:
I have updated the issue get, update and delete to use the hash of the issue get output as the etag.

dregad:
Isn't it is semantically wrong to factor the user_id into the Etag hash, considering that it is supposed to identify *[a specific version of a resource](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag)* ? IMO, the resource here is the issue, the *state* of which is independent from the user's profile. 

> imagine a service or a client that cache such issues and reuses the cache when it has an entry with the same ETag in the cache. Now, if such service or client is used by multiple users then they may see data that was cached from a higher privedge user.

Shouldn't it be the caching service's responsibility to handle this scenario ?

> updates that causes modifications on last update, which causes the issues rise on the view issue page, according to users the behaviour is not desirable. 

This is unrelated to this PR's scope, but I believe the only way to properly and fully deal with all scenarios would be to maintain two *last updated dates*: 
- a "system" one that is always updated whenever the issue changes, and 
- a "user" one that only gets bumped in specific (and possibly user-configurable) types of updates.

vboctor:
TODO: For updating and deleting issues, I should use `If-Match` instead of `If-None-Match` as the header request.  Will update.

@dregad 

> Isn't it is semantically wrong to factor the user_id into the Etag hash, considering that it is supposed to identify a specific version of a resource ? IMO, the resource here is the issue, the state of which is independent from the user's profile.

User id is no longer factored in `ETag` calculation.  It is now calculated based on the response generated by get issue request.

Having said that, the same issue will have different `ETag` value for different users because the response is different based on the user for aspects like visibility of private notes, email addresses of reporter, handler, note authors, and other aspects.  The new approach will take this into consideration, but will generate the same `ETag` for users that have the same access level.

 > Shouldn't it be the caching service's responsibility to handle this scenario ?

Yes, in case of a smart service or separate plugins, that would be handled.  However, in case of a browser being the client, it just looks at URL + ETag to decide what to do.

> This is unrelated to this PR's scope, but I believe the only way to properly and fully deal with all scenarios would be to maintain two last updated dates

That is what I was leaning towards initially, but I decided that this will not be enough due changes that impact output of Get Issue (as on example) that are impacted by changes that are not related directly to the issue:

- A user changes their email address - this will change the email address returned for such user as they are referenced as reporter, handler, note author, or user who trigger a history change.

- A project, issue, or issue note changing from public to private or the other way around.

- System configuration (e.g. access level to view private notes, access level to view email address, etc).

- A user added or removed from a project with some access level.

- A code fix or enhancement that changes the response for getting issues.

As you can see there are aspects that change the response that are not tied to this specific issue.  Hence, I elected to use a hash of the response as the basis for `ETag` calculation.

Hope this clarifies the reasoning behind the latest approach I have taken with this pull request.

dregad:
> Hope this clarifies the reasoning behind the latest approach I have taken with this pull request.

It does indeed. Thanks for the detailed feedback !

PR #1243, opened 2017-11-20
Fix Manage Plugins UI for protected plugins

atrol:
After setting a plugin to `protected` there are some UI issues

- the `protected` checkbox is not editable, thus you can't unprotect
- clicking the `Update` button resets all `protected` plugins to unprotected
- clicking the `Update` button resets `priority`of all protected plugins to 3 

The PR fixes these issues and reverts to the behavior in earlier versions, thus setting to `protected` means nothing more than `Hide the Uninstall button`.

Fixes #23658

### DIFF
--- a/manage_plugin_page.php
+++ b/manage_plugin_page.php
@@ -185,13 +185,13 @@ foreach ( $t_plugins_installed as $t_basename => $t_plugin ) {
 	} else {
 		echo '<td class="center">',
 			'<select name="priority_' . $t_basename . '"',
-				check_disabled( $t_protected ), ' class="input-sm">',
+				' class="input-sm">',
 				print_plugin_priority_list( $t_priority ),
 			'</select>','</td>';
 		echo '<td class="center">',
 		'<label>',
 			'<input type="checkbox" class="ace" name="protected_' . $t_basename . '"',
-				check_disabled( $t_protected ), check_checked( $t_protected ), ' />',
+				check_checked( $t_protected ), ' />',
 		'<span class="lbl"></span>',
 		'</label>',
 			'</select>','</td>';


dregad:
@atrol I was actually in the middle of reviewing this when you merged it... 

For the record http://www.mantisbt.org/wiki/doku.php/mantisbt:development_process#reviewing_a_work_item

dregad:
Anyway here's my review comment...

I don't think admins should be able to change a plugin's priority when it is protected.

atrol:
> @atrol I was actually in the middle of reviewing this when you merged it...

Sorry, but I got the impression that you are quite busy the last weeks, so I didn't expect your feedback.
e.g. no comment for https://github.com/mantisbt/mantisbt/pull/1241

> For the record, http://www.mantisbt.org/wiki/doku.php/mantisbt:development_process#reviewing_a_work_item

I am aware of it, but seeing that other devs merge larger changes without any approval, I thought merging such a small change with one approval is acceptable.

> I don't think admins should be able to change a plugin's priority when it is protected.

I had a short look but did not find any documentation of what `protected` is exactly meant for.
The main purpose might be to protect from unintended uninstall.

I can't invest more time for this at the moment, so I would like to keep the change as it is or revert if you want.

dregad:
> I got the impression that you are quite busy the last weeks

I am quite busy indeed, but I do try to review changes and follow up on the tracker as much as I can. 

> seeing that other devs merge larger changes without any approval

Sorry I did not mean to single you out, it just happened that the PR got closed just when I was typing, so I was surprised... but of course my comment is valid for everyone.

> I would like to keep the change as it is or revert if you want.

Keep it as it is, the PR fixed a blatant bug, I was only suggesting an improvement.

PR #1245, opened 2017-11-28
Fix the call to `custom_field_set_value()`

vboctor:
Fixes #23594


vboctor:
@atrol This change is based on your [comment](https://www.mantisbt.org/bugs/view.php?id=23594#c58211).

Before this fix, creating an issue with 3 custom fields will result in issue creation history entry + 3 custom field value change history entries.  With this change, only the new issue history entry will be created.

PR #1246, opened 2017-11-28
Add config helper functions for bugs and projects

vboctor:
The goals are:

- Helper functions to make sure the appropriate scope for config is always used.
- Helper functions to make config retrieval and access check more compact.

I have added helper functions and demonstrated their use in a couple of scripts.

### DIFF
--- a/bug_report.php
+++ b/bug_report.php
@@ -98,7 +98,7 @@ if( $t_project_id != helper_get_current_project() ) {
 	$g_project_override = $t_project_id;
 }
 
-access_ensure_project_level( config_get( 'report_bug_threshold' ) );
+access_ensure_project_level( project_config( $t_project_id, 'report_bug_threshold' ) );
 
 if( isset( $_GET['posted'] ) && empty( $_FILE ) && empty( $_POST ) ) {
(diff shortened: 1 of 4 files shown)


atrol:
We should discuss config_get in general before making this change.
I started to prepare some thoughts, but I am quite busy at the moment.
I will add a note to https://www.mantisbt.org/bugs/view.php?id=23665 when I am done with it.

vboctor:
@dregad 

> In my opinion, the new functions belong in config API, not in project/bug, and should be renamed accordingly.

The reasoning is:

1. These new APIs use config and access APIs.  These APIs don't belong in access and config should really be a lower layer than access and bug/project APIs.  Hence, it felt that project/bug API is a good option for this.  We had similar cases in the past where an API gets a config and returns it.

2. Less important, this allows a shorter and more descriptive name, e.g. `bug_config()`, `bug_has_access()`, etc.

dregad:
It's not a huge deal and I could live with it if you insist on having these function in the bug/project API, but I still feel it's confusing not to have the xxx_config functions in config API, and xxx_has_access in access API.

PR #1248, opened 2017-11-30
Fix SOAP API standard filters

vboctor:
Fixes #23561


atrol:
Looks good, diidn't test.

PR #1250, opened 2017-12-04
Fixes for core Time Tracking

EthraZa:
- Add report checkbox to select sub-projects. Fix [9007](https://mantisbt.org/bugs/view.php?id=9007&nbn=7)
- Add noprint class to button and export links to hide it in print mode
- Add g_time_tracking_cost_per_hour config parameter to be able to set a default value
- Removed useless collapse icon with duplicated title


billing_inc.php
```diff
@@ -218,7 +227,7 @@
 	</tr>
 <?php
 		foreach ( $t_bugnote_stats['issues'] as $t_issue_id => $t_issue ) {
-			$t_project_info = ( !isset( $f_bug_id ) && $f_project_id == ALL_PROJECTS ) ? '[' . project_get_name( $t_issue['project_id'] ) . ']' . lang_get( 'word_separator' ) : '';
+			$t_project_info = ( (!isset( $f_bug_id ) && $f_project_id == ALL_PROJECTS) || $f_subprojects ) ? '[' . project_get_name( $t_issue['project_id'] ) . ']' . lang_get( 'word_separator' ) : '';
```
vboctor: I don't think this expression is correct.  I would expect:

    ( !isset( $f_bug_id ) && ( ( $f_project_id == ALL_PROJECTS ) || $f_subprojects )

config_defaults_inc.php
```diff
@@ -3969,6 +3969,12 @@
  */
 $g_time_tracking_without_note = ON;
 
+/**
+ * default bugnote cost per hour
```
vboctor: How about: "Default billing rate per hour"?

core/billing_api.php
```diff
@@ -53,10 +53,11 @@ function billing_ensure_reporting_access( $p_project_id = null, $p_user_id = nul
  * @param string  $p_from          Starting date (yyyy-mm-dd) inclusive, if blank, then ignored.
  * @param string  $p_to            Ending date (yyyy-mm-dd) inclusive, if blank, then ignored.
  * @param integer $p_cost_per_hour Cost per hour.
+ * @param integer $p_subprojects   Get project and its subprojects.
```
vboctor: boolean instead of integer?

core/billing_api.php
```diff
@@ -53,10 +53,11 @@ function billing_ensure_reporting_access( $p_project_id = null, $p_user_id = nul
...
  * @return array array of bugnotes
  * @access public
  */
-function billing_get_for_project( $p_project_id, $p_from, $p_to, $p_cost_per_hour ) {
+function billing_get_for_project( $p_project_id, $p_from, $p_to, $p_cost_per_hour, $p_subprojects = null ) {
```
vboctor: Why not use `false` as the default?

core/billing_api.php
```diff
@@ -133,8 +139,8 @@ function billing_get_for_project( $p_project_id, $p_from, $p_to, $p_cost_per_hou
  * @return array The contains billing data grouped by issues, users, and total information.
  * @access public
  */
-function billing_get_summaries( $p_project_id, $p_from, $p_to, $p_cost_per_hour ) {
-	$t_billing_notes = billing_get_for_project( $p_project_id, $p_from, $p_to, $p_cost_per_hour );
+function billing_get_summaries( $p_project_id, $p_from, $p_to, $p_cost_per_hour, $p_subprojects ) {
```
vboctor: The subprojects parameter is missing from phpdoc.  Also I suggest renaming this parameter in general from `subprojects` to `include_subprojects`.

docbook/Admin_Guide/en-US/config/timetracking.xml
```diff
@@ -34,6 +34,15 @@
 			</listitem>
 		</varlistentry>
 
+		<varlistentry>
+			<term>$g_time_tracking_cost_per_hour</term>
+			<listitem>
+				<para>
+					Default cost per hour - Default is 0
```
vboctor: Default billing rate per hour.

core/billing_api.php
```diff
@@ -69,10 +70,15 @@ function billing_get_for_project( $p_project_id, $p_from, $p_to, $p_cost_per_hou
 	db_param_push();
 
 	if( ALL_PROJECTS != $p_project_id ) {
-		access_ensure_project_level( config_get( 'view_bug_threshold' ), $p_project_id );
-
-		$t_project_where = ' AND b.project_id = ' . db_param() . ' AND bn.bug_id = b.id ';
-		$t_params[] = $p_project_id;
+		if( $p_subprojects ) {
+			$t_project_ids = user_get_all_accessible_projects(null, $p_project_id);
```
vboctor: While you are here, the implementation of this method checks access level for `VIEWER` rather than `view_bug_threshold` as below.  It would be good if you update this method to match the check in the else block.

billing_inc.php
```diff
@@ -181,22 +189,23 @@
...
 
-		echo '<br />';
+		echo '<br /><div class="noprint">';
 
 		$t_exports = array(
 			'csv_export' => 'billing_export_to_csv.php',
 			'excel_export' => 'billing_export_to_excel.php',
 		);
-
+
```
vboctor: Remove spaces.

billing_inc.php
```diff
@@ -149,10 +145,22 @@
 		</tr>
 <?php
 	}
+?>
+<?php
+	if( $f_project_id != ALL_PROJECTS ) {
+?>
+		<tr class="row-1">
+			<td>
+				<input type="checkbox" name="subprojects" <?php if($f_subprojects) echo 'checked'; ?>>
```
vboctor: Here is an example of how we add the checked flag:

    <input type="checkbox" class="ace" id="bugnote_add_view_status" name="private" <?php check_checked( $t_default_bugnote_view_status, VS_PRIVATE ); ?> />

billing_inc.php
```diff
@@ -79,7 +79,9 @@
 $f_get_bugnote_stats_button = gpc_get_string( 'get_bugnote_stats_button', '' );
 
 # Retrieve the cost as a string and convert to floating point
-$f_bugnote_cost = floatval( gpc_get_string( 'bugnote_cost', '' ) );
+$f_bugnote_cost = floatval( gpc_get_string( 'bugnote_cost', config_get( 'time_tracking_cost_per_hour' ) ) );
+
+$f_subprojects = gpc_get_string( 'subprojects', '' );
```
vboctor: Shouldn't this be a boolean instead of a string?

billing_export_to_excel.php
```diff
@@ -42,12 +42,13 @@
 $f_cost = gpc_get_int( 'cost' );
 $f_from = gpc_get_string( 'from' );
 $f_to = gpc_get_string( 'to' );
+$f_subprojects = gpc_get_string( 'subprojects' );
```
vboctor: boolean?

billing_export_to_csv.php
```diff
@@ -41,6 +41,7 @@
 $f_cost = gpc_get_int( 'cost' );
 $f_from = gpc_get_string( 'from' );
 $f_to = gpc_get_string( 'to' );
+$f_subprojects = gpc_get_string( 'subprojects' );
```
vboctor: Boolean?

config_defaults_inc.php
```diff
@@ -3969,6 +3969,12 @@
...
+ * @global integer $g_time_tracking_without_note
+ */
```
atrol: wrong variable name

config_defaults_inc.php
```diff
@@ -3969,6 +3969,12 @@
...
+$g_time_tracking_cost_per_hour = 0;
```
atrol: The option should be added to the $g_public_config_names list

cproensa:
fyi, i had proposed some times before, removing timetracking from core, into a plugin.
See [21572](https://mantisbt.org/bugs/view.php?id=21572)
Maybe you want to collaborate into that too.

> Add g_time_tracking_cost_per_hour config parameter to be able to set a default value

The billing functionality is close to nothing (only a value to factor the showed time).
Do we really need to add a new config option, for something that ideally should be better implemented as separate plugin?

dregad:
> Do we really need to add a new config option, for something that ideally should be better implemented as separate plugin?

:+1:

billing_export_to_csv.php
EthraZa: Checkbox sends a string "on" by default and w3c says its value is a string. Should we tamper with its value in PHP side?

EthraZa:
I have tested the timetracking plugin in [21572](https://mantisbt.org/bugs/view.php?id=21572) before to decide to improve the core time tracking, but its detached from notes behavior, makes no sence to me, also it does not have any billing functionality.

I would say MantisBT needs to take care of its time tracking feature bacause I see it as a very important part of the workflow. Maybe it's not for OSS but sure it is in the enterprise.
The minimal billing functionallity it has is serving well our needs here, because we just need a preety print report with the times and costs to email our costumers that ask for some more detailed billing explanation, and that is better than try to use a separate tool as Kimai, for example, that adds too much burden to the anotation workflow. Also, the time tracking itself is good for ourselves to track where we spent time.

That said, I could understand if someone decides to strip core time tracking off and convert it to a plugin, but I can't understand why on earth would someone just trash it in favor of any other code instead of improve it.

core/billing_api.php
EthraZa: I don't know what "missing from phpdoc" means exactly. I searched for "f_bugnote_cost" and "p_cost_per_hour" for any examples and can't find it in other files.

core/billing_api.php
EthraZa: Should I just replace **VIEWER** for **config_get( 'view_bug_threshold' )** inside **user_get_all_accessible_projects**?
Or should I pass another param to tell it I want **view_bug_threshold** instead of **VIEWER** just in this case?

vboctor: > Should I just replace VIEWER for config_get( 'view_bug_threshold' ) inside user_get_all_accessible_projects?

Yes.

core/billing_api.php
vboctor: I meant the @param was missing from the function documentation.

config_defaults_inc.php
```diff
@@ -4615,6 +4621,7 @@
 	'time_tracking_view_threshold',
 	'time_tracking_with_billing',
 	'time_tracking_without_note',
+	'time_tracking_cost_per_hour',
```
atrol: Please insert before `time_tracking_edit_threshold` to keep the list alphabetically sorted.

vboctor:
Thanks @EthraZa for your contribution.   Merged via dd2478d3da0e9197833b2d76db34404b333500e4 and a566e3a17025755daa396fada0375dd721803798.

vboctor:
@EthraZa if you have time there seems to be remaining issue that may be worth fixing - [23725](https://www.mantisbt.org/bugs/view.php?id=23725)

PR #1253, opened 2017-12-10
Don't use deprecated __autoload

atrol:
Fixes #23710

### DIFF
--- a/core.php
+++ b/core.php
@@ -159,7 +159,7 @@ function http_is_protocol_https() {
  * @param string $p_class Class name being autoloaded.
  * @return void
  */
-function __autoload( $p_class ) {
+function autoload_mantis( $p_class ) {
 	global $g_class_path;
 	global $g_library_path;
 
@@ -179,7 +179,7 @@ function __autoload( $p_class ) {
 }
 
 # Register the autoload function to make it effective immediately
-spl_autoload_register( '__autoload' );
+spl_autoload_register( 'autoload_mantis' );
 
 # Load UTF8-capable string functions
 define( 'UTF8', $g_library_path . 'utf8' );


vboctor:
@atrol I have this included in https://github.com/mantisbt/mantisbt/pull/1240/commits/7053b0ec69bec00aa829a8b5e90578f3015cc37b

atrol:
@vboctor I thought it's better to have it independent from all those bigger changes that might need some more time to be merged.

dregad:
> better to have it independent from all those bigger changes 

+1
And since the fix is the same, it would not cause any merge conflict.

vboctor:
@atrol @dregad I'm planning to merge the other PR this week.  In that PR, this change is in its own commit referring to the issue, so don't see a need for this one.  If for some reason I don't end up merging, I can cherry-pick this commit before the next release.

There may be merge conflicts since I changed the auto-load code before doing this change.  Hence, I included this change in that PR.

atrol:
> I'm planning to merge the other PR this week

@vboctor please don't merge

Changing core functionality like error handling is a certain risk and needs time for stabilization.

Maybe I am too fearful, but from my experience, it will introduce regressions.
I don't have time to have a deeper look at this PR at the moment, maybe next week.

We don't have that much automated tests, so there is some manual testing needed.
E.g. we have to ensure it runs fine with all PHP versions we support.

We should merge such kind of PR's quite soon after cutting a release and run the code in a production environment. If we are not able to do so, we should think about creating an unstable branch.
We could merge some more larger PR's in it (especially those ones from @cproensa) with the goal to stabilize the branch and to merge it to master at a later time, but not later than early February 2018.

PR #1260, opened 2017-12-28
Issue 17475

kabadi:
Added parent, grandparent etc. to list of projects to check for config options being set

Fixes [#17475](https://mantisbt.org/bugs/view.php?id=17475)

### DIFF
--- a/core/config_api.php
+++ b/core/config_api.php
@@ -36,6 +36,7 @@ require_api( 'constant_inc.php' );
 require_api( 'database_api.php' );
 require_api( 'error_api.php' );
 require_api( 'helper_api.php' );
+require_api( 'project_hierarchy_api.php' );
 require_api( 'utility_api.php' );
 
 # cache for config variables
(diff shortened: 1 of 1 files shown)


cproensa:
Eventhough this is a desirable fearture, it's not that simple

First, you may not have noticed, that one project can be a subroject of more that one project. Since this is not a tree structure, inheritance in those cases is hard to define. This is, at the moment, a blocking issue for this implementation. I suggested to remove this "feature", but there's nothing agreed still on it.

Second. If implemented, this would be a breaking change for already configured instances, where the projects have been configured to behave correctly with current scenario, the new config inheritance would potentially deliver a different configuration result. It should be, at least, driven by a config option that enables it only in new installations

There are more related issues:
The current configuration pages show visually how the options are inherited from "all projects", etc. Probably, a better displaying of that, including the new "inheritance" model shoud replace current UI.
Also, a specific setting of "inherit", beside the current on/off settings for each option may be a sensible addition.

vboctor:
Closing due to lack of activity.  Feel free to re-open.

PR #1261, opened 2017-12-29
Allow users to choose the font family

syncguru:
Fixes #23758

This is part of bigger effort to address the "hard to read" complains regarding the modern UI. Font selection gives users more options for different look and feel of the text on the page. Depending on the language selected and other display settings, users could find a much better option than the default "Open Sans" that ships with ManitsBT.

Font selection drop down in user account preference page:

screen shot 2017-12-28 at 7 57 56 pm


account_prefs_update.php
```diff
@@ -84,6 +84,9 @@
 	$t_prefs->language = $t_lang;
 }
 
+$t_font = gpc_get_string( 'font_family' );
+config_set ( 'font_family', $t_font, $f_user_id );
```
vboctor: I would only call config_set() if it is different from the current effective value.  This way we don't save it unless necessary and hence later on changing the font family affects all users that didn't explicitly change it and just saved user preferences before.

account_prefs_inc.php
```diff
@@ -383,6 +383,16 @@ function edit_account_prefs( $p_user_id = null, $p_error_if_protected = true, $p
 					</select>
 		</td>
 	</tr>
+    <tr>
```
vboctor: Indentation seems inconsistent with code arround this change.

core/layout_api.php
```diff
@@ -285,6 +288,16 @@ function layout_head_css() {
 	echo "\n";
 }
 
+/**
+ * Print user font preference
+ * @return void
+ */
+function layout_user_font_preference() {
+	echo '<style>', "\n";
+    echo  '* { font-family: "' , config_get( 'font_family' ) , '"; } ', "\n";
```
vboctor: Indentation not consistent.

config_defaults_inc.php
```diff
@@ -942,6 +942,48 @@
...
+##########################
+# MantisBT Font Settings #
+##########################
+
+/**
+ * Name of one of google fonts available at https://fonts.google.com/
+ * 'Open Sans' font files are part of MantisBT code. Selecting any other font will work if download from CDN
+ * option is enabled.
+ * @global string $g_font_family
+ */
```
vboctor: It is worth adding a see also for the cdn setting.

vboctor:
Should we hide/disable this option when CDN is OFF?  Otherwise, maybe we have a couple of fonts that are distributed with MantisBT code.  Also do we have a better candidate for default font?

account_prefs_update.php
atrol: 👍 
 also to keep the number of rows small in config table, especially as we always read the **whole** config table (see function config_cache_all)  

A bit off topic at this place, but even with this optimization, we could end in severe performance and memory issues in long term, at least on installations with lots of users.

I am aware that there are already other user preference options stored in this table, e.g. *_columns and email_notifications_verbose
This might have been a wrong decision and was just to avoid schema changes in user_pref table

I know, this is just one small new setting, but our next similar setting will also be just one small new setting.

We could create a new table for user preferences similar to the config table, where we fetch just data filtered by the current user id.
After that, migrate settings from user_pref and config table, finally drop user_pref table.

config_defaults_inc.php
```diff
@@ -942,6 +942,48 @@
...
+$g_font_family = 'Open Sans';
```
atrol: Add the option to $g_public_config_names

config_defaults_inc.php
```diff
@@ -942,6 +942,48 @@
...
+
+/**
+ * List the google fonts that the users are allowed to choose from.
+ * Google offers over 800 fonts. The list below is limited to the ones tested on MantisBT UI
+ * @global array $g_font_choices_arr
+ */
+$g_font_choices_arr = array(
```
atrol: Add the option to $g_public_config_names

core/layout_api.php
atrol: Assign to variable and use it for the two calls
```
$t_font_family = config_get( 'font_family' );
```

Do we want/need a project specific setting?
If not, call config_get with ALL_PROJECTS.

config_defaults_inc.php
atrol: Missing documentation in Admin Guide

atrol:
It would be fine to have also an option that works like before (<= 1.3) without loading a font.
Maybe offer two font families (sans-serif, serif).

syncguru:
@vboctor, @atrol - Thanks for the review and feedback
- I added a couple of fonts on top of Open Sans for the non-CDN case. I tried sans-serif & serif but they were not satisfactory.

config_defaults_inc.php
```diff
@@ -942,6 +942,61 @@
...
+	'Ubuntu',
+	'Vollkorn'
+);
+
+/**
+ * List of fonts that are installed as part of MantisBT
+ * This list is used when using CDN option is disabled
+ * @global array $g_font_choices_arr
+ */
+$g_safe_fonts_arr = array(
```
atrol: What about naming $g_font_choices_arr_cdn and $g_font_choices_arr_no_cdn

config_defaults_inc.php
```diff
@@ -942,6 +942,61 @@
...
+	'Poppins',
+	'Raleway',
+	'Ubuntu',
+	'Vollkorn'
+);
+
+/**
+ * List of fonts that are installed as part of MantisBT
+ * This list is used when using CDN option is disabled
+ * @global array $g_font_choices_arr
```
atrol: should be $g_safe_fonts_arr

config_defaults_inc.php
```diff
@@ -4458,6 +4513,8 @@
 	'filter_by_custom_fields',
 	'filter_custom_fields_per_row',
 	'filter_position',
+	'font_family',
+	'font_choices_arr',
```
atrol: should be inserted before font_family to keep sorting

config_defaults_inc.php
```diff
@@ -4593,6 +4650,7 @@
 	'sponsor_threshold',
 	'sponsorship_currency',
 	'sponsorship_enum_string',
+	'safe_fonts_arr',
```
atrol: should be inserted before search_title to keep sorting

core/print_api.php
```diff
@@ -1107,6 +1107,26 @@ function print_language_option_list( $p_language ) {
...
 }
 
+/**
+ * Print option list of available font choices
+ * @param string $p_font The current font.
+ * @return void
+ */
+function print_font_option_list( $p_font ) {
+	if ( config_get_global( 'cdn_enabled' ) == ON ) {
+		$t_arr = config_get('font_choices_arr');
```
atrol: Missing blanks after `(` and before `)`

core/print_api.php
```diff
@@ -1107,6 +1107,26 @@ function print_language_option_list( $p_language ) {
...
+	} else {
+		$t_arr = config_get('safe_fonts_arr');
```
atrol: Missing blanks after `(` and before `)`

config_defaults_inc.php
vboctor: I also suggest dropping the `_arr` suffixes.  I know we may have used them before, but they a lot of our configs support arrays and don't have such suffix.

vboctor: How about `font_choices` and `font_choices_local`?  Such names are better and follow the approach of having a common prefix for related configs.

account_prefs_update.php
```diff
@@ -84,6 +84,13 @@
 	$t_prefs->language = $t_lang;
 }
 
+$t_font = gpc_get_string( 'font_family' );
+if( config_get( 'font_family', null, $f_user_id, ALL_PROJECTS ) != $t_font ) {
+	config_set( 'font_family', $t_font, $f_user_id, ALL_PROJECTS );
+}
+
+config_set ( 'font_family', $t_font, $f_user_id );
```
vboctor: This line should be removed.

config_defaults_inc.php
vboctor: Actually `font_family_choices` and `font_family_choices_local`.

core/layout_api.php
```diff
@@ -256,7 +259,8 @@ function layout_head_css() {
 		html_css_cdn_link( 'https://maxcdn.bootstrapcdn.com/font-awesome/' . FONT_AWESOME_VERSION . '/css/font-awesome.min.css' );
 
 		# theme text fonts
-		html_css_cdn_link( 'https://fonts.googleapis.com/css?family=Open+Sans:300,400' );
+		$t_font_family =  config_get( 'font_family', null, null, ALL_PROJECTS );
```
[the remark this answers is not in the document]
syncguru: I thought about this initially but decided against it. It complicates the logic since `layout_user_font_preference()` has to produce different inline style element then 'config_get()' has to include the correct font files in both CDN and non-CDN cases. This complexity is rather unnecessary.

If any of the scenarios you mentioned takes place, impacted users will their default browser font (most likely ' 'sans serif'). They can simply fix the selection in their own preferences form.

account_prefs_inc.php
```diff
@@ -383,6 +383,16 @@ function edit_account_prefs( $p_user_id = null, $p_error_if_protected = true, $p
 					</select>
 		</td>
 	</tr>
+	<tr>
+		<td class="category">
+			<?php echo lang_get( 'font_family' ) ?>
+		</td>
+		<td>
+			<select id="font_family" name="font_family" class="input-sm">
+				<?php print_font_option_list( config_get( 'font_family' ) ) ?>
```
atrol: should be `config_get( 'font_family', null, $p_user_id, ALL_PROJECTS );`

PR #1262, opened 2018-01-03
Correct constructor name of FilterConverter class

atrol:
Fixes #23765

### DIFF
--- a/core/classes/FilterConverter.class.php
+++ b/core/classes/FilterConverter.class.php
@@ -52,7 +52,7 @@ class FilterConverter {
 	 * @param integer $p_user_id The logged in user id.
 	 * @param string $p_lang The logged in user language.
 	 */
-	public function __construction( $p_user_id, $p_lang ) {
+	public function __construct( $p_user_id, $p_lang ) {
 		$this->user_id = (int)$p_user_id;
 		$this->lang = $p_lang;
 	}


atrol:
> How about targeting this to 2.10.1?

It's not a regression, but a bug in a new function of 2.10.0, so I thought 2.11.0 is the right one.

There would be no advantage to have it in 2.10.1, assuming that 2.10.1 and 2.11.0 would be released the same date or if there is no other need to release a 2.10.1 version.

If there should be an early 2.10.1 because of a major regression, we could still cherry pick from master.

vboctor:
> It's not a regression, but a bug in a new function of 2.10.0

So the bug didn't apply before, but started to apply in 2.10.0 and hence I think it passes as a bug that wasn't there for several releases.

> There would be no advantage to have it in 2.10.1, assuming that 2.10.1 and 2.11.0 would be released the same date

I think there is a benefit even if they get released at the same time.  More conservative users may decide to always use the `.1` releases instead of `.0` releases to get 1 month worth of bake time in the wild and the related fixes.

atrol:
@vboctor until now there is no other reason to cut a 2.10.1 release.
I will merge the PR to master and cherry pick to master-2.10 if there should be another issue targeted  to  2.10.1.
If there is no other issue, you can decide at the end of January if you want to invest additional time in cutting 2.10.1 that includes this fix.

PR #1263, opened 2018-01-03
Command and REST API improvements

vboctor:
This PR restructured command data into multiple sub-arrays.  These are:
- `query` - typically used for url and query string parameters.
- `payload` - used for the payload that is expected to be passed with a POST command.
- `options` - provides the caller of the command the ability to pass options that control the command behavior.

The separation provides the advantage of having better similarity with REST API and protecting against a REST API payload specifying a value that it shouldn't, e.g. the issue_id that was provided before, or an option to a command.

The commands were also updated so as not to modify the passed in data.  If they are to validate, normalize, and update passed in values, then such values are shared into command properties that are shared between validate() and process() methods.

The existing command was updated to follow the new model.

This PR also includes the following features:

- [23774](https://www.mantisbt.org/bugs/view.php?id=23774) Implement IssueNoteAddCommand to share code for adding notes
- [23772](https://www.mantisbt.org/bugs/view.php?id=23772) Support attachments when adding notes via REST API
- [23773](https://www.mantisbt.org/bugs/view.php?id=23773) Support time tracking when adding notes via REST API
- [23775](https://www.mantisbt.org/bugs/view.php?id=23775) Remove obsolete code that checks if PHP file info API is defined
- [23776](https://www.mantisbt.org/bugs/view.php?id=23776) Support adding attachments that were not uploaded via the browser
- [23780](https://www.mantisbt.org/bugs/view.php?id=23780) Return 429 status code for spam check errors
- [23784](https://www.mantisbt.org/bugs/view.php?id=23784) REST and SOAP API send two email notifications for mentioned users
- [23785](https://www.mantisbt.org/bugs/view.php?id=23785) Adding notes via SOAP and REST API with time tracking uses incorrect access check
- [23786](https://www.mantisbt.org/bugs/view.php?id=23786) Implement IssueNoteDeleteCommand for deleting notes


core/commands/IssueNoteAddCommand.php
```diff
@@ -0,0 +1,303 @@
...
+		"name": "private"
+	},
+	"reporter": {
+		"name": "vboctor"
+	},
+	"time_tracking": {
+		"duration": "00:45",
+	},
+	"files": [
+		{
```
vboctor: Will fix indentation.

vboctor: Fixed

core/file_api.php
```diff
@@ -922,22 +967,66 @@ function file_ensure_uploaded( array $p_file ) {
...
+	return $t_finfo->file( $p_file_path );
+}
+
+/**
+ * Get mime type for the specified content.
+ *
+ * @param string $p_file_path The file path.
+ * @return boolean|string The mime type or false on failure.
+ */
+function file_get_mime_type_for_content( $p_content ) {
```
atrol: The name of the previous function is `file_get_mime_type`.
Shouldn't the name of this function be `content_get_mime_type` ?

core/file_api.php
```diff
@@ -922,22 +967,66 @@ function file_ensure_uploaded( array $p_file ) {
...
+	$t_info_file = config_get_global( 'fileinfo_magic_db_file' );
+
+	if( is_blank( $t_info_file ) ) {
+		$t_finfo = new finfo( FILEINFO_MIME );
+	} else {
+		$t_finfo = new finfo( FILEINFO_MIME, $t_info_file );
 	}
```
atrol: The code block starting with `$t_info_file =  ...` is the same like the block in function `file_get_mime_type`

Instead of the whole block, use existing function `finfo_get_if_available`.

core/file_api.php
vboctor: Nope. `file_` is the the prefix for all file_api.  At the end of the day this is a file_ api that acts on file content rather than file path.

core/file_api.php
vboctor: I just pushed a refactored version of the code.

PR #1264, opened 2018-01-06
Fix unprotecting admin users

atrol:
user_delete_project_specific_access_levels must not be called when
unprotecting a user, as user_ensure_unprotected is called in it.

Fixes #23787

### DIFF
--- a/manage_user_update.php
+++ b/manage_user_update.php
@@ -85,11 +85,11 @@ $t_user = user_get_row( $f_user_id );
 $f_username	= trim( $f_username );
 
 $t_old_username = $t_user['username'];
+$t_old_access_level = $t_user['access_level'];
 
 if( $f_send_email_notification ) {
 	$t_old_realname = $t_user['realname'];
 	$t_old_email = $t_user['email'];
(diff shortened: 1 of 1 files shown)


manage_user_update.php
```diff
@@ -148,8 +148,9 @@
 
 # Project specific access rights override global levels, hence, for users who are changed
 # to be administrators, we have to remove project specific rights.
-if( ( $f_access_level >= $t_admin_threshold ) && ( !user_is_administrator( $f_user_id ) ) ) {
-	user_delete_project_specific_access_levels( $f_user_id );
+if( ( $f_access_level != $t_old_access_level ) && ( $f_access_level >= $t_admin_threshold )
+	&& ( !user_is_administrator( $f_user_id ) ) ) {
```
vboctor: I would have the && at the end of the previous line, use spaces to line up the remainder of the if statement, and use a single tab for indentation of `if` body.

atrol: I changed it, not sure if it's exactly what you recommended.
I looked at some code before I did it the previous way and found && at the beginning of lines.
We don't have a clear rule for it.

PR #1231, opened 2017-11-07
summary page enhancement with bugs ratio added.

rainflood:
In summary page, added two columns in by_project, by_enum, by_developer and by_reporter summary tables: resolved bug ratio, i.e. (resolved + closed)/total *100%, and item ratio, i.e. (item total)/(sum of total bugs).

Fixed #0023627: Summary page enhancement with bugs ratio support

PR #1266, opened 2018-01-09
Fix Filter links for resolved/closed custom statuses in Summary By Status report

dregad:
Fixes [#23796](https://mantisbt.org/bugs/view.php?id=23796)

### DIFF
--- a/core/summary_api.php
+++ b/core/summary_api.php
@@ -210,57 +210,73 @@ function summary_print_by_enum( $p_enum ) {
 		
 		summary_helper_build_bugcount( $t_cache, $t_enum, $t_status, $t_bugcount );
 	}
+
+	switch( $p_enum ) {
+		case 'status':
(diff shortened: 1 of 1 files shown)


dregad:
This fix is not complete, as pointed out in https://mantisbt.org/bugs/view.php?id=23796#c58515 I failed to fix processing of the last line (outside of while loop). I think I'll refactor the whole function to get rid of code duplication.

atrol:
> I think I'll refactor the whole function

Keep in mind that there is also a refactoring after your requested it https://github.com/mantisbt/mantisbt/pull/1231#issuecomment-343862062 so you might introduce merge conflicts.

dregad:
Damn I forgot about that one

dregad:
Following merge of #1231, I have rebased the branch and (hopefully) fixed the remaining issue raised in https://mantisbt.org/bugs/view.php?id=23796#c58515

PR #1267, opened 2018-01-11
Issue File Attachments

vboctor:
- Add command for adding files to issues without notes.
- Add REST API for attaching files to issues without notes.
- Update REST API to use the new command when web UI has files without notes to add to issue.
- Raise an exception rather than trigger error when attachments file path is not valid.
- Define `$t_type` for case where issue note has an unexpected type.

### DIFF
--- a/core/commands/IssueFileAddCommand.php
+++ b/core/commands/IssueFileAddCommand.php
@@ -104,10 +104,10 @@ class IssueFileAddCommand extends Command {
 		$t_files_included = !empty( $this->files );
 
 		if( !$t_files_included ) {
-            throw new ClientException(
-                'Files not provided',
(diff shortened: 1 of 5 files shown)


core/commands/IssueFileAddCommand.php
```diff
@@ -0,0 +1,185 @@
...
+      {
+        "name": "filename.ext",
+        "type": "application/...",
+        "tmp_name": "/tmp/php/phpRELws8",
+        "error": 0,
+        "size": 114
+      }
+    ]
+  }
+}
```
dregad: This is not valid PHPDoc format, lines should begin with ` * `

vboctor: Fixed.

PR #1271, opened 2018-01-21
Issue Relationships REST API

vboctor:
- [23865](https://www.mantisbt.org/bugs/view.php?id=23865) Implement IssueRelationshipAddCommand to add relationships
- [23867](https://www.mantisbt.org/bugs/view.php?id=23867) Implement IssueRelationshipDeleteCommand
- [23866](https://www.mantisbt.org/bugs/view.php?id=23866) Support adding relationships via REST API
- [23868](https://www.mantisbt.org/bugs/view.php?id=23868) Support deleting issue relationships via REST API

### DIFF
--- a/core/commands/IssueRelationshipAddCommand.php
+++ b/core/commands/IssueRelationshipAddCommand.php
@@ -1,0 +1,206 @@
+<?php
+# MantisBT - A PHP based bugtracking system
+
+# MantisBT is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# MantisBT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+require_api( 'authentication_api.php' );
+require_api( 'bug_api.php' );
+require_api( 'constant_inc.php' );
+require_api( 'config_api.php' );
+require_api( 'helper_api.php' );
+require_api( 'user_api.php' );
+
+use Mantis\Exceptions\ClientException;
+
+/**
+ * A command that adds a relationship to an issue.  If a relationship already
+ * exists with the target issue, it will be updated.
+ *
+ * Sample:
+ *
+ * {
+ *   "query": { "issue_id": 1234 },
+ *   "payload": {
+ *     "issue": {
+ *       "id": 1235
+ *     },
+ *     "type": {
+ *       "id": 1,
+ *       "name": "related-to"
+ *     }
+ *   }
+ * }
+ *
+ * Relation Type IDs: check "bug relationship constants" in core/constants_inc.php
+ */
+class IssueRelationshipAddCommand extends Command {
+	/**
+	 * The relationship type id
+	 *
+	 * @var integer
+	 */
+	private $typeId;
+
+	/**
+	 * The source issue
+	 *
+	 * @var BugData
+	 */
+	private $sourceIssue = null;
+
+	/**
+	 * The target issue
+	 *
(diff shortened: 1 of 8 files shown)


core/commands/IssueRelationshipAddCommand.php
```diff
@@ -0,0 +1,206 @@
...
+	 * @returns array Command response
+	 */
+	protected function process() {
+		if( $this->sourceIssue->project_id != helper_get_current_project() ) {
+			# in case the current project is not the same project of the bug we are
+			# viewing, override the current project. This to avoid problems with
+			# categories and handlers lists etc.
+			global $g_project_override;
+			$g_project_override = $this->sourceIssue->project_id;
+		}
```
syncguru: I see this block called multiple times. Consider refactoring out.

core/commands/IssueRelationshipDeleteCommand.php
```diff
@@ -0,0 +1,111 @@
...
+
+		# Retrieve the target issue of the relationship
+		$t_target_issue_id = relationship_get_linked_bug_id( $t_relationship_id, $t_issue_id );
+		$t_target_issue = bug_get( $t_target_issue_id, true );
+
+		# Ensure that user can view target issue
+		$t_view_threshold = config_get( 'view_bug_threshold', null, null, $t_target_issue->project_id );
+		if( !access_has_bug_level( $t_view_threshold, $t_target_issue_id ) ) {
+			throw new ClientException(
+				sprintf( "Issue %d is read-only", $t_target_issue_id ),
```
syncguru: Same as above, the error message is incorrect.

core/commands/IssueRelationshipAddCommand.php
vboctor: Yes, I'm planning to do some horizontal refactoring across commands, but I'm trying to get the commands checked in first, before doing refactoring that would conflict with others.

core/commands/IssueRelationshipAddCommand.php
```diff
@@ -0,0 +1,206 @@
...
+	 * @param array The relationship type reference with id, name or both.
+	 * @return integer relationship type id.
+	 */
+	private function getRelationTypeId( $p_relationship_type ) {
+		if( isset( $p_relationship_type['id'] ) ) {
+			$t_type_id = (int)$p_relationship_type['id'];
+		} else if( isset( $p_relationship_type['name'] ) ) {
+			global $g_relationships;
+			switch( $p_relationship_type['name'] ) {
+				case 'related-to':
```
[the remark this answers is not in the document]
vboctor: In the SOAP API we had a pattern where an entity can be referenced by its id or name.  If both are supplied, the id is used.

In the REST API, we use the same model.  However, when getting back a reference from the server, we often including extra information like `label` which is the localized language based on the user that issued the request.  That applies across the board.

It is a core part of the REST API design to not just dump internal structures to users and expects client apps to construct them when issuing requests.

core/commands/IssueRelationshipDeleteCommand.php
vboctor: Fixed.

core/commands/IssueRelationshipAddCommand.php
```diff
@@ -0,0 +1,206 @@
...
+				ERROR_BUG_READ_ONLY_ACTION_DENIED,
+				array( $t_source_issue_id )
+			);
+		}
+
+		# Ensure that user can view target issue
+		$t_view_threshold = config_get( 'view_bug_threshold', null, null, $this->targetIssue->project_id );
+		if( !access_has_bug_level( $t_view_threshold, $t_target_issue_id ) ) {
+			throw new ClientException(
+				sprintf( "Issue %d is read-only", $t_target_issue_id ),
```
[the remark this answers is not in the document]
vboctor: Fixed.

core/commands/IssueRelationshipAddCommand.php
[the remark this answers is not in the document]
vboctor: I have done some refactoring and fixing to existing code to address the following:

- Have the mapping in the relationship APIs and make it consistent with other similar transformations.
- Fix an issue where issue get was localizing the name rather than having a label field. Also added label field there.

PR #1272, opened 2018-01-22
Files (attachments) REST APIs

vboctor:
- Add `access` to attachment information to indicate user's access to download and delete attachment.
- Add `delete_url` for the attachment's delete url, similar to `download_url`.
- Add `icon` for fontowesome icon and `alt` for alternate text.
- Ability to download all attachments associated with an issue via one request.
- Ability to download one attachment associated with an issue.

### DIFF
--- a/api/rest/restcore/issues_rest.php
+++ b/api/rest/restcore/issues_rest.php
@@ -65,6 +65,10 @@ $g_app->group('/issues', function() use ( $g_app ) {
 	# Files
 	$g_app->post( '/{id}/files/', 'rest_issue_file_add' );
 	$g_app->post( '/{id}/files', 'rest_issue_file_add' );
+	$g_app->get( '/{id}/files/', 'rest_issue_files_get' );
+	$g_app->get( '/{id}/files', 'rest_issue_files_get' );
+	$g_app->get( '/{id}/files/{file_id}/', 'rest_issue_files_get' );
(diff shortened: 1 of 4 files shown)


api/rest/restcore/issues_rest.php
```diff
@@ -370,3 +374,59 @@ function rest_issue_monitor_add( \Slim\Http\Request $p_request, \Slim\Http\Respo
...
+			)
+		);
+
+		if( $t_internal_file['can_download'] ) {
+			$t_file['download_url'] = $t_internal_file['download_url'];
+		}
+
+		if( $t_internal_file['can_delete'] ) {
+			$t_file['delete_url'] = $t_internal_file['delete_url'];
+		}
```
syncguru: I did not expect to see this. I thought deleting a file is another REST api, no?

vboctor: > I did not expect to see this. I thought deleting a file is another REST api, no?

Yes, there is another REST API for deleting a file.  The URL here is for the web URL for the delete page for use in the web UI.  I can see how this can be confusing.  Will look again and re-consider.

vboctor: @syncguru Removed the `download_url` and `delete-url`.

syncguru:
Looks good

PR #1275, opened 2018-01-27
TICKET 0023894: Manage User Page - Search Section ALSO on PAGE TOP

fmancardi:
In addition, a help icon has been added to remember user using this feature that users can be searched by providing username or email.

### DIFF
--- a/manage_user_page.php
+++ b/manage_user_page.php
@@ -292,7 +292,7 @@ $t_user_count = count( $t_users );
 <?php 
 $t_filter_string = $f_filter . $t_hide_inactive_filter . $t_show_disabled_filter . "&amp;sort=$c_sort&amp;dir=$c_dir";
 
-draw_search_section($t_total_user_count,$p_per_page,$t_page_count, $f_page_number, $t_filter_string); 
+draw_search_section('top',$t_total_user_count,$p_per_page,$t_page_count, $f_page_number, $t_filter_string); 
 ?>
 
 <div class="widget-main no-padding">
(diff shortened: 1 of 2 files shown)


manage_user_page.php
```diff
@@ -362,7 +369,7 @@
 	<div id="manage-user-edit-div" class="form-inline pull-left">
 		<form id="manage-user-edit-form" method="get" action="manage_user_edit_page.php" class="form-inline"
 			<?php # CSRF protection not required here - form does not result in modifications ?>>
-			<label class="inline" for="username"><?php echo lang_get( 'search' ) ?></label>
+			<label class="inline" for="username"><?php echo lang_get( 'search' )?> <i class="fa fa-info-circle" title="<?php echo lang_get('user_search_hint')?>"></i> </label>
```
atrol: I prefer placeholder as is works better with mobile devices, e.g. we use it for username on login page https://github.com/mantisbt/mantisbt/blob/master-2.10/login_page.php#L239

manage_user_page.php
```diff
@@ -379,3 +386,33 @@
 </div>
 <?php
 layout_page_end();
+
+
+/**
+ *
+ */
```
atrol: PHPDoc missing

manage_user_page.php
fmancardi: OK, got. Will remember it. Do you need I change this as do a new Pull Request? please let me know

manage_user_page.php
```diff
@@ -379,3 +386,33 @@
...
+function draw_search_section($p_user_count,$p_per_page,$p_page_count, $p_page_number, $p_filter_string) {
```
atrol: You introduce a new function that is quite similar to existing code.
The goal should be to replace existing code by a call to the new function.

fmancardi: I've copied the existent code to create the function, OK if existent code can be replaced => two calls to this function. Please let me know how I've to proceed in order to provide work as will be better to be integrated on stable code base

atrol:
Thanks for your contribution.

Please consider our Coding Guidelines, e.g. `One space after the opening ( and before the closing ) `
https://www.mantisbt.org/wiki/doku.php/mantisbt:coding_guidelines

manage_user_page.php
atrol: > Do you need I change this as do a new Pull Request?

No, you can go on committing/pushing to this one.
As a final step, we can squash all your commits to a single one.

manage_user_page.php
```diff
@@ -358,24 +365,51 @@
...
+ * @param int    $p_page_count how many pages do we have.
+ * @param int    $p_page_number current page number.
+ * @param string $p_filter_string used by print_page_links() 
+ *
+ * @return void
+ */
+function draw_search_section($p_where,$p_user_count,$p_per_page,$p_page_count, $p_page_number, $p_filter_string) {
+
+	if( $p_user_count < $p_per_page ) {
+		return;
```
atrol: Before this change, we were always able to search for a user by name.
Now there is use case where I am missing this functionality.

Assume there are 60 user names starting with character 'A'.
Click on `A` right beside the `All` button.
The search field is available in this case and you will be able to search for any user, not just the ones starting with 'A'.

Assume there are 40 user names starting with character 'A'.
Click on `A` right beside the `All` button.
The search field is not available in this case.

fmancardi:
my fault.Will fix. check has to be done Only to hide search on TOP never on
Bottom.
Apologize

Il 29/Gen/2018 21:39, "Roland Becker" <notifications@github.com> ha scritto:

atrol:
@fmancardi there were some enhancements in Manage User UI starting in version 2.17.0.
I think that having the search section on two places is no longer needed.

Do you agree that we can close this PR?

atrol:
Closed as there was no more feedback.
@fmancardi feel free to reopen/rebase/resolve merge conflicts if you still see a need for the change after the enhancements starting in version 2.17.0.

PR #1277, opened 2018-01-30
Revise show realname code and uniqueness checks

vboctor:
- Remove user realname uniqueness check against usernames or real names.  Username is required to be unique.
- Remove `$g_differentiate_duplicates` config option and reference to `duplicate_realname` user field that doesn't exist.
- When displaying usernames always display username, then annotate with realname in tooltip or realname between brackets.  This guarantees uniquness.
- Always showing username makes @ mentioning users easier. #23375
- Refactor code to centralize crafting of display name.
- Refactor code to centralize crafting of name used for sorting.
- Make sure that show realname threshold is always checked.

Fixes #23909, #23900, #23375

### DIFF
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -937,7 +937,12 @@ function user_get_email( $p_user_id ) {
  * @return string
  */
 function user_get_username( $p_user_id ) {
-	return user_get_field( $p_user_id, 'username' );
+	$t_row = user_cache_row( $p_user_id, false );
+	if( false == $t_row ) {
+		return lang_get( 'prefix_for_deleted_users' ) . (int)$p_user_id;
+	}
+
+	return $t_row['username'];
 }
 
 /**
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -1027,21 +1027,15 @@ function user_get_name( $p_user_id ) {
 
 	if( false == $t_row ) {
 		return lang_get( 'prefix_for_deleted_users' ) . (int)$p_user_id;
-	} else {
-		if( ON == config_get( 'show_realname' ) ) {
-			if( is_blank( $t_row['realname'] ) ) {
-				return $t_row['username'];
-			} else {
-				if( isset( $t_row['duplicate_realname'] ) && ( ON == $t_row['duplicate_realname'] ) ) {
-					return $t_row['realname'] . ' (' . $t_row['username'] . ')';
-				} else {
-					return $t_row['realname'];
-				}
-			}
-		} else {
-			return $t_row['username'];
+	}
+
+	if( ON == config_get( 'show_realname' ) ) {
+		if( !is_blank( $t_row['realname'] ) ) {
+			return $t_row['realname'];
 		}
 	}
+
+	return $t_row['username'];
 }
 
 /**
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -987,14 +987,22 @@ function user_get_name( $p_user_id ) {
 }
 
 /**
+ * Show realnames be shown to logged in user?
+ *
+ * @return bool true to show, false otherwise.
+ */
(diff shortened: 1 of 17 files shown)


core/user_api.php
```diff
@@ -318,65 +326,42 @@ function user_ensure_email_unique( $p_email, $p_user_id = null ) {
...
-			$t_users[] = $t_row;
-		}
-		$t_duplicate_count = count( $t_users );
-
-		if( $t_duplicate_count > 0 ) {
-			# set flags for non-unique realnames
-			if( config_get( 'differentiate_duplicates' ) ) {
-				for( $i = 0;$i < $t_duplicate_count;$i++ ) {
-					$t_user_id = $t_users[$i]['id'];
-					user_set_field( $t_user_id, 'duplicate_realname', ON );
```
atrol: As `differentiate_duplicates` defaulted to `OFF` and was not completely implemented, we could drop the column now. If not, we should create an issue with tag `schema` to consider it when introducing other database schema changes.

vboctor: I could not find a `differentiate_duplicates` in `schema.php`, so not sure if this code ever worked.

atrol: You are right, the column does not exist.

dregad:
I believe there is something conceptually wrong with this whole thing.

Realnames are not unique. That's a fact. And we can't force them to be in Mantis, because there are scenarios where this is out of our control (e.g. LDAP).

dregad:
> We enforce user name uniqueness.

As well we should (and need to) since it's the user's identifier.

> A user can use their real name as their user name, since regex allows it (and regex is customizable) and I have seen this before.

That's fine, the uniqueness is guaranteed.

> We enforce realname uniqueness against usernames (so if a user has John Smith as username, others can't use it as a real name).

This is what I'm challenging. If my name is _John Smith_, why should I be known as _John Smith 2_ or whatever other way to bypass the unique check ? Moreover, as I was saying before, I may not even have the option to edit my realname (if it comes from an LDAP directory).

> When show real name is set to ON, we only display real name. For users who don't have a real name we use username.

That's fine. The only thing we need (to avoid confusing users bearing the same real name) is to provide a way to differentiate them when using the Realname (e.g. by tagging the username, something like _John Smith (jsmith)_ vs _John Smith (john_smith). An alternative is, as per your suggestion, to always use the realname (with Realname as an optional pop-up) and get rid of show_realname.

>LDAP doesn't enforce real name uniqueness against other real names or usernames. LDAP however enforces username uniqueness.

Exactly. It should not, and neither should we.

I'm not really sure what the regression was about and don't have time to look into it now. I'll let you decide how to move forward, but I'm concerned that the cure could be worse than the disease.

atrol:
> I'm not really sure what the regression was about

This is the regression https://www.mantisbt.org/bugs/view.php?id=23900

which seems to be caused by
https://github.com/mantisbt/mantisbt/commit/3f29dcc4a74c47ecb4329041f97599faadee5193#diff-e09451e48bf1704da38ace9e4b20c46dR376

changing from
```
if( 1 > user_is_realname_unique( $p_username, $p_realname ) ) {
```
to
```
$t_result = user_is_realname_unique( $p_username, $p_realname );
	if( $t_result == 0 || $t_result > 1 ) {
```

vboctor:
OK, I revise the PR to always show username with (real name) or bubble as appropriate.  Will remove the uniqueness check for real name.

vboctor:
I have implemented some fixes and refactoring related to display names, sort order, access check for realname display threshold, and changed realname not to require uniqueness, etc.  The PR description is updated.

vboctor:
I just checked in a targeted fix for the regression.  Will get this PR merged after releasing 2.11.0.

PR #1278, opened 2018-01-31
Make sure that the summary will always fit in it's column size

mikeSimonson:


atrol:
I think that silent truncation at this low level is no good idea.

The client is responsible to deal with it.
E.g. don't allow to enter more than 128 characters or provide a message if the description is too long.

Do you get any problem in Mantis because of this?
If so, please open a bug report at https://www.mantisbt.org/bugs for it.

Also if you want some new feature or enhancement, create a report and reference it in the body of your commit message, e.g.
`Fixes #<TheIdOfTheReport>`

mikeSimonson:
I get errors because a plugin use your api with a longer summary ( a mail plugin). Right now the application just crash because of this.

The application at the very least should return an error when it's used with parameters that would make it crash.
By the way the silent truncation is how it works if you use an old enough version of mysql (pre 5.7.something).

PR #1280, opened 2018-02-03
Fix regression that discloses file path in some errors

vboctor:
This was introduced as part of refactoring error handler and it happens with some errors even when show_detailed_errors is set to OFF.

Fixes #23925

### DIFF
--- a/api/rest/index.php
+++ b/api/rest/index.php
@@ -45,7 +45,9 @@ ApiObjectFactory::$soap = false;
 
 # Show SLIM detailed errors according to Mantis settings
 $t_config = array();
-if( ON == config_get_global( 'show_detailed_errors' ) ) {
+
+$t_show_detailed_errors = ON == config_get_global( 'show_detailed_errors' );
+if( $t_show_detailed_errors ) {
 	$t_config['settings'] = array( 'displayErrorDetails' => true );
 }
 
@@ -74,7 +76,11 @@ $t_container['errorHandler'] = function( $p_container ) {
 		$t_error_to_log =  $p_exception->getMessage() . "\n" . $t_stack_as_string;
 		error_log( $t_error_to_log );
 
-		return $p_response->withStatus( HTTP_STATUS_INTERNAL_SERVER_ERROR )->withJson( $t_data );
+		if( $t_show_detailed_errors ) {
+			$p_response = $p_response->withJson( $t_data );
+		}
+
+		return $p_response->withStatus( HTTP_STATUS_INTERNAL_SERVER_ERROR );
 	};
 };
 
--- a/api/rest/index.php
+++ b/api/rest/index.php
@@ -70,7 +70,11 @@ $t_container['errorHandler'] = function( $p_container ) {
 			return $p_response->withStatus( $p_exception->getCode(), $p_exception->getMessage() )->withJson( $t_data );
 		}
 
-		return $p_response->withStatus( HTTP_STATUS_INTERNAL_SERVER_ERROR, $p_exception->getMessage() )->withJson( $t_data );
+		$t_stack_as_string = error_stack_track_as_string( $p_exception );
+		$t_error_to_log =  $p_exception->getMessage() . "\n" . $t_stack_as_string;
+		error_log( $t_error_to_log );
+
+		return $p_response->withStatus( HTTP_STATUS_INTERNAL_SERVER_ERROR )->withJson( $t_data );
 	};
 };
 
--- a/api/rest/index.php
+++ b/api/rest/index.php
@@ -70,7 +70,7 @@ $t_container['errorHandler'] = function( $p_container ) {
 			return $p_response->withStatus( $p_exception->getCode(), $p_exception->getMessage() )->withJson( $t_data );
 		}
 
-		$t_stack_as_string = error_stack_track_as_string( $p_exception );
+		$t_stack_as_string = error_stack_trace_as_string( $p_exception );
 		$t_error_to_log =  $p_exception->getMessage() . "\n" . $t_stack_as_string;
 		error_log( $t_error_to_log );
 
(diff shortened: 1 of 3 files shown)


core/error_api.php
```diff
@@ -247,11 +247,16 @@ function error_handler( $p_type, $p_error, $p_file, $p_line, array $p_context )
 
 	$t_error_description = nl2br( $t_error_description );
 
+	# Make sure the file path is not disclosed via exception details
+	$t_error_description = str_replace( config_get_global( 'absolute_path' ), '.../', $t_error_description );
```
atrol: we have
```
define ('DIRECTORY_SEPARATOR', "/");
$g_absolute_path = dirname( __FILE__ ) . DIRECTORY_SEPARATOR;
```
Can't check myself at the moment, did you try if it works on Windows?

Should the path be removed if `ON == config_get_global( 'show_detailed_errors' )`?

vboctor: I initially had the code to only show the path when `show_detailed_errors` is ON.  But I then removed this code since I think it is not worth having the extra checks.  The real value is the relative path.

I don't have a Windows environment setup at the moment.

dregad:
I'm also concerned about the robustness of the proposed _str_replace()_-based solution. It works for sure on unix-like environments with a standard MantisBT installation, but what about the case where the admin customized the folder locations in config_inc.php ? Regardless of the directory separator, there are bound to be cases where a simple replacement won't do the trick.

I thought about using the `file` item in the exception's first stack trace entry instead of $g_absolute_path as string to replace, but it this still does not sound very solid.

IMO, PHP errors are not meant to be displayed to end users; I believe it would be preferable and safer to display a generic error, while writing the actual message and stack trace in the system log - at least when $g_show_detailed_errors = OFF.

core/error_api.php
dregad: I can test on Windows when I'm in the office on Monday.

vboctor:
Will look into having our exception displayed, but PHP exceptions or exceptions that didn't originate from Mantis show a generic message (e.g. internal error, while logging the proper error).  If show detailed errors is ON, then we will always show exception.  Will refine the approach as I look further on the code.

Will remove path replacement.

dregad:
Cheers. Please note that I will probably not have the opportunity to review/test your revised fix until tomorrow evening (~22:00 CET) as I won't be at home the whole day. I'll do my best to do it as soon as possible, so you can release 2.11 as planned, but if I can't manage I'd rather postpone the release by a few days.

vboctor:
I have implemented an alternative fix which:

- Logs php errors.
- Don't display php exception messages in UI or API.

api/rest/index.php
```diff
@@ -70,7 +70,11 @@
 			return $p_response->withStatus( $p_exception->getCode(), $p_exception->getMessage() )->withJson( $t_data );
 		}
 
-		return $p_response->withStatus( HTTP_STATUS_INTERNAL_SERVER_ERROR, $p_exception->getMessage() )->withJson( $t_data );
+		$t_stack_as_string = error_stack_track_as_string( $p_exception );
+		$t_error_to_log =  $p_exception->getMessage() . "\n" . $t_stack_as_string;
+		error_log( $t_error_to_log );
+
+		return $p_response->withStatus( HTTP_STATUS_INTERNAL_SERVER_ERROR )->withJson( $t_data );
```
dregad: Is it normal to return `->withJson( $t_data )` which still contains the exception's message ?

core/error_api.php
```diff
@@ -480,30 +488,47 @@ function error_print_context( array $p_context ) {
...
 /**
- * Print out a stack trace
+ * Get the stack trace as a string that can be logged or echoed to CLI output.
+ *
+ * @param Exception|null $p_exception The exception to print stack trace for.  Null will check last seen exception.
+ * @return string multi-line printout of stack trace.
  */
-function error_print_stack_trace() {
-	$t_stack = error_stack_trace();
+function error_stack_track_as_string( $p_exception = null ) {
```
dregad: Should be *error_stack_trac**e**_as_string*, no ?

core/error_api.php
```diff
@@ -480,30 +488,47 @@ function error_print_context( array $p_context ) {
...
+/**
+ * Print out a stack trace
+ *
+ * @param Exception|null $p_exception The exception to print stack trace for.  Null will check last seen exception.
+ */
+function error_print_stack_trace( $p_exception = null ) {
+	$t_stack = error_stack_trace( $p_exception );
+
+	if( php_sapi_name() == 'cli' ) {
+		echo error_stack_track_as_string( $p_exception );
```
dregad: *error_stack_trac**e**_as_string*

core/error_api.php
```diff
@@ -480,30 +488,47 @@ function error_print_context( array $p_context ) {
...
+	return $t_output;
+}
+
+/**
+ * Print out a stack trace
+ *
+ * @param Exception|null $p_exception The exception to print stack trace for.  Null will check last seen exception.
+ */
+function error_print_stack_trace( $p_exception = null ) {
+	$t_stack = error_stack_trace( $p_exception );
```
dregad: Move that after the _cli_ test, just before the foreach loop

core/error_api.php
```diff
@@ -480,30 +488,47 @@ function error_print_context( array $p_context ) {
...
+
+	if( php_sapi_name() == 'cli' ) {
+		echo error_stack_track_as_string( $p_exception );
 		return;
 	}
```
dregad: `# remove the call to the error handler from the stack trace` (a few lines down from here) looks like a leftover comment from older code which should probably be removed.

vboctor: Removed.

core/error_api.php
vboctor: Fixed typo.

core/error_api.php
vboctor: Fixed typo.

core/error_api.php
vboctor: Moved.

api/rest/index.php
vboctor: Changed code to include it only if show detailed errors is ON.

vboctor:
@dregad thanks for the review.  I applied all your review comments.  Let me know if all looks good.

dregad:
Looks good, but I didn't test in depth the behavior from the API.

PR #1281, opened 2018-02-04
Update admin checks to make finfo extension mandatory

vboctor:
Fixes #23930


atrol:
Another approach for it https://github.com/mantisbt/mantisbt/pull/1282

PR #1283, opened 2018-02-05
Remove utf8 library, enforce mbstring extension

atrol:
Fixes [#23214](https://www.mantisbt.org/bugs/view.php?id=23214)


vboctor:
+1 for removing the dependency.  However, I think we should consider whether we should have deprecated functions for background compatibility like utf8_substr and utf8_strlen that just call the mb_ equivalent.  This way if plugins are calling such APIs they get deprecated message rather than fail.  Thoughts?  Not strong opinion there, just wanted to make sure we are intentionally about it.

atrol:
> I think we should consider whether we should have deprecated functions

Added utf8_strlen(), utf8_substr() and utf8_strtolower().
These are all utf8_* functions I found when checking the 73 plugins under https://github.com/mantisbt-plugins.

I am a bit astonished that there are not that many plugins affected from this change.
Seems some of them are not clean in terms of UTF8 and use the standard PHP functions (strlen(), ...)

I will also provide PRs to remove usage of utf8_* for those plugins that support Mantis 2.x.

There are a few more plugins affected, but they do not support 2.x or are no longer needed

dregad:
Nice one @atrol I've had this on my todo list for ages, but never got around to doing it.

With regards to the provided fallback, deprecated functions - is there a reason for only providing a subset of 3 (utf8_strlen, utf8_substr and utf8_strtolower) from the old library's list ? I would expect the following list for full backwards-compatibility:

```
PHP> include 'library/utf8/utf8.php';
PHP> include 'library/utf8/str_pad.php';
PHP> print_r(get_defined_functions()['user']);
Array
(
    [0] => utf8_strlen
    [1] => utf8_strpos
    [2] => utf8_strrpos
    [3] => utf8_substr
    [4] => utf8_strtolower
    [5] => utf8_strtoupper
    [6] => utf8_substr_replace
    [7] => utf8_str_pad
)
```

So we're missing: utf8_strpos, utf8_strrpos, utf8_strtoupper, utf8_substr_replace.

atrol:
> is there a reason for only providing a subset of 3 (utf8_strlen, utf8_substr and utf8_strtolower) from the old library's list ?

Laziness and
> These are all utf8_* functions I found when checking the 73 plugins under https://github.com/mantisbt-plugins.

vboctor:
Just merged #1280 - I recall it moved some utf8 calls into user_api.  So you may have to rebase and convert these calls.

atrol:
> So you may have to rebase

Done, got no conflicts

> convert these calls

I don't see any utf8 related changes in https://github.com/mantisbt/mantisbt/pull/1280/files

atrol:
> So you may have to rebase and convert these calls.

Now true + merge conflicts after latest changes.

library/README.md
```diff
@@ -9,8 +9,7 @@ The version and status of each is summarized below:
 
 directory       | project         | version   | status
 ----------------|-----------------|-----------|---------------
-rssbuilder      | RSSBuilder      | 2.2.1     | patched [1]
-utf8            | phputf8         | 0.5       | unpatched
+rssbuilder      | RSSBuilder      | 2.2.1     | patched [2]
```
vboctor: Should this be `patched [1]`?

atrol: It should.
I fixed this some while ago
https://github.com/mantisbt/mantisbt/commit/e30f40ce74d414b9604461eae3de88be29cc539d
but got lost when rebasing this branch.
Will fix.

atrol:
Do not merge.
I am not happy how it works at the moment if mbstring is not enabled.
Will work on a better solution.

atrol:
@dregad @vboctor I would like to get your opinion especially on my latest commit https://github.com/mantisbt/mantisbt/pull/1283/commits/da34cc9855349fd3a5fc664ee6bbdb8076c05df8

I don't want to open a can of worms because of this in this PR and start rewriting the installer.
There is at least one conceptual problem in installer, as there is a different behaviour depending on a valid login cookie, so you get different results for config_get calls.
For another side effect of this, see https://mantisbt.org/forums/viewtopic.php?f=3&t=19112#p64562 
where we try to update user last_visit time. This fails if we are upgrading from earlier versions, where we stored dates with another data type than we store now.

Should we keep my latest commit, or should we live with an ugly error message when upgrading an installation where mbstring extension is not enabled?

dregad:
I don't have time to look into this now - should be able to check tomorrow evening if all goes well.

dregad:
I am not so keen on the approach to systematically falling back (non-utf8) strtolower, although in the case of 69539e (session_handler) and 27a830f (browser language) it is unlikely to cause issues as you've well explained in the commit messages. 

da34cc9 is a different story though because the function operates on user-provided strings that have a non-negligible chance of being customized, so there is a risk that someone will use UTF8 string.

Alternatives:
- detect early in core.php that mbstring extension is enabled, and either abort gracefully or implement some fallback mechanism in this case
- use another, more modern UTF8 library that will abstract this whole mb_XXX business for us (e.g. https://github.com/tchwork/utf8)

atrol:
Thanks @dregad for the feedback

I removed the former mb_strtolower -> strtolower changes and implemented

> detect early in core.php that mbstring extension is enabled

Independant from that, I plan to remove the session_handler option https://www.mantisbt.org/bugs/view.php?id=23966

atrol:
@dregad is the PR ok for you after my latest changes?

@vboctor this could be a breaking change for some of our users.
Maybe there should be a special hint in release announcement for it.

PR #1284, opened 2018-02-06
Added the "-" for the users with a dash in their name (mentions), fixes 0022509

elen-ayanyan:


dregad:
Your commit message needs to reference (i.e. add `Fixes #xxxx`) an issue in our [tracker](https://mantisbt.org/bugs) , create one if needed.

elen-ayanyan:
Issue ID: 0022509

core/mention_api.php
```diff
@@ -74,7 +74,7 @@ function mention_get_candidates( $p_text ) {
 			. '(?<!' . $t_quoted_tag . ')' . $t_quoted_tag
 			. ')'
 			# any word char or period, but must not end with period
```
dregad: Change comment to reflect revised regex

`# any word char, dash or period, must end with word char`

elen-ayanyan:
Is it ok now?

dregad:
> Is it ok now?

The code looks fine.

As for the commit message, I think you don't understand what I'm asking. You changed the *Pull request's description*, which is not the same thing as the *Commit message*. The latter is the text you type when you commit your changes, e.g. `Update mention_api.php` for 57892ec157bbde5eacae37532d22aa5276ce5245.

Commit messages should actually describe what the change is about - *Update mention_api.php* is not really useful in that respect. For detailed information about what we consider a "good" commit message, please read http://mantisbt.org/wiki/doku.php/mantisbt:git_commit_messages

At that point, using your git client, you would need to [squash your 3 commits into one](https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git), and update (amend) the commit message, then force-push your feature branch to update the pull request.

PR #1286, opened 2018-02-06
Error page improvements

dregad:
This PR includes several improvements to the "Detailed error page" (i.e. when $g_show_detailed_errors = ON)

- Fix HTML error causing page footer to be displayed behind the sidebar [#23926](https://mantisbt.org/bugs/view.php?id=23926)
- Stack trace should begin where error was triggered [#23944](https://mantisbt.org/bugs/view.php?id=23944)
- Improve page layout [#23943](https://mantisbt.org/bugs/view.php?id=23944)
- Remove error handler's "errcontext" parameter (deprecated in PHP 7.2) [#23942](https://mantisbt.org/bugs/view.php?id=23942)
- Minor tweaks (PHPDoc, generated HTML)

### DIFF
--- a/core/error_api.php
+++ b/core/error_api.php
@@ -105,11 +105,10 @@ function error_stack_trace( $p_exception = null ) {
 	}
 
 	if ( $p_exception === null ) {
-		$t_stack = debug_backtrace();
-
-		# remove this function and its caller from the stack trace.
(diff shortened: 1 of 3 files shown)


dregad:
## Before
before

## After
after

PR #1288, opened 2018-02-09
Fix REST API cookie auth for users with 32 char cookies

vboctor:
Fixes #23954

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -1113,11 +1113,6 @@ function auth_get_current_user_id() {
  * @access public
  */
 function auth_user_id_from_cookie( $p_cookie_string ) {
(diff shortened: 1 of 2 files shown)


core/authentication_api.php
```diff
@@ -1113,11 +1113,6 @@ function auth_get_current_user_id() {
  * @access public
  */
 function auth_user_id_from_cookie( $p_cookie_string ) {
-	# Save a db query if value provided doesn't look like a cookie
-	if( strlen( $p_cookie_string ) != AUTH_COOKIE_LENGTH ) {
```
atrol: `AUTH_COOKIE_LENGTH` is no longer used after this change.
You could remove it from constant_inc.php.

vboctor: Fixed

PR #1289, opened 2018-02-09
Remove admin check for Application Errors on Login page

atrol:
This test is no longer necessary, as we are now enforcing handling
method DISPLAY_ERROR_HALT for all error types since commit
c816588d2616225f33896e7f4f4cd5a45447bce6.

Fixes #23955


atrol:
Will force push later, as I choosed wrong branch to merge. (master instead of master-2.11)

PR #1290, opened 2018-02-10
Login page fixes

vboctor:
This login page was broken when:
- Show realname is ON.
- Show realname threshold not NOBODY.
- Anonymous access is disabled.

This was caused by a combination of an old and a new bug.

OLD Issue
- [23960](https://mantisbt.org/bugs/view.php?id=23960) EVENT_AUTH_USER_FLAGS should always be passed username rather than name

NEW Issue
- [23958](https://mantisbt.org/bugs/view.php?id=23958) Login page broken

### DIFF
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -997,7 +997,8 @@ function user_get_name( $p_user_id ) {
  * @return bool true to show, false otherwise.
  */
 function user_show_realname() {
-	return access_has_project_level( config_get( 'show_user_realname_threshold', null, null, ALL_PROJECTS ) );
+	return auth_is_user_authenticated() &&
+	       access_has_project_level( config_get( 'show_user_realname_threshold', null, null, ALL_PROJECTS ) );
 }
 
 /**
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -100,7 +100,7 @@ function auth_flags( $p_user_id = null, $p_username = '' ) {
 	}
 
 	if( $t_user_id ) {
-		$t_username = user_get_name( $t_user_id );
+		$t_username = user_get_username( $t_user_id );
 		$t_email = user_get_email( $t_user_id );
 	} else {
 		$t_username = $p_username;


core/user_api.php
```diff
@@ -997,7 +997,8 @@ function user_get_name( $p_user_id ) {
  * @return bool true to show, false otherwise.
  */
 function user_show_realname() {
-	return access_has_project_level( config_get( 'show_user_realname_threshold', null, null, ALL_PROJECTS ) );
+	return auth_is_user_authenticated() &&
```
atrol: Can you explain please why the additional check for auth_is_user_authenticated() is needed?
I would expect that access_has_project_level() returns false if user is not logged in.

atrol:
> This login page was broken when:
> - Show realname is ON.

Not true, it can also be OFF

This is enough for it

> - Show realname threshold not NOBODY.

core/user_api.php
vboctor: That doesn't seem to be the case.  At the time when I implemented the fix I didn't want to change at lower level, though I considered it.

vboctor:
I have just merged the EVENT_AUTH_USER_FLAGS fix which is the one necessary to fix this issue.

atrol:
@vboctor could you please double check.
It seems that after all those user name related changes, option show_realname is no longer used in code.

I am not sure, if we loose any former functionality with this, or if the option could be completely removed as `$g_show_realname = OFF;` and `$g_show_user_realname_threshold = NOBODY;` might be the same.

dregad:
> I am not sure, if we loose any former functionality with this

If I recall correctly, the idea was that we should always display the username, and if user access permits then the realname is shown via a tooltip. So there was indeed a loss in functionality (in the sense that it is no longer possible to always display the realname), but it was intentional on @vboctor's part.

The only issue that I see now, is that there is some cleanup that needs to be done; obsoleting `$g_show_realname` is one, and the other is to update the documentation (PHPDoc blocks, admin guide and some comments in the code (e.g. https://github.com/mantisbt/mantisbt/blob/master/core/user_api.php#L974)

PR #1293, opened 2018-02-10
Show Attachments events in Timeline

dregad:
This PR follows-up and replaces #1136.

The original PR was based on a fairly old commit in master branch, around release 2.5. There were several outstanding revision comments by @vboctor which had not been addressed or even acknowledged by @Zipher04.

Changes from the original submission include @atrol and @vboctor's feedback 
- use of local variables
- use fa-file-o icon
- removed the collapsing of Timeline events of the same type as it introduced a regression on Tag events
- Include file name in event message
- Add handling for attachment removal events

### DIFF
--- a/core/classes/IssueAttachmentTimelineEvent.class.php
+++ b/core/classes/IssueAttachmentTimelineEvent.class.php
@@ -27,7 +27,7 @@
  * @package MantisBT
  * @subpackage classes
  */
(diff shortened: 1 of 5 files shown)


dregad:
Sorry about the noise. I messed up the remote branch name when pushing. Correct PR is #1294

PR #1294, opened 2018-02-11
Show Attachments events in Timeline

dregad:
This PR follows-up and replaces #1136.

The original PR was based on a fairly old commit in master branch, around release 2.5. There were several outstanding revision comments by @vboctor which had not been addressed or even acknowledged by @Zipher04.

Changes from the original submission include @atrol and @vboctor's feedback 
- use of local variables
- use fa-file-o icon
- removed the collapsing of Timeline events of the same type as it introduced a regression on Tag events
- Include file name in event message
- Add handling for attachment removal events 

Fixes [#23161](https://mantisbt.org/bugs/view.php?id=23161)

### DIFF
--- a/core/classes/IssueAttachmentTimelineEvent.class.php
+++ b/core/classes/IssueAttachmentTimelineEvent.class.php
@@ -39,7 +39,7 @@ class IssueAttachmentTimelineEvent extends TimelineEvent {
 	 * @param integer $p_user_id   An user identifier.
 	 * @param integer $p_issue_id  A issue identifier.
 	 * @param string  $p_filename  Attachment's file name.
-	 * @param boolean $p_type      Event type (FILE_ADDED, FILE_DELETED)
+	 * @param integer $p_type      Event type (FILE_ADDED, FILE_DELETED)
 	 */
 	public function __construct( $p_timestamp, $p_user_id, $p_issue_id, $p_filename, $p_type ) {
 		parent::__construct( $p_timestamp, $p_user_id );
@@ -52,6 +52,7 @@ class IssueAttachmentTimelineEvent extends TimelineEvent {
 	/**
 	 * Returns html string to display
 	 * @return string
+	 * @throws Mantis\Exceptions\ServiceException
 	 */
 	public function html() {
 		switch( $this->type ) {


core/timeline_api.php
```diff
@@ -57,13 +57,15 @@ function timeline_events( $p_start_time, $p_end_time, $p_max_events, $p_filter =
 	$t_result = history_query_result( $t_query_options );
 	$t_count = 0;
 
+	$t_previous_history_event = null;
```
atrol: I don't see where this variable is used.

core/timeline_api.php
```diff
@@ -101,11 +103,22 @@ function timeline_events( $p_start_time, $p_end_time, $p_max_events, $p_filter =
...
+					$t_history_event['old_value'],
+					$t_type == FILE_ADDED
+				);
+ 				break;
 		}
 
 		# Do not include skipped events
 		if( $t_event != null && !$t_event->skip() ) {
 			$t_timeline_events[] = $t_event;
+			$t_previous_history_event = $t_history_event;
```
atrol: Same as above, I don't see where this variable is used.

core/timeline_api.php
dregad: Leftover of reverting event collapsing (3f3c024). Fixed.

core/timeline_api.php
dregad: Fixed.

core/classes/IssueAttachmentTimelineEvent.class.php
```diff
@@ -0,0 +1,70 @@
...
+	private $add;
+
+	/**
+	 * @param integer $p_timestamp Timestamp representing the time the event occurred.
+	 * @param integer $p_user_id   An user identifier.
+	 * @param integer $p_issue_id  A issue identifier.
+	 * @param string  $p_filename  Attachment's file name.
+	 * @param boolean $p_add       True if adding, false if removing attachment.
+	 */
+	public function __construct( $p_timestamp, $p_user_id, $p_issue_id, $p_filename, $p_add ) {
```
vboctor: Why not get the event type instance of an add flag? i.e. FILE_ADDED and FILE_DELETED?

dregad: You mean instead of the $p_add boolean ?

vboctor: Yes

core/classes/IssueAttachmentTimelineEvent.class.php
```diff
@@ -0,0 +1,70 @@
...
+	 * @return string
+	 */
+	public function html() {
+		$t_string = $this->add ? 'timeline_issue_file_added' : 'timeline_issue_file_deleted';
+		$t_bug_link = string_get_bug_view_link( $this->issue_id );
+
+		$t_html = $this->html_start( 'fa-file-o' );
+		$t_html .= '<div class="action">'
+			. sprintf( lang_get( $t_string ),
+				user_get_name( $this->user_id ),
```
vboctor: Use prepare username instead of user_get_name()

dregad:
Rebased and addressed both @vboctor's review comments.

core/classes/IssueAttachmentTimelineEvent.class.php
dregad: Fixed in 8b1e0080745f4dd22c1b0f8dfe80efb0136f8847

core/classes/IssueAttachmentTimelineEvent.class.php
dregad: Fixed in bdd0668aaa23f5b62cbe54c09ac57b02387f4dbe

core/classes/IssueAttachmentTimelineEvent.class.php
```diff
@@ -0,0 +1,82 @@
...
+
+		$this->issue_id = $p_issue_id;
+		$this->filename = $p_filename;
+		$this->type = $p_type;
+	}
+
+	/**
+	 * Returns html string to display
+	 * @return string
+	 */
```
atrol: missing `@throws`

dregad:
> It's a bit strange that we show removal of attachments but don't show removal of notes

I guess you're right. But that's not in this PR's scope, please open a separate issue to fix it.

atrol:
Like I wrote, it's *a bit* strange.

I am asking myself, if displaying that someone deleted a note is helpful information.

The difference between attachments and notes is, that you see the name of the deleted attachment, but there is no (and can't be) any additional information about the note.

So maybe displaying the deletion of a note would just make sense, if we don't delete notes, but just mark them as deleted.
After that, we could provide a link in Timeline to view the content of the deleted note.

dregad:
@atrol thanks for merging the PR. For next time, I think it's better to create a merge commit when there's a lot of commits in the branch and we want to keep the sausage-making visible; it helps showing the change as a single unit in the git history (I've managed to fix it, since there's been no more pushes since).

> After that, we could provide a link in Timeline to view the content of the deleted note.

I don't think that's such a good idea. In my experience, deleting a note usually happens because you don't want the contents to be seen, so making that available in the timeline defeats the purpose.

PR #1295, opened 2018-02-11
ADDED DETAILS REGARDING $g_handle_bug_threshold

fmancardi:
Added details about the situation when this config option is an array

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2697,7 +2697,17 @@ $g_private_bug_threshold = DEVELOPER;
 
 /**
  * access level needed to be able to be listed in the assign to field.
- * @global integer $g_handle_bug_threshold
+ * if a single value provided, this means that all users that have 
+ * access level >= this value, will be listed.
+ * example:
+ * $g_handle_bug_threshold = DEVELOPER;
+ * 
+ * if an array is provided, then user access level need to be in the set
+ * provided in the array, for the user to be listed.
+ * example:
+ * $g_handle_bug_threshold = array(DEVELOPER,MANAGER);
+ *
+ * @global mixed $g_handle_bug_threshold
  */
 $g_handle_bug_threshold = DEVELOPER;
 


dregad:
@fmancardi can you please avoid typing with ALL CAPS. It is more difficult to read, and gives the impression that you're shouting. Thanks.

config_defaults_inc.php
```diff
@@ -2697,7 +2697,17 @@
...
 /**
  * access level needed to be able to be listed in the assign to field.
- * @global integer $g_handle_bug_threshold
+ * if a single value provided, this means that all users that have 
+ * access level >= this value, will be listed.
+ * example:
+ * $g_handle_bug_threshold = DEVELOPER;
+ * 
+ * if an array is provided, then user access level need to be in the set
+ * provided in the array, for the user to be listed.
```
dregad: * Start sentence with capital letter.
* _then **the** user **'s** access level need**s**_
* provided -> defined

config_defaults_inc.php
```diff
@@ -2697,7 +2697,17 @@
 
 /**
  * access level needed to be able to be listed in the assign to field.
- * @global integer $g_handle_bug_threshold
+ * if a single value provided, this means that all users that have 
+ * access level >= this value, will be listed.
```
dregad: - It's a proper sentence, so start with capital letter.
- Remove _this means that_
- Missing _**is** provided_ 
- Remove comma before _will be listed_

cproensa:
Maybe that phpdoc hasn't been reviewed for a long time.

Shouldn't this option mean: access level needed to handle issue?
The fact that those users are listed in the "assign to" controls, is a consequence of that, but not it's primary function

I would not detail how a threshold works (being able to contain a numeric value, or an array of access levels). There are many options that are thresholds and work the same way, we wouldn't not want to replicate the same explanation for each of them.
It's be better if we make sure that the options are documented as being a  threshold "type", and explain in documentation (if it hasn't been already), in one place, how a threshold type works.

PR #1296, opened 2018-02-12
Minor performance and code enhancements of config functions

atrol:
Fixes #23969

### DIFF
--- a/core/config_api.php
+++ b/core/config_api.php
@@ -44,7 +44,6 @@ $g_cache_config_eval = array();
 $g_cache_config_access = array();
 $g_cache_bypass_lookup = array();
 $g_cache_filled = false;
-$g_cache_can_set_in_database = '';
 
 # cache environment to speed up lookups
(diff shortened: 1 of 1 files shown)


core/config_api.php
```diff
@@ -154,19 +148,18 @@ function config_get( $p_option, $p_default = null, $p_user = null, $p_project =
...
+							return (float)$t_raw_value;
 							break;
 						case CONFIG_TYPE_INT:
-							$t_value = (int)$t_raw_value;
+							return (int)$t_raw_value;
 							break;
 						case CONFIG_TYPE_COMPLEX:
-							$t_value = json_decode( $t_raw_value, true );
+							return json_decode( $t_raw_value, true );
 							break;
```
dregad: Remove the `break` in the above 3 cases - they are dead code due to the use of `return` statement.

atrol: Done

Changed also function config_get_access to the same logic like function config_is_set (return instead of break 2)

PR #1298, opened 2018-02-14
Remove unusable option session_handler

atrol:
Fixes #23966

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -270,15 +270,6 @@ $g_manual_url = 'doc/en-US/Admin_Guide/html-desktop/';
 ##############
 
 /**
- * Session handler.  Possible values:
- *  'php' -> Default PHP filesystem sessions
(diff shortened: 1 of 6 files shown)


atrol:
> I would only suggest rewording the commit message

I will add the additional information.

> `Remove unused option session_handler`

Not sure if this should be changed.
I thought that unusable is the better word than unused.
The option is used in code, but setting to any other value than the default 'php' does not work, thus it's unusable.
WDYT?

dregad:
Yes you're right

Dimas-sc:
Sorry if it isn't the place to ask it but.. Is it possible to remember the session after browser/PC restart? Every time I've to log in. Thx.

atrol:
> Sorry if it isn't the place to ask it

It isn't, this is the right place to ask such kind of questions https://www.mantisbt.org/forums

To answer your question: It is possible, there is a check box for it on the page where you enter the password.

PR #1299, opened 2018-02-15
Email for a new private bugnote was send to a non authorized reporter

wuttke:
see https://www.mantisbt.org/bugs/view.php?id=22898

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -1993,7 +1993,7 @@ function bug_update_date( $p_bug_id, $p_last_modified = 0 ) {
 	db_param_push();
 	$t_query = 'UPDATE {bug} SET last_updated=' . db_param() . ' WHERE id=' . db_param();
 	if ($p_last_modified == 0) {
-		$p_last_modified = db_now()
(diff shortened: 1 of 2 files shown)


core/bug_api.php
```diff
@@ -1989,10 +1989,13 @@ function bug_reopen( $p_bug_id, $p_bugnote_text = '', $p_time_tracking = '0:00',
  * @access public
  * @uses database_api.php
  */
-function bug_update_date( $p_bug_id ) {
+function bug_update_date( $p_bug_id, $p_last_modified = 0 ) {
```
dregad: Since you change the function's signature, you need to update the PHPDoc also.

dregad:
Also, please reword the commit message to explain what you changed and why, and adjust the issue reference (`Fixes #22898`) otherwise it won't be picked up by source integration.
Thanks

vboctor:
I have added a high level question to the bug.

vboctor:
The associated issue was closed with a commit from @dregad - closing PR.

PR #1300, opened 2018-02-15
Tag attach email notifications

srhesse:
1A Auto needed the ability to send email notifications when a tag is attached. (This had previously been mentioned in the forum as a request.)  The code added sets the notify_flags for a new  'tag_attached' action and then allows it to be overridden by the user preferences. 

Our use case is to have a single user set up as a viewer on a project for emailing another division when a tag is attached. Most other users have the notification  preference off by default.  There is not much need for the action to be set for the reporter, creator, etc except for possibly those monitoring.


cproensa:
There are two problems with this PR:
1) Schema changes can't be done that way. Schema must be modified incrementally with an upgrade step.
However, we have frozen schema changes at the moment. 
2) The dev team has been discussing to change the user preferences system. Instead of having explicit table columns like it's now, switch into a system similar to config table, to allow for easier improvements like this without having to upgrade schema.

vboctor:
Thanks @srhesse for your contribution.

The PR can be updated to INCLUDE:

- generating an email notification when a tag is attached.
- generating an email notification when a tag is detached.
- ability to configure email notifications with tag attached and tag detached.

The PR should NOT include:

- Email schema changes.
- Changes to user preferences page. -- this can be added later once we have a good story for supporting new user preferences.

I also suggest fixing indentations where spaces are used instead of tabs.

vboctor:
You should also think about the verbose vs. non-verbose version of such emails.

PR #1302, opened 2018-02-18
Enabled external authentication via _SERVER['REMOTE_USER'].

raspopov:
A web server authenticated user by self when 'login_method' is equal to 'HTTP_AUTH' and a 'REMOTE_USER' environment variable is present. Also added code to parse Windows user name in form of 'Domain\User'. Tested on latest MantisBT 2.11.0 and IIS 8.5 (Windows Server 2012) + PHP 5.6.

Task: [0022398: HTTP_AUTH not working](https://www.mantisbt.org/bugs/view.php?id=22398)


core/authentication_api.php
```diff
@@ -342,6 +342,13 @@ function auth_prepare_username( $p_username ) {
 			if( !auth_http_is_logout_pending() ) {
 				if( isset( $_SERVER['PHP_AUTH_USER'] ) ) {
 					$t_username = $_SERVER['PHP_AUTH_USER'];
+				} else if( isset( $_SERVER['REMOTE_USER'] ) ) {
```
vboctor: The processing of `REMOTE_USER` is currently done under `BASIC_AUTH` (see above).  Is there documentation for this and a reason to classify it under `HTTP_AUTH` vs. `BASIC_AUTH`?

core/authentication_api.php
```diff
@@ -342,6 +342,13 @@ function auth_prepare_username( $p_username ) {
...
+					if ( preg_match( '/^(.*)\\\\(.*)$/i', $_SERVER['REMOTE_USER'], $user_match ) ) {
```
vboctor: Replace `$user_match` with `$t_user_match`.

core/authentication_api.php
```diff
@@ -409,6 +416,8 @@ function auth_auto_create_user( $p_username, $p_password ) {
 
 	if( $t_login_method == BASIC_AUTH ) {
 		$t_auto_create = true;
+    } else if ( $t_login_method == HTTP_AUTH && isset( $_SERVER['REMOTE_USER'] ) ) {
```
vboctor: Is the `isset()` check necessary?

raspopov:
It's a pretty interesting question what providing an better solution: a patched HTTP_AUTH or patched BASIC_AUTH (which is a subset of HTTP_AUTH), or even adding a new authentication plugin? I need a little more time to prepare good answer including installing VMWare with test MantisBT installation.

But one aspect is already known - HTTP_AUTH patch is a smaller one of all variants.

core/authentication_api.php
raspopov: Yes, since original HTTP_AUTH execution path is not guaranteed to has a valid 'REMOTE_USER' variable which is required to rightful web-server authentication.

vboctor:
@raspopov I would love to enrich the auth plugin framework to:

- Support lookup of users from directory.
- Support authentication models where user doesn't need to enter username.

It would be great if we can also document an easy way to test such authentication models for developers that don't have the corporate environment.

raspopov:
I decided to replace this PR with a separate external authentication plugin [MantisExtAuth](https://github.com/raspopov/MantisExtAuth).

PR #1304, opened 2018-02-19
Notifications and EmailNotifications skeleton

vboctor:
This is not a functioning code but a proposal for starting to clean up notifications.

- Notifications - These are classes that capture information about a change that should trigger notifications like email notifications and possibly others in the future (e.g. plugin notifications).  A notification is a single logical change that can trigger N emails, a slack notification, etc.

- Dispatching - Once a notification is triggered, the first step is to determine who should receive the notification.  In the current change it is the N users that should be notified via email.  This is currently done implicitly in an email_(change_type)().  In the future, we may refactor to have some dispatcher.

- Email Generation - The class that should handle a specific notification type for a single target user.  The notification derived class would generate the subject, meta-data, text body, html body and produce an EmailData class.

- Email APIs - The email API would handle queuing of email and sending of emails based on EmailData.

Let me know your feedback.

### DIFF
--- a/core/constant_inc.php
+++ b/core/constant_inc.php
@@ -720,3 +720,6 @@ define( 'DB_TYPE_MYSQL', 1 );
 define( 'DB_TYPE_PGSQL', 2 );
 define( 'DB_TYPE_MSSQL', 3 );
 define( 'DB_TYPE_ORACLE', 4 );
+
+# Notification Types
+define( 'NOTIFICATION_TYPE_ISSUE_NOTE_ADDED', 'issue_note_added' );
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -970,18 +970,18 @@ function email_bugnote_add( $p_bugnote_id, $p_files = array(), $p_exclude_user_i
 
 	ignore_user_abort( true );
 
+	$t_notification_data = array(
+		'issue_note_id' => $p_bugnote_id,
+		'files' => $p_files
+	);
+
+	$t_notification = new IssueNoteAddedNotification( $t_data );
+
 	$t_bugnote = bugnote_get( $p_bugnote_id );
 
 	log_event( LOG_EMAIL, sprintf( 'Note ~%d added to issue #%d', $p_bugnote_id, $t_bugnote->bug_id ) );
 
 	$t_project_id = bug_get_field( $t_bugnote->bug_id, 'project_id' );
-	$t_separator = config_get( 'email_separator2' );
-	$t_time_tracking_access_threshold = config_get( 'time_tracking_view_threshold' );
-	$t_view_attachments_threshold = config_get( 'view_attachments_threshold' );
-	$t_message_id = 'email_notification_title_for_action_bugnote_submitted';
-
-	$t_subject = email_build_subject( $t_bugnote->bug_id );
-
 	$t_recipients = email_collect_recipients( $t_bugnote->bug_id, 'bugnote', /* extra_user_ids */ array(), $p_bugnote_id );
 	$t_recipients_verbose = array();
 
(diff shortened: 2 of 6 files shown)


cproensa:
I see a different layout for all this notification scenario.
I'd like to prepare for the future, and decouple notification generators, notification items, outgoing messages...

-> **Notification generators**, as the general application events that would create the individual notification items for each user, when an event occurs.

-> **Notification item**, as elemental, informational content targeted to one user, is independent of how it's going to be presented, or communicated.
This means, at some point we could have a notificactions queue for the N-items created for each user. We could:
- Show them in the web-UI
- Send alerts by email, or other methods
- Process related N-items in some smart ways: replace outdated ones with newer ones, unify several into one (eg: issue changes digest)

The N-items by themselves should not implement those though. That's why we should have:


-> **Notification processors**, that would take one or several N-items and perform some action or transformation:
- Email processor, that sends an email with the contents of the N-item
- Other communications, slack, whatever. Note that we could allow a plugin to provide a custom N-processor (implementing the standard interface)
- Unification. For example: merge several notifications relating to a bug update into only one with a change summary.
- etc

. . .

Of course, this is oversimplified, and there are many caveats... but my main points are:
- Keep a precise naming convention ("notification" is too generic)
- Separate data from logic and communication.

. . .

I'd rather have the system based on interfaces.
Even if we can have one or several abstract classes as core logic, as implementation of those interfaces, I generally prefer designing on interfaces instead of abstract classes for reasons:
1) PHP class inheritance can't be multiple. In the other hand, a class can implement multiple interfaces
2) According to (1), interfaces dont impose limitations on how the final classes are implemented, nor artificial limits forcing single purposed classes, where one class may implement several related behaviours.
3) Even arguing that (2) is not relevant in our own codebase, at some point we should allow plugins to provide custom behaviour through object injections. Such objects will be implementations of well defined, standard, interfaces.

vboctor:
Thanks @cproensa for your feedback.

> Notification generators, as the general application events that would create the individual notification items for each user, when an event occurs.

I haven't implemented this piece yet, but it is planned once notifications are used beyond email notifications.  A Notification is a logical change and is mainly a data container.  The generators would eventually get 1 notification and generate N notification emails, or other types of notifications.

> Notification item

This is the equivalent of the Notification class in my PR.  It is decoupled from how it is communicated as you mentioned.  It can result into 0..N actually communication events.

> Notification processors

I'm not sure how this will work.  What I was suggesting is having the dispatcher take the logical notification and create N emails, slack notifications, etc.  So we should be covered there

> Unification. For example: merge several notifications relating to a bug update into only one with a change summary.

Not sure how this would be done.

> Keep a precise naming convention ("notification" is too generic)

Not sure if `NotificationItem` is any better.  I kind of like Notification, but am open for other names, e.g. `LogicalNotification`.

> Separate data from logic and communication.

Do you have specific examples where this was not the case?  As you can see the data is in Notification class, and the logic to generate an email is in the notification email class type.

> I'd rather have the system based on interfaces.

I avoid interfaces for areas that are likely to be extended for the following reasons:

- Creating a new interface method breaks all plugins or extensions that inherit from the interface.
- You end up creating a base class that provide common logic anyway, and hence, classes will have to inherit from it or copy and paste.
- Abstract class provides a good mix of abstract functions and code that can wrap them to provide common usage of such functions, e.g. see Commands->execute() which follows a common pattern with commands just having to implemented validate and process.

It would help if you can suggest some more specific tweaks to the proposals.  I also don't want to implement a big framework to start with.  I want to have a good foundation for logical notifications and email notifications, then add more to enable other kinds of notifications or scenarios.  Of course, once we get the basics in, we can work together to drive this faster.

cproensa:
> A Notification is a logical change and is mainly a data container. The generators would eventually get 1 notification and generate N notification emails, or other types of notifications.

I mean:
 `...a notification is a logical change that [eventually] generates N notifications`
is ambiguous

A logical change is "a logical change", which could create "notifications" for N users. Each of those "notifications" could generate an "email", or other type of communication element, which may happen at a later time than the source "notification".
So as i see it, a notification is the data container, targeted to one user, about one piece of information. 
That's why, to avoid the ambiguity, i suggested a more precise naming convention (whatever it is)



>> Notification item

> This is the equivalent of the Notification class in my PR. It is decoupled from how it is communicated as you mentioned. It can result into 0..N actually communication events.

From you example:
email_bugnote_add() creates a single IssueNoteAddedNotification, which is a Notification
and then in the same function:
    $t_email = new IssueNoteAddedEmail( $t_notification, $t_user_id );
    $t_email_data = $t_email->getEmailData();
    email_store_email_data( $t_email_data );

As i described, the "notification(item)" is the end piece of information already targeted to each user.
    
> Separate data from logic and communication.
   
In my scenario:
- Adding the bugnote calls some "NotificationGenerator_for_bugnote( bugid, noteid,...)"
(maybe you could also factor in to work with the commands classes)
- The generator creates one NotificationItem for each user that shpoud be notified, those are sotred in a notification queue.

At a later point, with a batch or exit function (like current email processing), the NotificationItem's queue can be processed into emails.

I'm not determined whether the notification item should be responsible of generating the email texts for itself, or separate Processors should do the generation for each family of notifications.



>>  Unification. For example: merge several notifications relating to a bug update into only one with a change summary.
> Not sure how this would be done.

One user have several notifications queued in a row, for changes on the same bug. Those have been created in a short time frame, and are still unread, and undispatched to email, etc. Then you could merge into one single notification:
- One relevant summary (like we try to do now for a combination of changes)
- A detail of the merged changes. Eg, a timeline-like presentation of the changes


---


> Creating a new interface method breaks all plugins or extensions that inherit from the interface.

Interfaces should be simple enough to be well defined. A change mean a significant modification of behaviour, so it's not unreasonable to break the compatibility

> You end up creating a base class that provide common logic anyway, and hence, classes will have to inherit from it or copy and paste.

No, you can either
- inherit the base abstract class, and override the provided behaviour. this is godd for simple alterations and not having to rewrite much of the inner logic.
- implement the interface with a different code.

Abstract classes also raise problems if you update an existing function definition (add or change type of parameters), or change the behaviour.

> Abstract class provides a good mix of abstract functions and code that can wrap them to provide common usage of such functions, e.g. see Commands->execute() which follows a common pattern with commands just having to implemented validate and process.

Until the needed implementation clashes with an oversimplified abstract class.
If the core app is built expecting that inheriting class, you are in a tough situation.

What i mean is:
Abstract classes are good for providing ready to go implementations of common behaviours.
But Core app should talk only to the interfaces. Having both is not exclusive.

vboctor:
@cproensa If you think about it, the best terminology for a logical change is an "Event" that core and plugins can subscribe to (e.g. note added, issue assigned, issue status change, issue reported, etc).  These events are going to be produced by commands with all the needed before/after data.  There is 1 Event per logical change that is triggered from the method that did the change (e.g. `bugnote_add()`).  The Event should go to a processor that do whatever needed business logic part of which is to generate notifications which can be email, slack, text messages, etc.  A notification uses the source event as the data source, does its own formatting based on its type (e.g. email vs. slack), and is targeted for a single user.  Hence, 1 Event can trigger 0..N notifications of one or more types.

The queuing part since this is orthogonal to other stuff and can be added later as needed.  So makes sense, but I wouldn't worry about it for now.

For interfaces vs. abstract classes, I have learned this the hard way over time.  There is a lot of minor improvements that can be done on abstract class based implementations that would be breaking changes in case of interfaces.  So even though I agree they are not exclusive, the contract should be the abstract class and not the interface, which makes the interface less interesting.  I like using interfaces for non-extensibility scenarios, e.g. unit testing, and other usage.

PR #1305, opened 2018-02-25
Fix show realname check

vboctor:


core/user_api.php
```diff
@@ -997,11 +997,13 @@ function user_get_name( $p_user_id ) {
  * @return bool true to show, false otherwise.
  */
 function user_show_realname() {
-	return access_has_project_level( config_get( 'show_user_realname_threshold', null, null, ALL_PROJECTS ) );
+	return config_get( 'show_realname', null, null, ALL_PROJECTS ) == ON &&
+		access_has_project_level( config_get( 'show_user_realname_threshold', null, null, ALL_PROJECTS ) );
```
atrol: Why do you check for ALL_PROJECTS?
`show_user_realname_threshold` and `show_realname` could be set project specific in earlier versions.

vboctor: I can change it to use current project.  Though not sure this kind of setting makes sense to have on project level given that it impacts pages like View Issues, My View, etc which may span multiple projects.

atrol: > not sure this kind of setting makes sense to have on project level

Right, there are certainly more configuration options where project specific settings are questionable.
But is this the right moment to change it?

If you think we can change now, did you consider to change even more, what we discussed at
https://github.com/mantisbt/mantisbt/pull/1290#issuecomment-366711353 (obsoleting $g_show_realname)?

BTW, it's also questionable (but sometimes can be a nice feature), that a lot of our options can be set user specific.
E.g. you could configure that a single user of an access level is allowed to see the real names.

cproensa: >  like View Issues, My View, etc which may span multiple projects.

...and changelog, roadmap...

i think that formatting the issue accounts for bug level access. Columns in view_issues works that way iirc. 
Filters also take access levels by project and factor them into the query.
I'm not sure if this option applies to filters, but others (show handler, etc) have been recently modified to do so.

dregad: I agree that there is no particular reason to limit this to ALL_PROJECTS in the context of this change.

While I don't particularly care if we want to introduce such a restriction, if we do it should be part of a work item of its own.

dregad:
For the record, [on our tracker](https://mantisbt.org/bugs/my_view_page.php) which is now using latest master, as developers we see a regression from #1290 on the Timeline: 

screenshot-20180228-1230

@vboctor This PR needs to be merged before the next release is cut.

vboctor:
I have updated the change to not limit the check to all projects and now with this change also include check for show realname.  Deployed to official bug tracker which fixed the issue that @dregad pointed out in Timeline.

dregad:
Thank you !

PR #1306, opened 2018-02-25
Fixed case-sensitive LDAP attributes names

raspopov:
A **ldap_get_entries()** function returns attribute names as lowercase keys so next call of **array_key_exists()** fails because of it performs a case-sensitive comparison. For example if a **$g_ldap_realname_field** option is set to '_givenName_' (a well-known LDAP attribute display name) then it finds nothing.


raspopov:
Thank you for review!
New fixed PR is ready, and a new issue created: https://www.mantisbt.org/bugs/view.php?id=24043

vboctor:
Merged via 34978af2072f475050f64a2ad7c5ade8b490e1f2

PR #1307, opened 2018-02-25
Enabled an independent using of LDAP user name and email

raspopov:
Real name and email are now independent from LDAP authentication i.e. when **$g_login_method != LDAP**. It's very useful when no LDAP authentication used but used a web-server authentication (IIS for example) instead with Active Directory.


raspopov:
Re-formatted commit message and added issue: https://www.mantisbt.org/bugs/view.php?id=24044

dregad:
Replaced by #1308

PR #1308, opened 2018-02-26
Getting user name and email from LDAP without LDAP authentication

raspopov:
Removed unneeded LDAP authentication checks to enable independent retrieval of LDAP real name and email.

Fixes #[24044](https://mantisbt.org/bugs/view.php?id=24044).


raspopov:
Also fixed a lost real name field on the **account_page.php** page.

dregad (approved):
Looks good to me.

@vboctor your feedback on this one would be appreciated.

vboctor:
Thanks, @raspopov, for your contribution.

1. The decoupling of email / realname retrieval from LDAP from the authentication sounds good.
2. In `config_defaults_inc.php`, the comment `Note that MantisBT will update the database with the data retrieved from LDAP when ON.` is associated with both `$g_use_ldap_email` and `$g_use_ldap_realname`. However, I'm not sure where this happens - if that is not happening, we should fix the comments. -- not introduced by this PR but worth fixing.
3. Pages that manage user information like create/update user have checks that are specific to LDAP where I think they shouldn't care about that. They should use methods like `user_can_update_realname()` and `user_can_update_email()` and have the LDAP logic encapsulated in the `user_api.php`. -- not introduced by this PR but worth fixing.
4. One interesting aspect, is that the logic now fallsback to the DB when LDAP returns an empty email or realname. However, there is no way to set or unset these values. -- not introduced by this PR.

vboctor:
It also doesn't seem that the config options for retrieving email or realname from LDAP is documented in the manual.

raspopov:
> It also doesn't seem that the config options for retrieving email or realname from LDAP is documented in the manual.

This description isn't enough? https://github.com/mantisbt/mantisbt/blob/f0c25385387656602f0992f34ed9f2c9ca1588e8/docbook/Admin_Guide/en-US/config/auth.xml#L328

dregad:
@raspopov coming back to this PR following https://github.com/mantisbt/mantisbt/pull/2179#issuecomment-3863977667... At least I did (belatedly) review, and even approved it 😉 

Incidentally, the setup in my dev box's config_inc.php with $g_use_ldap_email = ON is probably a leftover from when I tested this code...

Anyway, it looks like it's been hanging following @vboctor's review, as you have not replied to his comments.

> This description isn't enough?

It is, but I think it would be worth improving to explain that *use_ldap_XXXX* configs can in fact be used independently from LDAP  authentication. The section's introduction paragraph needs updating too, as it states that *The parameters below are **only** used if $g_login_method is set to LDAP.* which would become misleading after this is merged.

https://github.com/mantisbt/mantisbt/blob/975b97cc582e5b70cd8c0b080b5f04073c838ba8/docbook/Admin_Guide/en-US/config/auth.xml#L77-L80

@vboctor regarding
> the comment `Note that  MantisBT will update the database with the data retrieved from LDAP when ON.` is associated with both $g_use_ldap_email and $g_use_ldap_realname. However, I'm not sure where this happens 

The update *is* happening
* when the user's account is updated (UserUpdateCommand)
* when the user logs in (ldap_authenticate_by_username()

raspopov:
> It is, but I think it would be worth improving to explain

Thank you for the clarification. Earlier, I confirmed that all the necessary information was included, but I didn't consider the possibility of excess or irrelevant information. 😔

dregad:
Thanks for updating the docs, I'll review that later.

raspopov:
Resolved merge conflict with latest master.

dregad (approved):

dregad:
Merged in 7a51286897e0e2d6e1da6c7fd100647003e3c0a4

PR #1309, opened 2018-03-01
Truncate values before writing to database history table

atrol:
TEXTAREA custom fields can contain more than 255 characters.
Changing such fields fails, as the new value can't be stored
in history table.

This change fixes the issue by truncating the string.
See #24056 for other options to fix it.

Fixes #24056


atrol:
> but let's merge after the upcoming release

Right, I didn't plan to do it before that.

PR #1310, opened 2018-03-01
Fix past participle agreement

MrBricodage:


atrol:
Translations are managed via TranslateWiki. Please refer our own Wiki [1] for the process to submit new translations or revise existing ones.

[1] http://www.mantisbt.org/wiki/doku.php/mantisbt:translationshowto

PR #1311, opened 2018-03-05
Delete failing emails from queue after N days

vboctor:
Fixes #16070

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -1216,8 +1216,13 @@ function email_send_all( $p_delete_on_failure = false ) {
 		}
 
 		if( !$t_email_sent ) {
-			if( $p_delete_on_failure ) {
-				email_queue_delete( $t_email_data->email_id );
+			# Delete emails that were submitted more than N days ago
+			$t_submitted = (int)$t_email_data->submitted;
+			$t_delete_after_in_days = (int)config_get_global( 'email_retry_in_days' );
+			$t_retry_cutoff = time() - ( $t_delete_after_in_days * 24 * 60 * 60 );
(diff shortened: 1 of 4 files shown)


core/email_api.php
```diff
@@ -1216,8 +1216,13 @@ function email_send_all( $p_delete_on_failure = false ) {
 		}
 
 		if( !$t_email_sent ) {
-			if( $p_delete_on_failure ) {
-				email_queue_delete( $t_email_data->email_id );
+			# Delete emails that were submitted more than 7 days ago (MantisHub change)
```
dregad: What is the point of the _MantisHub change_ comment ?

vboctor: Copy and paste error, since I originally implemented this as a MantisHub specific change without the config option.  Fixed.

core/email_api.php
```diff
@@ -1216,8 +1216,13 @@ function email_send_all( $p_delete_on_failure = false ) {
 		}
 
 		if( !$t_email_sent ) {
-			if( $p_delete_on_failure ) {
-				email_queue_delete( $t_email_data->email_id );
+			# Delete emails that were submitted more than N days ago
+			$t_submitted = (int)$t_email_data->submitted;
+			$t_delete_after_in_days = (int)config_get( 'email_retry_in_days' );
```
atrol: Is this an option where we should allow to set it in database?
I think no.
If yes
- should we allow this to be a project / user specific setting?
- should it be listed in  public_config_names?

I am aware that there are similar options that can be set in database (e.g. email_send_using_cronjob).
But I assume there are just historic reasons for that.

vboctor: Changed to use `config_get_global()`.  No need to make it a public config.

PR #1314, opened 2018-03-08
Correct access check for admin folder

atrol:
Fixes #24089
Fixes #23476

### DIFF
--- a/login_password_page.php
+++ b/login_password_page.php
@@ -170,7 +170,7 @@ if( $f_error || $f_cookie_error || $f_reauthenticate ) {
 }
 
 $t_upgrade_required = false;
-if( config_get_global( 'admin_checks' ) == ON && file_exists( dirname( __FILE__ ) .'/admin' ) ) {
+if( config_get_global( 'admin_checks' ) == ON && file_exists( dirname( __FILE__ ) .'/admin/.' ) ) {
 	# since admin directory and db_upgrade lists are available check for missing db upgrades
 	# if db version is 0, we do not have a valid database.
 	$t_db_version = config_get( 'database_version', 0 );


login_password_page.php
```diff
@@ -170,7 +170,7 @@
 }
 
 $t_upgrade_required = false;
-if( config_get_global( 'admin_checks' ) == ON && file_exists( dirname( __FILE__ ) .'/admin' ) ) {
+if( config_get_global( 'admin_checks' ) == ON && file_exists( dirname( __FILE__ ) .'/admin/.' ) ) {
```
vboctor: would using dir exist make sense here?

atrol: What do you mean with `dir exist`? I didn't find such a function or a similar one, like  `dir_exists`.

`file_exists` should do the job according http://php.net/manual/en/function.file-exists.php
> Checks whether a file or **directory** exists.

vboctor: I was wondering if dir_exists removes the need for `/.` at the end of the folder name.

atrol: > I was wondering if dir_exists ...

As I said, I didn't find such a function.
Can you provide a link to the documentation?

dregad: The use of trailing `/.` is correct, and to my knowledge the only way to do this check properly.

> > I was wondering if dir_exists ...
> As I said, I didn't find such a function.

Does not exist in PHP. It's either _file_exists()_ or _is_dir()_

cproensa:
Confirming it solves #23476

login_password_page.php
vboctor: I meant is_dir(), but if others confirmed that this is fine, I'm ok with it.

PR #1318, opened 2018-03-17
Check early for minimum PHP version

atrol:
Fixes #24128

### DIFF
--- a/admin/check/check_php_inc.php
+++ b/admin/check/check_php_inc.php
@@ -44,14 +44,15 @@ check_print_test_row(
 	'PHP version ' . phpversion() . ' is currently installed on this server.'
 );
 
+# $t_extensions_required lists the extensions required to run Mantis in general
+# The fileinfo extension which is needed for attachments is checked in check_attachments_inc.php
 $t_extensions_required = array(
 	'date',
 	'hash',
 	'pcre',
 	'Reflection',
 	'session',
-	'mbstring',
-	'fileinfo'
+	'mbstring'
 );
 
 foreach( $t_extensions_required as $t_extension ) {
--- a/core.php
+++ b/core.php
@@ -63,13 +63,29 @@ if( file_exists( 'mantis_offline.php' ) && !isset( $_GET['mbadmin'] ) ) {
 
 $g_request_time = microtime( true );
 
+# Load supplied constants
+require_once( dirname( __FILE__ ) . DIRECTORY_SEPARATOR . 'core' . DIRECTORY_SEPARATOR . 'constant_inc.php' );
+
+# Enforce our minimum PHP requirements
+if( version_compare( PHP_VERSION, PHP_MIN_VERSION, '<' ) ) {
+	echo '<strong>FATAL ERROR: Your version of PHP is too old. '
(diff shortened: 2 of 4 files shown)


atrol:
> Maybe consider enforcing `fileinfo` as well

I am not sure if it's a real use case, but a user could set `$g_allow_file_upload = OFF;` and use Mantis without `fileinfo` extension without seeing any problems.

dregad:
You're right, but in that case I think the current [admin check for fileinfo extension](https://github.com/mantisbt/mantisbt/blob/master/admin/check/check_php_inc.php#L54) is not correct as it does not take this scenario into consideration.

core.php
```diff
@@ -65,12 +65,30 @@
 
 ob_start();
 
-# Load Composer autoloader
-require_once( dirname( __FILE__ ) . DIRECTORY_SEPARATOR . 'vendor/autoload.php' );
-
 # Load supplied constants
 require_once( dirname( __FILE__ ) . DIRECTORY_SEPARATOR . 'core' . DIRECTORY_SEPARATOR . 'constant_inc.php' );
```
vboctor: Including core.php typically includes a lot of files.  If we have a dependency on a new version of PHP, there can be a syntax error there.  It would be great if we can check php version before including core.php.

atrol: > Including core.php typically includes a lot of files. If we have a dependency on a new version of PHP, there can be a syntax error there

That's exactly why I opened this PR. There was a syntax erorr in guzzle when trying to run Mantis on PHP 5.3

> It would be great if we can check php version before including core.php.

Before including? Wouldn't this mean that the checks had to be added to nearly all our php files?

atrol:
> You're right, but in that case I think the current admin check for fileinfo extension is not correct

Moved the check so that it's just executed if `$g_allow_file_upload = ON`

admin/check/check_php_inc.php
```diff
@@ -50,8 +50,7 @@
 	'pcre',
 	'Reflection',
 	'session',
-	'mbstring',
-	'fileinfo'
```
dregad: Please add a comment to document the fact that _fileinfo_ is checked in check_attachments_inc.php

dregad:
I think that http://mantisbt.org/docs/master/en-US/Admin_Guide/html-desktop/#admin.install.requirements should be adjusted to reflect that _fileinfo_ is now effectively required only if `$g_allow_file_upload = ON`

admin/check/check_php_inc.php
atrol: Done

atrol:
@dregad concerning Admin Guide

Isn't the current explanation clear enough?
We don't mention any configuration option in this section of Admin Guide at the moment, just functionality, e.g. _curl_ for Twitter and _GD_ for captcha.

Of course, while looking at the requirements, there is some general rework needed, but should not be  part of this PR.

dregad:
I'm fine with not referencing the actual config option, but based on the present PR's discussion, _fileinfo_ should no longer be under _mandatory_ extensions, but moved to _optional_ with a mention like `required for file attachments`

atrol:
> but moved to _optional_ with a mention like `required for file attachments`

Agree, will do.

PR #1319, opened 2018-03-20
Fix order by clause not in select list

cproensa:
Strict sql syntax requires the order by clause to appear in the select
list. Otherwise, some rdbms may fail.
Eg: pgsql, or mysql with sql_mode=ONLY_FULL_GROUP_BY

Fixes: #24140

### DIFF
--- a/core/classes/BugFilterQuery.class.php
+++ b/core/classes/BugFilterQuery.class.php
@@ -1518,11 +1518,13 @@ class BugFilterQuery extends DbQuery {
 			case 'category_id':
 				# This join will be reduced as unique, if category search is active
 				$this->add_join( 'LEFT JOIN {category} ON {bug}.category_id = {category}.id' );
+				$this->add_select( '{category}.name' );
 				return '{category}.name';
 				break;
 
 			case 'project_id':
 				# project table is already joined by default
+				$this->add_select( '{project}.name' );
 				return '{project}.name';
 				break;
 


core/classes/BugFilterQuery.class.php
```diff
@@ -1518,11 +1518,13 @@ protected function helper_sort_column_alias( $p_prop ) {
 			case 'category_id':
 				# This join will be reduced as unique, if category search is active
 				$this->add_join( 'LEFT JOIN {category} ON {bug}.category_id = {category}.id' );
+				$this->add_select( '{category}.name' );
 				return '{category}.name';
 				break;
```
vboctor: Why is break needed after return?  Applies to both cases included in the diff.

core/classes/BugFilterQuery.class.php
```diff
@@ -1518,11 +1518,13 @@ protected function helper_sort_column_alias( $p_prop ) {
 			case 'category_id':
 				# This join will be reduced as unique, if category search is active
 				$this->add_join( 'LEFT JOIN {category} ON {bug}.category_id = {category}.id' );
+				$this->add_select( '{category}.name' );
 				return '{category}.name';
```
vboctor: Why do you need to add `{category}.name` in the select and also return it?

core/classes/BugFilterQuery.class.php
dregad: The `break` is indeed dead code (also at line 1553)

cproensa: > Why is break needed after return?

I usually put always a `break` unless it's explicitly intended to fall through next case.
What is our guideline here?

core/classes/BugFilterQuery.class.php
cproensa: > Why do you need to add {category}.name in the select and also return it?

This function returns the alias that must be used in the sort clause. Note that this function does not build the order by clauses, however, it does manages the additional join/select stuff needed to be able to use that sorting alias.

atrol:
Closed via https://github.com/mantisbt/mantisbt/commit/74ca8bc6bc28acc28d594ee084df27666fef8ad6

PR #1321, opened 2018-03-27
Display usernames in history

atrol:
Fixes #24167


atrol:
@vboctor I will hardly be able to work on this before you planned to cut 2.12.1 / 2.13.0.
I think it's better at the moment than it was before.
Merge as is and open a new issue for further enhancements, or keep open and change in later release?

dregad:
> I think it's better at the moment than it was before.

Agreed.

I'm fine with merging as-is and improve later, but at the same time it's not a major issue so postponing would be OK as well.

PR #1322, opened 2018-03-27
Correct username display style

atrol:
Fixes #24161


atrol:
> we should likely remove it from everywhere.

Done

PR #1327, opened 2018-04-01
Fix rendering of @ mentions

atrol:
Fixes #24201

I am not complete sure if this the right approach to fix it.
I just moved the @ mention processing  after the markdown procession as `setSafeMode` of  Parsedown destroyed our @ mention links.


vboctor:
@atrol Did the safe mode impact other links to issues or notes or just @ mentions?

atrol:
Other links are also affected, I will provide a new PR

PR #1329, opened 2018-04-04
Correct default value of datetime_picker_format in Admin Guide

atrol:
Fixes #24220


atrol:
> Typo in commit message

Thanks for the hint, fixed

PR #1330, opened 2018-04-09
Send always usernames in email notifications

atrol:
Don't send realnames if $g_show_realname = ON;

This is a quick workaround as a clean solution needs some redesign.

At the moment $g_show_user_realname_threshold is considered based
on the current user.
This is wrong, as the option must be considered based on the
recipient of the notification.

Fixes #24239


atrol:
> Does this cover the names of users that are listed in issue history?

@vboctor no, but didnt't I fix this in https://www.mantisbt.org/bugs/view.php?id=24167 ?

There is https://www.mantisbt.org/bugs/view.php?id=24188 open, but don't expect me to work on it, even though you assigned me. ;-)

PR #1331, opened 2018-04-11
Use rgb color values for Markdown quote styling

atrol:
Workaround as using hex values for colors starting with # introduces
unwanted side effects.

Fixes #24233

Let's fix this one now independant from the other conceptual Markdown issues, as we don't know at the moment how (maybe have to wait for new Parsedown version) and when the other issues can be fixed.

### DIFF
--- a/plugins/MantisCoreFormatting/core/MantisMarkdown.php
+++ b/plugins/MantisCoreFormatting/core/MantisMarkdown.php
@@ -169,7 +169,7 @@ class MantisMarkdown extends Parsedown
 
 		if( $block = call_user_func( 'parent::' . $fn, $line, $block ) ) {
 			# TODO: To open another issue to track css style sheet issue vs. inline style.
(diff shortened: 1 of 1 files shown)


ajtruckle:
OK. All a bit over my head. All I know is that when I pasted my own XML into the ParseDown portal it all looks correct. So did not understand why MantisBT didn't.

atrol:
@ajtruckle this PR does _not_ fix the XML issue.
It fixes #24233 where I wrote that this issue will just be used to track the quoting issue
https://www.mantisbt.org/bugs/view.php?id=24233#c59504

The XML issue is tracked at https://www.mantisbt.org/bugs/view.php?id=24240

>  I pasted my own XML into the ParseDown portal it all looks correct. So did not understand why MantisBT didn't.

We do a bit more in addition to what Parsedown/Markdown does, e.g. we look also for links, # issues, ~ notes and @ mentions.
There are places where these two handlings conflict, that's why the issue occurs in MantisBT, but not the Parsedown portal.

PR #1333, opened 2018-04-15
Authenticator support with autoprovision; admin dir security

V1pr:
- Added authenticator page support (changes form destination in login_password_page)
- added autoprovisioning support for CredentialsPage
- added automatic admin directory security with .htaccess if a specific file (install.done) exists. This is created upon a successful install

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,9 @@ api/rest/swagger
 /vendor/
 composer.phar
 
+# Admin
+admin/install.done
+
 # Settings below are for convenience when switching branches on Dev machines
 
 # v1.2.x config files
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -811,8 +811,8 @@ function auth_process_plain_password( $p_password, $p_salt = null, $p_method = n
  * @return string 16 character random password
  * @access public
  */
-function auth_generate_random_password() {
-	return crypto_generate_uri_safe_nonce( 16 );
+function auth_generate_random_password($length=16) {
+	return crypto_generate_uri_safe_nonce( $length );
 }
 
 /**
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -241,9 +241,9 @@ function auth_credential_page( $p_query_string, $p_user_id = null, $p_username =
  * @param string $p_username The username
  * @return string The authenticator page with query string.
  */
-function auth_authenticator_page( $p_query_string, $p_user_id = null, $p_username = '' ) {
+function auth_authenticator_page( $p_user_id = null, $p_username = '' ) {
 	$t_auth_flags = auth_flags( $p_user_id, $p_username );
-	return $t_auth_flags->getAuthenticatorPage( $p_query_string );
+	return $t_auth_flags->getAuthenticatorPage();
 }
 
 /**
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -1194,7 +1194,7 @@ function auth_http_is_logout_pending() {
  * @return boolean
  * @access public
  */
-function authentiator_autoprov_capable() {
+function authenticator_autoprov_capable() {
 	$t_auth_flags = auth_flags();
 	return $t_auth_flags->getAutoprovisionCapability();
 }
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -234,6 +234,19 @@ function auth_credential_page( $p_query_string, $p_user_id = null, $p_username =
 }
 
 /**
+ * Gets the page that validates the user credentials based on the standard flow.
+ *
+ * @param string $p_query_string The query string, can be empty.
+ * @param int|null $p_user_id The user id or null for current logged in user.
+ * @param string $p_username The username
+ * @return string The authenticator page with query string.
+ */
+function auth_authenticator_page( $p_query_string, $p_user_id = null, $p_username = '' ) {
+	$t_auth_flags = auth_flags( $p_user_id, $p_username );
+	return $t_auth_flags->getAuthenticatorPage( $p_query_string );
+}
+
+/**
  * Gets the logout page to redirect to for logging out the user, it will return a relative url
  * @return string logout page (e.g. 'logout_page.php' )
  */
@@ -1173,4 +1186,15 @@ function auth_http_is_logout_pending() {
 	$t_cookie = gpc_get_cookie( $t_cookie_name, '' );
 
 	return( $t_cookie > '' );
+}
+
+/**
+ * Check if the plugin provides user autoprovisioning capability
+ *
+ * @return boolean
+ * @access public
+ */
+function authentiator_autoprov_capable() {
+	$t_auth_flags = auth_flags();
+	return $t_auth_flags->getAutoprovisionCapability();
 }
--- a/core/classes/AuthFlags.class.php
+++ b/core/classes/AuthFlags.class.php
@@ -70,6 +70,17 @@ class AuthFlags {
 	private $credentials_page = null;
 
 	/**
+	 * The page to use for validating credentials.  This can be the default validator page
+	 * provided by MantisBT, an auth plugin provided page that validates the credentials 
+	 * the credentials againts an external source (and creates user if autoprovision is set)
+	 *
+	 * @see $login_page
+	 * @see $logout_page
+	 * @var string|null
+	 */
+	private $authenticator_page = null;
+
+	/**
 	 * The logout page to use instead of the standard MantisBT logout page.  This can be
 	 * a plugin page.
 	 *
@@ -124,6 +135,13 @@ class AuthFlags {
 	 * @var int|null
 	 */
 	private $reauthentication_expiry = null;
+
+	/**
+	 * Indicates the autoprovisioning capability of the plugin
+	 * @var bool|false
+	 */
+	private $autoprovision_capable = false;
+
 
 	/**
 	 * AuthFlags constructor.
@@ -222,6 +240,27 @@ class AuthFlags {
 	function getCredentialsPage( $p_query_string ) {
 		$t_page = is_null( $this->credentials_page ) ? AUTH_PAGE_CREDENTIAL : $this->credentials_page;
 		return helper_url_combine( $t_page, $p_query_string );
+	}
+
+	/**
+	 * Sets auth page to use instead of the default login.php page that validates the username/password
+	 *
+	 * @param string $p_page The relative url of the page name.
+	 * @return void
+	 * @see getAuthenticatorPage()
+	 */
+	function setAuthenticatorPage( $p_page ) {
+		$this->authenticator_page = $p_page;
+	}
+
+	/**
+	 * Gets the auth page to use.
+	 *
+	 * @return string The relative login page name.
+	 * @see setAuthenticatorPage()
+	 */
+	function getAuthenticatorPage() {
+		return is_null( $this->authenticator_page ) ? AUTH_PAGE_AUTHENTICATOR : $this->authenticator_page;
 	}
 
 	/**
@@ -397,5 +436,31 @@ class AuthFlags {
 
 		return $this->reauthentication_expiry;
 	}
+	
+	/**
+	 * Sets if the auth pugin is capable of provision users
+	 *
+	 * @param bool capability
+	 * @return void
+	 * @see getAutoprovisionCapability()
+	 */
+	function setAutoprovisionCapability( $autoprov ) {
+		$this->autoprovision_capable = $autoprov;
+	}
+
+	/**
+	 * Gets the plugin autoprovision capability
+	 *
+	 * @return bool
+	 * @see setAutoprovisionCapability()
+	 */
+	function getAutoprovisionCapability() {
+		if( is_null( $this->autoprovision_capability ) ) {
+			return false;
+		}
+
+		return $this->autoprovision_capability;
+	}
+	
 }
 
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -535,9 +535,11 @@ function user_create( $p_username, $p_password, $p_email = '',
 
 	user_ensure_name_valid( $p_username );
 	user_ensure_name_unique( $p_username );
-	user_ensure_email_unique( $p_email );
-	email_ensure_valid( $p_email );
-	email_ensure_not_disposable( $p_email );
+	if ( ! is_blank($p_email) ) {
+	    user_ensure_email_unique( $p_email );
+	    email_ensure_valid( $p_email );
+	    email_ensure_not_disposable( $p_email );
+	}
 
 	$t_cookie_string = auth_generate_unique_cookie_string();
 
--- a/login_password_page.php
+++ b/login_password_page.php
@@ -206,7 +206,7 @@ if( config_get_global( 'admin_checks' ) == ON && file_exists( dirname( __FILE__ 
 					<?php echo $t_form_title ?>
 				</h4>
 				<div class="space-10"></div>
-	<form id="login-form" method="post" action="<?php echo auth_authenticator_page(); ?>">
+	<form id="login-form" method="post" action="<?php echo auth_authenticator_page($t_user_id, $t_username); ?>">
 		<fieldset>
 
 			<?php
--- a/login_password_page.php
+++ b/login_password_page.php
@@ -81,8 +81,13 @@ if( is_blank( $t_username ) ) {
 
 # Get the user id and based on the user decide whether to continue with native password credential
 # page or one provided by a plugin.
+#
+# OR
+#
+# if the plugin provides autoprovisioning, than we can also redirect to the plugin page
+#
 $t_user_id = auth_get_user_id_from_login_name( $t_username );
-if( $t_user_id !== false && auth_credential_page( '', $t_user_id ) != AUTH_PAGE_CREDENTIAL ) {
+if( ( $t_user_id !== false || authenticator_autoprov_capable() ) && auth_credential_page( '', $t_user_id ) != AUTH_PAGE_CREDENTIAL ) {
 	$t_query_args = array(
 		'username' => $t_username,
         'cookie_error' => $f_cookie_error,
@@ -201,7 +206,7 @@ if( config_get_global( 'admin_checks' ) == ON && file_exists( dirname( __FILE__ 
 					<?php echo $t_form_title ?>
 				</h4>
 				<div class="space-10"></div>
-	<form id="login-form" method="post" action="login.php">
+	<form id="login-form" method="post" action="<?php echo auth_authenticator_page(); ?>">
 		<fieldset>
 
 			<?php
--- a/admin/.htaccess
+++ b/admin/.htaccess
@@ -1,0 +1,3 @@
+<If "-f %{DOCUMENT_ROOT}.'/admin/install.done'">
+    Require all denied
+</If>
--- a/admin/install.php
+++ b/admin/install.php
@@ -1357,7 +1357,14 @@ if( 7 == $t_install_state ) {
 <?php } else { ?>
 		Please log in as the administrator and <a href="../login_page.php">create</a> your first project.
 		</span>
-<?php } ?>
+<?php } 
+
+	if ( ! touch('./install.done') ) {
+?>
+	<p class="red bolder">Please create a file named '/admin/install.done' or delete the admin folder!</p>
+<?php
+	}
+?>
 	</td>
 	<?php print_test_result( GOOD ); ?>
 </tr>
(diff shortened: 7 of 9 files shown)


.gitignore
```diff
@@ -52,6 +52,9 @@ api/rest/swagger
 /vendor/
 composer.phar
 
+# Admin
+admin/install.done
```
vboctor: The `install.done` changes seem to be unrelated.  @dregad can comment on these and we should decide if they should be part of this PR or a separate one.

core/authentication_api.php
```diff
@@ -798,8 +811,8 @@ function auth_process_plain_password( $p_password, $p_salt = null, $p_method = n
  * @return string 16 character random password
  * @access public
  */
-function auth_generate_random_password() {
-	return crypto_generate_uri_safe_nonce( 16 );
+function auth_generate_random_password($length=16) {
```
vboctor: Replace with `function auth_generate_random_password( $p_length = 16 )`

core/authentication_api.php
```diff
@@ -798,8 +811,8 @@ function auth_process_plain_password( $p_password, $p_salt = null, $p_method = n
  * @return string 16 character random password
```
vboctor: Update phpdoc to add `@param` for new parameter and update `@return` comment.  Also update the description of the function to reflect that it is not always 16 characters.

core/authentication_api.php
```diff
@@ -1174,3 +1187,14 @@ function auth_http_is_logout_pending() {
...
 	return( $t_cookie > '' );
 }
+
+/**
+ * Check if the plugin provides user autoprovisioning capability
+ *
+ * @return boolean
+ * @access public
+ */
+function authenticator_autoprov_capable() {
```
vboctor: Hoq about replacing `authenticator_autoprov_capable()` with `auth_can_auto_provision()`?

core/classes/AuthFlags.class.php
```diff
@@ -69,6 +69,17 @@ class AuthFlags {
...
+	/**
+	 * The page to use for validating credentials.  This can be the default validator page
+	 * provided by MantisBT, an auth plugin provided page that validates the credentials 
+	 * the credentials againts an external source (and creates user if autoprovision is set)
+	 *
+	 * @see $login_page
+	 * @see $logout_page
+	 * @var string|null
+	 */
+	private $authenticator_page = null;
```
vboctor: Is there a reason why this shouldn't be transparent from core and implemented as part of `$credentials_page` and potentially other pages related to it as per plugin implementation?  For example, you can imagine the credentials page asking for password, then redirecting to authenticator, then logging in the user.

core/classes/AuthFlags.class.php
```diff
@@ -125,6 +136,13 @@ class AuthFlags {
 	 */
 	private $reauthentication_expiry = null;
 
+	/**
+	 * Indicates the autoprovisioning capability of the plugin
+	 * @var bool|false
```
vboctor: Remove `|false`, this is used in case of null because null is considered a different type than string for example, and hence you will see `string|null`.  But false is a valid bool value.

login_password_page.php
```diff
@@ -81,8 +81,13 @@
...
 # Get the user id and based on the user decide whether to continue with native password credential
 # page or one provided by a plugin.
+#
+# OR
+#
+# if the plugin provides autoprovisioning, than we can also redirect to the plugin page
+#
 $t_user_id = auth_get_user_id_from_login_name( $t_username );
-if( $t_user_id !== false && auth_credential_page( '', $t_user_id ) != AUTH_PAGE_CREDENTIAL ) {
+if( ( $t_user_id !== false || authenticator_autoprov_capable() ) && auth_credential_page( '', $t_user_id ) != AUTH_PAGE_CREDENTIAL ) {
```
vboctor: Should we have a logic like the following before this block:

```
if( $t_user_id === false ) {
    $t_user_id = auth_provision_user( $p_username );
}

// then use the existing logic
```

In this case, we always give plugins the change to provision the user, but they make decide that a username doesn't exist in the identity system and hence it shouldn't be provisioned or they don't support auto-provision.  Hence, we replace `authenticator_autoprov_capable()` with `auth_provision_user( $p_username )`.  The auth_provision_user() returns user id created or false.

login_password_page.php
```diff
@@ -81,8 +81,13 @@
 
 # Get the user id and based on the user decide whether to continue with native password credential
 # page or one provided by a plugin.
+#
+# OR
+#
+# if the plugin provides autoprovisioning, than we can also redirect to the plugin page
```
vboctor: Please explain the expected flow(s) when user is not found.  Explain the case where use should be provisioned and the other where user wouldn't be provisioned because it doesn't exist in identity system.

core/user_api.php
```diff
@@ -535,9 +535,11 @@ function user_create( $p_username, $p_password, $p_email = '',
 
 	user_ensure_name_valid( $p_username );
 	user_ensure_name_unique( $p_username );
-	user_ensure_email_unique( $p_email );
-	email_ensure_valid( $p_email );
-	email_ensure_not_disposable( $p_email );
+	if ( ! is_blank($p_email) ) {
```
vboctor: Proper formatting is `if( !is_blank( $p_email ) ) {`.  But why do we need this check?  The existing methods should apply the right validation based on configuration like whether or not to allow empty email addresses via `$g_allow_blank_email`.

core/authentication_api.php
```diff
@@ -233,6 +233,19 @@ function auth_credential_page( $p_query_string, $p_user_id = null, $p_username =
...
 
+/**
+ * Gets the page that validates the user credentials based on the standard flow.
+ *
+ * @param string $p_query_string The query string, can be empty.
+ * @param int|null $p_user_id The user id or null for current logged in user.
+ * @param string $p_username The username
+ * @return string The authenticator page with query string.
+ */
+function auth_authenticator_page( $p_user_id = null, $p_username = '' ) {
```
vboctor: The term `authenticator` reflects a specific implementation.  We should probably consider `extended_auth`.  It can work with all kinds of 2FA or any other approach (e.g. security questions), etc.

I would expect such page to be visible only after user enters their correct credentials via the standard password page.  If user is using a custom credentials page, then this setting wouldn't be applicable.

core/classes/AuthFlags.class.php
V1pr: This is separate by intention. In the updated readme of SampleAuth I described it in detail. Main reason is, that `$credentials_page`  is mainly for SSO IMHO (with separate password request page), while `$authenticator_page` replaced the standard login.php for access validation - but uses the Mantis username/password aquisition pages.

login_password_page.php
V1pr: I think this will not work with SSO systems. I mean, if you set $credentials_page to support SSO/external password aqusition page, than you'll not get redirected to the plugin page.

login_password_page.php
V1pr: In my opinion, user provisioning must always  run afterwards authentication. If I use a plugin for auth and the user not yet exists in the system, than I want to do provisioning only, if the user authenticated himself correctly (that's why I set the flag for autoprov capable)

The flow:
- if user found -> go to plugin page and to the auth
- if user not found:
  - if plugin can do autoprov, than continue to auth
  - if pluging cannot do autoprov, than skip plugin auth and use standard flow

core/user_api.php
V1pr: I've added this to make the autoprov work. I've failed to notice the $g_allow_blank_email in the code, but there can be scenarios, when it's not possible to turn it on (=allowing blank e-mails); like if signup is enabled, but I want to use external auth besides it with autoprov.

My initial problem was, than if I supply an e-mail to user_create, than a signup e-mail will be sent to the user, which I want to avoid. I would also be possible to add a new parameter to the function, like 'p_dont_send_signup_email', but I did not want to do that not knowing your future ideas with user_create.

core/authentication_api.php
V1pr: The user is either using credentials_page OR authenticator_page. The former request the password from the user and authenticates it (or does SSO), while the later replaces the standard login.php and  handles the authentication (redirects to login.php if unsuccessful).

Reasoning: I wanted to use the standard mantis flow (username and password aquisition), but was not able to change login.php to support the plugin as a second auth mech, since the authflags got resetted to default before the redirect and changing this seemed to be a much deeper change. I named is `authenticator`, because the page actually does the authentication with the username/password POST-ed to it (see SampleAuth readme in PR)

core/authentication_api.php
V1pr: Changed.

core/authentication_api.php
V1pr: Changed.

core/authentication_api.php
V1pr: Replaced.

core/classes/AuthFlags.class.php
V1pr: Changed.

.gitignore
V1pr: Yes, it's unrelated, but me was a bit pissed of with the warning :) Maybe it would be nicer to use an install.RELASE for such purpose - making upgrades easier. Anyway, I'll wait for @dregad 's answer.

dregad: @V1pr I can hear your point of view with the warning.

The problem with the proposed approach is that not everyone uses Apache for their webserver; someone running Mantis on IIS would no longer see the warning about the admin directory, even though it is still present and accessible.

I would prefer to treat this as a separate work item.

I'm not sure in what context the warning pissed you off, but FYI you can always disable it with `$g_admin_checks = OFF` (that's what I usually do on my dev box; obviously not recommended on a production server).

V1pr: Ok, I get your point. I'll revert these change.

V1pr:
I can't seem to find the place, where I can create an issue#/feture req.#. What am I missing?

About your last question: this setup works with my extended version on SampleAuth plugin (also added a lot of comments to it, with a more detailed readme). However, I found that the auth plugin concept might needs some work. I would change it to a "stack like" method. A plugin registers itself to the core with a set of flags (including priority of the mech). Upon login the core checks the corresponding plugins in priority order taking the flag into consideration when checking.

aborderon:
Have you any news ?

V1pr:
Ping?

aborderon:
@vboctor Do you have any elements to communicate to advance this amendment? Because it interests us.
Thank you

aborderon:
@V1pr about creating problems in the bug tracker and referencing them from the pull request. You have to go here : https://www.mantisbt.org/bugs/view_all_bug_page.php

V1pr:
@aborderon thanks, I've opened a ticket.
@vboctor @dregad  is there anything else I need to do speed the merge up a bit?

dregad:
With the revert of the .gitignore change, I don't have any issues with your PR. I let you follow up with vboctor.

core/authentication_api.php
```diff
@@ -793,13 +806,14 @@ function auth_process_plain_password( $p_password, $p_salt = null, $p_method = n
...
+ * Generate a random password. If no lenth is supplied, than the default value of 16 will be used.
  * @todo create memorable passwords?
- * @return string 16 character random password
+ * @param int $p_length	Requested password length
+ * @return string random password
  * @access public
  */
-function auth_generate_random_password() {
-	return crypto_generate_uri_safe_nonce( 16 );
+function auth_generate_random_password($p_length=16) {
```
atrol: Missing spaces http://www.mantisbt.org/wiki/doku.php/mantisbt:coding_guidelines#braces_and_parentheses

core/classes/AuthFlags.class.php
```diff
@@ -69,6 +69,17 @@ class AuthFlags {
 	 */
 	private $credentials_page = null;
 
+	/**
+	 * The page to use for validating credentials.  This can be the default validator page
+	 * provided by MantisBT, an auth plugin provided page that validates the credentials 
+	 * the credentials againts an external source (and creates user if autoprovision is set)
```
atrol: two times `the credentials`
s/againts/against/

core/classes/AuthFlags.class.php
```diff
@@ -397,5 +436,31 @@ function getReauthenticationLifetime() {
 
 		return $this->reauthentication_expiry;
 	}
+	
+	/**
+	 * Sets if the auth pugin is capable of provision users
+	 *
+	 * @param bool capability
```
atrol: $p_autoprov

core/classes/AuthFlags.class.php
```diff
@@ -397,5 +436,31 @@ function getReauthenticationLifetime() {
...
+	 * @return void
+	 * @see getAutoprovisionCapability()
+	 */
+	function setAutoprovisionCapability( $autoprov ) {
+		$this->autoprovision_capable = $autoprov;
```
atrol: two times $p_autoprov

login_password_page.php
```diff
@@ -201,7 +206,7 @@
 					<?php echo $t_form_title ?>
 				</h4>
 				<div class="space-10"></div>
-	<form id="login-form" method="post" action="login.php">
+	<form id="login-form" method="post" action="<?php echo auth_authenticator_page($t_user_id, $t_username); ?>">
```
atrol: spaces after/before parentheses

core/classes/AuthFlags.class.php
```diff
@@ -397,5 +436,31 @@ function getReauthenticationLifetime() {
...
+	}
+
+	/**
+	 * Gets the plugin autoprovision capability
+	 *
+	 * @return bool
+	 * @see setAutoprovisionCapability()
+	 */
+	function getAutoprovisionCapability() {
+		if( is_null( $this->autoprovision_capability ) ) {
```
atrol: s/autoprovision_capability/autoprovision_capable

core/classes/AuthFlags.class.php
```diff
@@ -397,5 +436,31 @@ function getReauthenticationLifetime() {
...
+			return false;
+		}
+
+		return $this->autoprovision_capability;
```
atrol: s/autoprovision_capability/autoprovision_capable

core/user_api.php
```diff
@@ -535,9 +535,11 @@ function user_create( $p_username, $p_password, $p_email = '',
 
 	user_ensure_name_valid( $p_username );
 	user_ensure_name_unique( $p_username );
-	user_ensure_email_unique( $p_email );
-	email_ensure_valid( $p_email );
-	email_ensure_not_disposable( $p_email );
+	if( !is_blank( $p_email ) ) {
```
atrol: Why did you introduce this check?

After your change, blank email addresses can be set even if $g_allow_blank_email = OFF; (our default setting)

core/authentication_api.php
```diff
@@ -233,6 +233,19 @@ function auth_credential_page( $p_query_string, $p_user_id = null, $p_username =
 	return $t_auth_flags->getCredentialsPage( $p_query_string );
 }
 
+/**
+ * Gets the page that validates the user credentials based on the standard flow.
+ *
+ * @param string $p_query_string The query string, can be empty.
```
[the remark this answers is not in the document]
V1pr: sorry, I don't get this one. $p_query_string exists as an input parameter for auth_credential_page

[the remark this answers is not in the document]
atrol: Isn't this the PHPdoc for auth_authenticator_page ?

core/user_api.php
V1pr: If I remember correctly, I ran into issues upon automatic user creation, but I must have overlooked something. I'll revert this one and add a comment to SampleAuth,

V1pr: I've found the answer in SampleAuth: the reason was behind this, that Mantis will send a SignUP e-mail, if e-mail is provided (which I did not want), so I create the use with empty e-mail and assign the email afterwards. See [SampleAuth plugin](https://github.com/V1pr/SampleAuth/blob/master/core/CustomAuthPlugin.php). Anyway, I'll change the logic/doc in SampleAuth.

core/user_api.php
```diff
@@ -535,9 +535,17 @@ function user_create( $p_username, $p_password, $p_email = '',
 
 	user_ensure_name_valid( $p_username );
 	user_ensure_name_unique( $p_username );
+<<<<<<< HEAD
```
atrol: Unresolved conflict from https://github.com/mantisbt/mantisbt/pull/1333/commits/407a940200f2e4e6ff814d120dce8900cdbfcd8d

aborderon:
@V1pr news ?

V1pr:
@aborderon  none, that I know of :( I did what I can :(

@dregad @vboctor @atrol is there anything I can do, to speed the things up?

atrol:
I am a bit concerned, as this is security related authentication code and it seems that this code has not been tested that much.
e.g. getAutoprovisionCapability did not work due to access to non-existing `autoprovision_capability` before V1pr addressed my comments in https://github.com/mantisbt/mantisbt/commit/b036694

When I reviewed in June, I didn't test anything, just did some static code review.
I don't use auth plugin functionality myself and I am not able to invest time in having a deeper look and tests.

Sorry @V1pr, I will not be the one who approves the PR.

V1pr:
Thank you Gents! I'll wait for @vboctor 's approval/comments.

core/classes/AuthFlags.class.php
```diff
@@ -397,5 +436,31 @@ function getReauthenticationLifetime() {
...
 	}
+	
+	/**
+	 * Sets if the auth pugin is capable of provision users
+	 *
+	 * @param bool $p_autoprov True, if the plugin can create users
+	 * @return void
+	 * @see getAutoprovisionCapability()
+	 */
+	function setAutoprovisionCapability( $p_autoprov ) {
```
vboctor: Should we use `setAutoProvisioningCapability` and `getAutoProvisioningCapability`?  Remember to update the php-docs as well.

vboctor:
We can even have a single plugin event to verify password and provision if necessary.  The event can take the username and password and return false in case of failure or user id (matched or provisioned).

EVENT_AUTH_VERIFY_PASSWORD_AND_PROVISION

You can imagine multiple auth plugins subscribed to this event and the first one that succeeds would win.

Thanks @V1pr

core/classes/AuthFlags.class.php
```diff
@@ -397,5 +436,31 @@ function getReauthenticationLifetime() {
...
+	}
+
+	/**
+	 * Gets the plugin autoprovision capability
+	 *
+	 * @return bool
+	 * @see setAutoprovisionCapability()
+	 */
+	function getAutoprovisionCapability() {
+		if( is_null( $this->autoprovision_capable ) ) {
```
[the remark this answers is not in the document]
V1pr: not really needed, but a safeguard to return false instead of null (if anything did set it to null). IMHO 'null' is not a bool type.

core/authentication_api.php
```diff
@@ -233,6 +233,18 @@ function auth_credential_page( $p_query_string, $p_user_id = null, $p_username =
 	return $t_auth_flags->getCredentialsPage( $p_query_string );
 }
 
+/**
+ * Gets the page that validates the user credentials based on the standard flow.
+ *
+ * @param int|null $p_user_id The user id or null for current logged in user.
```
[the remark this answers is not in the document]
V1pr: If nothing is returned, than the login form "action" fill be empty, pointing to self, which I assume it's safe.

core/classes/AuthFlags.class.php
V1pr: I don't think so, since the config variable in the pluging called 'autoprovision' not 'autoprovisioning'.

login_password_page.php
```diff
@@ -81,8 +81,13 @@
...
 # Get the user id and based on the user decide whether to continue with native password credential
 # page or one provided by a plugin.
+#
+# OR
+#
+# if the plugin provides autoprovisioning, than we can also redirect to the plugin page
+#
 $t_user_id = auth_get_user_id_from_login_name( $t_username );
-if( $t_user_id !== false && auth_credential_page( '', $t_user_id ) != AUTH_PAGE_CREDENTIAL ) {
+if( ( $t_user_id !== false || auth_can_auto_provision() ) && auth_credential_page( '', $t_user_id ) != AUTH_PAGE_CREDENTIAL ) {
```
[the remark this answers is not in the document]
V1pr: I did assume, that the original auth_credential_page() will handle it, since in the params list p_user_id is set to null, if none given.

V1pr:
> We can even have a single plugin event to verify password and provision if necessary. The event can take the username and password and return false in case of failure or user id (matched or provisioned).
> 
> EVENT_AUTH_VERIFY_PASSWORD_AND_PROVISION
> 
> You can imagine multiple auth plugins subscribed to this event and the first one that succeeds would win.
> 
> Thanks @V1pr

In this case, you'll likely have an userid collision soon. Using multiple auth backends in a frontend application is imho never a good idea. Authenticator backends should be handled by the authenticator app (SSO or something like that), not an application.

PR #1335, opened 2018-04-17
Add new event for notifications settings

samiraguiar:
As suggested and discussed in #5637, a new event is signalled after fetching
user preferences for notifications when sending e-mails. This allows plugins
to overwrite those preferences and allow flexible customization of notifications
(on a per-project basis, for example).

Although we currently have signals EVENT_NOTIFY_USER_INCLUDE
and EVENT_NOTIFY_USER_EXCLUDE, those will only work when
notifications are enabled for users. However, enabling those notifications can
become a problem when we have a large number of users and projects,
since one would need to manually disable notifications for each user
that does not have them customized.

The new event is a chain event since more than one plugin may want to
overwrite those settings.


docbook/Developers_Guide/en-US/Events_Reference_Notify.xml
```diff
@@ -58,6 +58,38 @@
...
+					<listitem><para>&lt;Integer&gt;: Bug ID</para></listitem>
+					<listitem><para>&lt;String&gt;: Notification type</para></listitem>
+				</itemizedlist>
+
+				<itemizedlist>
+					<title>Return Value</title>
+					<listitem><para>
+					&lt;Array&gt;: Array with all the parameters in the same order as they were received.
+					The first element of the returned array will be used to overwrite the value set
+					in the user preferences for that particular notification. A value equals to the OFF
```
dregad: _equal_ (no `s`)

dregad:
@cproensa since you're more of an expert with plugins than I am, maybe you want to comment.

cproensa:
I don't feel this is a good approach.
We would create an event with the main goal to fix user preferences for a per-project configuration. Why don't we fix that properly, it's about time?
The pending discussion for that is where to store the user prefs. in config table, which already can be set by user and project? another table with similar structure as config? etc

vboctor:
I haven't reviewed the code, but I agree with @cproensa and I don't understand why the current design doesn't address the need by having a DEVELOPER plugin for those working on the project, and a lower access level (e.g. UPDATER) for those that have access but are not working on the project.  See the issue for more details about my thoughts.

samiraguiar:
@vboctor Please take a look at [the issue](https://www.mantisbt.org/bugs/view.php?id=5637), I've added a note explaining why the current design doesn't allow this.

PR #1336, opened 2018-04-18
Code Cleanup

atrol:
- Remove unused function filter_exists
- PHPdoc fixes
- Remove unused local variables

### DIFF
--- a/core/classes/DbQuery.class.php
+++ b/core/classes/DbQuery.class.php
@@ -390,6 +390,7 @@ class DbQuery {
 	 * to final ADOdb parameter syntax.
 	 * Will convert all labeled ":xxx", and anonymous "$n" parameters, and build
 	 * a values array suitable for ADOdb.
+	 * @param integer $p_counter_start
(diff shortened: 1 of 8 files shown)


core/print_api.php
```diff
@@ -1248,10 +1248,6 @@ function print_custom_field_projects_list( $p_field_id ) {
  * @return void
  */
 function print_plugin_priority_list( $p_priority ) {
-	if( $p_priority < 1 && $p_priority > 5 ) {
-		echo '<option value="', $p_priority, '" selected="selected">', $p_priority, '</option>';
-	}
-
```
dregad: I wouldn't qualify this as useless. Even though it's unlikely to happen as there is no path for that in Mantis core, the priority *could* technically have a value other than 1..5; this code ensures that whatever is stored in the database does not get silently overwritten when the form is submitted.

I recommend to not merge this commit.

dregad: Yes you're right, I didn't check the actual condition :blush:... Should be `||` then.

PR #1337, opened 2018-04-18
Correct default value of my_view_boxes in Admin Guide

atrol:
Fixes #24326

### DIFF
--- a/docbook/Admin_Guide/en-US/config/myview.xml
+++ b/docbook/Admin_Guide/en-US/config/myview.xml
@@ -13,13 +13,17 @@
 					to be shown. A box that is not to be shown can have its value set
 					to 0. The default is:
 					<programlisting>
-$g_my_view_boxes = array( 'assigned' =&gt; '1',
-                          'unassigned' =&gt; '2',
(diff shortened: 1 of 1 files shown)


docbook/Admin_Guide/en-US/config/myview.xml
```diff
@@ -13,13 +13,17 @@
...
 					<programlisting>
-$g_my_view_boxes = array( 'assigned' =&gt; '1',
-                          'unassigned' =&gt; '2',
-                          'reported' =&gt; '3',
-                          'resolved' =&gt; '4',
-                          'recent_mod' =&gt; '5',
-                          'monitored' =&gt; '6'
-                   );
+$g_my_view_boxes = array(
+	'assigned'      => '1',
```
vboctor: Isn't the escaping of `>` necessary as per the deleted code?

atrol: I tried this on one of my dev boxes and didn't see anything wrong in generated HTML.
Not sure if it will work also on our build server as I am using a newer OS / Publican, but I thought it's worth a try when changing it.

PR #1348, opened 2018-05-15
Consider show_user_realname_threshold when sorting by reporter or handler

atrol:
Fixes #24435

### DIFF
--- a/core/classes/BugFilterQuery.class.php
+++ b/core/classes/BugFilterQuery.class.php
@@ -1541,7 +1541,7 @@ class BugFilterQuery extends DbQuery {
 				$t_select = 'COALESCE(';
 				# Note: show_realname should only be set at global or all_projects
 				# Note: sort_by_last_name is not supported here
-				if( ON == config_get( 'show_realname' ) ) {
+				if( user_show_realname() ) {
 					$t_select .= 'NULLIF(' . $t_table_alias . '.realname, \'\'), ';
 				}
 				$t_select .= $t_table_alias . '.username, ';


cproensa:
IIRC, this was left out of the implementation due to the difficulties of having the option to show "last name/ first name". In that case the sorting would be messed.
However, having this sort implmented could be better than nothing.

atrol:
> this was left out of the implementation due to the difficulties of having the option to show "last name/ first name".

@cproensa I don't understand how this is related to my PR.

My PR deals with show_user_realname_threshold, but does not try to implement anything related to sort_by_last_name option.

cproensa:
> @cproensa I don't understand how this is related to my PR

OK. I had the confused thought that `sort_by_last_name` also displayed the realname in reversed order for last/first name. (has it never been that way?)
So i intentionally left out, when implementing that code, the changes from this PR under that assumption (because the last/first name is not stored in database).

So well, after realizing i was fool and wrong, this PR seems ok. (haven't tested yet)
Thanks

atrol:
No merge needed as the change is contained in https://github.com/mantisbt/mantisbt/pull/1351 as a minor code refactoring enhancement.

atrol:
I am closing ths PR for now to avoid merging, as the change is part of PR https://github.com/mantisbt/mantisbt/pull/1351 https://github.com/mantisbt/mantisbt/pull/1351/commits/edc63e04fb87fe70c0d31e591373b8f2375761be 

The change in context of https://github.com/mantisbt/mantisbt/pull/1351 is no longer dealing with show_user_realname_threshold but just a code cleanup as https://github.com/mantisbt/mantisbt/pull/1351 assumes that show_user_realname_threshold is not considered if show_realname = ON.

The PR should be reopened and merge in case we do not agree in concept of https://github.com/mantisbt/mantisbt/pull/1351

PR #1349, opened 2018-05-15
Display username as tooltip if realname is displayed

atrol:
Fixes #24436

Display realname if show_realname is set to ON and user is allowed to see realnames, display the username as tooltip.

This is more logical than what we have after https://mantisbt.org/bugs/view.php?id=23375 but is also better concerning @ mentions than what we had before implementation of # 23375


atrol:
No merge needed if we agree to merge https://github.com/mantisbt/mantisbt/pull/1351 as the same change is part of it.

atrol:
Replaced by https://github.com/mantisbt/mantisbt/pull/1351

PR #1351, opened 2018-05-19
Username realname proposal

atrol:
Proposal to fix https://mantisbt.org/bugs/view.php?id=24139

I don't think that my proposal https://mantisbt.org/bugs/view.php?id=24139#c59327 is something that can be done in short term, because of the open questions from https://mantisbt.org/bugs/view.php?id=24139#c59831

The general question is if we want to have again that $g_show_user_realname_threshold is not  considered if $g_show_realname = ON; 

If so,  this PR is a try to 
- keep all the refactoring done in https://mantisbt.org/bugs/view.php?id=23375
- partially keep enhancements of @ mentions
- get again the ability to hide usernames in e-mail notifications.

### DIFF
--- a/api/soap/mc_account_api.php
+++ b/api/soap/mc_account_api.php
@@ -42,12 +42,14 @@ function mci_account_get_array_by_id( $p_user_id ) {
 		# this deviates from the behaviour of view_user_page.php, but it is more intuitive
 		$t_is_same_user = $t_current_user_id === $p_user_id;
 
-		$t_can_see_realname = access_has_project_level( config_get( 'show_user_realname_threshold' ) );
-		$t_can_see_email = access_has_project_level( config_get( 'show_user_email_threshold' ) );
+		$t_can_see_realname = $t_is_same_user || $t_can_manage || user_show_realname() ||
+			access_has_project_level( config_get( 'show_user_realname_threshold' ) );
+		$t_can_see_email = $t_is_same_user || $t_can_manage ||
+			access_has_project_level( config_get( 'show_user_email_threshold' ) );
 
 		$t_result['name'] = user_get_username( $p_user_id );
 
-		if ( $t_is_same_user || $t_can_manage || $t_can_see_realname ) {
+		if ( $t_can_see_realname ) {
 			$t_realname = user_get_realname( $p_user_id );
 
 			if( !empty( $t_realname ) ) {
@@ -55,7 +57,7 @@ function mci_account_get_array_by_id( $p_user_id ) {
 			}
 		}
 
-		if ( $t_is_same_user || $t_can_manage || $t_can_see_email ) {
+		if ( $t_can_see_email ) {
 			$t_email = user_get_email( $p_user_id );
 
 			if( !empty( $t_email ) ) {
--- a/core/classes/BugFilterQuery.class.php
+++ b/core/classes/BugFilterQuery.class.php
@@ -1541,7 +1541,7 @@ class BugFilterQuery extends DbQuery {
 				$t_select = 'COALESCE(';
 				# Note: show_realname should only be set at global or all_projects
 				# Note: sort_by_last_name is not supported here
-				if( ON == config_get( 'show_realname' ) ) {
+				if( user_show_realname() ) {
 					$t_select .= 'NULLIF(' . $t_table_alias . '.realname, \'\'), ';
 				}
 				$t_select .= $t_table_alias . '.username, ';
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -683,7 +683,7 @@ function email_monitor_added( $p_bug_id, $p_user_id ) {
 
 	$t_opt = array();
 	$t_opt[] = bug_format_id( $p_bug_id );
-	$t_opt[] = user_get_username( $p_user_id );
+	$t_opt[] = user_get_name( $p_user_id );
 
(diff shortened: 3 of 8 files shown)


atrol:
>  may be confusing (realname == OFF, yet threshold causes realnames to be visible)

I don't think it's that much confusing, as the threshold is just used at a single place if realname == OFF
After merging my PR, the behavior is again what is documented and what worked for many years.
```
 * This specifies the access level that is needed to see realnames on user view
 * page
```


> if a user can see realname by default, they should see it everywhere in view issue page, email, etc

Have a look at the table below.
I think it is a good compromise between backward compatibility and adding some help to find out the username for @ mentions.


> If we want fine controls, between email and view page, then should these be their own config options, etc?

Right, but IMO we need a short term solution.
I am not able to invest even more time in this topic at the moment.

### Expected behavior after merging the PR

Display location |  show_realname  ON |  show_realname OFF
---------------- | -------------------- | ----------------------
"Reporter" and "Assigned To" on view.php | Realname + Username in tooltip | Username
"Reporter" and "Assigned To" in E-Mail Notifications | Realname | Username
"Reporter" and "Assigned To" in CSV Export | Realname | Username
"Reporter" and "Assigned To" in Excel Export | Realname | Username
 "Users monitoring this issue" on view.php | Realname + Username in tooltip | Username
History on view.php | Realname + Username in tooltip | Username
History in E-Mail Notifications  | Realname  | Username
Drop down lists to select a single user, e.g. "Assign To" or "Simple Filters" | Realname + (Username) | Username
Lists to select multiple users, e.g. "Send Reminder" or "Advanced Filters" | Realname + (Username) | Username
Timeline | Realname + Username in tooltip | Username
"Reporter" and "Assigned To" columns on view_all_bug_page.php | Realname + Username in tooltip | Username
"Status" column on view_all_bug_page.php if show_assigned_names = ON | Realname + Username in tooltip | Username
"Real Name" on view_user_page | Realname  | Realname or nothing, depending on show_user_realname_threshold

atrol:
> The implemented behavior here shows realname even if it is OFF which seems to be inconsistent with other configs

You mean this row with it

Display location |  show_realname  ON |  show_realname OFF
---------------- | -------------------- | ----------------------
"Real Name" on view_user_page | Realname  | Realname or nothing, depending on show_user_realname_threshold

Visit [this page](https://www.mantisbt.org/bugs/view_user_page.php?id=11111) on our bug tracker.
Do you really want to see any other thing than the real name of atrol in field `Real Name` on this page if you are logged in as user vboctor?

> Why wouldn't we require show_realname = ON and threshold greater or equal to show_realname_threshold?

As told before
- the behavior is what is documented `specifies the access level that is needed to see realnames on user view page`
-  the behavior is what we had for many years, which means that users don't have to change setshow_realname_threshold to see real names

> Should we show realname (username) in case of email notifications, csv export, and Excel export? 

Not at the moment, as there are users relying on the former behavior due to security reasons where usernames were not visible at various places if show_realname = ON.
Please read all the discussions in the affected and related issues.

Again: This is a compromise, as long as there is no one who implements https://mantisbt.org/bugs/view.php?id=24139#c59327 + adds a new option to protect user names (something like $g_show_user_username_threshold )+ deals with empty real names + deals with non unique real names.

All I could offer to replace this PR,  is a new PR that reverts even more to the old behavior but keeps just your refactoring to have less places where we deal with user/real names.

vboctor:
>> The implemented behavior here shows realname even if it is OFF which seems to be inconsistent with other configs

> You mean this row with it

What I meant that for other options we honor the ON/OFF feature first and only if ON we check associated threshold.  Or we don't have ON/OFF and just rely on threshold.  Of course, I expect that the Real Name field on the user view page to only show Real Name (if visible to logged in user).  I understand the inconsistent here is motivated by backward compatibility.

> the behavior is what we had for many years, which means that users don't have to set show_realname_threshold (default is NOBODY) to see real names if show_realname = ON

I think if the desired behavior is possible, it is OK to ask changing the value of the config options.  We have already changed the behavior for several releases, so it is not like it is a new change.  We can also make this clear in the release notes.

> This is a compromise

I understand that, and I'm generally OK with your proposal here.  I don't think we should revert.  The old code is a mess.  But I'm of the opinion that an inconsistent behavior should be fixed over time and that not every MantisBT user will be on board with the change.  So even with merging this PR shouldn't mean that we shouldn't have future changes that will drive consistency.  For cases where we want to provide flexibility, we can have config options, but we should eventually get to a point where we are consistent and forward looking by default.

In summary, I'm OK with this PR, but that doesn't exclude future changes to drive consistency even if there will be requests not to change the behavior.  I would also like with future changes to get to a point that calling code not have to replicate business logic of whether to show realname or not, but depend on call to user API.

Anyway, thanks @atrol for putting this together.

atrol:
> I don't think we should revert. The old code is a mess. 

Right, that's why I kept as much as possible of your refactoring in the PR.

> But I'm of the opinion that an inconsistent behavior should be fixed over time and that not every MantisBT user will be on board with the change.

Right, we cannot know if a user relies on an undocumented feature (e.g. don't display usernames in e-mail notifications). 
For most of the users (including myself) https://github.com/mantisbt/mantisbt/pull/1349 would have been enough.

> I'm OK with this PR, but that doesn't exclude future changes to drive consistency even if there will be requests not to change the behavior.

+1

PR #1353, opened 2018-05-21
Fix PHP 7.2 system warnings in SOAP API

atrol:
Fixes #24462

### DIFF
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -488,7 +488,7 @@ function mci_issue_get_custom_fields( $p_issue_id ) {
 		}
 	}
 
-	return count( $t_custom_fields ) == 0 ? null : $t_custom_fields;
+	return $t_custom_fields;
 }
 
 /**
@@ -600,7 +600,7 @@ function mci_issue_get_relationships( $p_issue_id, $p_user_id ) {
 		}
 	}
 
-	return (count( $t_relationships ) == 0 ? null : $t_relationships );
+	return $t_relationships;
 }
 
 /**
@@ -679,7 +679,7 @@ function mci_issue_get_notes( $p_issue_id ) {
 		$t_result[] = $t_bugnote;
 	}
 
-	return count( $t_result ) == 0 ? null : $t_result;
+	return $t_result;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


atrol:
I am not sure if this is the right fix, as I had to change also our tests.

If there is an issue without notes, attachments, custom fields  and relationships, this is an example extract of a response after the change
```
            <attachments SOAP-ENC:arrayType="ns1:AttachmentData[0]" xsi:type="SOAP-ENC:Array"/>
            <relationships SOAP-ENC:arrayType="ns1:RelationshipData[0]" xsi:type="SOAP-ENC:Array"/>
            <notes SOAP-ENC:arrayType="ns1:IssueNoteData[0]" xsi:type="SOAP-ENC:Array"/>
            <custom_fields SOAP-ENC:arrayType="ns1:CustomFieldValueForIssueData[0]" xsi:type="SOAP-ENC:Array"/>
            <due_date xsi:nil="true" xsi:type="xsd:dateTime"/>
            <monitors SOAP-ENC:arrayType="ns1:AccountData[0]" xsi:type="SOAP-ENC:Array"/>
            <sticky xsi:type="xsd:boolean">false</sticky>
            <tags SOAP-ENC:arrayType="ns1:ObjectRef[0]" xsi:type="SOAP-ENC:Array"/>
```
before the change we had
```
            <attachments SOAP-ENC:arrayType="ns1:AttachmentData[0]" xsi:type="SOAP-ENC:Array"/>
            <due_date xsi:nil="true" xsi:type="xsd:dateTime"/>
            <monitors SOAP-ENC:arrayType="ns1:AccountData[0]" xsi:type="SOAP-ENC:Array"/>
            <sticky xsi:type="xsd:boolean">false</sticky>
            <tags SOAP-ENC:arrayType="ns1:ObjectRef[0]" xsi:type="SOAP-ENC:Array"/>
```

Pros of the change
- uniform behavior if fields are not set

Cons of the change
- larger/slower response
- changed response might break clients

@vboctor I am not experienced in writing SOAP clients.
Do you think that the changed response could break a client?

If so, we would have to implement another approach to fix this.
Some `null` checks before calling `count` at some places should be enough for it.

vboctor:
@atrol It would be helpful to start with the PHP line that causes the warning.  I would go with the option of making sure that the code handle nulls rather than replacing nulls with empty arrays.  You can imagine code like our test cases that checks for null and assumes if not null, then the array won't be empty.  Very corner case, but possible. I would go to an extreme to accommodate such corner case, but avoid change is a better option.

atrol:
Replaced by https://github.com/mantisbt/mantisbt/pull/1354

PR #1357, opened 2018-05-30
Eliminate URL-encoding of root_namespace

rk4n3:
Fixes [#24496](https://mantisbt.org/bugs/view.php?id=24496)


dregad:
Add `Fixes #24496` to commit message prior to merge

PR #1359, opened 2018-06-05
Upgrade: improve handling of unserialize errors

dregad:
In MantisBT 1.3, the method used to store config, token and filter data 
changed from _serialize_ to JSON. Upgrade steps 193, 194 and 195 were
added to convert legacy data to the new format.

This frequently caused the upgrade process to fail when unserialize()
could not interpret the data, but the MantisBT installer just aborted
without providing any useful information to the admin.

This improves the error reporting, clearly identifying the offending
record so that the admin can take appropriate action to fix the problem.

Fixes #24416, #21376 (and maybe others too)

### DIFF
--- a/core/error_api.php
+++ b/core/error_api.php
@@ -404,6 +404,22 @@ function error_handler( $p_type, $p_error, $p_file, $p_line ) {
 	$g_error_proceed_url = null;
 }
 
+
+/**
+ * Error handler to convert PHP errors to Exceptions.
+ * This is used to temporarily override the default error handler, when it is
+ * required to catch a PHP error (e.g. when unserializing data in install
+ * helper functions).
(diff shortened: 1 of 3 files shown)


dregad:
Before:
image

After:
image

core/error_api.php
```diff
@@ -404,6 +404,22 @@ function error_handler( $p_type, $p_error, $p_file, $p_line ) {
...
 
+
+/**
+ * Error handler to convert PHP errors to Exceptions.
+ * This is used to temporarily override the default error handler, when it is
+ * required to catch a PHP error (e.g. when unserializing data in install
+ * helper functions).
+ * @param integer $p_type
+ * @param string $p_message
+ * @param string $p_filename
```
atrol: $p_file

PR #1360, opened 2018-06-09
Enhance PHP fileinfo extension checking

atrol:
Add a warning check for fileinfo extension if file uploads are disabled.

This is useful as a lot of plugins use function plugin_file()
which leads to a call of file_create_finfo() where the fileinfo
extension is needed.

Fixes #24501

### DIFF
--- a/admin/check/check_php_inc.php
+++ b/admin/check/check_php_inc.php
@@ -45,7 +45,6 @@ check_print_test_row(
 );
 
 # $t_extensions_required lists the extensions required to run Mantis in general
-# The fileinfo extension which is needed for attachments is checked in check_attachments_inc.php
 $t_extensions_required = array(
 	'date',
 	'hash',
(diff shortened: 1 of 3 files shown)


admin/check/check_php_inc.php
```diff
@@ -63,6 +62,43 @@
...
+		'Fileinfo PHP extension is available to support file uploads',
+		$t_fileinfo_loaded,
+		array( false => 'Ensure that the fileinfo extension is installed and enabled' )
+	);
+} else {
+	# most of the plugins need this extensions as they use functions plugin_file /  plugin_file_include
+	check_print_test_warn_row(
+		'Fileinfo PHP extension is available to support plugins',
+		$t_fileinfo_loaded,
+		array( false => 'Ensure that the fileinfo extension is installed and enabled if ' )
```
dregad: If what ?

atrol:
> but I'm wondering if we should not simply require fileinfo

I am fine with requiring fileinfo, but I don't know if we would rule out some users.

Are there any users that 
- have attachments disabled **and**
- don't use any of the plugins that need the extension (e.g. the bundled MantisGraph) **and**
- don't want / can for whatever reason install and enable the fileinfo extension ?

@dregad, @vboctor do you think this hardly a real use case?
If so, I will change the PR and always check for fileinfo as a mandatory extension.

admin/check/check_php_inc.php
atrol: > If what ?

leftover from my first internal try.
Will change, unless we agree to change the PR to simply require fileinfo.

PR #1363, opened 2018-06-25
Fix access checks in Timeline

atrol:
Fixes #24564


atrol:
> I wonder if setting project override is also a good idea.

I assume you mean `$g_project_override = $t_project_id;` instead of passing the project id as a parameter to config_get.

I prefer using parameters instead of global variables.

PR #1365, opened 2018-07-05
Allow more flexible customisation of the main menu sidebar from plugins by adding new chain event.

JamesAtChamSys:
This patch modifies the way the main sidebar menu is built; menu items are all assembled in an array first, which is then given the chance to be altered by plugins using a new chain filter EVENT_MENU_MAIN_FILTER.

This allows plugins to do a few things that currently isn't possible, especially to the hard-coded items (My View, View Bugs), such as re-order the items, change the icon, or remove them. It also allows plugins to insert new items anywhere in the menu (not just the start or end). Additionally, by returning an empty array, plugins can get rid of the sidebar entirely.

This is partly inspired by the way Wordpress allows plugins to alter admin menu item, and by our need to remove the My View item in certain circumstances.

I look forward to your comments!
James


JamesAtChamSys:
Thanks!

Have updated the code style, added documentation. Bug report is at https://mantisbt.org/bugs/view.php?id=24590

atrol:
I didn't try, but from what I see this would allow plugins to expand access rights by overriding core access checking.
E.g. plugins could change `access_level` so that users would see the `Change Log` menu entry  without having `view_changelog_threshold` access rights.
I don't think this should be possible.
Of course, it's just the menu entry, as there is again an access check on the page itself.

dregad:
> this would allow plugins to expand access rights by overriding core access checking

Not really, as you have rightly pointed out :

> it's just the menu entry, as there is again an access check on the page itself.

It's true that someone could add back a menu entry normally hidden by core thresholds, but it does not bypass the actual access checks. This would be the same as someone accessing http://path.to/mantis/changelog_page.php directly.

core/layout_api.php
```diff
@@ -793,34 +833,64 @@ function layout_print_sidebar( $p_active_sidebar_page = null ) {
...
+			$t_sidebar_items = array_merge( $t_sidebar_items, $t_config_menu_items );
+		}
 
-		# Ending sidebar markup
-		layout_sidebar_end();
+		# Allow plugins to alter the sidebar items array
+		$t_modified_sidebar_items = event_signal( 'EVENT_MENU_MAIN_FILTER', array( $t_sidebar_items ) );
+		if( is_array( $t_modified_sidebar_items ) && count( $t_modified_sidebar_items ) > 0 ) {
+			$t_sidebar_items = $t_modified_sidebar_items[0];
+		}
```
dregad: As a side note, I initially found it strange that the $t_sidebar_items was wrapped in an array in the event_signal() call, and then extracted back (line 872), until I realized that it's required due to the plugin API relying on call_user_func_array().

This also means that a plugin using this event needs to do the same as well when returning the updated menu  from the hook function, i.e. `return array( $p_menu_data );` and not `return $p_menu_data;`, as I found out while testing the new event.

@JamesAtChamSys I suggest to clarify this in the documentation.

dregad: I've done this myself.

dregad:
@JamesAtChamSys I rebased your PR on top of latest master, resolved merged conflicts and made some minor adjustments. Sorry for taking so long to finally get to this, and thanks again for your contribution.

PR #1368, opened 2018-07-22
Getters for REST fault

andersruneson:
Functions used in ApiObjectFactory::throwIfFault but non-existing, so currently a lot of errors in REST API returns Internal Server Error instead of the correct error.
https://www.mantisbt.org/bugs/view.php?id=24623


vboctor:
@andersruneson can you please open a bug with the SOAP API cases where you get an internal server error?  I suspect the correct fix will be different.

vboctor:
Closing since there is not enough information.

PR #1369, opened 2018-07-22
Added function REST project create

andersruneson:
https://www.mantisbt.org/bugs/view.php?id=24622


vboctor:
Closing in favor of #1376.

PR #1370, opened 2018-07-22
Added REST function project update

andersruneson:
https://www.mantisbt.org/bugs/view.php?id=24624


vboctor:
Closing in favor of #1376

PR #1372, opened 2018-07-25
Fix tag checking when adding tags on report bug page

atrol:
Fixes #24632

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -355,7 +355,8 @@ $(document).ready( function() {
 		var currentTagString = $('#tag_string').val();
 		var newTagOptionID = $(this).val();
 		var newTag = $('#tag_select option[value=' + newTagOptionID + ']').text();
-		if (currentTagString.indexOf(newTag) == -1) {
+		var tagArray = currentTagString.split(tagSeparator);
+		if (tagArray.indexOf(newTag) == -1) {
 			if (currentTagString.length > 0) {
 				$('#tag_string').val(currentTagString + tagSeparator + newTag);
 			} else {


js/common.js
```diff
@@ -355,7 +355,8 @@ $(document).ready( function() {
 		var currentTagString = $('#tag_string').val();
 		var newTagOptionID = $(this).val();
 		var newTag = $('#tag_select option[value=' + newTagOptionID + ']').text();
-		if (currentTagString.indexOf(newTag) == -1) {
+		var tagArray = currentTagString.split(tagSeparator);
+		if (tagArray.indexOf(newTag) == -1) {
```
vboctor: Spacing is incorrect.

atrol: @vboctor please explain. I don't see what's wrong with it.
e.g., isn't the spacing of the 2nd line the same as before? Just replaced `currentTagString` by `tagArray`

vboctor: @atrol It is not about the indentation, but about the spacing around if statement and parenthesis.  See below fixed example:

```
		var tagArray = currentTagString.split( tagSeparator );
		if( tagArray.indexOf( newTag ) == -1 ) {
```

atrol: I just used the style that I found around those lines. 
AFAIK we don't have Coding Guidelines for JavaScript.
@vboctor so we should use the PHP Coding Guidelines, at least for spacing rules?

dregad: > so we should use the PHP Coding Guidelines [for JavaScript], at least for spacing rules?

I don't think we should.

PR #1373, opened 2018-07-25
Enhance error handling when resolving/closing issues

atrol:
Trigger error at an early state if user is not allowed to resolve/close
issues due to unresolved child issues.

Issue #24633

### DIFF
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -153,6 +153,9 @@ layout_page_begin();
 			<?php
 				if( $f_new_status >= $t_resolved ) {
 					if( relationship_can_resolve_bug( $f_bug_id ) == false ) {
+						if( OFF == config_get( 'allow_parent_of_unresolved_to_close' ) ) {
+							trigger_error( ERROR_BUG_RESOLVE_DEPENDANTS_BLOCKING, ERROR );
+						}
 						echo '<tr><td colspan="2">' . lang_get( 'relationship_warning_blocking_bugs_not_resolved_2' ) . '</td></tr>';
 					}
 				}


bug_change_status_page.php
```diff
@@ -153,6 +153,9 @@
 			<?php
 				if( $f_new_status >= $t_resolved ) {
 					if( relationship_can_resolve_bug( $f_bug_id ) == false ) {
+						if( OFF == config_get( 'allow_parent_of_unresolved_to_close' ) ) {
+							trigger_error( ERROR_BUG_RESOLVE_DEPENDANTS_BLOCKING, ERROR );
+						}
 						echo '<tr><td colspan="2">' . lang_get( 'relationship_warning_blocking_bugs_not_resolved_2' ) . '</td></tr>';
```
vboctor: Is this warning now not used?  Should we remove the echo and possibly the associated string?

bug_change_status_page.php
```diff
@@ -153,6 +153,9 @@
 			<?php
 				if( $f_new_status >= $t_resolved ) {
 					if( relationship_can_resolve_bug( $f_bug_id ) == false ) {
+						if( OFF == config_get( 'allow_parent_of_unresolved_to_close' ) ) {
```
vboctor: Does this error apply for both web UI and APIs?

bug_change_status_page.php
atrol: The warning / associated string  is still used if `$g_allow_parent_of_unresolved_to_close == ON`.

bug_change_status_page.php
atrol: Just web UI.

Before this change, we had it just on bug update page.
The idea is to throw the error as early as possible to avoid unneeded entering of data.
e.g. before the change the user typically set `Resolution`, set `Fixed in Version` and added a note and got the error message after clicking the `Resolve Issue` button.

We might want to have this check also in function validate of BugData class, but that's another story.

PR #1377, opened 2018-08-02
Uniformize display of failed group updates

dregad:
Standard and custom bug group actions did not provide the same user
experience when displaying error messages. bug_actiongroup.php displays
errors in a table, while bug_actiongroup_ext.php prints them with `<p>`
elements.

This commit refactors the display code on both pages, by moving it to a
new API function bug_group_action_print_results().

Also fixes an issue with the page footer, which was displayed behind the
sidebar due to invalid HTML.

Fixes [#24643](https://mantisbt.org/bugs/view.php?id=24643), [#24644](https://mantisbt.org/bugs/view.php?id=24644)

### DIFF
--- a/bug_actiongroup.php
+++ b/bug_actiongroup.php
@@ -341,27 +341,10 @@ foreach( $f_bug_arr as $t_bug_id ) {
 
 form_security_purge( $t_form_name );
 
-$t_redirect_url = 'view_all_bug_page.php';
-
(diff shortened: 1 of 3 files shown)


atrol:
Closed via https://github.com/mantisbt/mantisbt/commit/d0d6b012c95c3eb7913aa0f496241f57e430965e

PR #1378, opened 2018-08-09
Add wildcard search in manage_user_page.php

CasN:
Added wildcard search on realname including the option to exclude a pattern.
So searchterm "jan" would lead to wildcard "%JAN%"
but searchterm "jan -dek" would lead to include  "%JAN%" and to exclude "%DEK%"

Fixes [#12677](https://mantisbt.org/bugs/view.php?id=12677)

### DIFF
--- a/manage_user_page.php
+++ b/manage_user_page.php
@@ -198,9 +198,9 @@ if( $f_filter === 'ALL' ) {
 		$c_prefix= trim(substr($c_prefix, 0,$t_pos-1));
 		$t_where_params[] ='%'.$c_prefix . '%';
 		$t_where = db_helper_like( 'UPPER(realname)' );
-		$t_where .= " and ";
+		$t_where .= " and not ";
 		$t_where_params[] ='%'.$t_exclude . '%';
-		$t_where .= db_helper_not_like( 'UPPER(realname)' );
+		$t_where .= db_helper_like( 'UPPER(realname)' );
 	} else{
 		$t_where_params[] ='%'.$c_prefix . '%';
 		$t_where = db_helper_like( 'UPPER(realname)' );
--- a/manage_user_page.php
+++ b/manage_user_page.php
@@ -193,17 +193,38 @@ if( $f_filter === 'ALL' ) {
 } else if( $f_filter === 'SEARCH' ) {
 	$t_pos = strpos($f_findname, '-');
 	if( $t_pos ) {
-		$t_exclude = trim( substr( $f_findname, $t_pos+1 ) );
+		$t_exclude = trim( substr( $f_findname,$t_pos+1 ) );
 		$f_findname = trim( substr( $f_findname, 0,$t_pos-1 ) );
 		$t_where_params[] = '%' . $f_findname . '%';
-		$t_where = db_helper_like( 'realname' );
-		$t_where .= " AND NOT ";
+		$t_where = " ( ";
+		$t_where .= db_helper_like( 'realname' );
+		$t_where .= " OR ";
+		$t_where_params[] .= '%' . $f_findname . '%';
+		$t_where .= db_helper_like( 'username' );
+		$t_where .= " OR ";
+		$t_where_params[] .= '%' . $f_findname . '%';
+		$t_where .= db_helper_like( 'email' );
+		$t_where .= " ) AND NOT (";
 		$t_where_params[] = '%' . $t_exclude . '%';
 		$t_where .= db_helper_like( 'realname' );
+		$t_where .= " OR ";
+		$t_where_params[] .= '%' . $t_exclude . '%';
+		$t_where .= db_helper_like( 'username' );
+		$t_where .= " OR ";
+		$t_where_params[] .= '%' . $t_exclude . '%';
+		$t_where .= db_helper_like( 'email' );
+		$t_where .= " ) ";
 	} else {
-		$t_where_params[] = '%' . $f_findname . '%';
+		$t_where_params[] 	= '%' . $f_findname . '%';
 		$t_where = db_helper_like( 'realname' );
+		$t_where .= " OR ";
+		$t_where_params[] 	= '%' . $f_findname . '%';
+		$t_where .= db_helper_like( 'username' );
+		$t_where .= " OR ";
+		$t_where_params[] 	= '%' . $f_findname . '%';
+		$t_where .= db_helper_like( 'email' );
 	}
+
 } else if( $f_filter === 'UNUSED' ) {
 	$t_where = '(login_count = 0) AND ( date_created = last_visit )';
 } else if( $f_filter === 'NEW' ) {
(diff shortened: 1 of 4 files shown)


manage_user_page.php
```diff
@@ -185,6 +190,17 @@
 $t_where_params = array();
 if( $f_filter === 'ALL' ) {
 	$t_where = '(1 = 1)';
+} else if( $f_findname<>"" ) {
+	$c_prefix = strtoupper(db_prepare_string($f_filter));
```
atrol: function db_prepare_string is deprecated and should not be used any longer, see https://github.com/mantisbt/mantisbt/blob/release-2.16.0/core/database_api.php#L620

manage_user_page.php
```diff
@@ -185,6 +190,17 @@
...
+	$pos = strpos($c_prefix, '-');
+	if ($pos){
+		$exclude = trim(substr($c_prefix,$pos+1));
+		$c_prefix= trim(substr($c_prefix, 0,$pos-1));
+		$t_where = "(upper(realname) like '%$c_prefix%' and upper(realname) not like '%$exclude%')";
```
atrol: There is a function db_helper_like that provides case-insensitive like phrase based on the current db type (e.g. works with MySQL and PostgreSQL, see https://github.com/mantisbt/mantisbt/blob/release-2.16.0/core/database_api.php#L726
I expect a better performance on installations with lot's of users as I assume that your implementation leads to a full table scan.

manage_user_page.php
```diff
@@ -185,6 +190,17 @@
...
+	} else{
+		$c_prefix= trim(substr($c_prefix, 0,$pos-1));
+		$t_where = "(upper(realname) like '%$c_prefix%')";
```
atrol: Same as above

atrol:
- There should be no difference of the buttons in terms of styling
managuser

- Bug when entering a single character in search field
You don't get the expected users, e.g. enter `a` but get a list of users without `a` in real name or empty real name

CasN:
Roland, I have adressed all your issues and fixed the bug. In additiojn I have added a function "db_helper_not_like" to the database_api to bring all inline.

cproensa:
> I have added a function "db_helper_not_like" to the database_api to bring all inline.

What is the difference between `A not like 'E'` and `not A like 'E'` ?
If none, using the second construct there is no need for the new function

dregad:
> Please be more specific 

http://www.mantisbt.org/wiki/doku.php/mantisbt:coding_guidelines#braces_and_parentheses
Quite a few space around parenthesis. Also add spaces around the operator in expressions.

manage_user_page.php
```diff
@@ -64,6 +64,11 @@
 $f_filter        = mb_strtoupper( gpc_get_string( 'filter', config_get( 'default_manage_user_prefix' ) ) );
 $f_page_number   = gpc_get_int( 'page_number', 1 );
 
+$f_findname = gpc_get_string( 'findname', '' ); 
+if($f_findname<>""){
```
dregad: here

manage_user_page.php
```diff
@@ -64,6 +64,11 @@
...
+	$f_filter=$f_findname;
```
dregad: here

manage_user_page.php
```diff
@@ -185,6 +190,21 @@
...
+	$c_prefix = strtoupper($f_filter);
+	$t_pos = strpos($c_prefix, '-');
+	if($t_pos){
+		$t_exclude = trim(substr($c_prefix,$t_pos+1));
+		$c_prefix= trim(substr($c_prefix, 0,$t_pos-1));
+		$t_where_params[] ='%'.$c_prefix . '%';
+		$t_where = db_helper_like( 'UPPER(realname)' );
+		$t_where .= " AND NOT ";
+		$t_where_params[] ='%'.$t_exclude . '%';
+		$t_where .= db_helper_like( 'UPPER(realname)' );
```
dregad: here and the above 10 lines

manage_user_page.php
```diff
@@ -64,6 +64,8 @@
 $f_filter        = mb_strtoupper( gpc_get_string( 'filter', config_get( 'default_manage_user_prefix' ) ) );
 $f_page_number   = gpc_get_int( 'page_number', 1 );
 
+$f_findname	 = mb_strtoupper( gpc_get_string( 'findname', '' ) );
```
atrol: mb_strtoupper not needed as db_helper_like provides case-insensitive comparison.

manage_user_page.php
```diff
@@ -185,6 +187,19 @@
 $t_where_params = array();
 if( $f_filter === 'ALL' ) {
 	$t_where = '(1 = 1)';
+} else if( $f_findname <> "" ) {
+	if( $t_pos ){
+		$t_exclude		= trim( substr( $f_findname,$t_pos+1 ) );
+		$f_findname		= trim( substr( $f_findname, 0,$t_pos-1 ) );
+		$t_where_params[]	= '%'.$f_findname . '%';
+		$t_where		= db_helper_like( 'UPPER(realname)' );
```
atrol: UPPER not needed as db_helper_like provides case-insensitive comparison.

manage_user_page.php
```diff
@@ -185,6 +187,19 @@
...
+		$t_where		.= " AND NOT ";
+		$t_where_params[] 	= '%'.$t_exclude . '%';
+		$t_where 		.= db_helper_like( 'UPPER(realname)' );
```
atrol: same as above

manage_user_page.php
```diff
@@ -185,6 +187,19 @@
...
+	} else {
+		$t_where_params[] 	= '%'.$f_findname . '%';
+		$t_where 		= db_helper_like( 'UPPER(realname)' );
```
atrol: same as above

atrol:
The functionality is broken after the latest change.

atrol:
After searching for a name and getting the result list, the UI does not work as expected.
E.g. the filter is lost when sorting columns, clicking the `Apply filter` button or clicking the `Next` button.

atrol:
@CasN using the master branch for PR's is no good idea in general.
Using the same branch for different PR's is even worse, as 
E.g. this PR contains now also the changes of your PR https://github.com/mantisbt/mantisbt/pull/1380
You should create a separate branches.
https://www.mantisbt.org/docs/master/en-US/Developers_Guide/html-desktop/#dev.contrib.prepare
https://www.mantisbt.org/docs/master/en-US/Developers_Guide/html-desktop/#dev.contrib.submit.github

atrol:
@cas let me know if you want to close this one in favor of https://github.com/mantisbt/mantisbt/pull/1381 where I

- rebased based on latest master
- squashed commits while keeping commits of cas and dregad
- minor code style corrections
- removed unrelated changes (config_defaults_inc.php)
- replaced `User Name` label by placeholder `Username, Real Name or E-mail`

atrol:
Closed in favor of https://github.com/mantisbt/mantisbt/pull/1381

PR #1381, opened 2018-08-13
 Enhance user search UI on page manage_user_page.php

atrol:


manage_user_page.php
```diff
@@ -185,14 +194,49 @@
...
+		$t_where .= " ) AND NOT (";
+		$t_where_params[] = '%' . $t_exclude . '%';
+		$t_where .= db_helper_like( 'realname' );
+		$t_where .= " OR ";
+		$t_where_params[] .= '%' . $t_exclude . '%';
+		$t_where .= db_helper_like( 'username' );
+		$t_where .= " OR ";
+		$t_where_params[] .= '%' . $t_exclude . '%';
+		$t_where .= db_helper_like( 'email' );
+		$t_where .= " ) ";
```
dregad: To make code more readable I would propose instead

``` php
$t_where = sprintf( ' ( %s OR %s OR %s ) AND NOT ( %s OR %s OR %s ) ',
    db_helper_like( 'realname' ),
    db_helper_like( 'username' ),
    db_helper_like( 'email' ),
    db_helper_like( 'realname' ),
    db_helper_like( 'username' ),
    db_helper_like( 'email' )
);
$t_where_params[] = '%' . $f_findname . '%';
$t_where_params[] = '%' . $f_findname . '%';
$t_where_params[] = '%' . $f_findname . '%';
$t_where_params[] .= '%' . $t_exclude . '%';
$t_where_params[] .= '%' . $t_exclude . '%';
$t_where_params[] .= '%' . $t_exclude . '%';
```

manage_user_page.php
```diff
@@ -185,14 +194,49 @@
 $t_where_params = array();
 if( $f_filter === 'ALL' ) {
 	$t_where = '(1 = 1)';
+} else if( mb_substr( $f_filter, 0, 6 ) === 'SEARCH' ) {
+	$t_pos = mb_strpos( $f_findname, '-' );
```
dregad: I did not actually test the code (no time, sorry) but I think this will not work as expected when searching for a string containing a `-` e.g. `jean-pierre`

CasN: Fair point the '-' character is used to indicate start of exclusion string. Do you have a suggestion for another character?

atrol: I have no time to have a deeper look at the moment but adding a space might work
`$t_pos = mb_strpos( $f_findname, ' -' );`

CasN: Would it be acceptable to use ' -'  (so space & '-')?

CasN: Have tested ' -' and that works. Since I cannot make a change here, below what it should be:
	$t_pos = strpos($f_findname, ' -');
	if( $t_pos ) {
		$t_exclude = trim( substr( $f_findname,$t_pos+2 ) );
		$f_findname = trim( substr( $f_findname, 0,$t_pos ) );

dregad: This is not very robust. What if I search for `-xxx` ?

samtuke:
@casN @atrol Once this is merged please contact me at sam at phplist dot com to claim the bounty offered on the Mantis issue tracker.

atrol:
> It does not play nice with the existing filter (A-Z buttons) because they reset each other. It would probably be better and less confusing to users to combine the filter criteria

I am not sure if this is a good idea, at least as long as we have `$g_default_manage_user_prefix`.
E.g. we have set it to `0` on mantisbt.org/bugs, so you first would have to click `All` if you want to search for any user where the username does not start with `0`.

> I would move the search input to the top of the page, so that all the filter-related controls are in a single place

You mean something like that?
user-filter

dregad:
>> It does not play nice with the existing filter (A-Z buttons) because they reset each other. It would probably be better and less confusing to users to combine the filter criteria
>
> I am not sure if this is a good idea, at least as long as we have `$g_default_manage_user_prefix`.
E.g. we have set it to `0` on mantisbt.org/bugs, so you first would have to click `All` if you want to search for any user where the username does not start with `0`.

IMHO this setting does not make any sense. On my dev box, the page loads in 0.7s for all users (~37K records) , and under 0.1s on mantisbt.org. I measure no significant difference when loading a smaller set of users (e.g. with $g_default_manage_user_prefix = 0), we only retrieve 50 records anyway.

As far as I'm concerned, we could obsolete `$g_default_manage_user_prefix`.

I still believe it makes sense to improve the UX by providing the possibility to restrict the filter by combining the prefix with the search criteria.

> I would move the search input to the top of the page, so that all the filter-related controls are in a 2single place
>
> You mean something like that?
user-filter

Yes, that's the idea.

atrol:
> As far as I'm concerned, we could obsolete `$g_default_manage_user_prefix`.

+1
The setting has been introduced in 2003 https://github.com/mantisbt/mantisbt/commit/2e2cb35d4c6e91f4c99ea25aac7256d20d696ac6
IMHO it could have been removed since the paging (hardcoded 50 records) has been implemented in 2008. https://github.com/mantisbt/mantisbt/commit/5e45f17aa9f0116d1ebed280c2ca6d9974ad12ff

Maybe we could even remove the whole clickable character list?
I don't see the advantage to combine this quite limited filter (**username** starts with a character) with the more powerful new filter.

All in all, I am asking myself, if we should merge to master what we have at the moment, especially as there is a user who offered a bounty is waiting for it.

Obsoleting the option and UI redesign could be done in a second step.

dregad:
> I don't see the advantage to combine this quite limited filter (username starts with a character) with the more powerful new filter.

I agree with you if we have the perspective of removing the clickable alphanum index. If not, we need to have consistent behavior.

This is nice-to-have, but maybe the new search functionality can be improved to restrict searching to some fields (e.g. only username), and also allow "starts with" (`xxx%`) or "search anywhere (`%xxx%`).

Anyway, I'm ok with postponing the following items :

- obsolete $g_default_manage_user_prefix
- remove the whole clickable character list (but we need to ensure we provide an alternative for the _Unused_ and _New_ filter buttons)

However I'd rather include moving the search field to the top as part of this PR.

atrol:
@dregad, I think I addressed all feedback.
Todo: Check / fix XSS when entering malicious search terms.

dregad:
I did not notice that (as I said, it was a quick test). But your approach makes sense and I'm fine with it. The new functionality is already a big improvement over what it was before.

Thanks !

samtuke:
Bounty paid.

PR #1382, opened 2018-08-17
Fix filter by date for the soap api

raimu:
In the soap endpoint “mc_filter_search_issues”, it is possible to
filter the search result by date. At least it is defined in the wsdl.
But as described in the issue #20101, this parameter has no
effect (yet).

Therefore handle the date parameters manually in the mc_filter_api,
so that it will be appended to the filter.

After this change it is possible to filter the search result by a
date range.


api/soap/mc_filter_api.php
```diff
@@ -341,6 +341,16 @@ function mci_filter_search_get_rows( $p_user_id, $p_filter_search, $p_page_numbe
 		}
 	}
 
+	// date fields
+	if( isset ( $t_filter[FILTER_PROPERTY_DATE_SUBMITTED_START_DAY] ) 
+		or isset ( $t_filter[FILTER_PROPERTY_DATE_SUBMITTED_START_MONTH] )
```
vboctor: We typically use `||` instead of `or`.

PR #1383, opened 2018-08-27
Add update_bug_threshold check for custom field bulk operations

atrol:
Fixes #24696

### DIFF
--- a/core/bug_group_action_api.php
+++ b/core/bug_group_action_api.php
@@ -212,11 +212,16 @@ function bug_group_action_get_commands( array $p_project_ids = null ) {
 		$p_project_ids = array( ALL_PROJECTS );
 	}
 
+	$t_user_id = auth_get_current_user_id();
+
 	$t_commands = array();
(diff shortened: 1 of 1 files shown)


core/bug_group_action_api.php
```diff
@@ -319,13 +319,13 @@ function bug_group_action_get_commands( array $p_project_ids = null ) {
 
 		$t_custom_field_ids = custom_field_get_linked_ids( $t_project_id );
 		foreach( $t_custom_field_ids as $t_custom_field_id ) {
-			if( !custom_field_has_write_access_to_project( $t_custom_field_id, $t_project_id ) ) {
-				continue;
+			if( custom_field_has_write_access_to_project( $t_custom_field_id, $t_project_id ) &&
+				access_has_project_level( config_get( 'update_bug_threshold', null, null, $t_project_id ), $t_project_id ) ) {
```
vboctor: Should this check be outside the foreach?

atrol: Fixed + some more code / performance enhancements

PR #1384, opened 2018-08-31
Add filter for the “last updated“ date in the soap api

raimu:
Currently the soap method `mc_filter_search_issues` allows to filter
the result set by the “sumbitted” date of the issue. But not with the
“last updated” date.

For a task, we need to get the latest updated issues, to execute some
business logic. Therefore expand the soap api, with this additional
filter, so that's now also posssible to filter the result-set of the
search with the last “updated date”.

- Add this new filter to the `mc_filter_api`, simular like the
  “submitted date”.
- Add it also the the wsdl, so that this new api is documented for
  other programms.

### DIFF
--- a/api/soap/mantisconnect.wsdl
+++ b/api/soap/mantisconnect.wsdl
@@ -296,6 +296,12 @@
    <xsd:element name="end_day" type="xsd:integer" minOccurs="0"/>
    <xsd:element name="end_month" type="xsd:integer" minOccurs="0"/>
    <xsd:element name="end_year" type="xsd:integer" minOccurs="0"/>
+   <xsd:element name="last_update_start_day" type="xsd:integer" minOccurs="0"/>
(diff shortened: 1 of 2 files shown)


atrol:
@rombert, is changing the WSDL a breaking change?

@raimu please open a report at https://www.mantisbt.org/bugs and amend commit message to add
```
Fixes #<IdOfTheIssue>
```

raimu:
The new issue for this feature was created, and amend to the commit message.

atrol:
Thanks @raimu,
I am fine with the change, but would like to get the feedback of your SOAP specialist rombert, to be sure we can change this in version 2.x.x without breaking anything.

rombert:
This WSDL change should not break older clients. The only breaking scenario is if the server somehow returns a `FilterSearchData` object with the new fields to older clients. But not sure if that can happen.

PR #1388, opened 2018-09-10
Always specify fallback fonts in case Open Sans cannot be loaded.

sagamusix:
This fixes https://www.mantisbt.org/bugs/view.php?id=24520

Maybe it would be more desirable to actually remove all those font definitions for headers, etc. because in the end the same font is used everywhere.


dregad:
@vboctor I'm not really sure what your expectation is with assigning the PR to @syncguru. Are there any objections with merging this ? Is there some more work / changes needed ?

syncguru:
Looks good. 
For some reason, I don't get notifications from github.

dregad:
> Looks good.

Thanks for the feedback, nice to see you're still around after all :wink: 

> For some reason, I don't get notifications from github.

Do you have the right "watch" settings ? Are the notifications caught by your spam filter ?
image

PR #1389, opened 2018-09-12
Fix BugData object to string conversion on string_get_bug_view_url() function

clagiordano:
Dear maintainer,

this PR fix annoying bug when you try to change bug status (e.g. change to resolve ) and add it a note,
in this case, $p_bug_id is an instance of BugData class instead a number or string causing following error:

> SYSTEM ERROR

> 'Object of class BugData could not be converted to string' in '/var/www/html/mantis/core/string_api.php' line 717

> Please use the "Back" button in your web browser to return to the previous page. There you can correct whatever problems were identified in this error or select another action. You can also click an option from the menu bar to go directly to a new section.

best regards.

### DIFF
--- a/core/string_api.php
+++ b/core/string_api.php
@@ -714,6 +714,10 @@ function string_get_bugnote_view_link( $p_bug_id, $p_bugnote_id, $p_detail_info 
  * @return string
  */
 function string_get_bug_view_url( $p_bug_id ) {
+    if ($p_bug_id instanceof BugData) {
+        $p_bug_id = $p_bug_id->id;
+    }
+
 	return 'view.php?id=' . $p_bug_id;
 }
 


atrol:
> fix annoying bug when you try to change bug status (e.g. change to resolve ) and add it a note

@clagiordano I was not able to reproduce this bug.
Please provide detailed step by step instructions to reproduce.

dregad:
@clagiordano you did not provide any feedback, so we're closing this PR. Feel free to reopen it with the requested information.

clagiordano:
Sorry for late response, I tried many times but I could not reproduce it on a clean environment but it occurs often on my production machine with different users with different access level.

Step by step procedure (**not on clean environment**):

- a user of group updater/reporter open a bug
- a developer (member of) see that bug and add a comment (sometimes occurs in this case)
- a developer change bug status to resolved with or without adding a resolution note (often occurs in this case)
- mantis show "system error" message
- back to bug and usually this has been updated but in some case bug does not been updated.

On my production machine I fixed that with suggested patch and this issue does not occurs anymore on similar scenarios.

atrol:
> but I could not reproduce it on a clean environment

- Do you use latest version 2.17.1 ?
- Did you change source code of MantisBT ?
- Did you install 3rd party plugins ?
- Do you use custom functions ?

To get some more debugging information you could change your code to
```
 if ($p_bug_id instanceof BugData) {
         $p_bug_id = $p_bug_id->id;
         $g_display_errors = array(
	         E_WARNING => 'halt',
	         E_NOTICE => 'halt',
	         E_USER_ERROR => 'halt',
	         E_USER_WARNING => 'halt',
	         E_USER_NOTICE => 'halt'
         );
         $g_show_detailed_errors = ON;
         trigger_error( ERROR_GENERIC, ERROR );
 }
```

clagiordano:
Thanks for the info, at moment I use 2.18-dev but this problems also occurs on previous versions.
I use only built in plugin except slack integration installed from https://github.com/mantisbt-plugins/Slack

- Avatars via Gravatar 2.18.0-dev
- Import/Export issues 2.18.0-dev
- Mantis Graphs 2.18.0-dev
- MantisBT Formatting 2.18.0-dev
- Slack Integration 1.0

Maybe this issue is caused by the last one ?

I'll try to enable extended debug information as suggested.

Thanks.

atrol:
> Slack Integration 1.0
> Maybe this issue is caused by the last one ?

Maybe, but I didn't have a deeper look.
Install it to your clean install and/or uninstall it from your production install to find out if it's causing the problem.

PR #1390, opened 2018-09-13
Improve presentation of temporary filter

cproensa:
Relevant changes:

commit 93bda1b693778cee4aef653155738c7d577e718b
    Improve presentation of temporary filters
    
    - Show a visual indication in the filter widget icon when the active
    filter is a temporary filter
    - Create a filter action to make the current temporary filter a
      persistent filter
    
    Fixes: #0024775

commit 0fc7bb7c275f70a181109354f21062674ac7c525
    Fix switching view mode for temporary filters
    
    Fix switching view mode for a temporary filter.
    Previously, switching the advanced or simple view for a temporary
    filter, would lose the filter.
    
    Fixes: #0024776

commit 151719b97bafa4a94999392967e0940efaeafe68
Add actions as buttons in filter dialog
    
    Add some actions as buttons placed in the filter dialog working area.
    Replace old buttons having verbose decription by new iconized buttons,
    for actions "reset filter" and "save filter".
    Include new action button for persisting a temporary filter.

commit 95a60b02707efbeafe6f22b18069e5de87ca3d22
Clean up filter dialog code
    
    Remove unused function parameters
    
    $p_page_number is not needed, the filter does not manage pagination
    $p_for_screen is not used
    $p_expanded is not used. If a reduced form variant is ever needed, it
    will be a better approach to write new code that manages the collapsed
    state.
    
    Remove unused form inputs
    
    view_all_set.php?f=3 the "f" parameter is not used
    "print" form parameter is not used
    "offset" form parameter is not used
    "page_number" form parameter is no used

commit 6e3d83cb2804ca33789153fbd0e10665fe2fd2a2
Create view_all_set.php constants
    
    Replace magic numbers with constants in view_all_set.php for filter
    actions.

commit 873603435bda0d50d179ffe678ec9c93b53e33a5
Rearrange filter dialog widget
    
    There were some issues withthe filter dialog:
    - The form tags are not placed correctly in the layout, crossing inner
      div boundaries.
    - The section at the bottom is not hidden when the widget is collapsed.
    - Only one form is used, but there are several actions that don't need
      to share all form fields.
    
    In order to improve some of the dialog options, the layour has been
    rearranged:
    
    - Move the buttons toolbar to the top of the widget, following the
      general style of other widgets.
    - Move the "text search" field inside the filter parameters area. This
      is part of the filter properties and should be managed and rendered by
    the dedicated code.
    - Assign each area its own form. Time ago, the filter submit would parse
      all the fields as a whole, but now there's the option to evaluate only
    the submitted fields, incrementally over the affected filter. This
    allows to separate the header search field, filter load dorwpdowns, etc,
    into individual forms.
    
    Fixes: #25109


cproensa:
Minimal changes to the filter dialog:
- a modifier for the filter icon when the filter is temporary
- a new option to make the filter persistent

menu_308

The option to make the filter persistent could be made to be faster access. For example, placing a pushpin (or other icon) button in the widget bar?

atrol:
Tested a bit and didn't encounter any error so far.

I am not sure if users without some background knowledge of the reported issue will understand the UI.

They might notice the small clock icon, they might get the tool tip (not on touch only devices) but will they understand what it means and will they find the new menu entry to influence it?

TBH I have no other/better idea at the moment.

core/filter_api.php
```diff
@@ -2526,22 +2526,37 @@ function filter_draw_selection_area( $p_page_number, $p_for_screen = true, $p_ex
 		<input type="hidden" name="view_type" value="<?php echo $t_view_type?>" />
 	<?php
 	$t_stored_queries_arr = filter_db_get_available_queries();
+	$t_is_temporary = filter_is_temporary( $t_filter );
 	if( $p_expanded ) {
 		$t_collapse_block = is_collapsed( 'filter' );
 		$t_block_css = $t_collapse_block ? 'collapsed' : '';
 		$t_block_icon = $t_collapse_block ? 'fa-chevron-down' : 'fa-chevron-up';
+
+		# further use of this icon must be inlined to avoid spaces in rendered html
```
atrol: What about using ternary operator?
```
$t_temporary_icon_html = $t_is_temporary ? '<i class="fa fa-clock-o fa-xs-top" title="' . lang_get( 'temporary_filter' ) . '"></i>' : '';
```

core/filter_api.php
```diff
@@ -2526,22 +2526,37 @@ function filter_draw_selection_area( $p_page_number, $p_for_screen = true, $p_ex
...
+				<i class="ace-icon fa fa-filter"><?php echo $t_temporary_icon_html ?>
+				</i>
 				<?php echo lang_get( 'filters' ) ?>
 			</h4>
+
 			<div class="widget-toolbar">
 				<?php
 					$t_view_filters = config_get('view_filters');
 
+					if( $t_is_temporary ) {
```
atrol: Again, what about using ternary operator?

lang/strings_english.txt
```diff
@@ -1259,6 +1259,8 @@ $s_manage_filter_page_title = 'Manage Filters';
 $s_manage_filter_edit_page_title = 'Edit filter';
 $s_delete_filter_button = 'Delete';
 $s_apply_filter_button = 'Apply';
+$s_temporary_filter = 'Temporary filter';
+$s_set_as_persistent_filter = 'Set as persistent filter';
```
atrol: Maybe a native speaker can comment.
Shouldn't it be `Set as persisted Filter``

view_all_set.php
```diff
@@ -168,8 +182,14 @@
 
 # If only using a temporary filter, don't store it in the database
 if( !$t_temp_filter ) {
-	# Store the filter string in the database: its the current filter, so some values won't change
-	filter_set_project_filter( $t_setting_arr );
+	# get project if it was specified
+	if( -1 == $f_project_id ) {
```
atrol: Again, ternary operator?

lang/strings_english.txt
dregad: I think _persistent_ is correct in this context. I would say _**Save** as persistent filter_ though.

dregad:
> The option to make the filter persistent could be made to be faster access. For example, placing a pushpin (or other icon) button in the widget bar?

@cproensa Following up on @atrol's remarks on UX, I agree that even though it works fine, the UI can be somewhat confusing for users who are not aware of the background. 

What about making the "filter with clock" icon active as a button to save the filter (with pop-up text like _This is a temporary filter, click to save it_ or something along those lines ? This could be a replacement or in addition to the menu option under the hamburger icon.

cproensa:
> I am not sure if users without some background knowledge of the reported issue will understand the UI.

I didn't intend to make any UI changes that may break current styling. That is a sensitive topic and i am not confident to push that one through by myself.
The relevant change is to enable the action, which now can be repeated with the same request as the option i added to the drop-down menu, in any other UI control that is proposed.

> What about making the "filter with clock" icon active as a button to save the filter (with pop-up text like This is a temporary filter, click to save it or something along those lines ? This could be a replacement or in addition to the menu option under the hamburger icon.

The clock icon is well suited as "informative" of filter's current nature. The action to persist it is not well represented by the clock. I thought of the thumb-tack icon, but placing that one on the widget bar is more commonly related to fixing or anchoring a widget/window, so it may be confusing.

dregad:
> The action to persist it is not well represented by the clock. 

Hence my suggestion to have an explanatory pop-up, both to enhance the meaning of the "filter with clock" icon, and to explain how to make the temp filter permanent.

I agree that adding a thumb-tack icon on the toolbar may be even more confusing than the way you've implement it here.

cproensa:
there has been some discussion on https://mantisbt.org/bugs/view.php?id=24549
about presentation.
Can you have a look to see if it's viable?

dregad:
As mentioned in the tracker, I'm OK with this change but would prefer to have the enhanced UI included as part of this PR.

cproensa:
Updated PR with a deeper changes in the filter dialog UI.
A considerable clean up was needed, in order to correctly implement the new changes. The layout for the widget was outdated and had some errors and  inconsistencies

Standard filter:


Temporary filter:


Collapsed widget:

atrol:
What about having the same UI for collapsed and non-collapsed view?
e.g. something like that
filtercollapsed

cproensa:
> What about having the same UI for collapsed and non-collapsed view?

The framework hides the widget body when collapsed. The fact that we had something still there was due to the broken structure of the form tags, and making the browser to render the bottom div outside of the proper body.
So, in order to achieve your suggestion of collapsing but still showing something as "body", i fear we would need to fight against the framework.

I understand the point against crowding the widget bar with a lot of items. Note also that those can be selectively hidden at different viewport sizes.

atrol:
> I understand the point against crowding the widget bar

It's not that much against crowding the widget bar, it's more about

- different UI's for the same functionality confuses user
- the additional line with the `Search label` looks not that good and wastes some vertical space

Just brainstorming, maybe it's an option that the Search / Apply Filter / Reset / Load, ... line is no longer part of the collapsable area, but a standalone line or part of the `Viewing Issues` section below.

cproensa:
>  the additional line with the Search label looks not that good and wastes some vertical space

yes... on one side, my reasoning is that the "text search" is part of the filter properties, so its logical place is with the rest of the filter fields. One objective of the filter form is to be reusable, however at the moment, we have to render the search field separately from the main filter fields (eg, see also editing a stored filter). Being part of the filter fields is one step towards modularity of the form.
We could style it in some other way to stand out differently from the other fields. The issue about vertical space: we could also rearrange some other fields around it.

> Just brainstorming, maybe it's an option that the Search / Apply Filter / Reset / Load, ... line is no longer part of the collapsable area, but a standalone line or part of the Viewing Issues section below.

i think that is even worse, as we would be crossing functions over different containers.

core/gpc_api.php
```diff
@@ -464,7 +464,13 @@ function gpc_make_array( $p_var_name ) {
  * @return boolean
  */
 function gpc_string_to_bool( $p_string ) {
-	if( 0 == strcasecmp( 'off', $p_string ) || 0 == strcasecmp( 'no', $p_string ) || 0 == strcasecmp( 'false', $p_string ) || 0 == strcasecmp( '', $p_string ) || 0 == strcasecmp( '0', $p_string ) ) {
+	if( 0 == strcasecmp( 'off', $p_string )
+		|| 0 == strcasecmp( 'no', $p_string )
+		|| 0 == strcasecmp( 'n', $p_string )
```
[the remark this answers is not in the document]
cproensa: iirc i added this as we already were using "temporary=y", which would resolve to true by default.

cproensa:
Updated to keep the search field as previously placed. For view issues filtering

imagen

atrol:
- when being logged in as anonymous user, persisting filters is offered (remove it?), but does not work
- list of filters contains not needed entry [Reset Filter] in collapsed mode

core/gpc_api.php
```diff
@@ -464,7 +464,13 @@ function gpc_make_array( $p_var_name ) {
...
+		|| 0 == strcasecmp( 'false', $p_string )
+		|| 0 == strcasecmp( '', $p_string )
+		|| 0 == strcasecmp( '0', $p_string )
```
dregad: One more thing... `strcasecmp` call is not needed for the last 2 items. Furthermore, to avoid multiple function calls, I would suggest to start by converting $p_string to lowercase; you can then perform simple string comparisons for all values.

Finally, wouldn't it be better to test against "truthy" values and default anything else to false, instead of the reverse as you've currently implemented it ?

cproensa:
>  I think that for consistency they should appear in the same overall sequence as when the filters section is expanded (Persist / Reset / Save / Load / Search / Apply).

Yes, that seems a good suggestion. 

> I'm not sure the separator between Load Filter and Search is necessary.

The order, and the separator, were already there, so did'nt think much about further changes. Probably we could get rid of that separator too.

> Also, what is the reason for not having a Save button in collapsed mode ?

My thinking was to not add too much stuff in the widget bar. Saving a filter is usually an action done after tweaking the filter parameters, so in that scenario, the widget should be expanded to do that. 
And in any case, is an action that should not be that frequently used to have its space in the collapsed bar.

core/gpc_api.php
cproensa: That function has not been changed, apart from adding the new option for parsing "y/n"

>  I would suggest to start by converting $p_string to lowercase

ok

> Wouldn't it be better to test against "truthy" values and default anything else to false, instead of the reverse

That makes sense. However note that it should be refactored with care. As i explained, adding the "y/n" case was needed because we already used "temporary=y" for a long time (see congif_defauls_inc), which defaulted to true even if it was not a supported boolean string. Other similar cases may exist in the wild.

view_all_set.php
```diff
@@ -72,16 +72,16 @@
 if( $f_isset_temporary ) {
 	# when only changing the temporary status of a filter and no action is specified
 	# we assume not to reset current filter
-	if( $f_type < 0 ) {
+	if( $f_type == -1 ) {
```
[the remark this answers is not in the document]
cproensa: -1 is not an actual value, it's a forced return by gpc_get_xxx to detect that the parameter has not been used.
We could use a different literal, but i don't think it should be it's own filter related constant.

core/gpc_api.php
dregad: In that case, let's keep it as it is

dregad:
> > Also, what is the reason for not having a Save button in collapsed mode ?
> 
> My thinking was to not add too much stuff in the widget bar. Saving a filter is usually an action done after tweaking the filter parameters, so in that scenario, the widget should be expanded to do that.
> And in any case, is an action that should not be that frequently used to have its space in the collapsed bar.

That makes sense. Let's keep it like that then.

cproensa:
i have adjusted strings, see https://github.com/mantisbt/mantisbt/pull/1390/commits/6ae02b9c85443dea9beba48e40f94946e940c3af
(after seeing this https://mantisbt.org/bugs/view.php?id=24549#c61152)

Note that i thought about generic keywords as "save" and "reset", because the original buttons placement had less width available. With the new location, we could keep using filter specific keywords: "save filter", "reset filter" (this one already exists).
I still feel it's good to have context-free common keywords, so i am ok with either option.

what do you think?

cproensa:
> atrol: Just brainstorming, no strong opinion at the moment.

Valid points. Let's hear @dregad @vboctor for some more input and finish this.

I'd rather have this finished asap, because i have another PR which depends directly on this one.

atrol:
> I'd rather have this finished asap,

Right, the discussion is no reason to prevent merging.

cproensa:
> the discussion is no reason to prevent merging.

Well, preferably, we should make any change to lang strings and its var names before merging

dregad:
In my opinion, there is no need to append "filter" on each button/label - it is redundant and quite clear from context what we are working with (it says so in the title bar right above the buttons / select). Unless these strings are also used elsewhere with a less obvious intention.

I agree with @cproensa's suggestion that the select's label can be shortened to `Load`; in that case the language string should probably be changed from `$s_load_filter` to `$s_load` (and maybe the same for `$s_persist_filter` too, for consistency).

Note that the "Reset Filter" language string (reset_query) is used as tooltip (see b54c1cac2696ed5fa766c8e9f15225fe2a0a73a4) so this one should probably be kept.

cproensa:
> I agree with @cproensa's suggestion that the select's label can be shortened to Load; 

it was atrol's suggestion...

Thanks for the feedback. I've deleted the previous commit and modified to the short "load" string.
So now it stays with the short, one word, action buttons/labels.

PR #1391, opened 2018-09-30
Support new tags when creating new issues

vboctor:
Fixes #24774

### DIFF
--- a/core/commands/IssueAddCommand.php
+++ b/core/commands/IssueAddCommand.php
@@ -217,6 +217,16 @@ class IssueAddCommand extends Command {
 			}
 		}
 
+		if( isset( $t_issue['tags'] ) && is_array( $t_issue['tags'] ) && !tag_can_create( $this->user_id ) ) {
+			foreach( $t_issue['tags'] as $t_tag ) {
+				if( $t_tag['id'] === -1 ) {
+					throw new ClientException(
+						sprintf( "User '%d' can't create tag '%s'.", $this->user_id, $t_new_tag ),
+						ERROR_TAG_NOT_FOUND );
+					}
+			}
+		}
+
 		$t_category = isset( $t_issue['category'] ) ? $t_issue['category'] : null;
 		$t_category_id = mci_get_category_id( $t_category, $t_project_id );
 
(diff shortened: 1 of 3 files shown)


core/commands/IssueAddCommand.php
```diff
@@ -332,7 +332,16 @@ protected function process() {
 
 		# Add Tags
 		if( isset( $t_issue['tags'] ) && is_array( $t_issue['tags'] ) ) {
-			mci_tag_set_for_issue( $t_issue_id, $t_issue['tags'], $this->user_id );
+			$t_tags = array();
+			foreach( $t_issue['tags'] as $t_tag ) {
+				if( $t_tag['id'] === -1 ) {
+					$t_tag['id'] = tag_create( $t_tag['name'] );
```
atrol: You could use `$this->user_id` as additional parameter to `tag_create`

atrol:
When trying to attach a non-existing tag without having tag_create_threshold right, you get just a `Access denied` message.
In versions < 2.13.0 there was a `Tag attachment failed` message.

core/commands/IssueAddCommand.php
vboctor: Added user_id to the call.

vboctor:
@atrol I have added validation to stop creation of the issue and prompt a tag not found error.  User can click back, fix the issue, and submit again.

An orthogonal issue we have is that the tag not found error doesn't include the tag name that isn't found.  I haven't changed that as part of this change since this the issue targeted by this issue.

core/tag_api.php
```diff
@@ -529,6 +529,25 @@ function tag_get_field( $p_tag_id, $p_field_name ) {
...
+ * @return bool true: can create, false: otherwise.
+ */
+function tag_can_create( $p_user_id = null ) {
+	return access_has_global_level( config_get( 'tag_create_threshold' ), $p_user_id );
+}
+
+/**
+ * Ensure specified user can create tags.
+ *
+ * @param integer $p_user_id     The user ID to link the new tag to.
```
atrol: add `@return void`

atrol: Wrong comment `The user ID to link the new tag to`

core/tag_api.php
```diff
@@ -529,6 +529,25 @@ function tag_get_field( $p_tag_id, $p_field_name ) {
 	}
 }
 
+/**
+ * Can the specified user create a tag?
+ *
+ * @param integer $p_user_id     The user ID to link the new tag to.
```
atrol: Wrong comment `The user ID to link the new tag to`

vboctor: Fixed.

core/tag_api.php
vboctor: Fixed.

vboctor: Fixed.

core/commands/IssueAddCommand.php
```diff
@@ -217,6 +217,22 @@ function validate() {
...
+				}
+			}
+
+			if( $t_new_tag != '' ) {
+				throw new ClientException(
+					sprintf( "User '%d' can't create tag '%s'.", $this->user_id, $t_new_tag ),
+					ERROR_TAG_NOT_FOUND );
+			}
+		}
+
```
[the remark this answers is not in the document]
vboctor: Done.  The old logic was remaining from an alternative solution that I was implementing.

PR #1392, opened 2018-09-30
Fix view state references

vboctor:
At the moment we have two configs that related to view state enum:

- `view_state` - which is used for issue and issue notes view state.
- `project_view_state` - which is used for projects.

This commit fixes cases where the wrong view state was used for the scenario at hand.
In the future, we may want to consider have a different view state enum for issues vs. issue notes.

Fixes #10411

### DIFF
--- a/bug_actiongroup_page.php
+++ b/bug_actiongroup_page.php
@@ -409,7 +409,7 @@ if( $t_multiple_projects ) {
 						<label class="lbl padding-6"><?php echo lang_get( 'private' ); ?></label>
 <?php
 			} else {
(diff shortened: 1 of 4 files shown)


dregad:
It's not really obvious what is being fixed and why. A more descriptive commit message would be useful.

vboctor:
@dregad I have added more details in the commit message.  Also the referenced issue provides further context.

atrol:
Closed via https://github.com/mantisbt/mantisbt/commit/25eec2cfba931249bce5952e5fabb49abe413dc7

PR #1393, opened 2018-10-01
Update strings_chinese_traditional.txt

flamerecca:
$s_timeline_no_activity should be "在時間範圍內沒有任何使用動作。"
instead of "在時間範圍內沒有任合使用動作。"


atrol:
Translations are managed via TranslateWiki. Please refer our own [Wiki](http://www.mantisbt.org/wiki/doku.php/mantisbt:translationshowto) for the process to submit new translations or revise existing ones.

PR #1396, opened 2018-10-08
Add EVENT_BUG_FILE_ATTACH

brlumen:
This patch allows plugins to track file attachment events.

### DIFF
--- a/bug_update.php
+++ b/bug_update.php
@@ -1,6 +1,5 @@
 <?php
 # MantisBT - A PHP based bugtracking system
-# 
 # MantisBT is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 2 of the License, or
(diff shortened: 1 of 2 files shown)


core/events_inc.php
```diff
@@ -135,6 +135,7 @@
 	'EVENT_BUGNOTE_DELETED' => EVENT_TYPE_EXECUTE,
 	'EVENT_TAG_ATTACHED' => EVENT_TYPE_EXECUTE,
 	'EVENT_TAG_DETACHED' => EVENT_TYPE_EXECUTE,
+	'EVENT_BUG_FILE_ATTACH' => EVENT_TYPE_EXECUTE,
```
vboctor: 1. Rename to `EVENT_FILE_ADDED` 
2. Add a comment `# File events` above it.
3. While you are at it, please also add `# Tag Events` about the tag attached/detached.

core/file_api.php
```diff
@@ -867,6 +869,9 @@ function file_add( $p_bug_id, array $p_file, $p_table = 'bug', $p_title = '', $p
 
 		# log file added to bug history
 		history_log_event_special( $p_bug_id, FILE_ADDED, $t_file_name );
+		
+		# Event integration
+		event_signal( 'EVENT_BUG_FILE_ATTACH', array( $p_bug_id, $t_file_id ) );
```
vboctor: In the case of a user adding multiple files in one go, we should ideally trigger this event once with an array of file ids.  This enables for example a plugin to trigger a single email notification with reference to N files added without too much work.  So the contract would be an issue id + array of file ids.

I have also been trying to move to an associative array approach over an array of parameters for new events.  This would look as follows:

```
event_signal( 'EVENT_FILE_ADD', array( array( 'issue_id' => $p_bug_id, 'file_ids' => array( $t_file_id, ... ) ) );
```

core/events_inc.php
brlumen: At the expense of "Tag Events" did not immediately understand. Now add!

core/events_inc.php
```diff
@@ -133,8 +133,14 @@
 	'EVENT_BUGNOTE_EDIT_FORM' => EVENT_TYPE_EXECUTE,
 	'EVENT_BUGNOTE_EDIT' => EVENT_TYPE_EXECUTE,
 	'EVENT_BUGNOTE_DELETED' => EVENT_TYPE_EXECUTE,
+	
+	# Tag Events
```
brlumen: Made changes on request @vboctor

core/events_inc.php
brlumen: I also renamed according to the proposed format.

core/file_api.php
vboctor: That is why we are moving to commands.  The commands attempt to do all validation ahead of time, then trigger the actions and events as appropriate.  This also enable the events to be logical rather than mechanical as part of some low level APIs.

Hence, I would add these events into commands like:

- `IssueAddCommand` - currently we trigger `EVENT_REPORT_BUG` after adding an issue and all its files.  In such case, file added is not super valuable since it can be retrieved given the issue id.
- `IssueFileAddCommand` - This supports adding N files and can trigger a single logical event.
- `IssueFileDeleteCommand` - This supports deleting a single file and can trigger a single event.  However, you can imagine support N files in the future, so having the event support N files would be good future proofing.

As for plugins or core cases where commands are not used, the fix is to move to using existing commands and add new ones as needed.  The goal would be to trigger a single logical command that gets the job done and not N commands, otherwise we end up with a bunch of uncoordinated events.

vboctor:
@brlumen Please also add documentation for the new events to the developer manual.

atrol:
@brlumen you are messing up this PR as you use the same branch for your other PR https://github.com/mantisbt/mantisbt/pull/1398

You should not use `master` branch for PRs, but create a separate branch for each PR.

brlumen:
@atrol @vboctor you are right. I realized that I had  messing my PR.
Please close my PR.
I will form two new PR from different branches.

PR #1397, opened 2018-10-10
Adding plugin filters management on search.php

elpoutro:
Before modification, you cannot use filters created through plugins (using MantisFilter class and EVENT_FILTER_FIELDS hook) in "Permalink" functionality.
Now, the data about filters mapped by EVENT_FILTER_FIELDS hook are managed by search.php witch is used to analyse permanent links.

### DIFF
--- a/search.php
+++ b/search.php
@@ -142,6 +142,12 @@ foreach( $_GET as $t_var_name => $t_var_value ) {
 
 $t_my_filter['custom_fields'] = $t_custom_fields;
 
+# Handle class-based filters defined as plugins
+$t_plugin_filters = filter_get_plugin_filters();
+foreach( array_keys( $t_plugin_filters ) as $s_filter_name ) {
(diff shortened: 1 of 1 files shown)


elpoutro:
I request this because I could not use the https://github.com/mantisbt-plugins/FilterBugList filters with "Permalink". I think that every plugin filters are impacted by this issue and my modification may resolve it in any cases

cproensa:
@elpoutro 
is there a bug open for this yet? If not please create one and reference the id in the commit message as `Fixes: #nnnnn`

On the other hand, the fix is not right. The parameter type is not always a string.
Using this with plugin filters that have other types causes an error.

You can see how the parsing is done for general filters [in filter_api::filter_gpc_get()](https://github.com/mantisbt/mantisbt/blob/release-2.18.0/core/filter_api.php#L3323)
Replicating the code from there should put this in the right path.

Thanks!

elpoutro:
OK: I close this PR and I open this one: https://github.com/mantisbt/mantisbt/pull/1423

PR #1398, opened 2018-10-16
Implement IssueUpdateCommand

brlumen:
For my plugin, I created a function conceptually similar to the Commands logic.

Accordingly, since I have not so much experience, I ask the community for help in moderation and comments.

Currently implemented the IssueUpdateCommand and bug_update.php class

PR #1400, opened 2018-10-17
Implement IssueUpdateCommand

brlumen:
While working on the update issue for my plugin, I found that it was similar to the Command interface available in mantisbt.
I hope that my work will be useful, because at the moment this team is not in the project.
Perhaps something else needs to be added or changed in the project code, please indicate in which direction to work in order for this PR to be adopted.

Replaces #1398

### DIFF
--- a/bug_update.php
+++ b/bug_update.php
@@ -74,7 +74,7 @@ $t_issue = array(
                           'last_updated' => gpc_get_string( 'last_updated' ),
 );
 
-$t_additional_info = gpc_get_string( 'additional_info', null );
+$t_additional_info = gpc_get_string( 'additional_information', '' );
 if( !is_blank( $t_additional_info ) ) {
     $t_issue['additional_information'] = $t_additional_info;
 }
@@ -218,12 +218,11 @@ foreach( $t_related_custom_field_ids as $t_id ) {
     $t_def = custom_field_get_definition( $t_id );
 
     # Produce an error if the field is required but wasn't posted
-    if( gpc_isset_custom_field( $t_id, $t_def['type'] ) ) {
-        $t_custom_fields[] = array(
-                                  'field' => array( 'id' => $t_id ),
-                                  'value' => gpc_get_custom_field( 'custom_field_' . $t_id, $t_def['type'], null )
-        );
-    }
+
+    $t_custom_fields[] = array(
+                              'field' => array( 'id' => $t_id ),
+                              'value' => gpc_get_custom_field( 'custom_field_' . $t_id, $t_def['type'], null )
+    );
 }
 
 if( !empty( $t_custom_fields ) ) {
--- a/bug_update.php
+++ b/bug_update.php
@@ -1,7 +1,7 @@
 <?php
 # MantisBT - A PHP based bugtracking system
 
-# # MantisBT is free software: you can redistribute it and/or modify
+# MantisBT is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 2 of the License, or
 # (at your option) any later version.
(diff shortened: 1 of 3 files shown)


bug_update.php
```diff
@@ -63,401 +63,186 @@
...
+$t_time_tracking = gpc_get_string( 'time_tracking', '' );
+if( !is_blank( $t_time_tracking ) ) {
+    $t_bugnote['time_tracking'] = $t_time_tracking;
 }
 
-# Add a duplicate relationship if requested.
-if( $t_updated_bug->duplicate_id != 0 ) {
-	relationship_upsert( $f_bug_id, $t_updated_bug->duplicate_id, BUG_DUPLICATE, /* email_for_source */ false );
+# Validate the custom fields before adding the bug.
+$t_related_custom_field_ids = custom_field_get_linked_ids( $t_project_id );
```
brlumen: The variable $t_project_id is not initialized.
I'll fix it.

bug_update.php
```diff
@@ -1,7 +1,7 @@
 <?php
 # MantisBT - A PHP based bugtracking system
 
-# MantisBT is free software: you can redistribute it and/or modify
+# # MantisBT is free software: you can redistribute it and/or modify
```
vboctor:
```suggestion
# MantisBT is free software: you can redistribute it and/or modify
```

core/commands/IssueUpdateCommand.php
```diff
@@ -0,0 +1,489 @@
...
+ * {
+ *   "query": {
+ *   },
+ *   "payload": {
+ *      "issue": ... see rest issue update documentation
+ *      "bugnote": ... see rest issue update documentation
+ *   },
+ *   "options: {
+ *      "action_type": BUG_UPDATE_TYPE_NORMAL,
+ *      "bug_id": 1234,
```
vboctor: Rename to `issue_id` and make this part of the `query`.  For the REST API that would wrap this command, the query would be what comes from the URL, payload from the body, and options for internals that shouldn't be exposed from a public API.  So `action_type` would make sense in options.

core/commands/IssueUpdateCommand.php
```diff
@@ -0,0 +1,489 @@
...
+use Mantis\Exceptions\ClientException;
+
+/**
+ * Sample:
+ * {
+ *   "query": {
+ *   },
+ *   "payload": {
+ *      "issue": ... see rest issue update documentation
+ *      "bugnote": ... see rest issue update documentation
```
vboctor: As we move to commands and REST API, we move to `issue` and `note`.

core/commands/IssueUpdateCommand.php
```diff
@@ -0,0 +1,489 @@
...
+
+        $f_update_type = $this->option( 'action_type', BUG_UPDATE_TYPE_NORMAL );
+        $f_bug_id      = $this->option( 'bug_id', 0 );
+
+        $t_update_data  = $this->payload( 'issue' );
+        $t_bugnote_data = $this->payload( 'bugnote' );
+
+        if( !isset( $t_update_data['last_updated'] ) || is_blank( $t_update_data['last_updated'] ) ) {
+            throw new ClientException(
+            'Last updated not specified', ERROR_EMPTY_FIELD, array( 'last updated' ) );
```
vboctor: Make one line or indent another tab.  Apply to other places you are throwing exceptions.

core/commands/IssueUpdateCommand.php
```diff
@@ -0,0 +1,489 @@
...
+        parent::__construct( $p_data );
+    }
+
+    /**
+     * Validate the data.
+     */
+    protected function validate() {
+        $this->user_id = auth_get_current_user_id();
+
+        $f_update_type = $this->option( 'action_type', BUG_UPDATE_TYPE_NORMAL );
```
vboctor: The code inside the function should use `$t_` and not `$f_` prefix, since in this case these variables are now initialized from the payload and not via `gpc_*`.

core/commands/IssueUpdateCommand.php
```diff
@@ -0,0 +1,489 @@
...
+            email_owner_changed( $this->updated_issue->id, $this->existing_issue->handler_id, $this->updated_issue->handler_id );
+        } else if( $this->existing_issue->status != $this->updated_issue->status ) {
+            $t_new_status_label = MantisEnum::getLabel( config_get( 'status_enum_string' ), $this->updated_issue->status );
+            $t_new_status_label = str_replace( ' ', '_', $t_new_status_label );
+            email_bug_status_changed( $this->updated_issue->id, $t_new_status_label );
+        } else {
+            email_bug_updated( $this->updated_issue->id );
+        }
+
+        return array( TRUE );
```
vboctor: The command's returns should depend on an option, e.g. `return_issue`.

- Web UI will set it to false to get an empty array.
- REST API would set it to true to get the update issue details.

See `rest_issue_update` method for the expected input and output format for the REST API.  This command should be updated to mirror what REST API needs and be used from there.

brlumen:
@vboctor 
Is it possible to output change operations of notes into a separate command (IssueNoteUpdateCommand)?
And is it possible to call a command from a command?

brlumen:
@vboctor 
Is it possible to output change operations of notes into a separate command (IssueNoteUpdateCommand)?
And is it possible to call a command from a command?

brlumen:
@vboctor 
Is it possible to output change operations of notes into a separate command (IssueNoteUpdateCommand)?
And is it possible to call a command from a command?

brlumen:
@vboctor 
Is it possible to output change operations of notes into a separate command (IssueNoteUpdateCommand)?
And is it possible to call a command from a command?

vboctor:
@brlumen Commands shouldn't call each other.  They should share code through other means, e.g. via calling common APIs, having a base class, etc.  This allows a command to have one semantic operation that can be exposed via an API and trigger the appropriate logical extensibility events.

bug_update.php
```diff
@@ -63,401 +63,187 @@
...
-	bugnote_process_mentions( $f_bug_id, $t_bugnote_id, $t_bug_note->note );
+$t_time_tracking = gpc_get_string( 'time_tracking', '' );
+if( !is_blank( $t_time_tracking ) ) {
+    $t_bugnote['time_tracking'] = $t_time_tracking;
 }
 
-# Add a duplicate relationship if requested.
-if( $t_updated_bug->duplicate_id != 0 ) {
-	relationship_upsert( $f_bug_id, $t_updated_bug->duplicate_id, BUG_DUPLICATE, /* email_for_source */ false );
+# Validate the custom fields before adding the bug.
```
brlumen: @vboctor 
I borrowed the custom fields from bug_report.php.
How can I get all the "custom fields" using the gpc_get_custom_field function?

core/commands/IssueUpdateCommand.php
brlumen: Fixed on offer.

core/commands/IssueUpdateCommand.php
brlumen: Fixed on offer.

core/commands/IssueUpdateCommand.php
brlumen: Fixed on offer.

bug_update.php
brlumen: Fixed code formatting.

core/commands/IssueUpdateCommand.php
brlumen: Fixed code formatting.

core/commands/IssueUpdateCommand.php
brlumen: Implemented.

brlumen:
@vboctor 
I made changes to the IssueUpdateCommand and bug_update.php considering your suggestions and error message.
I also made changes to bug_actiongroup.php.

core/commands/IssueUpdateCommand.php
```diff
@@ -0,0 +1,621 @@
...
+	protected function validate() {
+		$this->user_id = auth_get_current_user_id();
+
+		$t_update_type	 = $this->option( 'action_type', BUG_UPDATE_TYPE_NORMAL );
+		$t_issue_id	 = helper_parse_issue_id( $this->query( 'issue_id' ) );
+
+		$t_update_data	 = $this->payload( 'issue' );
+		$t_notes	 = $this->payload( 'notes' );
+
+		if( !isset( $t_update_data['last_updated'] ) || is_blank( $t_update_data['last_updated'] ) ) {
```
vboctor: In the rest api, this field is called `updated_at`.   Update accordingly everywhere.  Make sure the same is the case for notes.

core/commands/IssueUpdateCommand.php
```diff
@@ -0,0 +1,621 @@
...
+			throw new ClientException(
+				'Last updated not specified', 
+				ERROR_EMPTY_FIELD, 
+				array( 'last updated' ) );
```
vboctor: Match the field name here.

core/commands/IssueUpdateCommand.php
```diff
@@ -0,0 +1,621 @@
...
+	 * Validate the data.
+	 */
+	protected function validate() {
+		$this->user_id = auth_get_current_user_id();
+
+		$t_update_type	 = $this->option( 'action_type', BUG_UPDATE_TYPE_NORMAL );
+		$t_issue_id	 = helper_parse_issue_id( $this->query( 'issue_id' ) );
+
+		$t_update_data	 = $this->payload( 'issue' );
+		$t_notes	 = $this->payload( 'notes' );
```
vboctor: In the existing implementation of issue update in REST API, I believe we support providing notes inside the issue.  If they are new they are added.  If they have an id, then they are added.  We should be backward compatible with that.

core/commands/IssueUpdateCommand.php
```diff
@@ -0,0 +1,621 @@
...
+			foreach( $t_bugnotes as $t_bugnote ) {
+				$this->existing_notes[$t_bugnote->id] = $t_bugnote;
+			}
+
+			$t_reassign_on_feedback_trigger = false;
+
+			foreach( $t_notes as $t_note ) {
+				$t_note = ApiObjectFactory::objectToArray( $t_note );
+
+				if( isset( $t_note['view_state'] ) ) {
```
vboctor: Default view state is not specified should only be done for new notes, for existing notes, they view_state should remain as is.

core/commands/IssueUpdateCommand.php
vboctor: In REST API, we use etags rather than a `updated_at`.  See the current implementation in `issues_rest.php`.  Hence, this field shouldn't be allowed in the payload.  If needed for internal calls, it should be done via options array.

vboctor:
If this change fixes some bugs compared to old implementation, then we should find/add specific issues in the bug tracker.  For example, a change wasn't triggering an email or something similar.

vboctor:
Please also make sure that if fields like `created_at` or `updated_at` are in the payload, they are ignored.  Since these shouldn't be updated by the API caller.

dregad:
@brlumen are you planning to address Victor's review comments ? If so, please rebase the branch to fix the merge conflicts; otherwise we'll close the PR.

PR #1401, opened 2018-10-21
Fix IssueNoteAddCommand re-assign on feedback

vboctor:
The issue was caused by logged in user id being returned as a string.

Fixes #24877, #23712


vboctor:
I have pushed an updated commit to the source branch but for some reason Github PR didn't pick it up.
https://github.com/vboctor/mantisbt/commits/issue24877_fix_user_id_type

vboctor:
https://github.com/vboctor/mantisbt/commits/issue24877_fix_user_id_type

vboctor:
The PR is now caught up with the commits that I did during the Github outage.

PR #1402, opened 2018-10-25
Enable password managers detect usernames

cproensa:
Make the username field a locked non-visible field, instead of type
"hidden" input. This way the password managers present in browsers can
detect the different username values and remember them separately.
The problem is that inputs with type="hidden" are ignored by these
agents.

Fixes: #24896

### DIFF
--- a/login_password_page.php
+++ b/login_password_page.php
@@ -213,13 +213,12 @@ if( config_get_global( 'admin_checks' ) == ON && file_exists( dirname( __FILE__ 
 				echo '<input type="hidden" name="install" value="true" />';
 			}
 
-			echo '<input type="hidden" name="username" value="', string_html_specialchars( $t_username ), '" />';
 
 			echo sprintf( lang_get( 'enter_password' ), string_html_specialchars( $t_username ) );
 
 			# CSRF protection not required here - form does not result in modifications
 			?>
-
+			<input hidden readonly type="text" name="username" class="hidden" tabindex="-1" value="<?php echo string_html_specialchars( $t_username ) ?>" id="hidden_username" />
 			<label for="password" class="block clearfix">
 				<span class="block input-icon input-icon-right">
 					<input id="password" name="password" type="password" placeholder="<?php echo lang_get( 'password' ) ?>"


login_password_page.php
```diff
@@ -213,13 +213,12 @@
...
 			}
 
-			echo '<input type="hidden" name="username" value="', string_html_specialchars( $t_username ), '" />';
 
 			echo sprintf( lang_get( 'enter_password' ), string_html_specialchars( $t_username ) );
 
 			# CSRF protection not required here - form does not result in modifications
 			?>
-
+			<input hidden readonly type="text" name="username" class="hidden" tabindex="-1" value="<?php echo string_html_specialchars( $t_username ) ?>" id="hidden_username">
```
vboctor: Interesting that you have "hidden" still.  Should the tag be terminated by `/>`.

What password manager(s) did you test with?

vboctor:
Did you test this with the re-authentication flow?

login_password_page.php
```diff
@@ -213,13 +213,12 @@
...
 			}
 
-			echo '<input type="hidden" name="username" value="', string_html_specialchars( $t_username ), '" />';
 
 			echo sprintf( lang_get( 'enter_password' ), string_html_specialchars( $t_username ) );
 
 			# CSRF protection not required here - form does not result in modifications
 			?>
-
+			<input hidden readonly type="text" name="username" class="hidden" tabindex="-1" value="<?php echo string_html_specialchars( $t_username ) ?>" id="hidden_username" />
```
dregad: Is it not redundant to use both _hidden_ attribute and class ? Do we even need the class at all ?

cproensa: > Is it not redundant to use both hidden attribute and class ?

I'm not sure. I'd say the css style is provided as a fallback in case the property is not supported.
I'm open to change it if you think otherwise.

dregad: According to https://caniuse.com/#feat=hidden, the _hidden_ attribute is not supported under IE10, and that's still on our [official client requirements list](http://mantisbt.org/docs/master/en-US/Admin_Guide/html-desktop/#admin.install.requirements.client), so let's go for belt & braces then :wink:

PR #1404, opened 2018-11-02
Fix password checking when creating users

atrol:
Fixes #24925


manage_user_create.php
```diff
@@ -68,9 +68,9 @@
 
 # Password won't to be sent by email. It entered by the admin
 # Now, if the password is empty, confirm that that is what we wanted
-if( is_blank( $f_password ) &&
-	( ON != config_get( 'send_reset_password' ) ) ||
-	( ON != config_get( 'enable_email_notification' ) ) ) {
+if( is_blank( $f_password ) && (
+	ON != config_get( 'send_reset_password' ) ||
+	ON != config_get( 'enable_email_notification' ) ) ) {
```
dregad: I find the code more easy to read with the if's closing parenthesis and curly bracket on a new line to provide a clear separation between test condition and code, like this:

```
if( cond1 ||
    cond2
) {
    ...
```

manage_user_create.php
```diff
@@ -68,9 +68,9 @@
 
 # Password won't to be sent by email. It entered by the admin
```
dregad: I would suggest to take the opportunity to rephrase this in proper English
`Password won't be sent by email. It is entered by the admin`

PR #1405, opened 2018-11-04
Fix wrong display of Manage button

atrol:
Do no longer offer the "Manage" button if user does not have
manage_project_threshold rights for the currently selected project.

Fixes #24932


cproensa:
We have more issues here, for users that could manage tags, profiles, and arbitrary plugins that publish manage links there.

Should it be cleaner to refactor the print_manage_menu() logic to be used to determine if the main manage menu is enabled or not?
Also there is a level of complexity in choosing what is the landing page for the sidebar item, we could simplify that:
- by going always to manage overview
- by choosing the first one returned by that print_manage_menu() logic
- or the first one available in a presorted list, for example, to favour project over user

PR #1409, opened 2018-11-14
Fix sidebar collapse regression

atrol:
Regression introduced in version 2.13.0 commit b7b914549fe8fbdf4ee4053f475031df9d6a4674

Fixes [#24976](https://mantisbt.org/bugs/view.php?id=24976)

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -946,7 +946,7 @@ function layout_sidebar_end() {
 
 	$t_collapse_block = is_collapsed( 'sidebar' );
 
-	echo '<div id="sidebar" class="sidebar-toggle sidebar-collapse">';
+	echo '<div id="sidebar-btn" class="sidebar-toggle sidebar-collapse">';
 	if( layout_is_rtl() ) {
 		$t_block_icon = $t_collapse_block ? 'fa-angle-double-left' : 'fa-angle-double-right';
 		echo '<i data-icon2="ace-icon fa fa-angle-double-left" data-icon1="ace-icon fa fa-angle-double-right"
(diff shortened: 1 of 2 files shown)


atrol:
I could also remove the id in ` layout_sidebar_begin()`, or @dregad do you prefer introducing a new id, something like `sidebar-begin` ?

dregad:
It might work too, I didn't test. I'm fine either way (as long as it works :wink:) but it may indeed be better not to have an ID if it's not being used.

atrol:
> It might work too

Removing the ID works at first sight, updated the PR.
Will do some more tests on different browsers.

atrol:
Did some more tests.
The latest change (remove the ID in `layout_sidebar_begin()` seems to work quite good at first sight, but introduces another regression.
The burger menu that is used if horizontal size is small, does not work any longer.
As this is not acceptable (especially on mobile devices in portrait mode), I will change back to my first change.

Of course,  it will reintroduce the duplicate HTML element ID, but it seems to work (tested on various desktop browsers and various window sizes, no mobile tests).

Sorry @dregad I can't invest time for a clean solution.

dregad:
The problem with duplicated ID is that it is impossible to reliably predict which element will be selected e.g. when using `#sidebar` in CSS. So while this may work today, there is no guarantee that it does in all browsers or will do in the future. 

The erratic behavior you noticed during testing is a direct consequence of that - some parts of the code expect to reference the "button" div (e.g. common.js), while others are targetting the actual sidebar div.

The bug was introduced in 8a3eedcc6c2b4b9011815a097e3529807148564b (when @syncguru added sidebar collapse state persistence), and is a mix of referencing the wrong element (since they share the same ID) and inconsistent logic to retrieve the sidebar div's name to store it in the cookie.

I'll push a commit with what I believe is the proper fix shortly. Let me know if

PR #1418, opened 2018-12-08
Honor $g_bug_change_status_page_fields for form fields on the change …

cmfitch1:
…status page

Issue #14841

### DIFF
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -70,6 +70,7 @@ $t_mantis_dir = dirname( __FILE__ ) . DIRECTORY_SEPARATOR;
 $t_show_page_header = false;
 $t_force_readonly = true;
 $t_fields_config_option = 'bug_change_status_page_fields';
+$t_fields = config_get( $t_fields_config_option, null, null, $t_bug->project_id );
 
 if( $t_bug->project_id != helper_get_current_project() ) {
 	# in case the current project is not the same project of the bug we are viewing...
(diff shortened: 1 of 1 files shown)


atrol:
According documentation `$g_bug_change_status_page_fields` is
```
An array of optional fields to show on the bug change status page.
```

I would expect that it works similar to `$g_bug_update_page_fields` and `bug_update_page.php`.

Your PR allows just to exclude the few existing fields, but does not allow to include all the other fields that can be set in `$g_bug_change_status_page_fields`.

cmfitch1:
Based on this suggestion, wouldn't `bug_change_status_page.php` just be another version of the `bug_update_page.php` with a different set of default fields displayed? I'm guessing that was not the intent of the status update page.

I would compare `$g_bug_change_status_page_fields` to `$g_bug_view_page_fields` instead, as the page is already passing this value to `bug_view_inc.php` through `$t_fields_config_option`. It seems that it was simply an oversight that the handful of form fields needed to edit the bug status were not also honoring this option.

cmfitch1:
After a bit more thought, I'll go back a bit on my previous comment. I can see some value in being able to specify additional form fields to show when changing status. This did not seem to be the request of issue 14841, however. Based on issue 14841's steps to reproduce and several of the comments, it seemed to be centered around not being able to remove an existing form field using the config option (I had this same problem). This PR was a simple change to address that specific problem, and hopefully doesn't need to be held up in the interest of a longer term redesign of the page and its config option.

PR #1419, opened 2018-12-09
Add some more information to view_user_page

atrol:
Fixes #25042


dregad:
I added a minor cosmetic fix for [#25059](https://mantisbt.org/bugs/view.php?id=25059), which is not directly related to your change (issue was already present before), but I think it makes sense to include it as part of this PR.

PR #1421, opened 2018-12-09
Add config option for filtering on disabled users

cmfitch1:
Issue #10141

### DIFF
--- a/core/project_api.php
+++ b/core/project_api.php
@@ -612,7 +612,7 @@ function project_get_local_user_rows( $p_project_id ) {
  * @param boolean $p_include_global_users Whether to include global users.
  * @return array List of users, array key is user ID
  */
-function project_get_all_user_rows( $p_project_id = ALL_PROJECTS, $p_access_level = ANYBODY, $p_include_global_users = true ) {
+function project_get_all_user_rows( $p_project_id = ALL_PROJECTS, $p_access_level = ANYBODY, $p_include_global_users = true, $p_include_disabled_users = false ) {
 	$c_project_id = (int)$p_project_id;
 
 	# Optimization when access_level is NOBODY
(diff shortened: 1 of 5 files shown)


cproensa:
@vboctor looks good to me
but in my opinion:
- Default should be OFF. In the issue it's suggested to be on, but after long time, the current standard is off
- The proper solution is to integrate this functionality into a better UI element for user selector. At that point this should be part of the UI and not a config option. That's why i'd like this option to be a "provisional" config. This is, once we get a better UI, just remove the option without caring about backwards compatibility, etc. Maybe place a comment about this in the documentation?

Chris says in the issue:
> There are reasons to want to filter on disabled users, but I don't see the use case for wanting to assign, etc. to disabled users, nor does this bug ask for that capability.

I agree with that.

Additionally, having the parameter for disabled users in the core functions is suitable for an hypothetical rewrite of the UI, that will eventually use the same calls as this PR

core/project_api.php
```diff
@@ -679,9 +679,14 @@ function project_get_all_user_rows( $p_project_id = ALL_PROJECTS, $p_access_leve
...
 		$t_query = 'SELECT id, username, realname, access_level
 				FROM {user}
-				WHERE enabled = ' . db_param() . '
-					AND access_level ' . $t_global_access_clause;
-		$t_result = db_query( $t_query, array( $t_on ) );
+				WHERE access_level ' . $t_global_access_clause;
+
+		if ( !$p_include_disabled_users ) {
+			$t_query .= ' AND enabled = ' . db_param();
+			$t_result = db_query( $t_query, Array( $t_on ) );
```
cproensa: This query may have, or not, parameters based onthe value of $p_include_disabled_users.
For correctness, when the query does not have parameters, the parameter array is not used in db_query(), but it's also important to **not** issue db_param_push(). Doing so has historically causes problems in postgres... iirc.

I think that after the latest db_api rewrite, that's is no longer an issue, as db_query() executes a wrapper for a DbQuery object. BUT... in code like this, i usually rewrite the query in the new syntax, just in case...

PR #1422, opened 2018-12-10
Fix Generic Error with Anonymous login 

dregad:
Fixes [#25061](https://mantisbt.org/bugs/view.php?id=25061)

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -176,7 +176,7 @@ function auth_signup_access_level() {
  * @return bool true: enabled; false: otherwise.
  */
 function auth_anonymous_enabled() {
-	return config_get_global( 'allow_anonymous_login' );
+	return config_get_global( 'allow_anonymous_login' ) && auth_anonymous_account();
 }
 
 /**
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -82,7 +82,7 @@ $g_cache_current_user_id = NO_USER;
  * @return AuthFlags The auth flags object to use.
  */
 function auth_flags( $p_user_id = null, $p_username = '' ) {
-	if( is_null( $p_user_id ) ) {
+	if( !$p_user_id ) {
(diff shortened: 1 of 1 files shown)


core/authentication_api.php
```diff
@@ -176,7 +176,7 @@ function auth_signup_access_level() {
  * @return bool true: enabled; false: otherwise.
  */
 function auth_anonymous_enabled() {
-	return config_get_global( 'allow_anonymous_login' );
+	return config_get_global( 'allow_anonymous_login' ) && auth_anonymous_account();
```
vboctor: That change seems to change the semantics of what this function does.  It is supposed to tell if anonymous login is enabled and not whether the current account is the anonymous account.

atrol: > and not whether the current account is the anonymous account.

`auth_anonymous_account` does not check if the current account is the anonymous account.

```php
/**
 * Get the anonymous account username.
 * @return string Anonymous account username.
 */
function auth_anonymous_account() {
	return config_get_global( 'anonymous_account' );
}
```

dregad: @vboctor This is definitely not a change in semantics in my opinion. 

Anonymous login is *de facto* not operational (i.e. Mantis triggers a generic error) if *allow_anonymous_login = ON* but the anonymous account is not defined.

This is clearly documented http://mantisbt.org/docs/master/en-US/Admin_Guide/html-desktop/#admin.config.misc
> **$g_allow_anonymous_login**
    Enable anonymous access to Mantis. **You must also specify $g_anonymous_account** as the account which anonymous users will browse Mantis with. The default setting is OFF.

dregad: @vboctor are you satisfied with my explanation, or do you still think some changes are needed ? If so, let me know what you would like me to do.

vboctor: @dregad OK.

PR #1424, opened 2018-12-20
Fix cookie settings to use plugin AuthFlags

vboctor:
Fixes #25099

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -189,11 +189,12 @@ function auth_anonymous_account() {
 
 /**
  * Get the auth cookie expiry time.
+ * @param integer $p_user_id The user id to get session expiry for.
  * @param boolean $p_perm_login Use permanent login.
  * @return integer cookie lifetime or 0 for browser session.
  */
-function auth_session_expiry( $p_perm_login ) {
-	$t_auth_flags = auth_flags();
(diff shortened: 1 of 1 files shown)


core/authentication_api.php
```diff
@@ -190,10 +190,11 @@ function auth_anonymous_account() {
 /**
  * Get the auth cookie expiry time.
  * @param boolean $p_perm_login Use permanent login.
+ * @param int $p_user_id The user id to get session expiry for.
```
dregad: add _defaults to current user_

vboctor: @dregad This method is only called from one place what I changed.  I defaulted it to null just in case it is being called.  The null value means user not specified and hence will use default AuthFlags since there is no authenticated / known user at this time (this flow happens before auth cookie is set).

I can remove it or clarify in the comment.

dregad: > I can [...] clarify in the comment.

That's what I meant. Thanks

atrol: Concerning `@param` `int` vs. `integer`.
We don't have clear rules for that, but just one line below and most of the time`integer` is used.

vboctor:
I removed the defaulting to null to avoid confusion and because it is not needed.  I have also applied @atrol's comment about the integer vs. int, and I also swapped the order of the parameters.

PR #1433, opened 2019-01-01
Logic for users with only one accessible project

cproensa:
commit 6c20f7e25f03e8c8b9ca426d4fc83d0929d126a9
Rewrite hide project list and default project check
    
    Previously, when a user only has one project, some things happens:
    - The project selector was hidden, as it is not useful without more
      projects.
    - A check is performed where if current project is ALL_PROJECTS, then
      current and default project are updated to the only available project.
    
    To fix issues with that implementation, the logic has been separated:
    - Helper function to check if the user has more than one project. It's
      used in layout_api to decide if the project selector must be shown.
    - Helper function to check if the user has ALL_PROJECTS as
      current/default project when only one is available. In that case,
      update current and modify default in user preferences.
    
    Fixes: #25110, #25133

commit 9b21b09f32da11332ddca0c9e3c15ccfe5f059d6
Don't apply single project logic for some users
    
    Eventhough the project selector may be hidden when there are not
    multiple projects to choose from, it is still needed for editing manage
    and configuration pages that apply to either specific project or
    ALL_PROJECTS.
    Here we add a condition that if the user has access to manage pages, the
    project selector is never hidden, and both the specific project and
    ALL_PROJECTS are available to be used.
    
    Fixes: #20054

### DIFF
--- a/core/current_user_api.php
+++ b/core/current_user_api.php
@@ -272,14 +272,20 @@ function current_user_has_more_than_one_project() {
  * - If default project is not the visible one, modify it to be that project,
  *   or ALL_PROJECTS if the user has no accesible projects.
  *
+ * These changes only apply to user who can't use the project selection.
+ *
  * @return void
  */
 function current_user_modify_single_project_default() {
+	# The user must not be able to use the project selector
+	if( layout_navbar_can_show_projects_menu() ) {
+		return;
+	}
+	# The user must have one, or none, projects
 	$t_user_id = auth_get_current_user_id();
 	if( user_has_more_than_one_project( $t_user_id ) ) {
 		return;
 	}
-	# the user has one, or none, projects
 	$t_default = user_pref_get_pref( $t_user_id, 'default_project' );
 	$t_current = helper_get_current_project();
 	$t_projects = user_get_all_accessible_projects( $t_user_id );
--- a/core/current_user_api.php
+++ b/core/current_user_api.php
@@ -301,6 +301,6 @@ function current_user_modify_single_project_default() {
 		helper_set_current_project( $t_project_id );
 	}
 	if( $t_project_id != $t_default ) {
-		user_set_default_project( $t_user_id, $t_project_id );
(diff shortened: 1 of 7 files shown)


core/current_user_api.php
```diff
@@ -253,3 +253,54 @@ function current_user_get_bug_filter( $p_project_id = null ) {
...
+		$t_project_id = ALL_PROJECTS;
+	} else {
+		$t_project_id = reset( $t_projects );
+	}
+
+	if( $t_project_id != $t_current ) {
+		helper_set_current_project( $t_project_id );
+	}
+	if( $t_project_id != $t_default ) {
+		user_set_default_project( $t_user_id, $t_project_id );
```
atrol: This throws `APPLICATION ERROR #12` when trying with a `protected` (typically anonymous) user.

core/layout_api.php
```diff
@@ -1262,3 +1215,54 @@ function layout_login_page_logo() {
...
+function layout_manage_menu_link() {
+	static $t_link = null;
+	if( access_has_global_level( config_get( 'manage_site_threshold' ) ) ) {
+		$t_link = 'manage_overview_page.php';
+	} else {
+		if( access_has_global_level( config_get( 'manage_user_threshold' ) ) ) {
+			$t_link = 'manage_user_page.php';
+		} else {
+			if( access_has_any_project_level( 'manage_project_threshold' ) ) {
+				if( $t_current_project == ALL_PROJECTS ) {
```
atrol: $t_current_project is not defined at this place

cproensa: fixed

cproensa:
> This throws APPLICATION ERROR #12 This account is protected.... when trying with a protected (typically anonymous) user.

simple things, that turn into a rabbit hole...
i raised the protected user checks to higher level functions and added a way to skip that validation. Also refactor other api calls so the low level database operations are not called directly outside of the user_pref_api

PR #1437, opened 2019-01-08
issue#25170: Fix security breach

realitix:
When the request contains the X-Forwarded-Host HTTP header,
in the returned HTML page, many ressources are loaded using
this Header, mainly the css one.
It means an attacker can inject it's own code.

The problem comes from two paths:

    in core/html_api.php:245 string_sanitize_url must be set to false
    in core/layout_api.php:94 same problem

You can find the description here: 
https://mantisbt.org/bugs/view.php?id=25170

Thanks,
Jean-Sébastien

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -245,7 +245,7 @@ function html_css_link( $p_filename ) {
 	if( $p_filename == basename( $p_filename ) ) {
 		$p_filename = 'css/' . $p_filename;
 	}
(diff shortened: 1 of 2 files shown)


realitix:
Travis errors are not linked to my change but to the travis plateform.

dregad:
This can't be merged as it would break loading CSS from locations outside of MantisBT root (e.g. admin or plugin pages). Follow-up discussion in [#25170](https://mantisbt.org/bugs/view.php?id=25170).

PR #1438, opened 2019-01-08
Export FLOAT custom fields as numbers to Excel

atrol:
Fixes #25174


atrol:
> but if you don't mind please split the long line

@dregad I changed, but I am not sure if this is what you expected concering tabs, spaces, braces and operators.

At least it looks better than before in my IDE (better than the preview) concerning readability.

PR #1443, opened 2019-01-10
REST API: Add web.config (url rewrite) for IIS

obmsch:
To enable access to MantisBT through REST API under IIS, add an
.htaccess equivalent web.config for the url rewrite rule.
Otherwise 404 is returned on every request.
Generated with IIS-URL-Rewrite.Import(api/rest/.htaccess).

Fixes #24347

### DIFF
--- a/api/rest/web.config
+++ b/api/rest/web.config
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+  <system.webServer>
+    <rewrite>
+      <rules>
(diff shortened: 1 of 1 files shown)


atrol:
@obmsch I assume you are using this on your own production environment.
If so, I am also fine with the change.

obmsch:
@atrol Although access trough the REST API is still experimental on my side, yes I use my
production environment (IIS 10). I simply copy that file to api/rest, after I update to a new
release of MantisBT (currently 2.19).

PR #1448, opened 2019-01-23
Fix 25370. Add hosting folder /webstat to .gitignore

Kirill:
Yes. I know that use git repository on hosting not right idea. But this simple way for testing dev version of MantisBT


Kirill:
Don't need 
use `.git/info/exclude` file

PR #1449, opened 2019-01-23
Improve project users manage page and Sortable tables.

cproensa:
Manage project page, for user access is a cumbersome and difficult to mantain UI
As a start, improve that table, with inline editing and staging of changes, table can be sorted, filtered, paginated, done with list.js library.
In this implementation, the form also works without javascript.

seleccion_323

seleccion_324

If this PR goes well, next steps can be to integrate the "add users" form, and implement similar feature in the user manage page, where projects access can be also managed.


dregad:
Thanks for this @cproensa and sorry for the delay. I'll try to test this over the week-end.

PR #1450, opened 2019-01-25
Deliver all versions when querying projects via SOAP or REST API

atrol:
Before this change, obsolete versions were not part of the response.

Fixes #25381

### DIFF
--- a/api/soap/mc_project_api.php
+++ b/api/soap/mc_project_api.php
@@ -834,7 +834,9 @@ function mci_custom_field_type_name( $p_type_id ) {
  * @return array The project versions.
  */
 function mci_project_versions( $p_project_id ) {
-	$t_versions = version_get_all_rows( $p_project_id, /* inherit */ null );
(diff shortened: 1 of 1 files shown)


dregad:
I didn't check if this internal function is actually used in any public API endpoints, but if it is, then this change would break BC for existing callers expecting to only receive active versions.

If not, then I have no objections with this change.

atrol:
> I didn't check if this internal function is actually used in any public API endpoints, but if it is, then this change would break BC for existing callers expecting to only receive active versions.

The changed `function mci_project_versions` is just used in `function mci_project_get` if parameter `$p_detail` is true.

`function mci_project_get` with `$p_detail = true` is just called by REST API functions to create / update / read (the case reported by user obmsch) projects.

So all in all I would say that the change does not affect BC, but solves even more issues than the reported one.

Didn't try, but I expect that before the change, creating/updating projects with obsolete versions via REST does not deliver the obsolete versions in response.

PR #1451, opened 2019-01-26
Administrator can change user's password

callmecabman:
There is a plenty of local MantisBT installations where managing passwords with email reset is simply not feasible.

### DIFF
--- a/lang/strings_english.txt
+++ b/lang/strings_english.txt
@@ -356,6 +356,7 @@ $s_change_preferences_link = 'Preferences';
 $s_edit_account_title = 'Edit Account';
 $s_username = 'Username';
 $s_username_label = 'Username';
+$s_password_hash_notice = 'Password hash is being displayed. Enter plaintext to change the password.';
 $s_realname = 'Real Name';
 $s_realname_label = 'Real Name';
 $s_email = 'E-mail';
--- a/lang/strings_english.txt
+++ b/lang/strings_english.txt
@@ -356,7 +356,6 @@ $s_change_preferences_link = 'Preferences';
 $s_edit_account_title = 'Edit Account';
 $s_username = 'Username';
 $s_username_label = 'Username';
-$s_password_hash_notice = 'Password hash is being displayed. Enter plaintext to change the password.';
 $s_realname = 'Real Name';
 $s_realname_label = 'Real Name';
 $s_email = 'E-mail';
--- a/manage_user_edit_page.php
+++ b/manage_user_edit_page.php
@@ -136,7 +136,7 @@ print_manage_menu( 'manage_user_page.php' );
 					<?php echo lang_get( 'password' ) ?>
 				</td>
 				<td>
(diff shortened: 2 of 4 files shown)


atrol:
Thanks for your contribution.

I had just a cursory look, so no detailed but some general comments at the moment.

- works just with default setting `$g_login_method = MD5;` but not any of the other available login methods.

- contains non English localization. Translations are managed via TranslateWiki. This can be done after the English version is committed to our master branch. Please refer our own [Wiki](http://www.mantisbt.org/wiki/doku.php/mantisbt:translationshowto) for the process to submit new translations.

- commit message should reference an issue at our issue tracker. There is already [#15658](https://www.mantisbt.org/bugs/view.php?id=15658), so the message should contain `Fixes #15658`

callmecabman:
I've added the conditional statement for login method. AFAIK MD5 is the only method which uses mantis database as a source of truth, so this feature is not relevant for other methods.

callmecabman:
@atrol 
Thank you for your review. All the points that you mentioned have been fixed.

atrol:
> AFAIK MD5 is the only method which uses mantis database as a source of truth

Not true.

I don't have enough time at the moment to provide a lot of feedback.

Please have a look at account_page.php and function auth_process_plain_password in authentication_api.php to get some hints how to find out if password can be changed, the right length of the password field in form, the right password string to store, ...

manage_user_update.php
```diff
@@ -186,6 +199,10 @@
 		$t_changes .= lang_get( 'username_label' ) . ' ' . $t_old_username . ' => ' . $f_username . "\n";
 	}
 
+	if( $t_password_changed ) {
+		$t_changes .= lang_get( 'password' ) . ' ' . $t_old_password . ' => ' . "[REDACTED]" . "\n";
```
vboctor: `[REDACTED]` is not localized, and we shouldn't show old password.  I would just have a localized message here that password was changed.

manage_user_edit_page.php
```diff
@@ -127,6 +127,24 @@
 				</td>
 			</tr>
 
+			<!-- Password -->
+			<?php
+				if ( MD5 == $g_login_method ) {
```
vboctor: We shouldn't have password type specific logic here.  Look at account_update.php and see how it changes the password via standard APIs that handles all hashing models via `user_set_password()`.

lang/strings_english.txt
```diff
@@ -356,6 +356,7 @@ $s_change_preferences_link = 'Preferences';
 $s_edit_account_title = 'Edit Account';
 $s_username = 'Username';
 $s_username_label = 'Username';
+$s_password_hash_notice = 'Password hash is being displayed. Enter plaintext to change the password.';
```
vboctor: Why do we want to display the password hash?  That is a security issue.

dregad:
@callmecabman are you planning to address / respond to the above review comments ?

PR #1442, opened 2019-01-09
Filter summary

cproensa:
Implement filter functionality on Summary pages (and MantisGraph). There are not many changes in UI, but the core is there for future improvements.

As a start, a "summary" button is added to the view issues page, that opens the summary for the current applied filter. Once in the summary, all filters in use are temporary, to allow following the view links without effects on current persistent filters for view issues page.

seleccion_315

Summary shows an indication that a filter is applied:

seleccion_316

Additionally
- Some fixes and adjustments for summary and graphs presentation
- Some queries has been rewritten to improve performance

Example: performance in a project with 55k issues:
Before:
```
Unique queries executed: 61
Total queries executed: 62
Total query execution time: 19.8188 seconds
```
After:
```
Unique queries executed: 41
Total queries executed: 42
Total query execution time: 6.8595 seconds
```

PR #1452, opened 2019-01-26
Summary layout improvements

dregad:
While testing #1442, I found a few minor display/layout issues with Summary and MantisGraph pages (see [25385](https://mantisbt.org/bugs/view.php?id=25385), [25386](https://mantisbt.org/bugs/view.php?id=25386) and [25387](https://mantisbt.org/bugs/view.php?id=25387) for details). 

I would propose to merge this PR at the same time as @cproensa's #1442, but they are effectively independent and older issues, so it can also be dealt with separately if you prefer.

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -585,15 +585,14 @@ function print_summary_submenu() {
 
 	if( count($t_menu_options) > 0 ) {
 		echo '<div class="space-10"></div>';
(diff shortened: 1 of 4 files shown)


dregad:
Since there's ongoing discussion in #1442, I'll merge this now.

PR #1453, opened 2019-01-30
Add/Update/Delete project subproject via REST API

obmsch:
- Create a command for adding a project subproject.
- Create a REST API to add a project subproject.

Fixes #25400


core/commands/SubprojectAddCommand.php
```diff
@@ -0,0 +1,137 @@
...
+			throw new ClientException(
+				'Access denied to add subprojects',
+				ERROR_ACCESS_DENIED );
+		}
+
+		if( !project_exists( $this->project_id )) {
+			throw new ClientException(
+			'Project not found',
+			ERROR_PROJECT_NOT_FOUND,
+			array( $this->project_id ) );
```
dregad: Fix indentation

obmsch:
Thanks @dregad for the review and sorry for pushing the update/delete stuff so late.

api/rest/restcore/projects_rest.php
```diff
@@ -150,6 +154,83 @@ function rest_project_subproject_add( \Slim\Http\Request $p_request, \Slim\Http\
...
+ * @return \Slim\Http\Response The augmented response.
+ */
+function rest_project_subproject_update( \Slim\Http\Request $p_request, \Slim\Http\Response $p_response, array $p_args ) {
+	$t_project_id = isset( $p_args['id'] ) ? $p_args['id'] : $p_request->getParam( 'id' );
+	if( is_blank( $t_project_id ) ) {
+		$t_message = "Project id is missing.";
+		return $p_response->withStatus( HTTP_STATUS_BAD_REQUEST, $t_message );
+	}
+
+	$t_subproject_id = isset( $p_args['subproject_id'] ) ? $p_args['subproject_id'] : $p_request->getParam( 'subproject_id' );
```
dregad: Break long line

api/rest/restcore/projects_rest.php
```diff
@@ -150,6 +154,83 @@ function rest_project_subproject_add( \Slim\Http\Request $p_request, \Slim\Http\
...
+	$t_command = new SubprojectUpdateCommand( $t_data );
+	$t_result = $t_command->execute();
+
+	$t_subproject_id = (int)$t_result['id'];
+
+	return $p_response->withStatus( HTTP_STATUS_NO_CONTENT, "Subproject ($t_subproject_id) updated" );
+}
+
+/**
+/**
```
dregad: Extra `/**`

api/rest/restcore/projects_rest.php
```diff
@@ -150,6 +154,83 @@ function rest_project_subproject_add( \Slim\Http\Request $p_request, \Slim\Http\
...
+			'subproject_id' => $t_subproject_id
+		)
+	);
+
+	$t_command = new SubprojectDeleteCommand( $t_data );
+	$t_result = $t_command->execute();
+
+	$t_subproject_id = (int)$t_result['id'];
+
+	return $p_response->withStatus( HTTP_STATUS_NO_CONTENT, "Subproject ($t_subproject_id) deleted from project ($t_project_id)" );
```
dregad: Break long line

api/rest/restcore/projects_rest.php
```diff
@@ -150,6 +154,83 @@ function rest_project_subproject_add( \Slim\Http\Request $p_request, \Slim\Http\
...
+ * @return \Slim\Http\Response The augmented response.
+ */
+function rest_project_subproject_delete( \Slim\Http\Request $p_request, \Slim\Http\Response $p_response, array $p_args ) {
+	$t_project_id = isset( $p_args['id'] ) ? $p_args['id'] : $p_request->getParam( 'id' );
+	if( is_blank( $t_project_id ) ) {
+		$t_message = "Project id is missing.";
+		return $p_response->withStatus( HTTP_STATUS_BAD_REQUEST, $t_message );
+	}
+
+	$t_subproject_id = isset( $p_args['subproject_id'] ) ? $p_args['subproject_id'] : $p_request->getParam( 'subproject_id' );
```
dregad: Break long line

core/commands/SubprojectUpdateCommand.php
```diff
@@ -0,0 +1,130 @@
...
+			throw new ClientException(
+				'Access denied to delete subprojects',
+				ERROR_ACCESS_DENIED );
+		}
+
+		if( !project_exists( $this->project_id )) {
+			throw new ClientException(
+			'Project not found',
+			ERROR_PROJECT_NOT_FOUND,
+			array( $this->project_id ) );
```
dregad: Fix indentation

core/commands/SubprojectDeleteCommand.php
```diff
@@ -0,0 +1,115 @@
...
+			throw new ClientException(
+				'Access denied to delete subprojects',
+				ERROR_ACCESS_DENIED );
+		}
+
+		if( !project_exists( $this->project_id )) {
+			throw new ClientException(
+			'Project not found',
+			ERROR_PROJECT_NOT_FOUND,
+			array( $this->project_id ) );
```
dregad: Fix indentation

obmsch:
@vboctor As it stands, projects and project hierarchy do not have a simple relationship.
Any project might have 0..n parents (at arbitrary levels in the hierarchy). There is
no explicit root project (implicitly perhaps ALL_PROJECTS). Only loops are forbidden, aka
a project can't be a decendant of itself.
I am not against an additional, creating a project with a parent in one go (#25397) function.
But given the said above, I don't see a practical way to handle project hierarchy through a
parent project.

api/rest/restcore/projects_rest.php
```diff
@@ -114,6 +122,121 @@ function rest_project_version_add( \Slim\Http\Request $p_request, \Slim\Http\Res
...
+		'query' => array(
+			'project_id' => $t_project_id
+		),
+		'payload' => $t_subproject_to_add
+	);
+
+	$t_command = new SubprojectAddCommand( $t_data );
+	$t_subproject_added = $t_command->execute();
+
+	return $p_response->withStatus( HTTP_STATUS_CREATED,
```
vboctor: Adding a version returns and updating a sub-project returns no content status code and no json.  This seems inconsistent.

api/rest/restcore/projects_rest.php
```diff
@@ -114,6 +122,121 @@ function rest_project_version_add( \Slim\Http\Request $p_request, \Slim\Http\Res
...
+		)
+	);
+
+	$t_command = new SubprojectDeleteCommand( $t_data );
+	$t_result = $t_command->execute();
+
+	$t_subproject_id = (int)$t_result['id'];
+
+	return $p_response->withStatus( HTTP_STATUS_NO_CONTENT,
+		"Subproject ($t_subproject_id) deleted from project ($t_project_id)" );
```
vboctor: Use single quotes around project ids.

core/commands/SubprojectAddCommand.php
```diff
@@ -0,0 +1,139 @@
...
+class SubprojectAddCommand extends Command {
+	/**
+	 * $p_data['query'] is expected to contain:
+	 * - project_id (integer)
+	 *
+	 * $p_data['payload'] is expected to contain:
+	 * - subproject_id (integer)
+	 * - inherit_parent (bool)
+	 *
+	 * @param array $p_data The command data.
```
vboctor: This comment should be above the constructor.

core/commands/SubprojectAddCommand.php
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * A command that adds a project subproject.
+ */
+class SubprojectAddCommand extends Command {
+	/**
+	 * $p_data['query'] is expected to contain:
+	 * - project_id (integer)
+	 *
+	 * $p_data['payload'] is expected to contain:
+	 * - subproject_id (integer)
```
vboctor: This field should be `project` with value of json blob that can container `name`, `id`, or both.  If both are provided, the `id` takes precedence.  There are utility methods that resolve such json structures to an id.

core/commands/SubprojectAddCommand.php
```diff
@@ -0,0 +1,139 @@
...
+	function validate() {		
+		$this->project_id = helper_parse_id( $this->query( 'project_id' ), 'project_id' );
+
+		if( !access_has_project_level( config_get( 'manage_project_threshold' ), $this->project_id ) ) {
+			throw new ClientException(
+				'Access denied to add subprojects',
+				ERROR_ACCESS_DENIED );
+		}
+
+		if ( config_get( 'subprojects_enabled' ) == OFF ) {
```
vboctor: I would start this check that subprojects is enabled.

core/commands/SubprojectAddCommand.php
```diff
@@ -0,0 +1,139 @@
...
+			throw new ClientException(
+				'Access denied to add subprojects',
+				ERROR_ACCESS_DENIED );
+		}
+
+		if( !project_exists( $this->project_id )) {
```
vboctor: I would follow subprojects enabled check with check that both project and subproject exist.

core/commands/SubprojectAddCommand.php
```diff
@@ -0,0 +1,139 @@
...
+		parent::__construct( $p_data );
+	}
+
+	/**
+	 * Validate the data.
+	 */
+	function validate() {		
+		$this->project_id = helper_parse_id( $this->query( 'project_id' ), 'project_id' );
+
+		if( !access_has_project_level( config_get( 'manage_project_threshold' ), $this->project_id ) ) {
```
vboctor: We should check that both project and sub-project can be managed by the user.

core/commands/SubprojectAddCommand.php
```diff
@@ -0,0 +1,139 @@
...
+
+		$this->subproject_id = helper_parse_id( $this->payload( 'subproject_id' ), 'subproject_id' );
+		if( !project_exists( $this->subproject_id )) {
+			throw new ClientException(
+				"Project \"$this->subproject_id\" not found",
+				ERROR_PROJECT_NOT_FOUND,
+				array( $this->subproject_id ) );
+		}
+
+		if( $this->project_id == $this->subproject_id ) {
```
vboctor: We are better off checking this after that sub-projects is enabled and before checking access levels.  If core API does such check and loop checks, then we should remove it from here and depend on the core APIs bubbling exceptions.

core/commands/SubprojectAddCommand.php
```diff
@@ -0,0 +1,139 @@
...
+				array( $this->subproject_id, $this->project_id ) );
+		}		
+
+		if( in_array( $this->project_id, project_hierarchy_get_all_subprojects( $this->subproject_id, true ) ) ) {
+			throw new ClientException(
+				"Project can't be a descendant subproject of itself",
+				ERROR_PROJECT_RECURSIVE_HIERARCHY );
+		}		
+
+		$this->inherit_parent = $this->payload( 'inherit_parent', true );
```
vboctor: This can move to process() and hence not need an attribute in the class.  There is no validation in this step.

core/commands/SubprojectAddCommand.php
```diff
@@ -0,0 +1,139 @@
...
+			# in case the current project is not the same project of the bug we are
+			# viewing, override the current project. This to avoid problems with
+			# categories and handlers lists etc.
+			global $g_project_override;
+			$g_project_override = $this->project_id;
+		}
+
+		project_hierarchy_add( $this->subproject_id, $this->project_id , $this->inherit_parent );
+
+		return array( 'parent_id' => $this->project_id,
```
vboctor: Not sure if this is needed, but if it is then we should return "project": { "id": 1234, "name ": "xyz" }.  I believe there is a helper method that returns a project reference structure which contains these two fields.

core/commands/SubprojectDeleteCommand.php
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * A command that deletes a project subproject.
+ */
+class SubprojectDeleteCommand extends Command {
+	/**
+	 * $p_data['query'] is expected to contain:
+	 * - project_id (integer)
+	 * - subproject_id (integer)
+	 *
+	 * @param array $p_data The command data.
```
vboctor: This comment should be associated with constructor.

core/commands/SubprojectDeleteCommand.php
```diff
@@ -0,0 +1,116 @@
...
+		parent::__construct( $p_data );
+	}
+
+	/**
+	 * Validate the data.
+	 */
+	function validate() {		
+		$this->project_id = helper_parse_id( $this->query( 'project_id' ), 'project_id' );
+
+		if( !access_has_project_level( config_get( 'manage_project_threshold' ), $this->project_id ) ) {
```
vboctor: Similar to sub-project add, we should check if we should check that subproject is manageable by the user as well.

core/commands/SubprojectDeleteCommand.php
```diff
@@ -0,0 +1,116 @@
...
+			throw new ClientException(
+				'Access denied to delete subprojects',
+				ERROR_ACCESS_DENIED );
+		}
+
+		if ( config_get( 'subprojects_enabled' ) == OFF ) {
```
vboctor: Revise ordering to be consistent with revised ordering in sub-project add.

core/commands/SubprojectDeleteCommand.php
```diff
@@ -0,0 +1,116 @@
...
+			# in case the current project is not the same project of the bug we are
+			# viewing, override the current project. This to avoid problems with
+			# categories and handlers lists etc.
+			global $g_project_override;
+			$g_project_override = $this->project_id;
+		}
+
+		project_hierarchy_remove( $this->subproject_id, $this->project_id );
+
+		return array( 'id' => $this->subproject_id );
```
vboctor: Not sure we need to return anything here, but if we do, we should return a `project` structure and not an `id` field.

core/commands/SubprojectUpdateCommand.php
```diff
@@ -0,0 +1,131 @@
...
+class SubprojectUpdateCommand extends Command {
+	/**
+	 * $p_data['query'] is expected to contain:
+	 * - project_id (integer)
+	 * - subproject_id (integer)
+	 *
+	 * $p_data['payload'] is expected to contain:
+	 * - inherit_parent (bool)
+	 *
+	 * @param array $p_data The command data.
```
vboctor: Associate comment with constructor.

core/commands/SubprojectUpdateCommand.php
```diff
@@ -0,0 +1,131 @@
...
+
+		if ( config_get( 'subprojects_enabled' ) == OFF ) {
+			throw new ClientException(
+				'Access denied to delete subprojects',
+				ERROR_ACCESS_DENIED );
+		}
+
+		if( !project_exists( $this->project_id )) {
+			throw new ClientException(
+				"Project \"$this->project_id\"not found",
```
vboctor: Use single quotes instead of double quotes.

core/commands/SubprojectUpdateCommand.php
```diff
@@ -0,0 +1,131 @@
...
+				ERROR_PROJECT_NOT_FOUND,
+				array( $this->project_id ) );
+		}
+
+		$this->subproject_id = helper_parse_id( $this->query( 'subproject_id' ), 'subproject_id' );
+		if( !project_exists( $this->subproject_id )) {
+			throw new ClientException(
+				"Project \"$this->subproject_id\" not found",
```
vboctor: single quotes around project id.

core/commands/SubprojectUpdateCommand.php
```diff
@@ -0,0 +1,131 @@
...
+				ERROR_PROJECT_NOT_FOUND,
+				array( $this->subproject_id ) );
+		}
+		
+		if( !in_array( $this->subproject_id, project_hierarchy_get_subprojects( $this->project_id, true ) ) ) {
+			throw new ClientException(
+				"Project \"$this->subproject_id\" is not a subproject of \"$this->project_id\"",
```
vboctor: single quotes around project ids.

core/commands/SubprojectUpdateCommand.php
```diff
@@ -0,0 +1,131 @@
...
+			# in case the current project is not the same project of the bug we are
+			# viewing, override the current project. This to avoid problems with
+			# categories and handlers lists etc.
+			global $g_project_override;
+			$g_project_override = $this->project_id;
+		}
+
+		project_hierarchy_update( $this->subproject_id, $this->project_id, $this->inherit_parent );
+
+		return array( 'id' => $this->subproject_id );
```
vboctor: This is not needed and if it is, then it should be a full structure for sub-project with project that has id and name, and then an inherit_parent.

lang/strings_english.txt
```diff
@@ -1259,8 +1259,6 @@ $s_manage_filter_page_title = 'Manage Filters';
 $s_manage_filter_edit_page_title = 'Edit filter';
 $s_delete_filter_button = 'Delete';
 $s_apply_filter_button = 'Apply';
-$s_temporary_filter = 'Temporary filter';
```
vboctor: Maybe this is a rebase issue, but there shouldn't be filter related changes.

api/rest/restcore/projects_rest.php
```diff
@@ -39,6 +39,14 @@
 	# Project versions
 	$g_app->post( '/{id}/versions', 'rest_project_version_add' );
 	$g_app->post( '/{id}/versions/', 'rest_project_version_add' );
+
+	# Project subprojects
+	$g_app->post( '/{id}/subprojects', 'rest_project_subproject_add' );
+	$g_app->post( '/{id}/subprojects/', 'rest_project_subproject_add' );
+	$g_app->patch( '/{id}/subprojects/{subproject_id}', 'rest_project_subproject_update' );
```
vboctor: I wonder if we should make subproject_id optional in the patch since it should be specified in the body along with inherit_parent using subproject id or subproject name under the project field.

obmsch:
@vboctor hopefully adressed all your comments. Not sure about:

> I wonder if we should make subproject_id optional in the patch

You like to have an additional endpoint, or only reduce the existing one and
make it equivalent to post?

api/rest/restcore/projects_rest.php
```diff
@@ -114,6 +122,113 @@ function rest_project_version_add( \Slim\Http\Request $p_request, \Slim\Http\Res
...
+
+	$t_data = array(
+		'query' => array(
+			'project_id' => $t_project_id
+		),
+		'payload' => $p_request->getParsedBody()
+	);
+
+	$t_command = new SubprojectAddCommand( $t_data );
+	$t_result = $t_command->execute();
```
vboctor: $t_result isn't used, please remove it.

api/rest/restcore/projects_rest.php
```diff
@@ -114,6 +122,113 @@ function rest_project_version_add( \Slim\Http\Request $p_request, \Slim\Http\Res
...
+		'query' => array(
+			'project_id' => $t_project_id,
+			'subproject_id' => $t_subproject_id
+		),
+		'payload' => $t_subproject_update
+
+	);
+
+	$t_command = new SubprojectUpdateCommand( $t_data );
+	$t_result = $t_command->execute();
```
vboctor: Remove `$t_result`.

api/rest/restcore/projects_rest.php
```diff
@@ -114,6 +122,113 @@ function rest_project_version_add( \Slim\Http\Request $p_request, \Slim\Http\Res
...
+
+	$t_data = array(
+		'query' => array(
+			'project_id' => $t_project_id,
+			'subproject_id' => $t_subproject_id
+		)
+	);
+
+	$t_command = new SubprojectDeleteCommand( $t_data );
+	$t_result = $t_command->execute();
```
vboctor: Remove `$t_result`.

core/commands/SubprojectAddCommand.php
```diff
@@ -0,0 +1,146 @@
...
+	 * - project (array)
+	 * - inherit_parent (bool)
+	 *
+	 * @param array $p_data The command data.
+	 */
+	function __construct( array $p_data ) {
+		parent::__construct( $p_data );
+	}
+
+	private function get_subproject_project_id( $p_project ) {
```
vboctor: Use `mci_get_project_id()` similar to `IssueAddCommand`.  Also this method is missing phpdocs, but it should be removed anyway.

core/commands/SubprojectDeleteCommand.php
```diff
@@ -0,0 +1,114 @@
...
+require_api( 'constant_inc.php' );
+require_api( 'config_api.php' );
+require_api( 'helper_api.php' );
+require_api( 'project_api.php' );
+require_api( 'project_hierarchy_api.php' );
+
+use Mantis\Exceptions\ClientException;
+
+/**
+ * A command that deletes a project subproject.
```
vboctor: We should clarify that we are deleting the sub-project relationship rather than the delete the sub-project itself (as in delete project).  In the Postman APIs that I have added, I used "add/update/delete subproject relationship".

obmsch: Are you asking to simply clarify this in the comments, or reflect this in the commands too, by renaming them to something like `SubprojectRelationshipAdd/Update/DeleteCommand`?

core/commands/SubprojectAddCommand.php
obmsch: The only reason I added that private, is to distinguish neither `id` nor `name` are set from
only `name` is set but that project doesn't exist, and reflect that to the caller.

`IssueAddCommand` BTW raises 'Project not specified' in both cases, which is at least questionable for the latter.

core/commands/SubprojectDeleteCommand.php
vboctor: I was talking about the comments, but now that you mention it, reflecting in the command names is even better.

core/commands/SubprojectAddCommand.php
vboctor: I would rather improve `mc_get_project_id()` by adding an optional parameter for default project id ($p_default = ALL_PROJECTS).  You can call it with default value `false` and that will give you the distinction.  We can eventually update other callers to do the same.

obmsch:
@vboctor all your review comments should be resolved, including the high level ones. Except:

> I noticed that in Get Project, the sub-project information doesn't include the `inherit_paent` field.  We should add it there.

This should get its own issue in the tracker and not handled by this PR. From my point of view, Get Project has more issues. Let's say project (1) has three subprojects (2,3,4). For a request GetProject(1) I would expect a response with one project (1) including its subproject list. Actually returned is an array with four projects (1,2,3,4). This is at least not consistent, as for example GetIssue(123) only returns one issue, and not additionally those in relationships.

vboctor:
Merged via 4cc687a9c0ad987f6d43ad8d4fafa92e0d5c54e2.  Will open a bug to fix sub-projects in GET project or add GET subprojects.

vboctor:
@obmsch opened bug https://www.mantisbt.org/bugs/view.php?id=25514

PR #1454, opened 2019-01-31
Minor documentation fixes in Admin Guide and config_defaults_inc.php

atrol:


docbook/Admin_Guide/en-US/config/defaults.xml
```diff
@@ -249,11 +249,10 @@
 			</listitem>
 		</varlistentry>
 		<varlistentry>
-			<term>$g_default_email_on_status$g_default_email_on_priority</term>
+			<term>$g_default_email_on_status, $g_default_email_on_priority</term>
```
vboctor: Should these be split into two separate entries?

docbook/Admin_Guide/en-US/config/defaults.xml
```diff
@@ -249,11 +249,10 @@
...
 			<listitem>
 				<para>Default user preferences to enable receiving emails when
-					status or priority is changed. Default is ON. Note that this option
-					is not implemented.
+					status or priority is changed. Default is OFF.
```
vboctor: same here.

PR #1455, opened 2019-02-01
Fix 'Undefined variable' PHP notice in API REST

dregad:
Since the Slim Container's 'displayErrorDetails' setting is set to true
according to $t_show_detailed_errors's value, instead of referencing the
variable directly within the closure via a 'use' statement, we retrieve
the Container's settings to determine whether PHP exception details
should be shown or not.

Regression introduced by b2119ce0dc944f1db13813e863471e2bc3ae64f0.

Fixes #25429 (see #1280)

### DIFF
--- a/api/rest/index.php
+++ b/api/rest/index.php
@@ -46,8 +46,7 @@ ApiObjectFactory::$soap = false;
 # Show SLIM detailed errors according to Mantis settings
 $t_config = array();
 
-$t_show_detailed_errors = ON == config_get_global( 'show_detailed_errors' );
-if( $t_show_detailed_errors ) {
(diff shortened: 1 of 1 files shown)


atrol:
While you are on it, you might also want to change
```
$t_show_detailed_errors = ON == config_get_global( 'show_detailed_errors' );
if( $t_show_detailed_errors ) {
	$t_config['settings'] = array( 'displayErrorDetails' => true );
}
```
to
```
if( ON == config_get_global( 'show_detailed_errors' ) ) {
	$t_config['settings'] = array( 'displayErrorDetails' => true );
}
```

dregad:
> While you are on it, you might also want to change [...]

Good suggestion, done.

PR #1457, opened 2019-02-05
Fix mssql concurrent insert id

cproensa:
The implementation of db_insert_id() in mssql relied on IDENT_CURRENT
making it not thread safe.
Modify it to use ADOdb native implementation, based on SCOPE_IDENTITY()
which is session independent.

Fixes: #25442

### DIFF
--- a/core/database_api.php
+++ b/core/database_api.php
@@ -451,32 +451,41 @@ function db_result( $p_result, $p_row_index = 0, $p_col_index = 0 ) {
 }
 
 /**
- * return the last inserted id for a specific database table
+ * Return the last inserted ID after a insert statement.
+ * Warning: this function must be used immediately after the insert statement
+ *
+ * This relies on ADOdb to get the entity id when this functionality is available
+ * for the specific driver, and it makes sense in our model.
+ * Natively supported:
+ * - mysqli, using: mysqli_insert_id(connection).
+ * - mssqlnative, using SCOPE_IDENTITY().
+ * Not natively supported:
+ * - pgsql, oracle, using the underlying sequence for the table.
+ *
+ * Since the table is needed for those drivers where a sequence is used, the
+ * $p_table parameter is mandatory to ensure portability.
+ * Warning: $p_table is not expected to be a different table than the one used
+ * for the previous insert. Note that it's not even used by some drivers.
+ *
  * @param string $p_table A valid database table name.
  * @param string $p_field A valid field name (default 'id').
  * @return integer last successful insert id
  */
-function db_insert_id( $p_table = null, $p_field = 'id' ) {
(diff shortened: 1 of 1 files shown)


core/database_api.php
```diff
@@ -451,32 +451,43 @@ function db_result( $p_result, $p_row_index = 0, $p_col_index = 0 ) {
 }
 
 /**
- * return the last inserted id for a specific database table
+ * Return the last inserted ID after a insert statement.
+ * Warning: this function must be used immediately after the insert statement
+ *
+ * This relies in AdoDB to get the entity id when this functionality is available
```
dregad: on ADOdb

core/database_api.php
```diff
@@ -451,32 +451,43 @@ function db_result( $p_result, $p_row_index = 0, $p_col_index = 0 ) {
...
+ * for the specific driver, and it makes sense in our model.
+ * Natively supported:
+ * - mysqli, using: mysqli_insert_id(connection).
+ * - mssqlnative, using SCOPE_IDENTITY().
+ * Not natively supported:
+ * - pgsql, oracle, using the underlying sequence for the table.
+ *
+ * Since the table is needed for those drivers where a sequence is used, the
+ * $p_table parameter is mandatory to ensure portability.
```
dregad: I assume you've checked that the function is not called anywhere without the table parameter

core/database_api.php
```diff
@@ -451,32 +451,43 @@ function db_result( $p_result, $p_row_index = 0, $p_col_index = 0 ) {
...
-		}
 	}
-	return $g_db->Insert_ID();
+	if( isset( $t_query ) ) {
+		$t_result = db_query( $t_query );
+		return (int)db_result( $t_result );
+	} else {
+		return $g_db->Insert_ID();
+	}
 }
```
dregad: Avoid the if() with

``` php
switch( $g_db_functional_type ) {
    ...
    default:
        return $g_db->Insert_ID();
}
$t_result = db_query( $t_query );
return (int)db_result( $t_result );
```

cproensa: thanks

cproensa:
> I assume you've checked that the function is not called anywhere without the table parameter

Yes, all usages from core call it with the table parameter.

core/database_api.php
dregad: This wasn't fixed... `in AdoDB` -> `on ADOdb`

dregad: Sorry for picking nits, it's *rely **on***, not *in*
https://dictionary.cambridge.org/dictionary/english/rely-on-upon-sb-sth

cproensa: > _**on**_, not _in_

sorry i din't notice that

the corrections are welcome, thanks

PR #1458, opened 2019-02-05
Hide "Report Issue" button in navbar on bug_report_page

atrol:
Fixes #23045


atrol:
@dregad it would be good if you could check/approve again, as I added the change for the bug_update_page.

vboctor:
I would rather if the button is always visible, but disabled when we don't want it to be pressed rather than the left bar changing all the time.

PR #1460, opened 2019-02-07
 Allow show_queries_count to be set in database

atrol:
The 2nd commit `Remove unused function helper_show_query_count` could be removed from the PR if someone thinks that it could be used in 3rd party code.


cproensa:
Tested and looks good to me.

> The 2nd commit `Remove unused function helper_show_query_count` could be removed from the PR if someone thinks that it could be used in 3rd party code.

I'd say to delete it, since i doubt this has any point to be used by external code...
If in doubt, we can mark as deprecated, and delete at some point in the future

atrol:
@vboctor do you agree to merge after my [latest note]( https://www.mantisbt.org/bugs/view.php?id=25446#c61468)?

vboctor:
@atrol I'm not convinced that is the right path.  See my comments in the associated issues.

atrol:
Closed in favour of https://github.com/mantisbt/mantisbt/pull/1466

PR #1461, opened 2019-02-07
Fixes for Config report page

cproensa:
Fix adm_config_report user list
    
    The code to build the user dropdown list have some issues:
    - Missing db_param_push()
    - Calls to get user names are not cached generating a lot of separate
      queries for individual user ids
    - User names are not properly ordered in the dropdown list, as it's
      sorting with a case sensitive mode.

Fix adm_config_report query
    
    The main query has a missing db_param_push(), but because of how the
    query is built, it should only be used only when any of the conditions
    are actually adding a parameter.
    To simplyfy, the query has been rewritten with the new syntax.

Fixes: #25454, #25455, #25456

### DIFF
--- a/adm_config_report.php
+++ b/adm_config_report.php
@@ -233,20 +233,27 @@ $t_edit_action = in_array( $f_edit_action, $t_valid_actions )
 # Apply filters
 
 # Get users in db having specific configs
-$t_query = new DbQuery( 'SELECT DISTINCT user_id FROM {config} WHERE user_id <> :all_users',
-		array( 'all_users' => ALL_USERS ) );
-# build result as: array( [id]=>id, ... )
-$t_user_ids = array_column( $t_query->fetch_all(), 'user_id', 'user_id' );
-
+$t_sql = 'SELECT id, username, realname FROM {user} WHERE {user}.id IN ('
+		. ' SELECT user_id FROM {config} WHERE user_id <> :all_users )';
+$t_query = new DbQuery( $t_sql, array( 'all_users' => ALL_USERS ) );
+$t_users_list = array();
+$t_users_ids = array();
+$t_sort = array();
+while( $t_row = $t_query->fetch() ) {
+	$t_users_ids[] = (int)$t_row['id'];
+	$t_users_list[] = user_get_name_from_row( $t_row );
+	$t_sort[] = user_get_name_for_sorting_from_row( $t_row );
+}
 if( $t_filter_user_value != META_FILTER_NONE && $t_filter_user_value != ALL_USERS ) {
 	# Make sure the filter value exists in the list
-	$t_user_ids[$t_filter_user_value] = $t_filter_user_value;
+	$t_row = user_get_row( $t_filter_user_value );
+	$t_users_ids[] = $t_filter_user_value;
+	$t_users_list[] = user_get_name_from_row( $t_row );
+	$t_sort[] = user_get_name_for_sorting_from_row( $t_row );
 }
-user_cache_array_rows( $t_user_ids );
-
-# build array for user names
-$t_users_list = array_map( 'user_get_name', $t_user_ids );
-asort( $t_users_list, SORT_NATURAL | SORT_FLAG_CASE );
+user_cache_array_rows( $t_users_ids );
+array_multisort( $t_sort, SORT_ASC, SORT_NATURAL | SORT_FLAG_CASE, $t_users_list, $t_users_ids );
+$t_users_list = array_combine( $t_users_ids, $t_users_list );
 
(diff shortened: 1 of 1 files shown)


cproensa:
This page would benefit from table sorting and pagination like proposed in #1449

cproensa:
> After: Total queries executed: 61

that number still looks a bit high, can you detail if those come from this page? that is aproximately after: `project_api.php:149 project_cache_array_rows()`
or what conditions are on your filter, request, etc.

> I noticed that there is not much difference if the filter shows all users/all projects/all configs, or specific user for example.  The number of queries and time is about the same.  However, that was also the case before the changes.

That may be explained by the config and user tables join... since there is need to get the real/sorting names,i thought retrieving directly the user "rows" would save on later calls to user_api for individual users. That's how is done in other user dropdowns, and calling the row based functions saves several subsequent calls.
But it may be faster to keep single table queries. your times also seems comparatively high, so specific database version or performance can be at play there.
I will try the other approach and compare.

> While I was testing, I noticed that if I enabled log level to DATABASE and log to file, PHP always runs out of memory.  That happens even if I go to a page that doesn't do much queries, like the default Manage page which just shows MantisBT site information.  Are you able to reproduce this on your side as well?  I wonder if it is related to DbQuery changes.  I disabled pushing to `$g_queries_array` but still didn't help.

No, i did't have those issues. Could that be related to [#25432](https://www.mantisbt.org/bugs/view.php?id=25432) ? where some 5.x php version have significantly bad memory behaviour for preg_match.

atrol:
Concerning memory usage:

About 8MB  with 10 users
About 55MB with 34500 users

No big difference between PHP 5.6 and 7.x

vboctor:
Some info on my environment:

1. MAMP PRO
2. MacOS
3. PHP 7.1.12
4. MySQL 5.6.38
5. Apache

A DB with large number of users, and a small number of projects.

cproensa:
And compared to previous implementation? time & memory?

The query i was talking about, for building the dropdown will allways have the same cost, regardless of showinf the config report fitlered or not, becasue all users are rendered in the list. Anyway, you can look at this other version if you like it better:
https://github.com/cproensa/mantisbt/commit/1de9bafc262da4532f91b3e6e6b33b2cb0c774a3
Afaict, there is a very small improvement in the query time, and a small downgrade in execution time (more recurrent function calls), averaging into the same.

> About 55MB with 34500 users

That means, you have at least 34500 config records?

> The time goes down significantly (but still high) when I disable xdebug.

I never noticed any significant time penalty by having xdebug enabled, but that may be because the module is not "active" until a debug session is called from the IDE. I have: `xdebug.remote_autostart => Off` 
Once the debug session is started, however, everything its noticeably slow.

vboctor:
@cproensa Two ideas come to mind for improving performance:

1. Should we have the username as an text box?  This is similar to the text box we have in Manage Users page to search for user by username or email.

2. Should we figure out the config options that will be displayed, then get the associated information, rather than looking at all configs in the db.

atrol:
> That means, you have at least 34500 config records?

No, just about 1300

cproensa:
> Should we have the username as an text box?  This is similar to the text box we have in Manage Users page to search for user by username or email.

Ideally, most if not all, of the user related inputs would benefit from an autocomplete, active, text box.
As for the specific search-enabled text box from manage-users, this case is a bit different, since it's useful to know which users, and only those, have an option in this table.
As opposed to, eg, the user list in the bottom "create/edit" form, where you can choose from any user.

> Should we figure out the config options that will be displayed, then get the associated information, rather than looking at all configs in the db.

you mean, for building the filter inputs? the selected filter should not limit the other filter parameters. otherwise you wouldn't be able to generalize the search.


in my system i have 700 options for 70 unique users.
so i could insert some sore records to stress it a bit more

vboctor:
@cproensa Had another look and the issue is not related at all to the filter or the listing of the configs.  The bulk of the time (i.e. 80-90%) is spent populating the combo-box of users in the add config option form below the list.  On my local setup it takes 6-7s with it and 1 or less without it.

Query stats WITHOUT populating the add-config-form users combo box
Total queries executed: 21     Total query execution time: 0.0325 seconds

Query stats WITH populating the add-config-form users combo box
Total queries executed: 61     Total query execution time: 0.415 seconds

Note that the list of users in the filter part is small since it only contains users with specific configs, but the list in the add-config-option form has all of the users and hence the high cost.  This is similar to the "Reporter" options in the update-issue page, which we worked around it by only populating it when user specified that they want to change the reporter.

I think we could easily switch the user dropdown in the config dialog with an text box and cut all this cost and save a bunch of memory and db load too.

PR #1463, opened 2019-02-11
Dropzone improvements

cproensa:
Several fixes and improvements for Dropzone.
- Update version
- Improve validation of file size (at client side)
- Fix image thumbnails for previews
- Redesign the previews

I added "bootbox" library as a helper to display bootstrap dialogs. I think using the framework is better than a native "alert", but the choose of helper library can be reconsidered. This one looks simple enough.

seleccion_341

seleccion_342


atrol:
Not able to test, I get
```
Dropzone.js does not support older browsers!
```

cproensa:
Have you cleared the js cache?
if persisting, which browser?

atrol:
Any browser.
Just found that it happens just if `$g_cdn_enabled = ON;`

cproensa:
> it happens just if `$g_cdn_enabled = ON;` 

thanks, the integrity hash was wrong. I updated with the correct one, please refresh your local branch

bug_report_page.php
```diff
@@ -622,7 +622,9 @@
 <?php
 	# File Upload (if enabled)
 	if( $t_show_attachments ) {
-		$t_max_file_size = (int)min( ini_get_number( 'upload_max_filesize' ), ini_get_number( 'post_max_size' ), config_get( 'max_file_size' ) );
+		$a1 = ini_get_number( 'upload_max_filesize' );
```
vboctor: Are `a1` and `a2` still needed?

atrol:
> but the choose of helper library can be reconsidere

Adding a library and load it with each request is quite a lot of overhead for a better look of a single error message.

vboctor:
I agree with @atrol.

cproensa:
> I assume the same browsers are supported by the new libraries.

will check.

> Adding a library and load it with each request is quite a lot of overhead for a better look of a single error message.

Ok. So are you fine with this message being a native alert() then?

A new issue: 
Dropzone allows adding several files in one step. When there are more that one file that surpass the size limit, i guess the expected result is one alert with a message referencing several dropped files, instead of N alerts.
This also means the current message is not directly reusable, so it will need new text strings and translations.

cproensa:
https://www.dropzonejs.com/#compatibility

> Browser Support
Chrome 7+
Firefox 4+
IE 10+
Opera 12+ (Version 12 for MacOS is disabled because their API is buggy)
Safari 6+
For all the other browsers, dropzone provides an oldschool file input fallback.

atrol:
> Ok. So are you fine with this message being a native alert() then?

I am fine with it.
Maybe using something that comes with Bootstrap is also an option (didn't have a deeper look)
https://www.w3schools.com/bootstrap/bootstrap_modal.asp

cproensa:
All comments have been addressed

> Maybe using something that comes with Bootstrap is also an option

This libray is a wrapper for bootstrap modal. With raw bootstrap, you still have to write all the html for a div, buttons, ....

PR #1465, opened 2019-02-14
Version improvements

cproensa:
Some improvements for version related apis

- Clean up of queries and cache. Less database queries.
- Fix some issues due to wrong cache.
- Fix including inherited versions in filters

https://www.mantisbt.org/bugs/view.php?id=12261
https://www.mantisbt.org/bugs/view.php?id=22100
https://www.mantisbt.org/bugs/view.php?id=23245
https://www.mantisbt.org/bugs/view.php?id=24821

### DIFF
--- a/core/prepare_api.php
+++ b/core/prepare_api.php
@@ -92,9 +92,10 @@ function prepare_user_name( $p_user_id ) {
  * A function that prepares the version string for outputting to the user on view / print issue pages.
  * This function would add the version date, if appropriate.
  *
- * @param integer $p_project_id The project id.
- * @param integer $p_version_id The version id.  If false then this method will return an empty string.
- * @param boolean $p_show_project Force prepending project name to version
+ * @param integer $p_project_id         The project id to use as context.
+ * @param integer $p_version_id         The version id. If false then this method will return an empty string.
+ * @param boolean|null $p_show_project  Whether to include the project name or not,
+ *                                      null means include the project if different from current context.
  * @return string The formatted version string.
  */
 function prepare_version_string( $p_project_id, $p_version_id, $p_show_project = null ) {
--- a/core/prepare_api.php
+++ b/core/prepare_api.php
@@ -94,14 +94,15 @@ function prepare_user_name( $p_user_id ) {
  *
  * @param integer $p_project_id The project id.
  * @param integer $p_version_id The version id.  If false then this method will return an empty string.
+ * @param boolean $p_show_project Force prepending project name to version
  * @return string The formatted version string.
  */
-function prepare_version_string( $p_project_id, $p_version_id ) {
+function prepare_version_string( $p_project_id, $p_version_id, $p_show_project = null ) {
 	if( $p_version_id === false ) {
 		return '';
 	}
 
-	$t_version_text = version_full_name( $p_version_id, null, $p_project_id );
+	$t_version_text = version_full_name( $p_version_id, $p_show_project, $p_project_id );
 
 	if( access_has_project_level( config_get( 'show_version_dates_threshold' ), $p_project_id ) ) {
 		$t_short_date_format = config_get( 'short_date_format' );
--- a/core/prepare_api.php
+++ b/core/prepare_api.php
@@ -107,8 +107,10 @@ function prepare_version_string( $p_project_id, $p_version_id, $p_show_project =
 	if( access_has_project_level( config_get( 'show_version_dates_threshold' ), $p_project_id ) ) {
 		$t_short_date_format = config_get( 'short_date_format' );
 
-		$t_version = version_get( $p_version_id );
-		$t_version_text .= ' (' . date( $t_short_date_format, $t_version->date_order ) . ')';
+		$t_version = version_cache_row( $p_version_id );
+		if( 1 == $t_version['released'] ) {
+			$t_version_text .= ' (' . date( $t_short_date_format, $t_version['date_order'] ) . ')';
+		}
 	}
 
 	return $t_version_text;
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -846,34 +846,34 @@ function print_os_build_option_list( $p_os_build, $p_user_id = null ) {
 }
 
 /**
- * Print the option list for versions
+ * Print the option list for versions.
+ * All versions related for each project will be printed. Those include, for each
+ * project, the directly linked versions and the inherited versions if applicable.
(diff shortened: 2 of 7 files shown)


core/print_api.php
```diff
@@ -846,34 +846,34 @@ function print_os_build_option_list( $p_os_build, $p_user_id = null ) {
...
 
 /**
- * Print the option list for versions
+ * Print the option list for versions.
+ * All versions related for each project will be printed. Those include, for each
+ * project, the directly linked versions and the inherited versions if applicable.
+ *
  * @param string  $p_version       The currently selected version.
- * @param integer $p_project_id    Project id, otherwise current project will be used.
+ * @param integer $p_project_ids   Array of project ids, of null to use current project.
```
dregad: PHPdoc type is wrong based on description, should be `array|null`

core/prepare_api.php
```diff
@@ -94,20 +94,23 @@ function prepare_user_name( $p_user_id ) {
  *
  * @param integer $p_project_id The project id.
  * @param integer $p_version_id The version id.  If false then this method will return an empty string.
+ * @param boolean $p_show_project Force prepending project name to version
```
dregad: PHPDoc says boolean, but param defaults to _null_. Change PHPDoc type to `boolean|null` and document what null means, or change default param value to _false_ (whatever is appropriate)

core/print_api.php
```diff
@@ -846,34 +846,34 @@ function print_os_build_option_list( $p_os_build, $p_user_id = null ) {
...
-function print_version_option_list( $p_version = '', $p_project_id = null, $p_released = null, $p_leading_blank = true, $p_with_subs = false ) {
-	if( null === $p_project_id ) {
-		$c_project_id = helper_get_current_project();
-	} else {
-		$c_project_id = (int)$p_project_id;
+function print_version_option_list( $p_version = '', $p_project_ids = null, $p_released = null, $p_leading_blank = true ) {
+	if( null === $p_project_ids ) {
+		$p_project_ids = helper_get_current_project();
 	}
+	$t_project_ids = is_array( $p_project_ids ) ? $p_project_ids : array( $p_project_ids );
```
vboctor: How about moving this conversion to array be moved inside `version_get_all_rows()`?

core/print_api.php
```diff
@@ -846,34 +846,34 @@ function print_os_build_option_list( $p_os_build, $p_user_id = null ) {
...
 	# Note: Filter API specifies selected versions as an array
 	if( !is_array( $p_version ) ) {
 		if( !empty( $p_version ) ) {
-			$t_version_id = version_get_id( $p_version, $c_project_id );
-			if( $t_version_id !== false ) {
-				$t_versions[] = version_cache_row( $t_version_id );
+			foreach( $t_project_ids as $t_project_id ) {
+				$t_version_id = version_get_id( $p_version, $t_project_id );
+				if( $t_version_id !== false ) {
+					$t_versions[] = version_cache_row( $t_version_id );
```
vboctor: Shouldn't methods that are not part of `version_api.php` use a more public method than `version_cache_row()`? For example, `version_get()` or something along these lines?

core/print_api.php
```diff
@@ -900,8 +902,8 @@ function print_version_option_list( $p_version = '', $p_project_id = null, $p_re
 			$t_listed[] = $t_version_version;
 			echo '<option value="' . $t_version_version . '"';
 			check_selected( $p_version, $t_version['version'] );
-
-			$t_version_string = string_attribute( prepare_version_string( $c_project_id, $t_version['id'] ) );
+			$c_project_id = $t_version['project_id'];
```
vboctor: Use `$t_project_id` instead of `$c_project_id`.

core/version_api.php
```diff
@@ -202,6 +213,47 @@ function version_cache_row( $p_version_id, $p_trigger_errors = true ) {
...
+	}
+
+	$t_query = new DbQuery();
+	$t_sql = 'SELECT * FROM {project_version} WHERE ' . $t_query->sql_in( 'project_id', $t_ids_to_fetch );
+	$t_query->sql( $t_sql );
+	while( $t_row = $t_query->fetch() ) {
+		$c_project_id = (int)$t_row['project_id'];
+		$c_version_id = (int)$t_row['id'];
+		if( !isset( $g_cache_versions[$c_version_id] ) ) {
+			$g_cache_versions[$c_version_id] = $t_row;
```
vboctor: Why not refresh the cache anyway since the version was retrieved from the db?

core/version_api.php
```diff
@@ -438,156 +490,70 @@ function version_remove_all( $p_project_id ) {
...
-		$g_cache_versions_project[(int)$t_project_id] = $t_row;
-	}
-}
-
 /**
- * Return all versions for the specified project
- * @param integer $p_project_id A valid project id.
+ * Return all versions for the specified project or projects list
+ * Returned versions are ordered by reverse 'date_order'
+ * @param integer|array $p_project_ids  A valid project id, or array of ids
```
vboctor: Looks like you already support a single version or an array, then doing the same from print_api is redundant.

core/print_api.php
cproensa: $t_project_ids is used later in the method, and having it normalized to an array is cleaner.
`version_get_all_rows()` may be called from other places, so it also has to normalize the input separately.

core/print_api.php
cproensa: `version_get()`returns a VersionData object. Most of the places we user directly version "rows", but the only method to get a row is `version_cache_row()`.
Your comment is in the line of why we have `bug_get_row()` which is a passthrough wrapper for `bug_cache_row()`

core/version_api.php
cproensa: Is this related to previous comment?: https://github.com/mantisbt/mantisbt/pull/1465#discussion_r257476614
about the `is_array( $p_project_ids ) ? $p_project_ids : array( $p_project_ids )`

core/print_api.php
vboctor: Makes sense to have `version_get_row()` then.  Not a blocker, but nice to do.

core/version_api.php
vboctor: Yep, seems you have reasons for duplication, so I'm ok here.

core/print_api.php
vboctor: ok

PR #1467, opened 2019-02-18
Update Chart.js to 2.7.3

atrol:
Fixes #25488


cproensa:
Maybe it's a matter of tuning the parameters for the graphs, but with this new version as is, the legends on the bar charts are significantly more sparse.
The usernames are blured but can be seen here:

Before:

seleccion_348

After:

seleccion_349

atrol:
Reporter Charts look quite good on my test system.
reportercharts
Developer charts are worse than before the chart library update, some other charts (e.g. status) are better than before.
developercharts
@cproensa while testing I noticed that top developers are no longer limited to 20 users since your Graph changes.

cproensa:
> Reporter Charts look quite good on my test system.

Mine is with less window width. With less space, the distribution of labels is different from previous version

>  I noticed that top developers are no longer limited to 20 users since your Graph changes.

you are right. I limited the reporter graph, but forgot about the developers...

dregad:
It is possible to force labels to display for each bar on the chart, as follows:

``` diff
diff --git a/plugins/MantisGraph/files/MantisGraph.js b/plugins/MantisGraph/files/MantisGraph.js
index b603e5465..cce46d5f4 100644
--- a/plugins/MantisGraph/files/MantisGraph.js
+++ b/plugins/MantisGraph/files/MantisGraph.js
@@ -14,6 +14,11 @@ $(document).ready( function() {
             },
             options: {
                 scales: {
+                    xAxes: [{
+                        ticks: {
+                            autoSkip: false
+                        }
+                    }],
                     yAxes: [{
                         ticks: {
                             beginAtZero: true
```

This might cause cramped display of labels or even overlap if there are many categories on the chart, though.

atrol:
> I limited the reporter graph, but forgot about the developers...

@cproensa will you fix this regression before we release 2.20.0?

@cproensa @dregad please check if labels are acceptable after my latest change.

atrol:
> As a side note indirectly related to this PR, I was wondering while was testing, if we shouldn't limit the _By Category_ pie chart to 10-15 slices and regroup the rest under "others" - for our tracker (> 50 categories), the chart is ugly (only a few categories have colors) and difficult to read.

Good idea.

There is one more problem, as we use the status colors for categories.
At the moment, If there are more categories than status values, a lot of categories get the same default light grey color.
If we go on using status colors for categories (and also for other pie charts), we could at least repeat the status colors until the number of colors is equal to the number of categories. E.g. with default settings start again with `new / red` after having reached `closed / grey`.

cproensa:
> will you fix this regression before we release 2.20.0?

Here: #1468

dregad:
> please check if labels are acceptable after my latest change.

It's better, especially on smaller displays, but unfortunately there are still some occasional issues even though there should enough space to display labels... e.g.

image

dregad:
> > As a side note indirectly related to this PR, I was wondering while was testing, if we shouldn't limit the _By Category_ pie chart to 10-15 slices and regroup the rest under "others" - for our tracker (> 50 categories), the chart is ugly (only a few categories have colors) and difficult to read.
> 
> Good idea.

OK, I'll open an issue in the tracker and work on that.

> There is one more problem, as we use the status colors for categories.
> At the moment, If there are more categories than status values, a lot of categories get the same default light grey color.
> If we go on using status colors for categories (and also for other pie charts), we could at least repeat the status colors until the number of colors is equal to the number of categories. E.g. with default settings start again with `new / red` after having reached `closed / grey`.

Instead of generating colors ourselves on PHP side (which only makes sense for the _By Status_ charts anyway), we should set them dynamically; a nice and simple option to achieve that would be to use a chartjs plugin e.g. https://github.com/nagix/chartjs-plugin-colorschemes. I'll make some tests.

dregad:
Please find a work-in-progress for the above-mentioned issues at https://github.com/dregad/mantisbt/tree/graph-category. 

My branch is based on @atrol's branch for this PR (as the plugin I used for color schemes requires Chart.js 2.5.0). 

This needs some testing and refining, I will provide a PR later on but until then feel free to have a look and let me know your feedback.

dregad:
> > please check if labels are acceptable after my latest change.
> 
> It's better, especially on smaller displays, but unfortunately there are still some occasional issues even though there should enough space to display labels... e.g.

@atrol I'm sure you have more important things to do, but if you can spare 5 minutes maybe you could explain if there is a particular reason for not applying my proposed fix https://github.com/mantisbt/mantisbt/pull/1467#issuecomment-467007515 for this issue ?

atrol:
> @atrol I'm sure you have more important things to do, but if you can spare 5 minutes maybe you could explain if there is a particular reason for not applying my proposed fix [#1467 (comment)](https://github.com/mantisbt/mantisbt/pull/1467#issuecomment-467007515) for this issue ?

IIRC I tried, but was not that happy with the result in all cases and tried other options.
Maybe it does no longer apply after @cproensa limited also the developer chart and after you changed category chart to vertical.

Would be good if you could think about if a hardcoded color scheme is worth to introduce / load a 3rd party component that is maintained by a single person.
Maybe introducing an own hardcoded scheme is the better option at the moment, at least as long as we don't offer users to select a color scheme.
Sorry, no time at the moment to look/comment even more. :-(

dregad:
> IIRC I tried, but was not that happy with the result in all cases and tried other options.
Maybe it does no longer apply after @cproensa limited also the developer chart and after you changed category chart to vertical.

OK, I'll make another round of comparative tests.

> think about if a hardcoded color scheme is worth to introduce / load a 3rd party component that is maintained by a single person.

I just went with the easy approach... I think it's much more convenient and simple to apply a preset with a plugin config when generating the graph (at the cost of an extra ~50K download), than having to manually maintain our own hardcoded list of colors.

True it's a risk that the plugin's development could die, but considering it's just a set of colors, IMO such risk is quite low.

cproensa:
> > @atrol I'm sure you have more important things to do, but if you can spare 5 minutes maybe you could explain if there is a particular reason for not applying my proposed fix [#1467 (comment)](https://github.com/mantisbt/mantisbt/pull/1467#issuecomment-467007515) for this issue ?
> 
> IIRC I tried, but was not that happy with the result in all cases and tried other options.
> Maybe it does no longer apply after @cproensa limited also the developer chart and after you changed category chart to vertical.

After limiting the graphs, with Damien's patch, i find it more consistent. Still there some strange cases, but overall it looks better in most of the cases..

vboctor:
@dregad is this going to be closed in favor of #1472 ?

dregad:
> @dregad is this going to be closed in favor of #1472 ?

As mentioned there, #1472 is built on top of this one, so will be merged together.

PR #1468, opened 2019-02-25
Fix MantisGraph limit for developers graph

cproensa:
Fix regression where the graph for "Developers by Fixed Issues" was not
being limited to top 20 results.

### DIFF
--- a/plugins/MantisGraph/core/graph_api.php
+++ b/plugins/MantisGraph/core/graph_api.php
@@ -279,6 +279,7 @@ function create_developer_resolved_summary( array $p_filter = null ) {
 		'status_resolved' => (int)$t_resolved_status_threshold,
 		'resolution_fixed' => FIXED
 		) );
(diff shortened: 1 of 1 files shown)


atrol:
Ok, it was hardcoded 20 and it's still hardcoded.

But why did you introduce parameter `$p_limit` in `function create_reporter_summary` instead of choosing the same simple approach?
There is no call of the function where the parameter is used.

If you think about future use, you should introduce the parameter also in this function.
If not, use the same simple hardcoded approach in `function create_reporter_summary`.

cproensa:
> But why did you introduce parameter `$p_limit` in `function create_reporter_summary` instead of choosing the same simple approach?

The reporter graph had its width modified, so seemed good that the presentation dictates the limits, not the summary logic.
Unless more work is done with the presentation, you are right, so i have changed to simple hardcoded numbers in both cases.

PR #1469, opened 2019-02-27
login failed login attempts

tgulacsi:
To allow fail2ban to ban the rouge IP, thus slowing down brute force attacks
considerably.

### DIFF
--- a/login.php
+++ b/login.php
@@ -40,6 +40,7 @@ require_api( 'gpc_api.php' );
 require_api( 'print_api.php' );
 require_api( 'session_api.php' );
 require_api( 'string_api.php' );
+require_api( 'logging_api.php' );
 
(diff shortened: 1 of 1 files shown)


dregad:
@tgulacsi While I can see the benefits of what you're trying to do, the events written to the log under `LOG_LDAP` effectively have absolutely nothing to do with LDAP.

The MantisBT logging facility was generally meant as a troubleshooting tool, it was not intended for general system purposes as you're submitting here. If we were to do this, I think we would probably need another logging level category. Maybe other devs could voice their opinion on that.

Also, your PR is not following MantisBT coding guidelines.

Finally, PRs need to reference an issue in our tracker.

As it stands, I believe this can't be merged into core.

PR #1470, opened 2019-02-27
Trim space chars in destination bug_id when create relationship.

Kirill:
Fix #25532

https://mantisbt.org/bugs/view.php?id=25532

### DIFF
--- a/bug_relationship_add.php
+++ b/bug_relationship_add.php
@@ -54,7 +54,7 @@ foreach( $t_dest_bug_id_array as $t_dest_bug_id ) {
 		'query' => array( 'issue_id' => $f_src_bug_id ),
 		'payload' => array(
 			'type' => array( 'id' => $f_rel_type ),
-			'issue' => array( 'id' => $t_dest_bug_id )
+			'issue' => array( 'id' => trim( $t_dest_bug_id ) )
 		)
 	);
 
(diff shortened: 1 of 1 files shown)


bug_relationship_add.php
```diff
@@ -54,7 +54,7 @@
 		'query' => array( 'issue_id' => $f_src_bug_id ),
 		'payload' => array(
 			'type' => array( 'id' => $f_rel_type ),
-			'issue' => array( 'id' => $t_dest_bug_id )
+			'issue' => array( 'id' => trim( $t_dest_bug_id ) )
```
dregad: Since this is supposed to be a bug id (i.e. a number), it would be better to typecast to `int` IMO

dregad:
@vboctor The Command delegates issue id retrieval and validatin to *helper_parse_id()*. That is the source of the exception causing the error reported by Kirill in [#25532](https://mantisbt.org/bugs/view.php?id=25532) when there is leading/trailing whitespace (since it is not numeric).

So the question is, how would you like to handle this ?
- modify *IssueRelationshipAddCommand::validate()*, or
- modify *helper_parse_id()* to trim the $p_id string before checking its type

vboctor:
@dregad I would prefer having the fix in `helper_parse_id()` since it will apply more broadly across multiple commands.

dregad:
@Kirill I would apply the fix per @vboctor's suggestion right on this PR, but since you used your master branch for it, I'd rather not push to it. Next time I suggest you create a dedicated branch to submit pull requests. 

I'll include the fix as part of #1471, which I'm planning to merge shortly.

PR #1472, opened 2019-02-27
MantisGraph plugin improvements

dregad:
Fixes the following issues 

- [0025522](https://mantisbt.org/bugs/view.php?id=25522) - MantisGraph: limit number of slices in By Category pie chart
- [0025523](https://mantisbt.org/bugs/view.php?id=25523) - MantisGraph: improve handling of colors in Pie charts
- [0025524](https://mantisbt.org/bugs/view.php?id=25524) - MantisGraph: improve display of By Category Bar chart
- Code cleanup

This is built on top of PR #1467, which should be merged first.


cproensa:
I get these errors in browser: console
```
Refused to load the script 'https://cdn.jsdelivr.net/npm/chartjs-plugin-colorschemes@0.3.0/dist/chartjs-plugin-colorschemes.min.js' because it violates the following Content Security Policy directive: "script-src https://cdnjs.cloudflare.com 'self' ajax.googleapis.com maxcdn.bootstrapcdn.com cdnjs.cloudflare.com". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.

plugin_file.php?file=MantisGraph/MantisGraph.js:3 Uncaught TypeError: Cannot set property 'scheme' of undefined
    at HTMLDocument.<anonymous> (plugin_file.php?file=MantisGraph/MantisGraph.js:3)
    at i (jquery.min.js:2)
    at Object.fireWith [as resolveWith] (jquery.min.js:2)
    at Function.ready (jquery.min.js:2)
    at HTMLDocument.J (jquery.min.js:2)
```

cproensa:
And with CDN=OFF, also results in error.
It's caused by case sensitive filenames:
includes: `chart-2.7.3.min.js`
filename is: `Chart-2.7.3.min.js`

cproensa:
> New _get_status_color_by_label()_ helper function to get status color based on label instead of id

if i change language other than english, the status colors are grey. 
Seems like the labels are already translated, and don't match to the keys in `status_colors` config

dregad:
Thanks for testing @cproensa 

> Refused to load the script 'https://cdn.jsdelivr.net/ [...] /chartjs-plugin-colorschemes.min.js' because it violates the following Content Security Policy directive [...]

I forgot to add jsdelivr.net to the CSP exceptions, must have had disabled CSP while testing. Will fix.
> And with CDN=OFF, also results in error. It's caused by case sensitive filenames

I didn't notice that. Will teach me for testing under Windows :confounded: Fix belongs in #1467 though.

> if i change language other than english, the status colors are grey.
Seems like the labels are already translated, and don't match to the keys in `status_colors` config

Will look into it.

dregad:
@cproensa the latest commits and rebase I just pushed should address the 3 issues you've raised.

dregad:
@cproensa did you get a chance to re-test this after my fixes ? I would like to merge before @vboctor cuts the next release sometime this week-end.

cproensa:
> @cproensa the latest commits and rebase I just pushed should address the 3 issues you've raised.

They're working fine now: cdn, file include and status colors.
thanks

Howeer, there's something wrong in this branch with the file `docbook/Developers_Guide/Makefile`, probably that's what failing in the travis tests

dregad:
I think I know what that is - I had an issue with the symlink on my workstation yesterday.

PR #1473, opened 2019-03-02
Fix minor regression with PR 1464

cproensa:
When defining a standard filter that returns "any" issue, don't specify the
project id. Advanced and simple filters both works in the same way,
regarding inclusion of subprojects, as long as the project_id property
is not filled explicitly.

Prevous to PR #1464, the api would include subprojects always, but after
the PR, it would have different behaviour if, for example, having
configured $g_view_filters = ADVANCED_ONLY

Fixes: #25515

### DIFF
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -3171,18 +3171,13 @@ function filter_create_recently_modified( $p_days, $p_filter = null ) {
 }
 
 /**
- * Create a filter for getting issues assigned to the specified project.
- * @param integer $p_project_id The project id or ALL_PROJECTS.
+ * Create a filter for getting any issues without restrictions
  * @return mixed A valid filter.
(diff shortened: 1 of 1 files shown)


cproensa:
@obmsch please have a look to see if it works as expected. 

As explained in https://www.mantisbt.org/bugs/view.php?id=25515, this fix will keep the previous behavior, but does not state if that behavior is good or bad.

dregad:
I'll merge this now without waiting for @obmsch's feedback (sorry). This will ensure the fix gets bundled with upcoming 2.20.0 (together with the commit that introduced the regression), and hopefully avoid having to release 2.20.1 in a few days...

obmsch:
@cproensa thanks, fix works as expected.

PR #1474, opened 2019-03-05
Fix column width in manage_proj_edit_page

cproensa:
This is related to PR #1449 
Clicking on the edit area for user access level makes the column jump in width and it's a bit confusing.
No tracker issue, since hasn't been released yet?

In manage_proj_edit_page.php, in the redesigned section for user
assignments the column width varies when showing and hiding the access
level dropdowns.
By making this column a fixed width, the location of contents is more
predictable.

### DIFF
--- a/js/manage_proj_edit_page.js
+++ b/js/manage_proj_edit_page.js
@@ -83,6 +83,9 @@ function try_hide_input( div ) {
 
 
 $(document).ready( function() {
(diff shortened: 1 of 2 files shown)


cproensa:
> The behavior is indeed much better, but still noticeable on narrower screens. Since there is some extra space in other columns, maybe col-md-4 would further improve things

thanks

i have updated that, and added some more minor fixes

PR #1476, opened 2019-03-07
Add autofocus to the project-list searchbar

keessonnema:
- Add autofocus to the project-list searchbar

We use this in our Mantis application and it saves us some time. Instead of clicking twice you only have to click once to search (open the dropdown list).


keessonnema:
Pushed the wrong branch, fixed in new PR #1477

PR #1477, opened 2019-03-07
Add autofocus to the project-list searchbar

keessonnema:
- Add autofocus to the project-list searchbar
We use this in our Mantis application and it saves some time when searching, so I figured it'd be a good feature.

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -386,7 +386,9 @@ $(document).ready( function() {
 
 	$(document).on('shown.bs.dropdown', '#dropdown_projects_menu', function() {
 		$(this).find(".dropdown-menu li.active a").focus();
-	 });
+		/* Add autofocus support for IE */
+		$('[autofocus]:not(:focus)').eq(0).focus();
+	});
 
 	/**
 	 * Manage visiblity on hover trigger objects
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -602,7 +602,7 @@ function layout_navbar_projects_list( $p_project_id = null, $p_include_all_proje
 	echo '<li>';
 	echo '<div id="projects-list">';
 	echo '<div class="projects-searchbox">';
-	echo '<input class="search form-control input-md" placeholder="' . lang_get( 'search' ) . '" />';
+	echo '<input class="search form-control input-md" autofocus="autofocus" placeholder="' . lang_get( 'search' ) . '" />';
 	echo '</div>';
 	echo '<ul class="list dropdown-yellow no-margin">';
 


js/common.js
```diff
@@ -386,7 +386,9 @@ $(document).ready( function() {
 
 	$(document).on('shown.bs.dropdown', '#dropdown_projects_menu', function() {
 		$(this).find(".dropdown-menu li.active a").focus();
-	 });
+		/* Add autofocus support for IE */
+		$('[autofocus]:not(:focus)').eq(0).focus();
```
dregad: According to Can I use [autofocus attribute is supported since IE 10](https://caniuse.com/#feat=autofocus), and our minimum req is IE11 so I'm not sure this is actually needed.

keessonnema: Sadly Internet Explorer does not support autofocus at all. Despite it saying so in 'Can I Use'. So this is indeed needed.

keessonnema:
I'll create an issue in the tracker after the weekend and make sure to update the commit message with the issue.

keessonnema:
@dregad Updated commit message, fixes file permissions and created the issue in the bug tracker.

cproensa:
I don't favour this change.
This same behaviour was changed to focus on the current selected project
https://www.mantisbt.org/bugs/view.php?id=23248

It's a nuisance when there's a long list of projects, that your current selection is out of view. And the reasoning is also that it breaks the standard behavior of a dropdown list.

This request is very similar to: https://www.mantisbt.org/bugs/view.php?id=23037

I suggest, to render the search box outside of the scrollable area, so it's always visible, and kkep the focus for the current selection in the actual list content. This way, the user doesn't need to scroll to the top to get to the search box.

keessonnema:
@cproensa I agree that the search-field in the scrollable portion is a bad practice in the first place. For the fact that the focused element is outside of the view, sure. Nothing we can do about that right now. We use 300+ projects in our installation and never had problems with scrolling up/down (using this code). Can't this be tested? For us it works perfectly fine.

cproensa:
> We use 300+ projects in our installation and never had problems with scrolling up/down (using this code). Can't this be tested? For us it works perfectly fine.

I don't say there's anything broken about scrolling:

> It's a nuisance when there's a long list of projects, that your current selection is out of view. And the reasoning is also that it breaks the standard behavior of a dropdown list.

keessonnema:
> I believe that the proper solution, that satisfy both request, is to fix the search box outside of the scrollable list. IIRC this is a custom div, so it looks feasible
I could see what I can do to change the layout up a bit to fix the problem (that already exists really) of the search-field going out of the view, and the autocomplete with it.

cproensa:
See this PR: #1478 
That seems to address both of our use cases.

I'm not sure about the actual "autofocus" intent.
Do you mean to focus visually the item?, or also to focus the edit cursor, so you cant type without further clicking on the search box (related: https://www.mantisbt.org/bugs/view.php?id=23037?

js/common.js
dregad: There was also an MSDN article saying that it was supported on IE10. Anyway, I don't use IE so I'll trust you on that one.

dregad:
> It's a nuisance when there's a long list of projects, that your current selection is out of view. And the reasoning is also that it breaks the standard behavior of a dropdown list.

@cproensa Right, I didn't realize that because on trackers I use there are not so many projects.

> This request is very similar to: https://www.mantisbt.org/bugs/view.php?id=23037

Indeed. In that case there would be no need to open a separate one, just reference #23037.

I'll have a look at proposed alternative #1478

keessonnema:
> See this PR: #1478
> That seems to address both of our use cases.
Correct. Allthough I can't test it right now.

> I'm not sure about the actual "autofocus" intent.
> Do you mean to focus visually the item?, or also to focus the edit cursor, so you cant type without further clicking on the search box (related: https://www.mantisbt.org/bugs/view.php?id=23037?

Yes that issue is similar to mine. You could still use my commit with your code as far as I can see. I can reference to the related issue you linked.

dregad:
> You could still use my commit with your code as far as I can see

I tested that, but actually it causes unexpected behavior for a selection list, i.e. it is no longer possible to change selected project in the list via keyboard (up/down/enter).

keessonnema:
Makes sense. We're not using that functionality, so I guess we'll need to stick to custom changes. Now that we're discussing this. Is there a way to use a plugin to override core code? (or would it be possible?) So that I can update without much effort but still keep our custom changes?

cproensa:
@keessonnema  in the other PR i have added the functionality to focus when typing an actual string.

keessonnema:
> @keessonnema in the other PR i have added the functionality to focus when typing an actual string.

Can't really say I like this approach. It's certainly not clear for the user that you can type in a project-name. (the blinking cursor is missing). But I guess it's something getting used to...

keessonnema:
As far as I see, this is not going to happen. So I'm closing this PR.

PR #1478, opened 2019-03-09
Fix projects searchbox outside of scrollable list

cproensa:
In the navbar menu for project selection, fix the search box, and "all
projects" item, outside of the scrollable area for the list of
individual projects.

Fixes: [#23037](https://mantisbt.org/bugs/view.php?id=23037)

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -592,13 +592,6 @@ function layout_navbar_projects_list( $p_project_id = null, $p_include_all_proje
 	$t_project_ids = user_get_accessible_projects( $t_user_id );
 	$t_can_report = true;
 
-	if( $p_include_all_projects && $p_filter_project_id !== ALL_PROJECTS ) {
-		echo ALL_PROJECTS == $p_project_id ? '<li class="active">' : '<li>';
-		echo '<a href="' . helper_mantis_url( 'set_project.php' ) . '?project_id=' . ALL_PROJECTS . '">';
-		echo lang_get( 'all_projects' ) . ' </a></li>' . "\n";
-		echo '<li class="divider"></li>' . "\n";
-	}
-
 	echo '<li>';
 	echo '<div class="projects-searchbox">';
 	echo '<input class="search form-control input-md" placeholder="' . lang_get( 'search' ) . '" />';
@@ -608,6 +601,13 @@ function layout_navbar_projects_list( $p_project_id = null, $p_include_all_proje
 	echo '<li>';
 	echo '<div class="scrollable-menu">';
 	echo '<ul class="list dropdown-yellow no-margin">';
+
+	if( $p_include_all_projects && $p_filter_project_id !== ALL_PROJECTS ) {
+		echo ALL_PROJECTS == $p_project_id ? '<li class="active">' : '<li>';
+		echo '<a href="' . helper_mantis_url( 'set_project.php' ) . '?project_id=' . ALL_PROJECTS . '">';
+		echo lang_get( 'all_projects' ) . ' </a></li>' . "\n";
+		echo '<li class="divider"></li>' . "\n";
+	}
 
 	foreach( $t_project_ids as $t_id ) {
 		if( $p_can_report_only ) {
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -528,7 +528,7 @@ function layout_navbar_projects_menu() {
 	echo ' <i class="ace-icon fa fa-angle-down bigger-110"></i>' . "\n";
 	echo '</a>' . "\n";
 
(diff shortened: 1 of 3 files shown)


dregad:
> Fixes: #25584

Since that's a duplicate of 23037, commit message reference should be adjusted accordingly.

dregad:
Seems fine except that the scrollable div is now bigger, and the bottom is outside of the browser's viewport. Will push a commit adjusting the CSS accordingly.

cproensa:
> After testing, this actually breaks the search functionality

yes, it's good now

cproensa:
> Since that's a duplicate of 23037, commit message reference should be adjusted accordingly.

After the fact, 23037 also mentions "focusing to edit", which is not exactly what it's being done here, an you said that would break the keyboard navigation in the list.
But also may be 0025584, so i'm not sure if this should be an actual fix for any of those

cproensa:
I added focus to the searchbox when a key is pressed (typing a search term), and it's not a list navigation key.
That should cover the original requests too.

cproensa:
>  that the scrollable div is now bigger, and the bottom is outside of the browser's viewport

we could place the "all projects" entry in the scrollable area too, having then one less fixed row,

atrol:
A bit off topic note: While testing, I found that MantisBT can't be no longer used with deactivated JavaScript, as you can't select a project. Didn't have a deeper look, might have been introduced with search functionality.

While you are on it, it would be good if you could check if there is a simple fix for it.

cproensa:
>  with deactivated JavaScript, as you can't select a project

Without js enabled, i can't open the dropdown, in the first place.
But also can't open the user menu, the fot's are not rendered properly, the fa-icons are not rendered...

I think at this point we should require js

keessonnema:
Just to make sure. Is the focus outside of the search-field? I'll quote myself from my original PR:

> Can't really say I like this approach. It's certainly not clear for the user that you can type in a project-name. (the blinking cursor is missing). But I guess it's something getting used to...

This could end up causing more problems that we currently have with searching.

cproensa:
>  Is the focus outside of the search-field? 

The focus is switched based on the keyboard inputs... so the focus is effectively in the active element after each interaction.
Additionally, with the latest commit, the focus is placed in the searchbox by default, when opening the dropdown

atrol:
> the font's are not rendered properly, the fa-icons are not rendered...

@cproensa I don't see this is related to JavaScript. Fonts and fa-icons are working in my environment if JavaScript is disabled. Just tried Safari and Chrome until now.

keessonnema:
> Additionally, with the latest commit, the focus is placed in the searchbox by default, when opening the dropdown

This is fine. Thanks for adding that.

cproensa:
> Fonts and fa-icons are working in my environment if JavaScript is disabled

you're right, i went disabling too much things...
but yes none of the navbar dropdown works without javascript, and this probably exists since v2.0


With this PR, when testing on mobile browser, i find that having the focus in search box when opening the dropdown, makes the phone keyboard to appear on screen. This is bad because it hides most of the screen, and my intention in the phone is primarily to browse the list, instead of enter with the keyboard, so i have to explcitly close the keyboard everytime i open the project list.

cproensa:
So, reconsidering the issue with the search box focus that degrades the mobile interaction, i have removed that part, so it's reverted to initially focus to the list. However, the keyboard events will still switch the focus between the searchbox and the list.
Added some more details to the script. The latest commit resumes it all:

    Changes and clean up for the navbar project list dropdown:
    
    - When the dropdown is displayed:
      - Apply focus to the active list element to position the
        scrollable area over that element.
      - If the list is empty, apply focus to the searchbox.
    
    - Manage key events in the dropdown elements:
      - When using the list and any key not related to list navigation is
        pressed (for example, typing a text for searching), switch the focus
        to the search box for it to receive the input.
      - When using the searchbox and any key of up, down, pageup, pagedown,
        is pressed, switch the focus to the list to allow navigating with
        keyboard.
      - Escape key in the search box will close the dropdown.
    
    - Fix hiding the searchbox when only 10 or less items are displayed.
    
    - Code clean-up:
      - Move all the code related to the navbar projects menu to the same
        place in the js file.
      - Rename variables for less probability of collision with other
        unrelated code.

thE-iNviNciblE:
@Mantis Projekt Navigation needed for a CIO or Projektmanager: 

i've another feature request, maybe i could contribute source to mantis

Desktop-Only feature: https://bludau-media.de/de/1/Blog-von-Bludau-Media/ (look that the left side, the arrow you can mouseover it, and a menü will pop-out.

https://www.mantisbt.org/bugs/view.php?id=25535

PR #606, opened 2015-05-11
Fixes #19642: If log file is not writable, log_event() fails silently

kevin556:
Added a warning in log_event() when the log file
is not writable and prevented from calling error_log with
the same not writable file which caused the page to crash.

PR #1483, opened 2019-03-21
Display error when Mantis log file is not writable

dregad:
Replaces PR #606

Fixes [#19642](https://mantisbt.org/bugs/view.php?id=19642)

Here's how it looks...
image


dregad:
This PR is built on top of #1482, which should be merged first.

dregad:
Thanks for the feedback @vboctor. 

I'm not sure I understand what kind of testing you expect me to do, and what you mean with needing exceptions, as the new code does not trigger any errors (see [changes in logging_api.php](https://github.com/mantisbt/mantisbt/pull/1483/files#diff-badb9a686244714227de05948a79b3bcR122)), it just adds a message to the queue for display at page bottom, and prints it to the system log as well (which would be how the problem would be logged in the case of an API call).

Are you saying that I should instead be throwing an exception to interrupt the execution flow ? I can certainly do that, but my thinking was that a failure to log is not a blocking error.

dregad:
@vboctor you have not responded to https://github.com/mantisbt/mantisbt/pull/1483#issuecomment-475529665.

dregad:
@vboctor without feedback from you, I'll go ahead and merge this coming week-end.

vboctor:
@dregad The thing I wanted to make sure doesn't happen is that if:

- logging is enabled.
- Log file is not writable.
- An API call is made

The result shouldn't be that execution stops, but API is successful.  This may not be an issue due to different error handlers, but it is worth making sure that is the case.

dregad:
@vboctor thanks for your feedback, I understand what you mean now. Will test and confirm that the system behaves per your expectation.

dregad:
@vboctor the test is **successful** - the system behaves as per your expectations, just as I thought it would.

## Testing scenario

1. set config_inc.php
    ```php
    $g_log_level = LOG_EMAIL;
    $g_log_destination = 'file:/tmp/mantisbt.log';
    ```
2. Make an API call: add a bugnote (note: assumes that the action would actually trigger an e-mail notification, e.g. the issue's reporter is not the same as the user making the call)
    ```bash
    curl -sS -o /dev/null -w "Status: %{response_code}\n" --request POST \
      --url https://path.to/mantisbt/api/rest/issues/1/notes \
      --header 'Authorization: {{token}}' \
      --header 'Content-Type: application/json' \
      --data '{"text": "test note", "view_state": {"name": "public"}}'
    ```
3. Output is **Status: 201**, a new bugnote has been added to the issue and the MantisBT log file shows:
    ```
    2019-04-01 12:21 UTC MAIL email_api.php:980 email_bugnote_add() Note ~31 added to issue #1
    ```
4. Make MantisBT log-file read-only 
    ```bash
    chmod 444 /tmp/mantisbt.log
    ```
5. Repeat the same API call
6.  Output is **Status: 201**, a new bugnote has been added to the issue, nothing was added to the MantisBT log file, and the PHP error log shows:
    ```
    [01-Apr-2019 12:27:55 UTC] MantisBT - The file specified in $g_log_destination "/tmp/mantisbt.log" is not writable.
    ```

vboctor:
Thanks @dregad for testing this.

PR #1484, opened 2019-03-22
Status color boxes should not have background css set

dregad:
Changes:
- add new html_get_status_css_fg() / html_get_status_css_bg() functions
- define new, distinct status color classes (_status-XX-fg_ and _status-XX-bg_) to replace the single _status-XX-color_ used previously
- update MantisBT pages to use the new classes / API functions
- deprecate html_get_status_css_class() (could be used by 3rd party plug-ins)
- code cleanup (renaming variables)

Fixes [#23550](https://mantisbt.org/bugs/view.php?id=23550)

Note: this PR is build on top of #1482, which should be merged first.

### DIFF
--- a/account_sponsor_page.php
+++ b/account_sponsor_page.php
@@ -186,9 +186,9 @@ if( $t_sponsor_count === 0 ) {
 		}
 
 		# choose color based on status
-		$t_status_label = html_get_status_css_class( $t_bug->status, auth_get_current_user_id(), $t_bug->project_id );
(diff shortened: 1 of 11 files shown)


atrol:
Contains changes that are not related to the issue, at least core/error_api.php and core/layout_api.php

dregad:
> Contains changes that are not related to the issue, at least core/error_api.php and core/layout_api.php

This is correct, but intentional. As mentioned in the PR's opening text 
_Note: this PR is build on top of #1482, which should be merged first._

This is because the last commit *Deprecate html_get_status_css_class()* requires those changes otherwise the deprecation error is not displayed.

PR #1486, opened 2019-03-22
Simplify Configuration Report page

cproensa:
To save page space, don't show complex values in adm_config_report page. Use an on-demand retrieval of the content through a rest api call.
Create a new page for viewing and editing configuration options from adm_config_report page. This also serves as fallback for viewing complex values that now are not displayed in the main report, in case javascript is not available to populate it through rest api calls.

### DIFF
--- a/js/adm_config_report.js
+++ b/js/adm_config_report.js
@@ -1,0 +1,49 @@
+/*
+# Mantis - a php based bugtracking system
+
+# Copyright 2000 - 2002  Kenzaburo Ito - kenito@300baud.org
+# Copyright 2002 MantisBT Team   - mantisbt-dev@lists.sourceforge.net
+
+# Mantis is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# Mantis is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Mantis.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+$(document).ready( function() {
+	$('.adm_config_expand .expand_hide').hide().removeClass('hidden');
+	$('.adm_config_expand .expand_hide .toggle')
+			.click(function(event) {
+				event.preventDefault();
+				var container = $(this).closest('.adm_config_expand');
+				container.find('.expand_hide').hide();
+				container.find('.expand_show').show();
+				container.find('.expand_content').hide();
+			});
+	$('.adm_config_expand .expand_show .toggle')
+			.click(function(event) {
+				event.preventDefault();
+				var container = $(this).closest('.adm_config_expand');
+				container.find('.expand_hide').show();
+				container.find('.expand_show').hide();
+				// Load contents if not already done and display
+				var content = container.find('.expand_content');
+				if (content[0].childElementCount == 0) {
+					var req_data = {
+						'config_id': container.data('config_id'),
+						'project_id': container.data('project_id'),
+						'user_id': container.data('user_id')
+					};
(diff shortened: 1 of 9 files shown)


js/adm_config_report.js
```diff
@@ -0,0 +1,45 @@
...
+				var container = $(this).closest('.adm_config_expand');
+				container.find('.expand_hide').show();
+				container.find('.expand_show').hide();
+				var req_data = {
+					'config_id': container.data('config_id'),
+					'project_id': container.data('project_id'),
+					'user_id': container.data('user_id')
+				};
+				container.find('.expand_content')
+						.load('api/rest/internal/config_display',req_data);
```
dregad: Avoid unnecessary requests, only call the AJAX once. Something like

```JavaScript
		// Load contents if not already done and display
                var content = container.find('.expand_content');
		if (content[0].childElementCount == 0) {
                    var req_data = {
                        'config_id': container.data('config_id'),
                        'project_id': container.data('project_id'),
                        'user_id': container.data('user_id')
                    };
                    content.load('api/rest/internal/config_display', req_data);
                }
                content.show();
```

cproensa:
> Maybe now would be a good time to implement an alternative to this huge and useless select ?

What is your suggestion? Pagination, or other approach?

dregad:
What about a text field with lookup-as-you-type, similar to https://jqueryui.com/autocomplete/ or what GitHub does when you type `@` ?

cproensa:
>  only call the AJAX once. 

done

> (> 35K users) on a slow machine, and adm_config_page took more than 45 seconds to load on average...

with that perspective, i have removed the commit to make this a sortable table (on client side)
I have kept, however, commit 4200d3833734f93832e825350fdb9463658de936 (Support non sorting columns in sortable tables), since it's a useful addition to that script.

cproensa:
> What about a text field with lookup-as-you-type, similar to https://jqueryui.com/autocomplete/ or what GitHub does when you type @ ?

That's nice, but i'm not sure how to factor having multiple option values, per user, per project, etc.
Would you open an issue to track this, and discuss better in detail the possible UIs for that?

dregad:
> removed the commit to make this a sortable table 

Not sure that was necessary, as the performance issue was caused by loading the user selection list (i.e. in adm_config_page.php), not with the main page with the list. Anyway I'll re-test with the new code.

> I have kept, however, commit 4200d38 (Support non sorting columns in sortable tables), since it's a useful addition to that script.

Probably good to reference an issue ID in this commit, either the one where you introduced sortable tables, or a new one if you consider this as a new feature.

> Would you open an issue to track this, and discuss better in detail the possible UIs for that?

Done https://www.mantisbt.org/bugs/view.php?id=25666

cproensa:
> Done https://www.mantisbt.org/bugs/view.php?id=25666

i thought you meant to replace the whole page for "configuration report" with a different UI, not just the user selector input.
Because, currently this is a whole page for all the config, and can be considerably large. That's why i asked you about pagination.

>> removed the commit to make this a sortable table

> Not sure that was necessary, as the performance issue was caused by loading the user selection list (i.e. in adm_config_page.php), not with the main page with the list. Anyway I'll re-test with the new code.

Ok. There is processing for this script is done in client side, so it's impacted by having a big table on page. 
This approach is a quick way to have sortable tables without server side support, suitable for manage pages, etc. 
We could put a hard limit on it to ignore tables with > N rows (1000?), to play safe.

dregad:
@cproensa I was just about to merge this, but noticed that the PR's commits do not no reference any issue in the bugtracker. Is there one ? If not, could you please create it ?

Cheers

dregad:
> @cproensa I was just about to merge this, but noticed that the PR's commits do not no reference any issue in the bugtracker. Is there one ? If not, could you please create it ?

ping

cproensa:
I have reintroduced the sorting for the table, but adding a hard limit on the number of rows to the script.

>  I was just about to merge this, but noticed that the PR's commits do not no reference any issue in the bugtracker. Is there one ?

i have created [25910](https://www.mantisbt.org/bugs/view.php?id=25910), [25911](https://www.mantisbt.org/bugs/view.php?id=25911) to document both the main change for the page and the script modifications

PR #1487, opened 2019-03-23
Add ProjectHierarchyGetCommand and REST API

obmsch:
Add ProjectHierarchyGetCommand, extend the project hierarchy rest api and use the new command in GET project to retrieve the list of subprojects.

Fixes #25514

### DIFF
--- a/core/commands/ProjectHierarchyGetCommand.php
+++ b/core/commands/ProjectHierarchyGetCommand.php
@@ -1,0 +1,156 @@
+<?php
+# MantisBT - A PHP based bugtracking system
+
+# MantisBT is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# MantisBT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+require_api( 'constant_inc.php' );
+require_api( 'config_api.php' );
+require_api( 'helper_api.php' );
+require_api( 'authentication_api.php' );
+require_api( 'user_api.php' );
+require_api( 'project_api.php' );
+require_api( 'project_hierarchy_api.php' );
+
+use Mantis\Exceptions\ClientException;
+
+/**
+ * A command that gets project subprojects from the project hierarchy.
+ */
+class ProjectHierarchyGetCommand extends Command {
+
+	/**
+	 * @var integer
+	 */
+	private $project_id;
+
+	/**
+	 * @var boolean
+	 */
+	private $all_descendents;
+
+	/**
+	 * @var boolean
+	 */
+	private $include_disabled;
+
+	/**
+	 * return an array of user accessible subprojects of a project (recurse if requested)
+	 * @param integer $p_project_id	A valid project identifier.
+	 * @param boolean $p_all_descendents Include only childs or all descendents.
+	 * @param boolean $p_include_disabled Include disabled projects in the resulting array.
+	 * @param array $p_subprojects The current array of subprojects.
+	 * @return array
+	 */
+	private function get_subprojects( $p_project_id, $p_all_descendents, $p_include_disabled, $p_subprojects  = array() ) {
+		$t_subprojects = $p_subprojects;
+		$t_subproject_ids = user_get_accessible_subprojects(
+			auth_get_current_user_id(), $p_project_id, $p_include_disabled );
+		foreach( $t_subproject_ids as $t_subproject_id ) {
+			$t_subprojects[] = array(
+				'id' => $t_subproject_id,
+				'name' => project_get_name( $t_subproject_id ),
+				'parent_id' => $p_project_id,
+				'inherit_parent' => project_hierarchy_inherit_parent(
+					 $t_subproject_id, $p_project_id, $p_include_disabled )
+			);
+
+			if( true === $p_all_descendents ) {
+				$t_subprojects += $this->get_subprojects(
+					$t_subproject_id, $p_all_descendents, $p_include_disabled, $t_subprojects );
+			}
+		}
+
+		return $t_subprojects;
+	}
+
+	/**
+	 * Constructor
+	 *
+	 * $p_data['query'] is expected to contain:
+	 * - project_id (integer)
+	 *
+	 * $p_data['payload'] is expected to contain:
+	 * - all_descendents (boolean)
+	 * - include_disabled (boolean)
+	 *
+	 * @param array $p_data The command data.
+	 */
+	function __construct( array $p_data ) {
+		parent::__construct( $p_data );
+	}
+
+	/**
+	 * Validate the data.
+	 */
+	function validate() {		
+		if ( config_get( 'subprojects_enabled' ) == OFF ) {
+			throw new ClientException(
+				'Project hierarchy (subprojects) is disabled',
+				ERROR_PROJECT_HIERARCHY_DISABLED );
+		}
+
+		$this->project_id = helper_parse_id( $this->query( 'project_id' ), 'project_id' );
+		if( !project_exists( $this->project_id ) ) {
+			throw new ClientException(
+				"Project '$this->project_id' not found",
+				ERROR_PROJECT_NOT_FOUND,
+				array( $this->project_id ) );
+		}
+
+		$this->all_descendents = $this->payload( 'all_descendents' );
+		if( !isset( $this->all_descendents ) ) {
+			throw new ClientException(
+				"'all_descendents' not provided",
+				ERROR_EMPTY_FIELD,
+				array( 'all_descendents' ) );
+		} else if( !is_bool( $this->all_descendents ) ) {
+			throw new ClientException(
+				"Invalid value for 'all_descendents', not bool",
+				ERROR_INVALID_FIELD_VALUE,
+				array( 'all_descendents' ) );
+		}
+
+		$this->include_disabled = $this->payload( 'include_disabled' );
+		if( !isset( $this->include_disabled ) ) {
+			throw new ClientException(
+				"'include_disabled' not provided",
+				ERROR_EMPTY_FIELD,
+				array( 'include_disabled' ) );
+		} else if( !is_bool( $this->include_disabled ) ) {
+			throw new ClientException(
+				"Invalid value for 'include_disabled', not bool",
+				ERROR_INVALID_FIELD_VALUE,
+				array( 'include_disabled' ) );
+		}
+	}
+
+	/**
+	 * Process the command.
+	 *
+	 * @returns array Command response
+	 */
+	protected function process() {
+		if( $this->project_id != helper_get_current_project() ) {
+			# in case the current project is not the same project of the bug we are
+			# viewing, override the current project. This to avoid problems with
+			# categories and handlers lists etc.
+			global $g_project_override;
+			$g_project_override = $this->project_id;
+		}
+
+		return $this->get_subprojects(
+			 $this->project_id, $this->all_descendents, $this->include_disabled );
+	}
+}
+
--- a/api/rest/restcore/projects_rest.php
+++ b/api/rest/restcore/projects_rest.php
@@ -41,6 +41,8 @@ $g_app->group('/projects', function() use ( $g_app ) {
 	$g_app->post( '/{id}/versions/', 'rest_project_version_add' );
 
 	# Project hierarchy (subprojects)
+	$g_app->get( '/{id}/subprojects', 'rest_project_hierarchy_get' );
+	$g_app->get( '/{id}/subprojects/', 'rest_project_hierarchy_get' );
 	$g_app->post( '/{id}/subprojects', 'rest_project_hierarchy_add' );
 	$g_app->post( '/{id}/subprojects/', 'rest_project_hierarchy_add' );
 	$g_app->patch( '/{id}/subprojects/{subproject_id}', 'rest_project_hierarchy_update' );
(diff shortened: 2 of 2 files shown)


cproensa:
I'm not familiar with the new Commands scenario, but does it really need a Command for something that does not make changes?

Also, i have proposed #1480 that is making direct changes in project hierarchy.

core/commands/ProjectHierarchyGetCommand.php
```diff
@@ -0,0 +1,156 @@
...
+		$this->project_id = helper_parse_id( $this->query( 'project_id' ), 'project_id' );
+		if( !project_exists( $this->project_id ) ) {
+			throw new ClientException(
+				"Project '$this->project_id' not found",
+				ERROR_PROJECT_NOT_FOUND,
+				array( $this->project_id ) );
+		}
+
+		$this->all_descendents = $this->payload( 'all_descendents' );
+		if( !isset( $this->all_descendents ) ) {
```
vboctor: When fields are not specified, use a default rather than error out.  I think we can have a reasonable default for `recursive` and `include_disabled`.

core/commands/ProjectHierarchyGetCommand.php
```diff
@@ -0,0 +1,156 @@
...
+		return $t_subprojects;
+	}
+
+	/**
+	 * Constructor
+	 *
+	 * $p_data['query'] is expected to contain:
+	 * - project_id (integer)
+	 *
+	 * $p_data['payload'] is expected to contain:
```
vboctor: Since this is modeled as a GET, I would expect the fields to show up in the query rather than the payload.

core/commands/ProjectHierarchyGetCommand.php
```diff
@@ -0,0 +1,156 @@
...
+	 * @param integer $p_project_id	A valid project identifier.
+	 * @param boolean $p_all_descendents Include only childs or all descendents.
+	 * @param boolean $p_include_disabled Include disabled projects in the resulting array.
+	 * @param array $p_subprojects The current array of subprojects.
+	 * @return array
+	 */
+	private function get_subprojects( $p_project_id, $p_all_descendents, $p_include_disabled, $p_subprojects  = array() ) {
+		$t_subprojects = $p_subprojects;
+		$t_subproject_ids = user_get_accessible_subprojects(
+			auth_get_current_user_id(), $p_project_id, $p_include_disabled );
```
vboctor: How about adding `$p_user_id` as a parameter rather than getting it each time?

core/commands/ProjectHierarchyGetCommand.php
```diff
@@ -0,0 +1,156 @@
...
+		foreach( $t_subproject_ids as $t_subproject_id ) {
+			$t_subprojects[] = array(
+				'id' => $t_subproject_id,
+				'name' => project_get_name( $t_subproject_id ),
+				'parent_id' => $p_project_id,
+				'inherit_parent' => project_hierarchy_inherit_parent(
+					 $t_subproject_id, $p_project_id, $p_include_disabled )
+			);
+
+			if( true === $p_all_descendents ) {
```
vboctor: Do we really need `true ===` rather than just checking of the parameter is truthy?

core/commands/ProjectHierarchyGetCommand.php
```diff
@@ -0,0 +1,156 @@
...
+	 * @param boolean $p_include_disabled Include disabled projects in the resulting array.
+	 * @param array $p_subprojects The current array of subprojects.
+	 * @return array
+	 */
+	private function get_subprojects( $p_project_id, $p_all_descendents, $p_include_disabled, $p_subprojects  = array() ) {
+		$t_subprojects = $p_subprojects;
+		$t_subproject_ids = user_get_accessible_subprojects(
+			auth_get_current_user_id(), $p_project_id, $p_include_disabled );
+		foreach( $t_subproject_ids as $t_subproject_id ) {
+			$t_subprojects[] = array(
```
vboctor: Since we are filtering based on `enabled` flag, does it make sense to include it, specially if `include_disabled` is true.

core/commands/ProjectHierarchyGetCommand.php
```diff
@@ -0,0 +1,156 @@
...
+				'id' => $t_subproject_id,
+				'name' => project_get_name( $t_subproject_id ),
+				'parent_id' => $p_project_id,
```
vboctor: I would make parent an array with `id` and `name`.  In the REST API, I have been avoiding just having ids on their own.

core/commands/ProjectHierarchyGetCommand.php
```diff
@@ -0,0 +1,156 @@
...
+require_api( 'user_api.php' );
+require_api( 'project_api.php' );
+require_api( 'project_hierarchy_api.php' );
+
+use Mantis\Exceptions\ClientException;
+
+/**
+ * A command that gets project subprojects from the project hierarchy.
+ */
+class ProjectHierarchyGetCommand extends Command {
```
vboctor: I assume that a project X can be returned multiple times with different parents in the hierarchy.

vboctor:
> I'm not familiar with the new Commands scenario, but does it really need a Command for something that does not make changes?

@cproensa I haven't been doing it partially relying on SOAP API helper functions.  However, I think it is a good practice to have a command for each REST API.  If we don't have such commands, the same functionality will have to live elsewhere which won't be consistent across all APIs.

cproensa:
> @cproensa I haven't been doing it partially relying on SOAP API helper functions. However, I think it is a good practice to have a command for each REST API. If we don't have such commands, the same functionality will have to live elsewhere which won't be consistent across all APIs

i don't know, as i haven't been looked the commands in depth.
I had the impression that commands would be some common, atomic, operation that would be common to web UI, rest api, etc.
As far as i can see, the output of this command is specifically written for the rest api.
And i don't see any benefit over calling directly current core apis and format the result within the rest api.

vboctor:
> I had the impression that commands would be some common, atomic, operation that would be common to web UI, rest api, etc.
> As far as i can see, the output of this command is specifically written for the rest api.
And i don't see any benefit over calling directly current core apis and format the result within the rest api.

Commands are designed after REST API patterns, but it is re-used from other consumers like SOAP API and Web UI.  They are also achieve a single logical goal that can trigger corresponding plugin events, notifications, logs, trigger webhooks, etc.

Once a command is available, the consumer's job is to format the input of the command, run the command, and format the output.  This applies to REST API methods, web pages, etc.  That should be the only logic available in these calling methods.  They shouldn't even get into validation of these values.

As functionality is modeled as Commands, it is easy to expose them as REST APIs that are used by javascript to call into backend to get work done.  i.e. you can imagine more ajax functionality building on these rather than round tripping to the server for action page pattern.

cproensa:
As another note:
I am thinking that disabled projects should not be published. Form a user point of view, disabled projects don't exist and should not be operated upon. That's what i infer from the current behaviour in the web ui.
If you need to inspect disabled projects, probably it's from the point of view of an administrator (actual hierarchy), and those request/methods for administration purposes should not be mixed with plain user (visible hierarchy) client methods

vboctor:
Good point @cproensa 

I think we shouldn't list disabled projects.  This will also avoid cases where a disabled project has children that are enabled and showing up in the results as enabled.  Hence, client having to do a bunch of work to get the correct list.

cproensa:
After working with the hierarchy refactor, iirc, the only places we are (currently) using any call that needs disabled projects are:
- manage projects, pages and api commands
- category and version related stuff, for computing the inherited versions from parents, etc.

One of my goals in PR 1480 is to remove the "disabled" parameter from the hierarchy/user-get-projects group of functions. That would need still a second iteration of clean up. But my impression is that disabled projects, aside from hierarchy management, are only needed by the most internal core. 

Similar issue for returning a "inherit parent" value in the rest api.
Would you need the raw property, aside from a hierarchy management scenario?


> This will also avoid cases where a disabled project has children

That point is open to discussion here: [0025563](https://www.mantisbt.org/bugs/view.php?id=25563)
From the fact that we allow multiple parents, a subproject of a disbaled project should be still usable.

obmsch:
@vboctor thanks for the review, will tackle your comments after we agree on common ground, especially on the questions/issue raised by @cproensa.

From my point of view the rest api is an alternative way (to the WebUI) for a user to access mantisbt and its functionality. The rest client is nothing else, but a 'normal' mantisbt user, with all access restrictions that user has applied. In principle all operations this user has access to in the WebUI, should be available for him through the rest api as well. This doesn't necessarily mean that the representation of data needs to be the same, nor that the rest api doesn't require other/additional steps to achieve the same result.

All that said, I wouldn't care how we achieve it in this case, but at least a rest api client with sufficient rights should have means to retrieve the complete project hierarchy of any project, including disabled ones. Or in other words, this rest client should be able to present its own view of the 'manage_proj_page' out of the delivered data.

cproensa:
@obmsch 
i think i see it clearer now with your comment.
All those api routes in projects_rest.php were added as intended to project _administration_. In such case:
If you want to follow 'manage_proj_page', the returned projects should be only the projects manageable* by the user. In this context, there is no need to filter out disabled projects, imo.

I think, and what led me to my initial confusion, is that 'rest_projects_get' is intended to return visible projects for a user, which is a different context than the administration of projects.
Still, my suggestion goes into having separate queries for "visible projects" and "pure hierarchy", and each one would return a different set of information, and check different set of permissions.
...or... should all the hierarchy related api stuff go into a specific /hierarchy route?

*note: currently, a user with 'manage project' permission at project level, but not at global level, gets an empty listing in _manage_proj_page_, but can effectively manage the project from _manage_proj_edit_page_. This should be fixed (show those projects in the list) at some time.

vboctor:
> From the fact that we allow multiple parents, a subproject of a disbaled project should be still usable.

@cproensa The scenario that I was referring to is the following:

Example 1:

- Project A -> Project B (disabled) -> Project C
- Project A -> Project D

In this case I would include Project A and D.

Example 2:

- Project A -> Project B (disabled) -> Project C
- Project A -> Project D -> Project C

In this case, A, C, D should be included.

core/commands/ProjectHierarchyGetCommand.php
```diff
@@ -0,0 +1,156 @@
...
+	 */
+	function __construct( array $p_data ) {
+		parent::__construct( $p_data );
+	}
+
+	/**
+	 * Validate the data.
+	 */
+	function validate() {		
+		if ( config_get( 'subprojects_enabled' ) == OFF ) {
```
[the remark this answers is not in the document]
obmsch: This might be a semantic nitpick, but `subprojects_enabled` doesn't tell if there are subprojects at all. Perhaps created and never cleaned up before subprojects were disabled. I'd rather stick with erroring out than returning anything not matching current state.

core/commands/ProjectHierarchyGetCommand.php
```diff
@@ -0,0 +1,156 @@
...
+
+		$this->project_id = helper_parse_id( $this->query( 'project_id' ), 'project_id' );
+		if( !project_exists( $this->project_id ) ) {
+			throw new ClientException(
+				"Project '$this->project_id' not found",
+				ERROR_PROJECT_NOT_FOUND,
+				array( $this->project_id ) );
+		}
+
+		$this->all_descendents = $this->payload( 'all_descendents' );
```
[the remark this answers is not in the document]
obmsch: Ok

core/commands/ProjectHierarchyGetCommand.php
obmsch: No, I disagree. The command should operate on well defined params. It's the command callers responsibility to ensure reasonable defaults.

core/commands/ProjectHierarchyGetCommand.php
obmsch: Ok

core/commands/ProjectHierarchyGetCommand.php
obmsch: Ok

core/commands/ProjectHierarchyGetCommand.php
obmsch: It's a boolean (true|false), truethy is a nightmare.

core/commands/ProjectHierarchyGetCommand.php
obmsch: Ok

core/commands/ProjectHierarchyGetCommand.php
obmsch: Yes, anything wrong about this?

cproensa:
> Example 1:
    Project A -> Project B (disabled) -> Project C
    Project A -> Project D
> In this case I would include Project A and D.

I think this specific examples should be discussed as part of: [0025563](https://www.mantisbt.org/bugs/view.php?id=25563)
However, in your example:
I a user has access only to project C, then that project will be visible.
Why should C not be included if the user also had access to A and B?

dregad:
> Fixes #25514

@obmsch I think you referenced the wrong issue Id in the PR's opening note, as well as in the original commit message... [#25614](https://www.mantisbt.org/bugs/view.php?id=25614) is about _Missing file (api/rest/web.config) in installer_ maybe you meant [#25641](https://www.mantisbt.org/bugs/view.php?id=25641) ?

core/commands/ProjectHierarchyGetCommand.php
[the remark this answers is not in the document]
obmsch: Perhaps an alternative might be to omit the check for `subprojects_enabled` and just return anything in there. This would be most likely `array()` anyway.

obmsch:
@dregad no, #25514 is correct.

obmsch:
I'd really like to move this forward. I tried to cover all possible cases in the command. That might be to much (not needed at all). But a rest api client should have means to acquire the complete hierarchy of a project he has access to, independently from there state (enabled|disabled).
As I have added an enabled flag to the subproject list and if the command is only intended to be used from rest, I wouldn't mind to drop `recursive` and `include_disabled`, if the command always returns a complete list of subprojects (recursive, including disabled). But I still think, the pr in its current state is the better solution (more flexible, wider scope).

cproensa:
I still have doubts in:
- `/projects/`GET is intended to return visible information about the project, for _any_ user. Having specific information about hierarch, like "inherit_parents", and "inherit global categories (which should also be included somewhere, to be compelte as your requeriment _"this rest client should be able to present its own view of the 'manage_proj_page' out of the delivered data"_ )
- `/subprojects/`GET, if we assume it's directly related to the actual project hierarchy data, it should be restricted by manage project permissions, just like the web UI.
- i can't ses as an optimum solution to return a single, one dimension, array for _all_ subprojects. Note that a project with multiple parents may appear duplicated in this array. That's not wrong, but still needs a more complex processing by the client as opposed as (IIRC) the soap api that builds a nested, recursive array, with the actual tree representation.

I can't say how to solve those issues, since it seems we need to return different kind of project related information targeted for different purposes: visibility/public data vs hierarchy/administration data.

obmsch:
@cproensa leaving aside for a moment the question of representation (of the delivered data), your doubts begs the question what data any GET should deliver. In a broader sense this might be a security issue and we need to review all of them carefully.
In case of project/subproject GET accessibility is twofold:
- Visibility, what information the user can get
- Manageability, what operations the user is allowed to perform on the subject

From my point of view the latter rules over the former. For the subproject GET we could add the same check (access_has_project_level) as for add/delete/update in the command to achieve this. The project GET should not contain 'subprojects' in this case at all.

dregad:
> @dregad no, #25514 is correct.

Nevermind me, just now I realize that I had a typo in the bug ID, and I was actually looking at #25614 :astonished:

PR #1488, opened 2019-03-23
Handle disabled projects in user_get_all_accessible_(sub)projects

obmsch:
Add param `$p_show_disabled = false` and propagate. Preparation of fix for #25639.

Fixes #25637


dregad:
@vboctor @cproensa any objections to me merging this ?

cproensa:
I think we have an open discussion in PR #1487 that can lead to this being unnecessary, or not...

Otherwise, this change as is, is correct, so no objections if approved.

dregad:
> I think we have an open discussion in PR #1487 that can lead to this being unnecessary, or not...

OK, so let's delay merging until something comes out of your discussion then.

PR #1490, opened 2019-03-23
Fix missing boxes in my view page

cproensa:
In My View page, some boxes are discarded if the user don't have
certain permissions. Those checks are performed against the current
project only, hoewever, the boxes may show issues from subprojects
where the user can have those permissions.

With this change, the checks are made for the actual list of projects
that will be included in the filters for those sections.

Fixes: #23333

### DIFF
--- a/my_view_page.php
+++ b/my_view_page.php
@@ -89,7 +89,20 @@ $t_boxes = config_get( 'my_view_boxes' );
 asort( $t_boxes );
 reset( $t_boxes );
 
-$t_project_ids_to_check = $t_current_project_id == ALL_PROJECTS ? null : array( $t_current_project_id );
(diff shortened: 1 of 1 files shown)


cproensa:
> As you are cleaning up sub-projects scenarios. If there are queries that can be cheaper when subprojects are off, please leverage the config option.

is that comment intended for this PR?

vboctor:
@cproensa it is just a general note, not specific to this PR.  However, if you take this PR as an example, the code to support sub-projects is more involved than the one before.  It calls methods like `filter_get_included_projects` and others.  If some of these methods generate queries that can be short-circuited when sub-projects are disabled, then it is worth doing so.

PR #1491, opened 2019-03-28
Added an ID to the row of a custom field

theStudent001:


cproensa:
@theStudent001 
Can you review my comment for issue [25654](https://www.mantisbt.org/bugs/view.php?id=25654#c61790)
as it seems an id is already present in input fields, and the approach to add ids to TRs, and only in custom fields, would be at least inconsistent.

dregad:
Closing PR as explained in https://mantisbt.org/bugs/view.php?id=25654#c61822

PR #1492, opened 2019-03-28
Fix show obsolete versions

cproensa:
Fix version_get_all_rows for a default parameter value of "null".
Regression caused by 46c77ace319216aa6403af4130b83b4deb801b3c

Fixes: [#25661](https://www.mantisbt.org/bugs/view.php?id=25661)

### DIFF
--- a/core/version_api.php
+++ b/core/version_api.php
@@ -492,7 +492,11 @@ function version_remove_all( $p_project_id ) {
  * Return all versions for the specified project or projects list
  * Returned versions are ordered by reverse 'date_order'
  * @param integer|array $p_project_ids  A valid project id, or array of ids
- * @param boolean $p_released   Whether to include released versions.
+ * @param boolean $p_released   Whether to show only released, unreleased, or both.
+ *                  For this parameter, use constants defined as:
+ *                  VERSION_ALL (null): returns any
+ *                  VERSION_FUTURE (false): returns only unreleased versions
+ *                  VERSION_RELEASED (true): returns only released versions
  * @param boolean $p_obsolete   Whether to include obsolete versions.
  * @param boolean $p_inherit    True to include versions from parent projects,
  *                              false not to, or null to use configuration
@@ -530,7 +534,7 @@ function version_get_all_rows( $p_project_ids, $p_released = null, $p_obsolete =
 		if( !empty( $g_cache_versions_project[$t_project_id]) ) {
 			foreach( $g_cache_versions_project[$t_project_id] as $t_id ) {
 				$t_version_row = version_cache_row( $t_id );
-				if( $p_obsolete == false && (int)$t_version_row['obsolete'] == 1 ) {
+				if( $p_obsolete === false && (int)$t_version_row['obsolete'] == 1 ) {
 					continue;
 				}
 				if( $p_released !== null ) {
--- a/core/version_api.php
+++ b/core/version_api.php
@@ -577,7 +577,7 @@ function version_get_id( $p_version, $p_project_id = null, $p_inherit = null ) {
 		$c_project_id = (int)$p_project_id;
 	}
 
-	$t_versions = version_get_all_rows( $c_project_id, null /* released: any */, true /* incl. obsolete */ ,  $p_inherit );
+	$t_versions = version_get_all_rows( $c_project_id, VERSION_ALL, true /* incl. obsolete */ ,  $p_inherit );
 	foreach( $t_versions as $t_version ) {
 		if( $t_version['version'] === $p_version ) {
 			return $t_version['id'];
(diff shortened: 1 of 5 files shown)


cproensa:
consider applying this to 2.20.x branch as well?

core/version_api.php
```diff
@@ -530,7 +530,7 @@ function version_get_all_rows( $p_project_ids, $p_released = null, $p_obsolete =
 		if( !empty( $g_cache_versions_project[$t_project_id]) ) {
 			foreach( $g_cache_versions_project[$t_project_id] as $t_id ) {
 				$t_version_row = version_cache_row( $t_id );
-				if( $p_obsolete == false && (int)$t_version_row['obsolete'] == 1 ) {
+				if( $p_obsolete === false && (int)$t_version_row['obsolete'] == 1 ) {
```
atrol: Didn't test, but isn't the real problem that there are places where `version_get_all_rows` is called with `null` for parameter `$p_obsolete`, but should be `true` or `false`?
```
 * @param boolean $p_obsolete   Whether to include obsolete versions.
```
e.g. in `manage_proj_edit_page.php` there is
```
$t_versions = version_get_all_rows( $f_project_id, null, null );
```

cproensa: $p_released is a 3 way selector (i have updated the doc):
```php
 @param boolean $p_released   Whether to show only released, unreleased, or both.
                  For this parameter, use contants defined as:
                  VERSION_ALL (null): returns any
                  VERSION_FUTURE (false): returns only unreleased versions
                  VERSION_RELEASED (true): returns only released versions
```
In the previous implementation, $p_obsolete worked in the same way:
```php
	if( $p_obsolete !== null ) {
		$t_query .= ' AND obsolete = ' . db_param();
		$t_query_params[] = (bool)$p_obsolete;
	}
```
But having a default value of `false`, it was never used with a value of `true` (meaning, return only obsolete versions)
After rewriting this function, now it's a true/false parameter meaning precisely "include, or not, obsolete versions"


> but isn't the real problem that there are places where version_get_all_rows is called with null for parameter $p_obsolete, but should be true or false?

probably, those should be cleaned up to use true/false.
since this is a regression fix, i'd rather not include it in this PR

dregad: @cproensa So what you're saying is that it is no longer possible to retrieve only obsolete versions ? OK, fine. 

Adding the strict-type check against *false* as you are proposing is a good thing to do as a safety net, but I would agree with @atrol that since the semantics of the `$p_obsolete` parameter have changed, then the **proper fix** for this regression, is to change the callers to use *true* instead of null. 
```
$ git grep version_get_all_rows\(.*,.*,\ null
api/soap/mc_project_api.php:839:    $t_versions = version_get_all_rows( $p_project_id, null, null );
core/print_api.php:865: $t_versions = version_get_all_rows( $t_project_ids, $p_released, null );
manage_proj_edit_page.php:495:  $t_versions = version_get_all_rows( $f_project_id, null, null );
```

What could be done outside of this PR if you prefer, is to make sure the callers are referencing the *VERSION_xxx* constants instead of true/false/null. That's only 5 occurences AFAICT
```
$ git grep -P '(?<!function )version_get_all_rows\(.*,\ (?!VERSION)'
api/soap/mc_project_api.php:839:    $t_versions = version_get_all_rows( $p_project_id, null, null );
changelog_page.php:248: $t_version_rows = version_get_all_rows( $t_project_id, null, false );
core/print_api.php:865: $t_versions = version_get_all_rows( $t_project_ids, $p_released, null );
core/version_api.php:576:   $t_versions = version_get_all_rows( $c_project_id, null /* released: any */, true /* incl. obsolete */ ,  $p_inherit );
manage_proj_edit_page.php:495:  $t_versions = version_get_all_rows( $f_project_id, null, null );
```

dregad:
> consider applying this to 2.20.x branch as well?

Definitely yes. I updated target version on the tracker

dregad:
@cproensa you have not responded to https://github.com/mantisbt/mantisbt/pull/1492#discussion_r270359921. Specifically, I'd like your feedback on 

> since the semantics of the $p_obsolete parameter have changed, then the proper fix for this regression, is to change the callers to use true instead of null.

I let you decide if you want to reference the VERSION_xxx constants in this PR later.

core/version_api.php
dregad: @cproensa replied via chat
> 19-Apr-2019 11:10
> ok. for #1492 i agree with your suggestions, and will try to include them.

Since he's busy, I'll include them myself :wink:

dregad:
Branch rebased on top of master-2.20 to match target of 2.20.1 release.

Added 2 commits to address discussion in https://github.com/mantisbt/mantisbt/pull/1492#discussion_r270359921.

@cproensa This should to merge as-is, unless you have any objection.

PR #1496, opened 2019-04-04
Fix plugin_get_current() when used in MantisColumn classes

brlumen:
Added field 'plugin_name' in MantisColumn.
Added initialization of the field 'plugin_name' to columns_get_plugin_columns()
Concluded between plugin_push_current() and plugin_pop_current() functions are clear_cache(), cache(), display(), value(), sortquery() wherever these functions are called.

At first glance, and after tests, including on my plug-in, I believe that this PR will not cause compatibility errors with already released plug-ins.

Fixes #25677

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -2220,7 +2220,9 @@ function bug_clear_cache_all( $p_bug_id = null ) {
 
 	$t_plugin_objects = columns_get_plugin_columns();
 	foreach( $t_plugin_objects as $t_plugin_column ) {
+                plugin_push_current($t_plugin_column->plugin_name);
 		$t_plugin_column->clear_cache();
+                plugin_pop_current();
 	}
 	return true;
 }
@@ -2252,7 +2254,9 @@ function bug_cache_columns_data( array $p_bugs, array $p_selected_columns ) {
 
 		if( column_is_plugin_column( $t_column ) ) {
 			$plugin_objects = columns_get_plugin_columns();
+                        plugin_push_current($plugin_objects[$t_column]->plugin_name);
 			$plugin_objects[$t_column]->cache( $p_bugs );
+                        plugin_pop_current();
 			continue;
 		}
 
--- a/core/classes/BugFilterQuery.class.php
+++ b/core/classes/BugFilterQuery.class.php
@@ -1643,7 +1643,9 @@ class BugFilterQuery extends DbQuery {
 				$t_plugin_columns = columns_get_plugin_columns();
 				$t_column_object = $t_plugin_columns[$c_sort];
 
+                                plugin_push_current( $t_column_object->plugin_name );
 				$t_clauses = $t_column_object->sortquery( $c_dir );
+                                plugin_pop_current();
 				if( is_array( $t_clauses ) ) {
 					if( isset( $t_clauses['join'] ) ) {
 						$this->add_join( $t_clauses['join'] );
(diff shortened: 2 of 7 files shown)


dregad:
Thanks for your contribution.

I didn't look in depth or test your code yet, but I get the feeling that it is only half-fixing the problem. Adding plugin push/pop calls in several places in the core, amounts to reporting the burden to remember doing it on MantisBT core devs, should these APIs need to be changed in the future. I think that the logic should somehow be encapsulated within the class.

The only issue is to ensure that $plugin_name is properly initialized whenever it needs to be used (are we sure that _columns_get_plugin_columns()_ is always called first ?), and fail elegantly if not.

Just thinking out loud here, but wouldn't it be better to achieve the same goal by defining "wrapper" final methods in the MantisColumn class for each of _sortquery()_, _cache()_, _clear_cache()_, _display()_ and _value()_ that would be called by the MantisBT API, e.g.

_core/classes/MantisColumn.class.php_
``` php
final function call_xxx( ... ) {
    plugin_push_current( $this->plugin_name );
    $this->xxx( ... );
    plugin_pop_current();
}
```

Then, wherever _MantisColumn::xxx()_ function is used in core, replace the call by _MantisColumn::call_xxx()_ 

Note: I just used `call_` as prefix here for the wrapper functions, as I can't think of a better name at the moment. That being said, a cleaner approach would be to let the MantisColumn class's public method remain the same as it is today, and define a protected, _internal_xxx()_ function instead to be overridden by child classes. The problem with this approach however is that it would break backwards-compatibility, but that can be considered as an option.

Let me know your thoughts. Maybe @cproensa has some opinion on this as well.

cproensa:
IIRC, the plugin push is executed by the plugin api when any event is executed, and so "plugin current" context is prepared before the execution is handled to the plugin code.

However, the issue reported seems to be related to the MantisColumn, MantisFIlter objects, which have a different context than events.
This is, the plugin return a MantisColum object to core, and this object is executed in the context of the core api, not in the plugin context. I don't think that is wrong. In any case, maybe that could be better explained in dev guide.

If we agree on that convention, you need to:
- explicitly do the plugin push within the MantisColumn methods
- or use plugin_config_get, plugin_lang_get, etc instead, which accept a specific plugin name as parameter.

Unless i am missing any other details, i think that this should not be changed.

brlumen:
@cproensa @dregad 
Regarding the implementation of logic within the class, I unfortunately do not see a solution to the compatibility issue.
In fact, I did not take into account that it is possible to transfer not only the finished object, but also the class. In the second case, this means that the object will be initialized during the process of creating a class object in columns_get_plugin_columns (), which in its own way implies initialization of the plugin_name field at this stage. Unfortunately, my PR offers exactly that.

I see my solution as if from the developer of plug-ins, and not from the developer of the kernel. I see that in terms of architecture, my solution is not entirely correct.
On the other hand, it works, and solves the problem, and also has backward compatibility.

My main idea is that the plugin developer does not expect that the kernel will be blind while processing objects belonging to its plugin. And what is most interesting, all functions that use columns_get_plugin_columns () do not work correctly. They also do not expect this.

If the issue with backward compatibility is not so acute, then I propose to reject my PR, and transfer the task to the category “future”. And perhaps in the future to return to this issue.

In any case, I was glad to participate in this discussion.

brlumen:
@cproensa 
>or use plugin_config_get, plugin_lang_get, etc instead, which accept a specific plugin name as parameter.

At the moment, they do not accept such a parameter. 
Did you mean that we need to change these functions?

dregad:
This PR has merge conflicts. Considering that there has been no activity on it for over a year and a consensus was never reached, I see no reason to keep it open.

PR #1498, opened 2019-04-04
add active class in summary menu for custom tabs

shaozi:
This is to fix issue [23418](https://mantisbt.org/bugs/view.php?id=23418)

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -897,7 +897,7 @@ function print_summary_menu( $p_page = '', array $p_filter = null ) {
 		$t_active = $p_page && strpos( $t_menu_item, $p_page ) !== false
 			? ' class="active"'
 			: '';
-		echo "<li{$t_active}>" . $t_menu_item . '</li>';
+		echo "<li{$t_active}>{$t_menu_item}</li>";
 	}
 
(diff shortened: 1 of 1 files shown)


core/html_api.php
```diff
@@ -894,7 +894,10 @@ function print_summary_menu( $p_page = '', array $p_filter = null ) {
 
 	# Plugins menu items - these are cooked links
 	foreach ( $t_menu_options as $t_menu_item ) {
-		echo '<li>' . $t_menu_item . '</li>';
+		$t_active = $p_page && strpos( $t_menu_item, $p_page ) !== false
+			? ' class="active"'
+			: '';
+		echo "<li{$t_active}>" . $t_menu_item . '</li>';
```
dregad: Since there's already a substitution, might as well do this instead

    echo "<li{$t_active}>{$t_menu_item}</li>";

shaozi: just did a commit

dregad: That's perfect, thanks

vboctor:
Thanks @shaozi for your contribution.  Merged via 113a270b69b0299a240c746c525467ec21689bed

PR #1499, opened 2019-04-05
Fix redirect after bug report process ending

brlumen:
Fix the behavior of the redirect after creating a issue.

This is not the final PR.
May need to add comments.

Fixes [#25680](https://mantisbt.org/bugs/view.php?id=25680)


brlumen:
We are closing as the question has moved to another plane.

dregad:
Replaced by #1505

PR #1501, opened 2019-04-10
Fix link to user page fom bug view page.

brlumen:
Fixes [#25686](https://mantisbt.org/bugs/view.php?id=25686)

### DIFF
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -230,6 +230,26 @@ function print_user_with_subject( $p_user_id, $p_bug_id ) {
 }
 
 /**
+ * Print email icon with fills in the subject with the bug summary
+ *
+ * @param integer $p_user_id A user identifier.
+ * @param integer $p_bug_id  A bug identifier.
+ * @return void
+ */
+function print_email_icon_with_subject( $p_user_id, $p_bug_id ) {
+	if( NO_USER == $p_user_id ) {
+		return;
+	}
+
+	if( user_exists( $p_user_id ) && user_get_field( $p_user_id, 'enabled' ) ) {
+		$t_email = user_get_email( $p_user_id );
+		print_email_link_with_subject( $t_email, '', '', $p_bug_id, true );
+	} else {
+                return;
+	}
+}
+
+/**
  * print out an email editing input
  *
  * @param string $p_field_name Name of input tag.
@@ -1654,7 +1674,7 @@ function get_email_link( $p_email, $p_text ) {
  * @param string $p_bug_id The bug identifier.
  * @return void
  */
-function print_email_link_with_subject( $p_email, $p_text, $p_tooltip, $p_bug_id ) {
+function print_email_link_with_subject( $p_email, $p_text, $p_tooltip, $p_bug_id, $p_icon_only = false ) {
 	if( !is_blank( $p_tooltip ) && $p_tooltip != $p_text ) {
 		$t_tooltip = ' title="' . $p_tooltip . '"';
 	} else {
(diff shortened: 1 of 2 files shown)


core/print_api.php
```diff
@@ -1654,7 +1674,7 @@ function get_email_link( $p_email, $p_text ) {
  * @param string $p_bug_id The bug identifier.
  * @return void
  */
-function print_email_link_with_subject( $p_email, $p_text, $p_tooltip, $p_bug_id ) {
+function print_email_link_with_subject( $p_email, $p_text, $p_tooltip, $p_bug_id, $p_icon_only = false ) {
```
dregad: Parameter needs to be documented (PHPDoc).
I also suggest to rename it to `$p_show_tooltip = true` to better reflect its purpose.

core/print_api.php
```diff
@@ -1663,8 +1683,12 @@ function print_email_link_with_subject( $p_email, $p_text, $p_tooltip, $p_bug_id
 
 	$t_bug = bug_get( $p_bug_id, true );
 	if( !access_has_project_level( config_get( 'show_user_email_threshold', null, null, $t_bug->project_id ), $t_bug->project_id ) ) {
-		echo $t_tooltip != '' ? '<a' . $t_tooltip . '>' . $p_text . '</a>' : $p_text;
-		return;
+		if( $p_icon_only == true ) {
```
dregad: `== true` is redundant

With the renamed parameter, the logic is reversed, so 
```php
if( $p_show_tooltip ) {
    echo ....
}
return;
```

core/print_api.php
```diff
@@ -1678,7 +1702,11 @@ function print_email_link_with_subject( $p_email, $p_text, $p_tooltip, $p_bug_id
 	$t_mailto = string_attribute( 'mailto:' . $t_email . '?subject=' . $t_subject );
 	$t_text = string_display( $p_text );
 
-	echo '<a href="' . $t_mailto . '"' . $t_tooltip . '>' . $t_text . '</a>';
+        if( $p_icon_only == true ) {
```
dregad: See previous comment

core/print_api.php
```diff
@@ -1678,7 +1702,11 @@ function print_email_link_with_subject( $p_email, $p_text, $p_tooltip, $p_bug_id
...
+                echo '<a href="' . $t_mailto . '"' . $t_tooltip . ' class="fa fa-envelope-o" style="text-decoration: none;"></a>';
```
dregad: What's the purpose of the inline style ? We try to avoid these.

dregad:
@vboctor I very rarely use this feature myself so I don't have a strong opinion against removing it. I would tend to agree that discussion should normally take place on the issue itself and using direct e-mail should be avoided. 

That being said, other people may think otherwise, and since this feature has been present for a very long time, they could be relying on it. As for adding a config, I'm not sure that's worth it, too many of these already.

With regards to combining the functions, the way @brlumen did it is consistent with how the API currently works: *print_user_with_subject()* -> *print_email_link_with_subject()*; the latter is a lower-level API, and the new *print_email_icon_with_subject()* would be at the same level as the former.  *print_user_with_subject()* is still used in print_all_bug_page_word.php so it would make sense to keep things as is.

core/print_api.php
brlumen: Used to disable the underscore fa fa-envelope-o when you hover the pointer.
Unfortunately, I am still very weak in the structure of css mantisbt.
I would be very grateful if you tell me how to implement it correctly.

dregad:
Closing this, please follow up in #1519

PR #1502, opened 2019-04-12
fix issue 23694

AnatolyKabakov1983:
[issue#23694](https://www.mantisbt.org/bugs/view.php?id=23694)

### DIFF
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -1452,7 +1452,7 @@ function print_form_button( $p_action_page, $p_label, array $p_args_to_post = nu
  * @return void
  */
 function print_bracket_link_prepared( $p_link ) {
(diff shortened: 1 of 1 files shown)


dregad:
Thanks for your contribution. 

After reviewing the PR, I came to the conclusion that function `print_bracket_link_prepared()` should be left unchanged, and marked as obsolete. As of this writing, the only usage of this function in the MantisBT codebase, is when processing the _EVENT_MENU_ISSUE_ hook in bug_view_inc.php.

The correct way of preventing links to be wrapped with `[ ]` in this context, is to change the plugin hook to return an associative array of the form `label => link_target`, instead of cooked links.

I will provide an alternative PR.

PR #1508, opened 2019-04-29
Fix updating user access level for projects

cproensa:
Fix an error from PR 1449 where updating access level for a user in a
project, updates the access levels on all projects assigned to that
user. Missing project_id in the update query.

Fixes: #25722


cproensa:
This should be also updated to 2.20.x

dregad:
Rebased onto master-2.21 branch

PR #1509, opened 2019-05-04
log_event() create log file if it does not exist

dregad:
This is a regression from issue [#19642](https://mantisbt.org/bugs/view.php?id=19642) (PR #1483). When MantisBT tries to write an
event to a log file that does not already exist, a warning is printed
and the file is not created even if PHP process has write access to the
directory. The event is added to the PHP system log.

Instead of relying on is_writable(), the code now calls error_log() with
errors suppressed, then checks the result of the function call to
determine if the operation was successful or not.

Fixes [#25734](https://mantisbt.org/bugs/view.php?id=25734)

### DIFF
--- a/core/logging_api.php
+++ b/core/logging_api.php
@@ -107,18 +107,24 @@ function log_event( $p_level, $p_msg ) {
 			break;
 		case 'file':
 			if( isset( $t_modifiers ) ) {
-				# Log the event, suppress errors in case the file is not writable
-				$t_log_writable = @error_log( $t_php_event . PHP_EOL, 3, $t_modifiers );
+				static $s_log_writable = null;
 
-				if( !$t_log_writable ) {
-					# Display a one-time warning message and write it to PHP system log as well.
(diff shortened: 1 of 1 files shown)


atrol:
I set `$g_log_destination` to a non-existing file in a non-writable directory and `$g_log_level = LOG_DATABASE;`

After that, trying to display any page gives just an empty page.

I am getting the expected `ERROR_LOGFILE_NOT_WRITABLE` message in PHP error log, 
The next entry in log is `PHP Fatal error:  Maximum execution time of 30 seconds exceeded`

dregad:
Good catch @atrol I should have tested with LOG_DATABASE... I can confirm the behavior, although on my dev box, with xdebug it fails instantly with _Fatal error: Uncaught Error: Maximum function nesting level of '256' reached, aborting!_

The code gets into an infinite loop when retrieving the user's default language... The offending sequence is lang_get_default() -> auth_is_user_authenticated() -> auth_is_cookie_valid() which runs a db query.

I'll push a fix shortly.

PR #1513, opened 2019-05-19
Allow adding monitors when reporting an issue

dregad:
This was not extensively tested. Feedback welcome.

Fixes [#6128](https://mantisbt.org/bugs/view.php?id=6128)


bug_report_page.php
```diff
@@ -203,7 +203,10 @@
 $t_show_severity = in_array( 'severity', $t_fields );
 $t_show_priority = in_array( 'priority', $t_fields );
 $t_show_steps_to_reproduce = in_array( 'steps_to_reproduce', $t_fields );
-$t_show_handler = in_array( 'handler', $t_fields ) && access_has_project_level( config_get( 'update_bug_assign_threshold' ) );
+$t_show_handler = in_array( 'handler', $t_fields )
+	&& access_has_project_level( config_get( 'update_bug_assign_threshold' ) );
+$t_show_monitors = in_array( 'monitors', $t_fields )
+	&& access_has_project_level( config_get( 'monitor_add_others_bug_threshold' ) );
```
atrol: Not sure if we should add also
`	&& access_has_project_level( config_get( 'show_monitor_list_threshold' )`
as
- it's a bit strange to be able to add users to the monitor list, but not being able to see the list afterwards, if user does not have show_monitor_list_threshold rights
- adding monitor users on view.php is not possible without show_monitor_list_threshold rights

but
- addinig monitor users on view.php is possible without show_monitor_list_threshold rights by using the `Send a Reminder` functionality

vboctor: It is fair to assume that "add" threshold will be higher than "view" threshold for adding others to monitor list.  We can be explicit about this in the documentation.  However, I assume if we poke at a lot of the other configuration and do such conversion, there will be similar odd situations.

dregad: > It is fair to assume that "add" threshold will be higher than "view" threshold for adding others to monitor list

That was my assumption indeed.

> adding monitor users on view.php is not possible without show_monitor_list_threshold rights

That's kind of strange, since view.php page does not actually offer a list of users... 

Furthermore, I fail to see the correlation between being able to add new monitors to an issue, vs seeing the list of users currently monitoring it (which is what *show_monitor_list_threshold* is about... and that's not exactly the same thing as being able to see the list of users who are authorized to monitor issues)...

Having said that, it's easy to add a check on *show_monitor_list_threshold* to be consistent, if this is the desired behavior (I think not). Anyway let me know what you want me to do. 

>  We can be explicit about this in the documentation

Improving the documentation is always good, but in that specific case and considering my above remark, do we actually need to ?

atrol: > That's kind of strange, since view.php page does not actually offer a list of users...

Depending on `show_monitor_list_threshold` view.php page displays the monitoring users in section `Users monitoring this issue`.

> Furthermore, I fail to see the correlation between being able to add new monitors to an issue, vs seeing the list of users currently monitoring it 

RIght, but that's what we have on view.php at the moment.
You are not able to add monitor users if you are not allowed to see monitor users.

dregad:
I have added additional commits, taking into consideration the discussion following @atrol's review.

- In view.php, users can now add monitors even if they can't see the list 
image
- Improved documentation (Admin Guide & PHPDoc in config_defaults_inc.php)
- Added missing _add/remove monitors_ configs in manage workflow thresholds page

With these changes, the system's behavior is more consistent, and clearly documented.

PR #1514, opened 2019-05-19
check affected rows before update history

shaozi:
I have tested on mysql and it worked.


dregad:
@vboctor I think your suggestion is what I proposed in #1035 but as the discussion there shows, it's not as simple as it sounds.

marvin-SL:
Hello @shaozi dregad 
Why not :

```
$old_value = custom_field_database_to_value( $t_row[$t_value_field], $t_type );
$new_value = custom_field_database_to_value( $t_value, $t_type );
     
history_log_event_direct( $p_bug_id, $t_name, $old_value, $new_value );
```

Works for me with list custom file and for integer/float

PR #1518, opened 2019-06-05
Add newline after <textarea> tags

dregad:
The HTML 5 specification states that "Newlines at the start of textarea
elements are ignored as an authoring convenience." [1]

To avoid altering user data (e.g. issue description, steps to reproduce,
etc.) having leading newline(s) when editing and saving, the markup of
all pages displaying user content in textarea tags has been modified to
add a newline after the opening tag.

Fixes [#25839](https://mantisbt.org/bugs/view.php?id=25839)

[1]: https://www.w3.org/TR/html52/syntax.html#the-in-body-insertion-mode


vboctor:
I have added a comment to the issue.

PR #1520, opened 2019-06-10
Disable Dropzone timeout

atrol:
Fixes #25763


vboctor:
How about merging this into `master-2.21`?

dregad:
> How about merging this into master-2.21?

Technically this is a regression from 2.20.0, not 2.21.0, but anyway that would be fine with me.

vboctor:
Merged via bcb1334ff95268adba1202406484c0a276bece38

PR #1525, opened 2019-07-08
Block URLs in Public Notes from Anonymous Accounts

RealityRipple:
Example Pull for [Bug #25908](https://www.mantisbt.org/bugs/view.php?id=25908)


RealityRipple:
https://github.com/RealityRipple/SafeAnon

PR #1522, opened 2019-06-28
Fix helper_generate_cache_key() in re-auth process

brlumen:
Fixes [#25892](https://mantisbt.org/bugs/view.php?id=25892)

PR #1526, opened 2019-07-09
Don't cache status_config.php when it's not populated

cproensa:
When the dynamic css file `status_config.php` can't be generated, force a no-caching header to avoid the client caching a wrong version of this file.

Fixes: #24189


cproensa:
Alternate fix from PR #1522

PR #1527, opened 2019-07-12
Add support for pasting images to forms with dropzone

syncguru:
Fixes #21797

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -409,7 +409,19 @@ $(document).ready( function() {
 	$( 'form .dropzone' ).each(function(){
 		var classPrefix = 'dropzone';
 		var autoUpload = $(this).hasClass('auto-dropzone');
-		enableDropzone( classPrefix, autoUpload );
+		var zoneObj = enableDropzone( classPrefix, autoUpload );
+		if( zoneObj ) {
+			/* Attach image paste handler to add-note form */
+			$( '#bugnoteadd' ).bind( 'paste', function( event ) {
+				var items = ( event.clipboardData || event.originalEvent.clipboardData ).items;
+				for( index in items ) {
+					var item = items[index];
+					if( item.kind === 'file' ) {
+						zoneObj.addFile( item.getAsFile() )
+					}
+				}
+			});
+		}
 	});
 
 	$('.bug-jump').find('[name=bug_id]').focus( function() {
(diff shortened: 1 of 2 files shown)


syncguru:
You are right - fixed.

atrol:
The following behavior is convenient, but a bit strange.
Nut sure if it's intended.

- Copy an image to clipboard
- Click `Report Issue`
- Enter Ctrl+V in `Summary` field

As a result, the image is added as attachment.

atrol:
Does not work in Safari.

vboctor:
I have done few test cases:

1. Copy one file and paste.
2. Copy multiple files and paste
3. Copy an image (e.g. screenshot) then paste.

I have had issues with copying and pasting multiple issues.

- In Firefox - all file names are pasted in text box, but none added to dropzone.
- In Chrome - all file names are added, but only one image added to dropzone.

syncguru:
@atrol: Thanks for giving this a try.

1- The behavior is intended for paste functionality. I capture any paste event over the surface area of the form. Other implementations I've seen covering the entire document. Reducing the surface area will make pasting less useful.
2- It works for me on Safari.  I am not sure about IE since I cannot test. Notice that if we have issues with a browser or version the paste functionality won't work but you still can do what you do today.

syncguru:
@atrol: Thanks for giving this a try.
The copy/paste functionality appears to be intended for taking a snapshot or copying image from a html page. Copying multiple files from a file manager use case seems to be nonuniformly supported as a drag and drop. Browsers behave differently if you try to copy & paste and that is what you are experiencing:
See [this](https://stackoverflow.com/questions/18055422/how-to-receive-php-image-data-over-copy-n-paste-javascript-with-xmlhttprequest) StackOverflow thread from more details.

cproensa:
For me it works fine.
I can point:
- By doing repeated ctl-v, the same file is duplicated N times. It would be good if this could be detected to avoid duplication
- I miss some visual tool, button, etc that allows to paste from the ui. At least to make the user aware that this possibility exists.

syncguru:
> I miss some visual tool, button, etc that allows to paste from the ui. At least to make the user aware that this possibility exists.

@cproensa do you have example of what that would look like

cproensa:
>> I miss some visual tool, button, etc that allows to paste from the ui. At least to make the user aware that this possibility exists.

> @cproensa do you have example of what that would look like

Probably just modifying the dropzone text is enough: "drop files here to upload", adding a hint that pasting images is possible

vboctor:
@syncguru I would like to merge this.  Are you planning any tweaks?

syncguru:
@vboctor Yes, I will make a text change as suggested by @cproensa ... give me a few mins to update the PR.

syncguru:
@vboctor done.

I used the same expression as in github
Screen Shot 2019-08-03 at 3 42 03 PM

PR #1530, opened 2019-07-25
ADD MAIN MENU DROPDOWN BUTTON

JeremyOWO:
Since mantis 2.X if we want to add a plugin button for non manager, we need to use EVENT MAIN MENU.
But we are limited by the number of button that our resolution can integrate.

So this pull request allow to integrate a dropdown button and so allow un bigger number of button

This pull request is possible thanks to the new EVENT MAIN MENU FILTER.
With this event you can get the array of all main menu button and so integer inside an array of array for each plugin.

Picture available : https://ibb.co/m9mwM4p

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -886,6 +886,7 @@ function layout_options_for_sidebar( $p_menu_options, $p_active_sidebar_page ) {
 	foreach( $p_menu_options as $t_menu_option ) {
 		$t_icon = isset( $t_menu_option['icon'] ) ? $t_menu_option['icon'] : 'fa-plug';
 		if( !isset( $t_menu_option['url'] ) || !isset( $t_menu_option['title'] ) ) {
+			layout_sidebar_dropdown_menu($t_menu_option);
 			continue;
 		}
 
@@ -899,6 +900,38 @@ function layout_options_for_sidebar( $p_menu_options, $p_active_sidebar_page ) {
 	}
 }
 
+/**
+ * Print dropdown menu item
+ * @param array  $p_menu_options menu item of group
+ * @return void
+ */
+function layout_sidebar_dropdown_menu( $p_menu_options ) {
+	echo '<li class="dropdown">' . "\n";
+   	echo '<a class="dropdown-toggle" data-toggle="dropdown" href="#">' . "\n";
+   	echo '<i class="menu-icon fa fa-plug"></i>' . "\n";
+   	echo '<span class="menu-text">Plugins</span>' . "\n";
+   	echo '<span class="caret"></span>';
+   	echo '</a>' . "\n";
+        echo '<ul class="dropdown-menu" >';
+	foreach( $p_menu_options as $t_menu_option ) {
+		
+			$t_icon = isset( $t_menu_option['icon'] ) ? $t_menu_option['icon'] : 'fa-plug';
+			if( !isset( $t_menu_option['url'] ) || !isset( $t_menu_option['title'] ) ) {
+				continue;
+			}
+
+			if( isset( $t_menu_option['access_level'] ) ) {
+				if( !access_has_project_level( $t_menu_option['access_level'] ) ) {
+					continue;
+				}
+			}
+
+			layout_sidebar_menu( $t_menu_option['url'], $t_menu_option['title'], $t_icon, $p_active_sidebar_page );
+		
+	}
+	echo '</ul>';
+	echo '</li>' . "\n";
+}
 /**
  * Print sidebar opening elements
  * @return void
--- a/js/common.js
+++ b/js/common.js
@@ -35,6 +35,7 @@ if (a!= -1) {
 style_display = 'block';
 
 $(document).ready( function() {
+    $('.dropdown-toggle').dropdown();
     $('.collapse-open').show();
     $('.collapse-closed').hide();
     $('.collapse-link').click( function(event) {


core/layout_api.php
```diff
@@ -886,6 +886,7 @@ function layout_options_for_sidebar( $p_menu_options, $p_active_sidebar_page ) {
 	foreach( $p_menu_options as $t_menu_option ) {
 		$t_icon = isset( $t_menu_option['icon'] ) ? $t_menu_option['icon'] : 'fa-plug';
 		if( !isset( $t_menu_option['url'] ) || !isset( $t_menu_option['title'] ) ) {
+			layout_sidebar_dropdown_menu($t_menu_option);
```
dregad: coding guidelines

core/layout_api.php
```diff
@@ -899,6 +900,38 @@ function layout_options_for_sidebar( $p_menu_options, $p_active_sidebar_page ) {
...
+/**
+ * Print dropdown menu item
+ * @param array  $p_menu_options menu item of group
+ * @return void
+ */
+function layout_sidebar_dropdown_menu( $p_menu_options ) {
+	echo '<li class="dropdown">' . "\n";
+   	echo '<a class="dropdown-toggle" data-toggle="dropdown" href="#">' . "\n";
+   	echo '<i class="menu-icon fa fa-plug"></i>' . "\n";
+   	echo '<span class="menu-text">Plugins</span>' . "\n";
```
dregad: Hardcoded string - use lang_get() 
Maybe this should be a defaulted parameter ?

JeremyOWO:
I changed layout to use Ace theme.

dregad:
The markup using ACE theme does not work (or, to be more accurate, does not display properly). I don't know how to fix that, hopefully @syncguru can help...

- the submenu has a transparent background (see screenshots)
- icons are only visible when hovering over the menu item
- with expanded sidebar, the icon overlaps the caption 

image
image

Attached is the dummy plugin I used for testing. Save it in `./plugins/Test1530/Test1530.php`
```php
<?php
class Test1530Plugin extends \MantisPlugin
{
	public function register() {
		$this->name = 'Test1530';
		$this->description = "Test plugin for PR 1530";
		$this->url = "https://github.com/mantisbt/mantisbt/pull/1530";
		$this->version = '1.0';
	}

	public function hooks() {
		return parent::hooks() + [ "EVENT_MENU_MAIN_FILTER" => "submenu" ];
	}

	public function submenu( $p_event, $p_menu ) {
		$p_menu[] = [
			[ 'title' => 'Test menu', 'url' => 'http://example.com' ],
			[ 'title' => 'No icon', 'url' => 'http://example.com', 'icon' => '' ],
			[ 'title' => 'Google', 'url' => 'http://google.com', 'icon' => 'fa-google' ],
		];
		return [$p_menu];
	}
}
```

dregad:
Suggested enhancement: instead of using a hardcoded, generic _Plugins_ string and _fa-plug_ icon, I believe it would be better to let the plugin set the main menu item's entry as appropriate.

cproensa:
> Attached is the dummy plugin I used for testing

Is that the intended usage for this feature?
I'd expect that this submenu feature should be integrated with the current EVENT_MENU_MAIN event.

> Suggested enhancement: instead of using a hardcoded, generic Plugins string and fa-plug icon, I believe it would be better to let the plugin set the main menu item's entry as appropriate.

Thats what i think i would expect for submenus provided by a plugin
The current EVENT_MENU_MAIN structure is 
```
array( 
title => xxx
access_level => xxx
url => xxx
icon => xxx
)
```
So i'd go in the way of adding a new property, eg: submenu, or group, etc:
```
array( 
title => xxx
access_level => xxx
icon => xxx
submenu => array(
  [title, acces_level, url, icon],   [title, acces_level, url, icon], ...
)
```
so, each menu item may contain an url or a submenu array with more items

dregad:
> Is that the intended usage for this feature?

I just implemented a quick&dirty test setup to review the PR, and do not necessarily endorse such use.

My main point here is, that the submenu feature does not render properly with our customized ACE admin template, hence my question to @syncguru .

> I'd expect that this submenu feature should be integrated with the current EVENT_MENU_MAIN event.

I agree this would be a much more appropriate event to use for that.

The new proposed structure makes sense. I assume you mean the code would render the menu item as a single "button" (for lack of a better word) if there is only one submenu item, and as a submenu if there are more.

We probably need to move away from dumb arrays, and start using objects for this..

cproensa:
> My main point here is, that the submenu feature does not render properly with our customized ACE admin template, hence my question to @syncguru .

Right, in syncguru's screenshot looks nice though

> The new proposed structure makes sense. I assume you mean the code would render the menu item as a single "button" (for lack of a better word) if there is only one submenu item, and as a submenu if there are more.

My suggestion is that each menu item, with a structure like we have now, may contain a new property 'submenu', containing an array of more menu items.
i'm not sure of the layout capabilities: whether it should only be allowed 1 level of submenu, o can be more.

> We probably need to move away from dumb arrays, and start using objects for this..

yes, that makes the structure more clear and enforced by the object declarations.

core/layout_api.php
```diff
@@ -899,6 +900,39 @@ function layout_options_for_sidebar( $p_menu_options, $p_active_sidebar_page ) {
...
+				continue;
+			}
+
+			if( isset( $t_menu_option['access_level'] ) ) {
+				if( !access_has_project_level( $t_menu_option['access_level'] ) ) {
+					continue;
+				}
+			}
+
+			layout_sidebar_menu( $t_menu_option['url'], $t_menu_option['title'], $t_icon, $p_active_sidebar_page );
```
cproensa: SYSTEM NOTICE: 'Undefined variable: p_active_sidebar_page' in '/media/sdb/www/html/git/mantis2/core/layout_api.php' line 930

cproensa:
with the small screen, expanded icons, the submenu does not work:

imagen

JeremyOWO:
Hello All,

I'mnot understanding situation.
What is the next step ?

Because without ace them, the submenu corks correctly on small screen.
Do you think that this pull request could be implemented a day after work or never ?

In this case I'm just going to use a personnal patch.

Thanks again

dregad:
@JeremyOWO 
> I'mnot understanding situation.

It's quite simple, really : This PR is currently in no condition to be merged. It contains errors (see review comments) and just does not work properly based on my and @cproensa's testing.

I understand that you tried to do it with ACE theme per @syncguru's suggestion, but he's not responding to my [request for help](https://github.com/mantisbt/mantisbt/pull/1530#issuecomment-524872073) despite offering his expertise, I'm afraid I can't do much more at this time.

syncguru:
@dregad / @JeremyOWO Sorry for the delay, I will take a look at this tomorrow.

syncguru:
@JeremyOWO 
The only difference in the markup for submenu items that it does not include `<span class="menu-text">` around the text part. You are calling `layout_sidebar_menu` which add this markup by default. Try to have a new implementation to render submenu items without the use of this `<span>` 
 element and see if that fixes the issue in mentioned above.

dregad:
@syncguru removing the `<span>` does not fix the problem at all.

I think it's a problem in the ACE CSS (or the way we use it), because if you set `<body class="no-skin">` instead of `<body class="skin-3">` ( as set in layout_page_header_end() function) then the submenu's background color is no longer transparent.

As a side note, do we have some official documentation on ACE template ? Also, do we have complete source of ACE ? Wondering, because I don't see any _skin-1_ or _skin-2_ classes

syncguru:
@dregad yes, we are not using it correctly. I have the original ACE package which includes a demo site to show how different components in the theme are used. My license allows me to share the package with Mantis core team but not to make it open source. We managed to open source the files that we attach with mantis only. Happy to email you a copy.

core/layout_api.php
JeremyOWO: I corrected this.

JeremyOWO:
Ok, So if I understand, that works correctly with your ACE them because you have a specific licence. But on mantis we are juste using an open source licence of this theme which not include all CSS style ?

Could the missing CSS be integer in mantis ?

dregad:
> we are not using it correctly

So what would be the correct way to use it, then ?

> Happy to email you a copy

If you could, that would be nice. Thanks !

> My license allows me to share the package with Mantis core team but not to make it open source. We managed to open source the files that we attach with mantis only. 

I apologize for going off-topic, and for reopening that can of worms, but as you may recall I have never been comfortable with ACE template's commercial, closed-licence origins. I accepted it because I was under the impression that it had been released under GPL, and until now I never realized that we only had a "partial" open-sourcing... 

Considering that the original code is no longer maintained by its author (who by the way never responded to any of my sollicitations), and that Bootstrap 4 has been released early 2018, maybe we should start thinking about replacing ACE with a fully open-source, well-maintained template ?

core/layout_api.php
```diff
@@ -905,7 +905,7 @@ function layout_options_for_sidebar( $p_menu_options, $p_active_sidebar_page ) {
  * @param array  $p_menu_options menu item of group
  * @return void
```
dregad: Missing PHPDoc for the new parameter

c2pil:
Hello,

I am interested in this PR.
I have the feeling that all the remaining changes requested were fixed in the last commits of JeremyOWO.

Would it be possible to take them into account and to merge the PR ? Or there are still changes to make ?

priocomsolutions:
> Hello,
> 
> I am interested in this PR.
> I have the feeling that all the remaining changes requested were fixed in the last commits of JeremyOWO.
> 
> Would it be possible to take them into account and to merge the PR ? Or there are still changes to make ?


Also, would you kindly show some of us budding coders an example of implementing this? WIll be much much appreciated.

maksun:
Hello,

Just to say I would love to see this one implemented ;)

Thanks!

dregad:
> I have the feeling that all the remaining changes requested were fixed in the last commits of JeremyOWO.

The only things that were addressed with the latest commits are minor cosmetic issues (tab indentation and missing PHPDoc).

I will gladly merge the PR, when it works. Unfortunately, this is still not the case today.

The problems I outlined in https://github.com/mantisbt/mantisbt/pull/1530#issuecomment-524872073 are still present.

PR #1533, opened 2019-08-02
Update Chart.js library to v2.8.0 + define constants in MantisGraph plugin

dregad:
Fixes
- [#25951](https://mantisbt.org/bugs/view.php?id=25951): MantisGraph: update Chart.js library to v2.8.0
- [#25952](https://mantisbt.org/bugs/view.php?id=25952): MantisGraph: define Chart.js-related constants in the plugin

### DIFF
--- a/plugins/MantisGraph/MantisGraph.php
+++ b/plugins/MantisGraph/MantisGraph.php
@@ -100,8 +100,8 @@ class MantisGraphPlugin extends MantisPlugin  {
 	/**
 	 * Add the RESTful routes that are handled by this plugin.
 	 *
-	 * @param $p_event_name The event name
-	 * @param $p_event_args The event arguments
+	 * @param string $p_event_name The event name
+	 * @param array  $p_event_args The event arguments
 	 * @return void
 	 */
 	function routes( $p_event_name, $p_event_args ) {
@@ -143,7 +143,7 @@ class MantisGraphPlugin extends MantisPlugin  {
 			return array( '<a class="btn btn-sm btn-primary btn-white btn-round" href="' .
 				plugin_page( 'issues_trend_page.php' ) . '">' . plugin_lang_get( 'issue_trends_link' ) . '</a>', );
 		} else {
-			return '';
+			return array();
 		}
 	}
 


plugins/MantisGraph/MantisGraph.php
```diff
@@ -140,13 +154,20 @@ function graph_filter_menu() {
...
+				html_javascript_cdn_link( $t_link . self::CHARTJSBUNDLE_FILE, self::CHARTJSBUNDLE_HASH );
+
+				# Chart.js color schemes plugin
+				$t_link = sprintf( $t_cdn, self::CHARTJS_COLORSCHEMES_LIB, self::CHARTJS_COLORSCHEMES_VERSION );
+				html_javascript_cdn_link( $t_link . self::CHARTJS_COLORSCHEMES_FILE, self::CHARTJS_COLORSCHEMES_HASH );
 			} else {
-				echo '<script type="text/javascript" src="' . plugin_file('Chart-' . CHARTJS_VERSION . '.min.js') . '"></script>';
-				echo '<script type="text/javascript" src="' . plugin_file('Chart.bundle-' . CHARTJS_VERSION . '.min.js') . '"></script>';
-				echo '<script type="text/javascript" src="' . plugin_file( 'chartjs-plugin-colorschemes.min.js' ) . '"></script>';
+				echo '<script type="text/javascript" src="' . plugin_file( self::CHARTJS_FILE ) . '"></script>';
```
atrol: Isn't it better to go on using the versions also for non-CDS to prevent caching side effects after upgrade?

dregad: I did not think of that. Good point, will change.

library/README.md
```diff
@@ -23,8 +23,8 @@ ace-admin theme                   | 1.4.0     | customized
 moment.js                         | 2.15.2    | unpatched
 bootstrap-datetimepicker          | 4.17.47   | unpatched
 dropzone.js                       | 5.5.0     | unpatched
-chart.js                          | 2.7.3     | unpatched
-chartjs-plugin-colorschemes       | 0.3.0     | unpatched
+chart.js                          | 2.8.0     | unpatched
```
vboctor: Given that the library is scoped to the plugin, should we have it listed there?  I can see benefits to both models.

plugins/MantisGraph/MantisGraph.php
```diff
@@ -140,15 +149,28 @@ function graph_filter_menu() {
 	function resources() {
 		if( current( explode( '/', gpc_get_string( 'page', '' ) ) ) === $this->basename ) {
 			if( config_get_global( 'cdn_enabled' ) == ON ) {
```
vboctor: Initialize the `$t_scripts` array in this branch.

plugins/MantisGraph/MantisGraph.php
```diff
@@ -140,15 +149,28 @@ function graph_filter_menu() {
...
-				html_javascript_cdn_link('https://cdnjs.cloudflare.com/ajax/libs/Chart.js/' . CHARTJS_VERSION . '/Chart.min.js', CHARTJS_HASH);
-				html_javascript_cdn_link('https://cdnjs.cloudflare.com/ajax/libs/Chart.js/' . CHARTJS_VERSION . '/Chart.bundle.min.js', CHARTJSBUNDLE_HASH);
-				html_javascript_cdn_link('https://cdn.jsdelivr.net/npm/chartjs-plugin-colorschemes@' . self::CHARTJS_COLORSCHEMES_VERSION . '/dist/chartjs-plugin-colorschemes.min.js', self::CHARTJS_COLORSCHEMES_HASH );
+				$t_cdn = 'https://cdn.jsdelivr.net/npm/%s@%s/dist/';
```
atrol: Would be good to have a constant for it and to use it also in function `csp_headers`

plugins/MantisGraph/MantisGraph.php
dregad: Not necessary. `$a[] = 1;` will properly initialize the array :
``` 
php > var_dump($a);
PHP Notice:  Undefined variable: a in php shell code on line 1
NULL
php > $a[]=1;
php > var_dump($a);
array(1) {
  [0] =>
  int(1)
}
```

library/README.md
dregad: I thought about it, but considering that MantisGraph is an official, bundled plugin, I left it there. I'm fine with removing it from this list if you prefer.

plugins/MantisGraph/MantisGraph.php
dregad: Done

library/README.md
dregad: @vboctor let me know if I you want me to remove the entries from the library/README.md file. Without reply from you, I'll leave them in.

PR #1534, opened 2019-08-02
New event_is_declared() function

dregad:
Fixes [#25953](https://mantisbt.org/bugs/view.php?id=25953)

### DIFF
--- a/core/event_api.php
+++ b/core/event_api.php
@@ -38,6 +38,17 @@ require_api( 'plugin_api.php' );
 $g_event_cache = array();
 
 /**
+ * Determine if a given event has been declared
+ * @param string $p_name Event name
+ * @return bool True if the even has been declared
+ */
+function event_is_declared( $p_name ) {
+	global $g_event_cache;
+
+	return isset( $g_event_cache[$p_name] );
+}
+
+/**
  * Declare an event of a given type.
  * Will do nothing if event already exists.
  * @param string  $p_name Event name.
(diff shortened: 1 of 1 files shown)


core/event_api.php
```diff
@@ -68,6 +68,17 @@ function event_declare_many( array $p_events ) {
...
 
+/**
+ * Determine if a given event has been declared
+ * @param string $p_name Event name
+ * @return bool True if the even has been declared
+ */
+function event_is_declared( $p_name ) {
+	global $g_event_cache;
+
+	return array_key_exists( $p_name, $g_event_cache );
```
atrol: Other functions use `isset` instead of `array_key_exists` to check if an event is declared.
e.g.

```
	if( !isset( $g_event_cache[$p_name] ) ) {
		error_parameters( $p_name );
		trigger_error( ERROR_EVENT_UNDECLARED, WARNING );
		return null;
	}
```
Worth to consider also the difference concerning `null` value treatment, e.g. check [this](https://coderwall.com/p/q9erfw/isset-vs-array_key_exists).

cproensa: > Other functions use isset instead of array_key_exists 

right, all the items in `$g_event_cache` are internally built and null is not an expected value.

dregad: I generally use array_key_exists() by default to be on the safe side, but considering @cproensa's comment I'll switch to isset.

dregad:
Force-pushed after updating the original commit taking review comments into consideration.

I also replaced usage of isset( $g_event_cache[...] ) elsewhere in event API, by calls to the new function.

dregad:
And force-pushed again because I accidentally removed a `!` in event_hook() function...

PR #1535, opened 2019-08-03
Fix Travis build

atrol:


dregad:
I merged #1536 for now, as it fixes the Travis builds for all PHP versions (including 5.5).

Your fix will still be needed when we switch to Xenial, but maybe it can be merged anyway now, I'll check if it works on Travis' trusty dist.

I have opened [#25956](https://mantisbt.org/bugs/view.php?id=25956) in the tracker to discuss increasing our minimum PHP requirement.

dregad:
Confirmed that adding the *services* setting in travis.yml works fine with `dist: trusty` as well
https://travis-ci.org/dregad/mantisbt/builds/567313782

vboctor:
I don't have context on this, so will defer to both of you.

dregad:
If we move forward with increasing minimum PHP version to 7.0, then it would make sense to include this change as part of that PR, instead of a standalone work item.

atrol:
No longer needed after 078044e5b3be354a9128d5e306c54b8aa7e537bb

PR #1538, opened 2019-08-05
Fix test case to expect one history entry

vboctor:
NOTE: not tested since SOAP tests not working on my dev machine at the moment.

When creating an issue with non-default status/resolution, check that a single entry is created.

Fixes #25962


dregad:
See my note https://mantisbt.org/bugs/view.php?id=25962#c62509

vboctor:
Closing this PR since it is now included in #1540

PR #1539, opened 2019-08-06
Don't force caching of form pages

cproensa:
These pages were explicitly setting a flag to make the pages cacheable.
Before the changes in 97b745dc102323c312ca27b6fcb8f838c3e50b8f
the expiration headers were not being correctly set, however after that
commit, the issue is fixed and these pages have become cacheable
This causes undesired effects.

Since the previos status of this scenario is that the pages were not
being cached anyway, we are removing the explicit $g_allow_browser_cache
flag.

Fixes: #25969

### DIFF
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -41,8 +41,6 @@
  * @uses version_api.php
  */
 
-$g_allow_browser_cache = 1;
-
 require_once( 'core.php' );
 require_api( 'access_api.php' );
 require_api( 'authentication_api.php' );
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -49,8 +49,6 @@
  * @uses version_api.php
  */
 
-$g_allow_browser_cache = 1;
-
 require_once( 'core.php' );
 require_api( 'access_api.php' );
 require_api( 'authentication_api.php' );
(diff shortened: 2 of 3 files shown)


dregad:
After this change, `$g_allow_browser cache` will no longer be referenced anywhere in the code, except in :

- *config_defaults_inc.php*, where the global variable's declaration is commented out (ever since it was introduced, back in 2005 actually... see 8f220341cefff44fb6322830166d54c8985dcb0e
- *html_api.php* - http_caching_headers() function

The question is, should this config (and possibly `$g_allow_file_cache` as well) be removed entirely ?

dregad:
Note: I have no objections with this change, which AFAICT is indeed fixing the caching issue (but I only performed some casual tests).

cproensa:
>  should this config (and possibly $g_allow_file_cache as well) be removed entirely ?

I'd say we can remove `$g_allow_browser_cache`
But `$g_allow_file_cache` can be seen as a more useful option, so there is a bigger chance that some users have it enabled at the moment.

And, while we are on this, what is your thoughts on my [comment about expiration date](https://mantisbt.org/bugs/view.php?id=25969#c62523)?
Should we set it to a shorter span, eg 1 day?

vboctor:
@cproensa @dregad I have merged this since it fixes the regression.  Let's follow up with a separate PR for cleanup suggestions.

PR #1540, opened 2019-08-09
Add missing PHPUnit tests to Travis CI

dregad:
Travis CI builds are not executing all of the defined PHPUnit test suites : _IssueHistoryTest_ and _REST tests_ are not executed.

This PR makes changes to both the PHPUnit scripts and the Travis configuration, to ensure all tests are run and we don't have skipped tests either.

### DIFF
--- a/core/api_token_api.php
+++ b/core/api_token_api.php
@@ -84,6 +84,24 @@ function api_token_hash( $p_token ) {
 }
 
 /**
+ * Checks that the specified token name is unique to the user.
+ *
+ * @param string $p_token_name The token name.
+ * @param string $p_user_id    The user id.
+ *
+ * @return bool True if unique, False if token already exists
+ */
+function api_token_name_is_unique( $p_token_name, $p_user_id ) {
+	db_param_push();
+	$t_query = 'SELECT * FROM {api_token} WHERE user_id=' . db_param() . ' AND name=' . db_param();
+	$t_result = db_query( $t_query, array( $p_user_id, $p_token_name ) );
+
+	$t_row = db_fetch_array( $t_result );
+
+	return $t_row === false;
+}
+
+/**
  * Ensure that the specified token name is unique to the user, otherwise,
  * prompt the user with an error.
  *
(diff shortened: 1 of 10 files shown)


dregad:
This includes @vboctor's fix in PR #1538.

dregad:
@vboctor I need some help with this one.

One of the REST tests is still failing (RestIssueAddTest::testCreateIssueWithTags) and I do not know how to fix that. I do not understand why the issue body is returned as NULL, when as far as I can tell we're executing the same command as in other test cases (e.g. _testCreateIssueWithVersionObjectId_), and the body is set correctly in that case.

Also, I had to comment out some of the assertions in _testCreateIssueWithTagIdNotFound_ and _testCreateIssueNoSummary_, because the REST API returns a 404 as expected, but I have no way of retrieving the created issue so I have nothing to compare against.

dregad:
@vboctor awating your feedback on this

vboctor:
@dregad This is probably caused by a bug in the `IssueAddCommand`.  The command expects the array of tags to be an array of arrays, each one containing an id and a name.  The sample expected usage can be found in `bug_report.php`.  The fix is to detect when ids are not provided and lookup the tag id from the provided name.

dregad:
@vboctor that was it indeed ! Thanks for pointing me in the right direction.

The code in IssueAddCommand is accessing $t_tag['id'] without first checking that the key exists, so when a tag is provided only by name, PHP throws a NOTICE causing the API to return an invalid JSON response (there is an HTML error message appended to it).

I logged https://mantisbt.org/bugs/view.php?id=25997 to track the bug.

You have not responded to the other issue I am facing
> I had to comment out some of the assertions in testCreateIssueWithTagIdNotFound and testCreateIssueNoSummary, because the REST API returns a 404 as expected, but I have no way of retrieving the created issue so I have nothing to compare against.

IMO, this is a design flaw of the API's current implementation, which first creates the issue, then creates/attaches tags to it as a separate, second step.

dregad:
Would it be the correct approach, to execute in IssueAddCommand::validate(), the same checks on tags to attach, as in mci_tag_set_for_issue() ?

dregad:
This is now ready to merge (after #1542).

PR #1541, opened 2019-08-10
Prevent email about private note to unprivileged users

dregad:
In email_collect_recipient(), the logic to exclude users who can't see
bugnotes relied on comparing the issue's last updated timestamp with the
bugnote's date.

Since these dates are not necessarily equal as they are updated
separately when a bugnote is added, this may result in a race condition
causing a notification e-mail about a new private bugnote to be sent to
users not authorized to see them.

Since email_collect_recipient()'s $p_bugnote_id parameter is always null
except for 'bugnote' notifications, the date check is not necessary; it
is sufficient to check that $p_bugnote_id is not null.

Fixes [#22898](https://mantisbt.org/bugs/view.php?id=22898)
Replaces PR #1299

### DIFF
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -331,13 +331,6 @@ function email_collect_recipients( $p_bug_id, $p_notify_type, array $p_extra_use
 	}
 
 	# add users who contributed bugnotes
-	$t_bugnote_id = ( $p_bugnote_id === null ) ? bugnote_get_latest_id( $p_bug_id ) : $p_bugnote_id;
-	if( $t_bugnote_id !== 0 ) {
-		$t_bugnote_date = bugnote_get_field( $t_bugnote_id, 'last_modified' );
-	}
-	$t_bug = bug_get( $p_bug_id );
-	$t_bug_date = $t_bug->last_updated;
-
 	$t_notes_enabled = ( ON == email_notify_flag( $p_notify_type, 'bugnotes' ) );
 	db_param_push();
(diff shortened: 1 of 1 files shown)


vboctor:
@dregad This shows my gravatar changes as part of this PR, can you adjust that.

dregad:
I probably messed up while rebasing, no idea how I did that :confused: 
Fixed.

vboctor:
@dregad I don't think this fix works.

It breaks the following case:
- Developer adds a private note.
- Sometime later the issue is updated.

Ideally, we need `$p_bugnote_id` to be passed in properly or infer the relationship via the the private note and issue having exactly the timestamp or within a small delta.

Preference for solutions:
1. Pass the `$p_bugnote_id` correctly in cases where bug update included a note.
2. Create the bugnote and update the issue with exactly same timestamp rather than now.
3. Implement a hack where the relationship is inferred via a small time delta.  This would break if developer creates a private note and then right after update the issue status.

dregad:
Thanks for the feedback @vboctor,

> I don't think this fix works.

You don't *think*, or did you actually test that it does fail in the described scenario ?

I did not re-test, but I believe I checked this scenario and If I'm not mistaken, in such case Mantis generates 2 distinct notifications :
- one for bugnote add (so *email_collect_recipients()* is called with $p_bugnote_id as expected), and 
- another one for issue update (where $p_bugnote_id is null, but it's a different $p_notify_type (`updated`) so it generates the default notification, which does not include the private notes contents

Let me know.

dregad:
@vboctor ping

PR #1542, opened 2019-08-14
RestIssueAdd fixes

dregad:
- [#25996](https://mantisbt.org/bugs/view.php?id=25996) Missing tag name in error message when creating issue via REST API
- [#25997](https://mantisbt.org/bugs/view.php?id=25997) Invalid JSON response when creating issue with tag by name via REST API

This is a pre-requisite to merge #1540


core/commands/IssueAddCommand.php
```diff
@@ -219,9 +219,9 @@ function validate() {
 
 		if( isset( $t_issue['tags'] ) && is_array( $t_issue['tags'] ) && !tag_can_create( $this->user_id ) ) {
 			foreach( $t_issue['tags'] as $t_tag ) {
-				if( $t_tag['id'] === -1 ) {
+				if( isset( $t_tag['id'] ) && $t_tag['id'] === -1 ) {
```
vboctor: We need to implement a method that takes a tag element and return an id or -1 if not found.

- If id supplied - verify it exists and return it.
- If name supplied (case insensitive), lookup the name and return its id.

The method can be designed to throw Client exception with appropriate error message based on what is supplied, (e.g. id, name, id + name, none).  The method should be used during validation and execution.

A similar, but not exactly the same (doesn't throw and used by SOAP hence transforms objects to array) method is `mci_get_project_id`.

dregad:
@vboctor following up on https://github.com/mantisbt/mantisbt/pull/1542#discussion_r314928269
> We need to implement a method that takes a tag element and return an id or -1 if not found.

see https://github.com/mantisbt/mantisbt/pull/1542/commits/66ee2db97f9776634fe7bb3193973e5c762de53c and let me know if this is what you have in mind.

dregad:
NOTE: I have not tested yet, but I think this should take care of my question in https://github.com/mantisbt/mantisbt/pull/1540#issuecomment-521247309

vboctor:
@dregad Note that if a user doesn't have access to create a tag and provides a tag name that doesn't exist, I believe we should treat it like providing a category that doesn't exist or equivalent, and return 404. Since the goal of the API is to refer to the tag and not explicitly to create one.

dregad:
OK that makes sense, will change accordingly.

dregad:
Branch has been updated as per our discussion. and PR should be ready to merge now. I'll confirm that after updating #1540.

The latest push fixes *bug_report.php*, which was failing to create new tags (because tag id == -1 is now considered invalid). The payload it sends to IssueAddCommand no longer contains the tag id, just the name.

The PR also includes a couple additional minor fixes / improvements for issues I stumbled upon while testing ([#26074](https://mantisbt.org/bugs/view.php?id=26074), [#26075](https://mantisbt.org/bugs/view.php?id=26075)).

dregad:
Merged via 6b774eac4893ae109b11138337ce8a50e3755691 for the main fix for the Command and c6727a8146fc90d9d7bdb6a3f2d2c03a77001c43, fa7262286cde0fded7d6331ceafa117a8dadb6d2 for the additional improvements.

PR #1543, opened 2019-08-15
Add email on monitor to config page

cproensa:
Email on "monitor" is an existing action, which is already configured in
config_defaults_inc, but it's not displayed in the email configuration
page.

Fixes: [#26002](https://www.mantisbt.org/bugs/view.php?id=26002)

### DIFF
--- a/lang/strings_english.txt
+++ b/lang/strings_english.txt
@@ -1479,6 +1479,7 @@ $s_email_on_deleted = 'E-mail on Deleted';
 $s_email_on_sponsorship_changed = 'E-mail on Sponsorship changed';
 $s_email_on_relationship_changed = 'E-mail on Relationship changed';
 $s_email_on_updated = 'E-mail on Updated';
+$s_email_on_monitor = 'E-mail when user starts monitoring';
 
 # Dynamic filters (load filter controls via AJAX upon request)
 $s_loading = 'Loading...';
--- a/core/email_api.php
+++ b/core/email_api.php
@@ -2032,7 +2032,7 @@ function email_shutdown_function() {
  * @return array List of actions
  */
 function email_get_actions() {
-	$t_actions = array( 'updated', 'owner', 'reopened', 'deleted', 'bugnote', 'relation' );
+	$t_actions = array( 'updated', 'owner', 'reopened', 'deleted', 'bugnote', 'relation', 'monitor' );
 
 	if( config_get( 'enable_sponsorship' ) == ON ) {
 		$t_actions[] = 'sponsor';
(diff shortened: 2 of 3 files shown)


lang/strings_english.txt
```diff
@@ -1479,6 +1479,7 @@ $s_email_on_deleted = 'E-mail on Deleted';
 $s_email_on_sponsorship_changed = 'E-mail on Sponsorship changed';
 $s_email_on_relationship_changed = 'E-mail on Relationship changed';
 $s_email_on_updated = 'E-mail on Updated';
+$s_email_on_monitor = 'E-mail on Monitor Added';
```
vboctor: How about something like `Email on User Monitoring Changes`?  We can brainstorm on the appropriate text, but the one in the PR is not ideal.  Is it also only when a user monitors an issue, or is it also trigger when a user unmonitors issue?  Does it apply to cases where a user adds others?

cproensa: >  Is it also only when a user monitors an issue, or is it also trigger when a user unmonitors issue? Does it apply to cases where a user adds others?

Called from the function bug_monitor() in bug_api, so whenever a new monitoring user is added.
Not when it is removed

> How about something like Email on User Monitoring Changes?

You are better with the english descriptions.
Based on it's usage, is it better as `Email on User Monitoring Added`?

dregad: "Email when someone starts monitoring"
"Email when user starts monitoring"

vboctor: I would personally prefer "Email on User Monitoring Changes" and make it trigger emails when `bug_monitor()` or `bug_unmonitor()` are called.  We should also update related documentation.

cproensa: > I would personally prefer "Email on User Monitoring Changes" and make it trigger emails when bug_monitor() or bug_unmonitor() are called. We should also update related documentation.

I didn't plan changing the underlying behavior. However, this may be the less useful notification type (and is all OFF by default), so that wouldn't have much impact.
But note that, in case of going that way, the email sent to the user that is added as monitor is not managed with this setting (it's sent as type "explicit"). Do you suggest to also email explicitly the target user when is removed from monitoring?
I'd say, to not change that, becasue unmonitoring usually means i don't want more notifications from that issue.

dregad: > unmonitoring usually means i don't want more notifications from that issue.

True, but user wanting to be notified about unmonitoring event probably depends on whether it's the user's own action, or if another person (eg. admin) removed them from monitoring list...

dregad: @vboctor I don't think we should change the system's behavior (i.e. send notifications on "unmonitoring") as part of this PR. If this is indeed a requirement (and I don't believe it is), then it should be tracked as a separate work item. Let's stick to the PR's current scope for now. 

Considering that there was no update on the label discussion for nearly a month, do you agree to keep @cproensa's original wording, or would you prefer one of my suggestions ?

My favorite is "Email when user starts monitoring"

cproensa: I have updated with @dregad suggestion: `E-mail when user starts monitoring`

PR #1544, opened 2019-08-15
Several fixes related to custom fields

cproensa:
Fixes for:
[0025972: Use custom field regular expression in the html input](https://www.mantisbt.org/bugs/view.php?id=25972)
[0025975: Manage custom fields page does not show fields in order](https://www.mantisbt.org/bugs/view.php?id=25975)


cproensa:
@dregad 
i have removed from this PR the changes related to: [0025973: custom_field_set_value() does not apply $p_log_insert](https://www.mantisbt.org/bugs/view.php?id=25973)
To not block other changes, and then move the discussion to the tracker.

PR #1545, opened 2019-08-20
Improve UX & consistency for inline action icons

syncguru:
Fixes #25905

Before:
delete-tag

email-reporter

image-actions

Screen Shot 2019-08-19 at 7 30 54 PM

After:
Screen Shot 2019-08-19 at 7 24 33 PM

Screen Shot 2019-08-19 at 7 26 19 PM

Screen Shot 2019-08-19 at 7 25 18 PM

Screen Shot 2019-08-19 at 7 35 19 PM

Screen Shot 2019-08-19 at 7 33 20 PM

### DIFF
--- a/bug_monitor_list_view_inc.php
+++ b/bug_monitor_list_view_inc.php
@@ -93,12 +93,12 @@ if( $t_can_see_monitors || $t_can_add_others ) {
 			$t_can_delete_others = access_has_bug_level( config_get( 'monitor_delete_others_bug_threshold' ), $f_bug_id );
 			if( $t_can_delete_others ) {
 				$t_button = '&nbsp;'
-					. '<a class="btn btn-xs btn-primary btn-white btn-round" '
+					. '<a class="red noprint zoom-130" '
 					. 'href="' . helper_mantis_url( 'bug_monitor_delete.php' )
 						. '?bug_id=' . $f_bug_id . '&amp;'
 						. 'user_id=%s'
 						. htmlspecialchars( form_security_param( 'bug_monitor_delete' ) )
-					. '"><i class="fa fa-times"></i></a>';
+					. '"><i class="ace-icon fa fa-trash-o bigger-115"></i></a>';
 			}
 
 			foreach( $t_users as $t_user ) {
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -376,8 +376,8 @@ if( $t_show_attachments ) {
 					<?php print_profile_option_list( auth_get_current_user_id(), $f_profile_id ) ?>
 				</select>
 			<?php } ?>
-			<?php collapse_open( 'profile' ); collapse_icon( 'profile' ); ?>
-			<?php echo lang_get( 'or_fill_in' ); ?>
+			<?php collapse_open( 'profile' ); ?>
+			<?php echo lang_get( 'or_fill_in' ); collapse_icon( 'profile' ); ?>
 			<table class="table-bordered table-condensed">
 				<tr>
 					<th class="category" width="30%">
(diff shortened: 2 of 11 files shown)


vboctor:
@syncguru can you please rebase on master.

syncguru:
@vboctor done. 
I think the new markup is breaking travis-ci validation.

cproensa:
Files that don't have a preview are shown differently, at least: different padding for the attachement element. They look skewed when compared.
I'm not sure if not having a lighter background box is intended (as not having a preview).

imagen

other impressions:
- I feel the collapse arrow is a bit too close to the filenames/size text .
- And i tend to go to click in the empty space between the arrow and the trashcan, as expecting that it would also expand the preview (but only the arrow is clickable to expand)
- some kind of hover effect for the attachment area? it would help when aiming for the right trashcan on several stacked attachments.
- The whole horizontal area for the attachment looks too wide/narrow in a full screen. But this is a general layout issue, not specific to this PR. However, it can be bad when the trashcan icon is so far to the right that it gets lost. Maybe some limitation in the grid for lg, xl widths can be applied?

syncguru:
@cproensa:
> Files that don't have a preview are shown differently

Good catch

> I'm not sure if not having a lighter background box is intended

Yes, it is intended and so it addresses your need of a hover of the raw

>I feel the collapse arrow is a bit too close to the filenames/size text

Will add more spacing


@dregad:
> I don't see the benefit of putting the trashcan to the right

I am not sure how much use `delete` action gets. I suspect that most of the usage is for expanding and collapsing the preview. Having a `delete` very close to expand\collapse felt like giving prominence to unlikely action which can easily be clicked by mistake due to proximity. Happy to revert it back but I thought I should present my case first.

> we're not deleting the tag, just detaching it from the issue

Good point. I will switch back. I have a bigger change to tagging in the works so I will defer any improvements until then.

cproensa:
> I am not sure how much use delete action gets. I suspect that most of the usage is for expanding and collapsing the preview. Having a delete very close to expand\collapse felt like giving prominence to unlikely action which can easily be clicked by mistake due to proximity.

I also dont like it placed next to the expand icon.
That's why i though of reducing the width of the boxes, so the icon is still at the right side, but is easier to locate it visually.

> Yes, it is intended and so it addresses your need of a hover of the raw

The hover suggestion was related to having the thrash icon so far to the right. If the box is more compact, that would not be necessary because everything is more focused in the field of view

syncguru:
> If the box is more compact
Since we preview the image inline (in the expanded block), making the box compact will reduce the preview space. If we move the image prereview to something like lightbox, then attachments could be just a series of small boxed lined beside each other.

cproensa:
>  If we move the image prereview to something like lightbox, then attachments could be just a series of small boxed lined beside each other.

The "inline" placing of boxes would be more compact. Slightly, resembling, the dropzone way of placing attachment boxes.
But i'm not suggesting going to that extreme. You are right, that a narrow box limits the preview space. In this scenario i was thinking of the boxes still being one per row, but the box is expanded to full width when the preview is enabled

syncguru:
@dregad ping.

dregad:
@syncguru sorry for the delayed response.

>> I don't see the benefit of putting the trashcan to the right
>
> [...] I suspect that most of the usage is for expanding and collapsing the preview. 

That's probably true.

> Having a delete very close to expand\collapse felt like giving prominence to unlikely action which can easily be clicked by mistake due to proximity. 

Fair enough, but while this layout if fine on smaller screens, on a "normal" one (say, 1280x1024) with a maximized browser window, the trashcan icon is too far on the right, making it feel "disconnected" from the attachment in my opinion.

image

I stand by my preference for having it closer to the other controls, but considering that I seem to be in minority here since @cproensa agrees with your perspective, I guess I'll learn to live with it :wink:.


Before the PR can be merged, please **REBASE** it on top of latest master and resolve the conflicts. Kindly note that merging is not equivalent as we try to maintain a more linear Git history.

syncguru:
Thanks, @dregad - rebase is done.

dregad:
Thanks. This pull request has been integrated manually.

PR #1550, opened 2019-08-23
 PHP 7.4 compatibility: remove deprecated usage of implode() 

dregad:
And some code cleanup...

Fixes [#26063](https://mantisbt.org/bugs/view.php?id=26063)

### DIFF
--- a/manage_plugin_page.php
+++ b/manage_plugin_page.php
@@ -137,7 +137,7 @@ foreach ( $t_plugins_installed as $t_basename => $t_plugin ) {
 
 	if( !empty( $t_author ) ) {
 		if( is_array( $t_author ) ) {
-			$t_author = implode( $t_author, ', ' );
+			$t_author = implode( ', ', $t_author );
 		}
 		if( !is_blank( $t_contact ) ) {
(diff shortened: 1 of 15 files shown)


manage_plugin_page.php
```diff
@@ -172,7 +172,7 @@ function ( $p_p1, $p_p2 ) {
 	}
 
 	if( 0 < count( $t_depends ) ) {
-		$t_depends = implode( $t_depends, '<br />' );
+		$t_depends = implode( '<br>', $t_depends );
```
vboctor: Why are you changing the br tag here?

manage_plugin_page.php
```diff
@@ -302,7 +302,7 @@ function ( $p_p1, $p_p2 ) {
 		}
 
 		if( 0 < count( $t_depends ) ) {
-			$t_depends = implode( $t_depends, '<br />' );
+			$t_depends = implode( '<br>', $t_depends );
```
vboctor: Why are you changing the br tag here?

dregad:
`<br>` is HTML while `<br />` is XHTML syntax. Since we moved away from that in 1.3.0-beta.1 and we serve our content as HTML5, it makes sense to use the simpler syntax.

PR #1551, opened 2019-08-25
Fix filter for date custom field

cproensa:
When a custom field of type date is used in the filter dialog, and there
are no values, for any issue, for this custom field, the filter shows an
error and don't populate the date picker.

Fixes: #26062

### DIFF
--- a/core/filter_form_api.php
+++ b/core/filter_form_api.php
@@ -2112,20 +2112,14 @@ function print_filter_custom_field_date( $p_field_id, array $p_filter = null ) {
 	$t_included_projects = filter_get_included_projects( $p_filter );
 	$t_values = custom_field_distinct_values( $t_cfdef, $t_included_projects );
 
-	# Resort the values so there ordered numerically, they are sorted as strings otherwise which
-	# may be wrong for dates before early 2001.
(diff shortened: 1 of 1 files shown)


core/filter_form_api.php
```diff
@@ -2112,20 +2112,14 @@ function print_filter_custom_field_date( $p_field_id, array $p_filter = null ) {
...
 	$t_sel_end_year = null;
-	if( isset( $t_values[0] ) ) {
-		$t_sel_start_year = date( 'Y', $t_values[0] );
-	}
-	$t_count = count( $t_values );
-	if( isset( $t_values[$t_count - 1] ) ) {
-		$t_sel_end_year = date( 'Y', $t_values[$t_count - 1] );
+	if( is_array( $t_values ) && !empty( $t_values ) ) {
+		# Resort the values so there ordered numerically, they are sorted as strings otherwise which
+		# may be wrong for dates before early 2001.
```
dregad: Suggest rewording 
_Resort the values so they are ordered numerically (otherwise they are treated as strings, which may be wrong for dates before early 2001_

cproensa: done

PR #1552, opened 2019-08-25
configure background and foreground color for custom fields

fcaspar:
https://mantisbt.org/bugs/view.php?id=26080
https://mantisbt.org/forums/viewtopic.php?f=9&p=68706


admin/schema.php
```diff
@@ -884,6 +884,11 @@
 
 # Release marker: 1.3.0
 
+$g_upgrade[210] = array( 'AddColumnSQL', array( db_get_table( 'mantis_custom_field_table' ), "
+	color_background     C(8)		NOTNULL DEFAULT \" '' \",
```
atrol: Please explain why you set length to 8. I would have expected 7.

atrol: Why are NULL values not allowed?
I would use NULL / empty for not defined colors.

core/custom_field_api.php
```diff
@@ -571,6 +571,16 @@ function custom_field_update( $p_field_id, array $p_def_array ) {
 		trigger_error( ERROR_CUSTOM_FIELD_NAME_NOT_UNIQUE, ERROR );
 	}
 
+  # if foreground color is set to default (black), do not store it
+  if ( $p_def_array['color_foreground'] == '#000000') {
```
atrol: Do not use magic strings to find out if a value is set or not.
See comment above concerning NULL values.
Keep also in mind that black/white are not always the default colors.

manage_custom_field_edit_page.php
```diff
@@ -108,6 +108,22 @@
 			</select>
 		</td>
 	</tr>
+  <tr>
+      <td class="category">
+          <?php echo lang_get( 'custom_field_color_foreground' ) ?>
+      </td>
+      <td>
+          <input type="color" id="custom-field-color-foreground" name="color_foreground" size="8" value="<?php echo $t_definition['color_foreground'] != '' ? $t_definition['color_foreground'] : '#000000' ?>" />
```
atrol: Didn't try, but seems that color input type is not supported in IE, which is still supported in Mantis until now.
https://caniuse.com/#feat=input-color

Maybe you can try, or I will try later.
I think it's ok if there is a text input field instead of a color picker in IE.

lang/strings_german.txt
```diff
@@ -1010,6 +1010,8 @@ $s_custom_field = 'Feld';
 $s_custom_field_label = 'Feld:';
 $s_custom_fields_setup = 'Benutzerdefinierte Felder';
 $s_custom_field_name = 'Name';
+$s_custom_field_color_foreground = 'Vordergrundfarbe';
+$s_custom_field_color_background = 'Hintergrundfarbe';
```
atrol: Don't change any other language file than English in code repository.

Translations are managed via TranslateWiki, which means that TranslateWiki checks about every two weeks for changes in English files.
New messages will appear on TranslateWiki as not translated, giving Mantis users the chance to translate them.
The translated messages are commited to our repository by TranslateWiki.

Please refer our own [Wiki](http://www.mantisbt.org/wiki/doku.php/mantisbt:translationshowto) for the process to submit new translations or revise existing ones.

admin/schema.php
```diff
@@ -884,6 +884,11 @@
 
 # Release marker: 1.3.0
 
+$g_upgrade[210] = array( 'AddColumnSQL', array( db_get_table( 'mantis_custom_field_table' ), "
```
atrol: Should not be `mantis_custom_field_table` but `custom_field`

admin/schema.php
fcaspar: changed to 7

fcaspar: changed. NULLs are allowed now.

lang/strings_german.txt
fcaspar: German language file reverted.
Didn't know that. Thanks for explanation.

admin/schema.php
fcaspar: changed to `custom_field`

core/custom_field_api.php
fcaspar: removed input type color. set to plain text input.
therefore, magic strings got obsolete and were also removed.

manage_custom_field_edit_page.php
fcaspar: Correct. Color input doesn't work in IE.
Removed input type color. Replaced by plain text input.

fcaspar:
> Thanks @fcaspar for your contribution
Thanks for reviewing. ;)

>     * conflicts with latest code have to be resolved
Is that a task for me? Could you please clarify?

>     * missing documentation in Admin Guide
Please recheck other points. If they are fine, I'll update documentation.

>     * why do you use color_background / color_foreground instead of the usual background_color / Changed.

>     * the colors are not applied to the `View Issues` page, maybe even more (didn't check)
I changed bug_view_inc to consider the colors. Did I miss something?

>     * we use different colors for field labels and field values. After your change, they are the same.
Correct. Open point.

>     * usage of inline styles
>       Inline styles should be avoided.
Inline Styles: I have no other idea how to affect a field's style individually. Do you have a suggestion?

atrol:
It seems something that went wrong with your latest changes.
E.g. there are changes in lang/strings_english.txt and manage_custom_field_edit_page.php that are not related to the new functionality.

Please make sure that your submissions adhere to our Coding Guidelines http://www.mantisbt.org/wiki/doku.php/mantisbt:coding_guidelines
E.g. add a space before closing braces 

>>     * conflicts with latest code have to be resolved

> Is that a task for me? Could you please clarify?

Typically yes.


>>     * why do you use color_background / color_foreground instead of the usual background_color / 

> Changed.

Not changed in schema.php

>>     * the colors are not applied to the `View Issues` page, maybe even more (didn't check)

> I edited only bug_view_inc to consider the colors. Did I miss something?

Custom fields can also be displayed on "View Issues" page if they are included via columns configuration.

>>       Inline styles should be avoided.

> Inline Styles: I have no other idea how to affect a field's style individually. Do you have a suggestion?

It could be implemented similar to the status color, but I am not sure if it's worth the effort.
That's why I hoped to get some feedback of other devs.

@dregad @vboctor what's your opinion concerning usage of inline styles ?
At the moment we allow them, but it will not work any longer if CSP headers should be changed back to what we had in Mantis 1.3
https://mantisbt.org/bugs/view.php?id=21908

syncguru:
-1 - I am strongly against this change for the following reasons:
1. Why treating custom fields any different from default ones. This seems arbitrary and not articulated at all
2. UX consistency is as important as code if not more important. Allowing any random color in the page will clash with the ACE theme and impact UX. ACE comes with a set of compatible colors that we can use (we use across the product) - the selection is less than 10 colors.
3. No inline CSS - please
4. Storing colors in the DB?! Really??

atrol:
> -1 - I am strongly against this change

@syncguru  I am also not 100% convinced that the current approach of the PR is what should be included in standard.
But strongly against?

> Why treating custom fields any different from default ones.

Right, in best case default fields are nothing more than preconfigured custom fields.
But current Mantis is quite far away from this approach.

> Allowing any random color in the page will clash with the ACE theme and impact UX.
> ACE comes with a set of compatible colors that we can use (we use across the product) - the selection is less than 10 colors.

We allow random colors for status squares.

We can't prevent users from changing whatever they want.
I have seen JavaScript plugins and other tweaks that lead to an ugly Mantis 2.x in my opinion, but there are users that like it.

Keep also in mind that there are users that don't like the ACE theme / colors due to various reasons.
e.g. it violates https://www.w3.org/TR/WCAG20/ concerning contrast ratio
https://mantisbt.org/bugs/view.php?id=22253

So in best case we offer different themes and some more predefined colors to choose for other purposes, e.g. coloring of (custom) fields.

> No inline CSS - please

+1 that's why I opened https://mantisbt.org/bugs/view.php?id=21908
Search the code for "style=" and "<style>>. You will find the dev who introduced most of the inline styles in 2.x ;-)

> Storing colors in the DB?! Really??

It's not that unusual in Mantis. We allow storing status colors in configuration file and DB.

atrol:
Closed based on feedback of other devs.
@fcaspar sorry, but you have to maintain your own fork / branch to keep this functionality.

For other users that might think about using these changes on their own Mantis instance:
There are database schema changes that conflict with those ones in standard Mantis (admin/schema.php), so you will get problems when trying to upgrade Mantis.

PR #1555, opened 2019-08-29
Support inline playing of audio/video attachments

vboctor:
Fixes for:

- 26095: Support inline playing of audio attachments
- 26102: Support inline playing of video attachments
- 26096: preview_*_extensions config options not documented
- Show proper icons for video/audio files.
- Show proper icon for markdown files.

Note the following behaviors:

- Audio/Video/Images are expanded by default.
- If audio/video file size is greater than preview size, show the control but don't preload the content.
- If audio/video file size is smaller than preview size, then preload.

Audio file attachments while collapsed:

image

Audio file attachments while expanded in Firefox:

image

Audio file attachments while expanded in Chrome:

image

Audio file attachments while expanded in Safari:

image


atrol:
While you are on it, what about implementing also `<video>` , `$g_preview_video_extensions`?

https://caniuse.com/#feat=video
https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_video

dregad:
Instead of multiplying the configs for each type (preview_XXX_extensions), what about refactoring by regrouping them under an array, with the type as key ?
```
[
  'text' => [ 'txt', ... ],
  'image' => [ 'jpg', ... ],
  ...
```

core/print_api.php
```diff
@@ -2006,6 +2008,23 @@ function print_bug_attachment_preview_image( array $p_attachment ) {
 	echo '</a></div>';
 }
 
+/**
+ * Prints the preview of an audio file attachment.
+ * @param array $p_attachment An attachment array from within the array returned by the file_get_visible_attachments() function.
+ * @return void
+ */
+function print_bug_attachment_preview_audio( array $p_attachment ) {
+	$t_file_url = $p_attachment['download_url'] . '&show_inline=1' . form_security_param( 'file_show_inline' );
```
syncguru: It would be great to update the `print_bug_attachment_header` function to show font-awesome icon `fa-file-audio-o` for audio files.
https://fontawesome.com/v4.7.0/icon/file-audio-o

vboctor:
@atrol I can add support for video as well.

@syncguru I will change the code to show the appropriate icon for video/audio.

@dregad I also thought about the same thing.  Here are the options:

1. Deprecate the old configs and add a new one with the suggested format.
2. Use the mime type instead of the extensions (e.g. text/*, image/*, audio/*, video/*) and deprecate the config options or at least not add new ones.

I'm leaning towards using mime type at least for video and audio, or we can do for all.  We can in the future add a "don't preview extensions" config if there are ones that we don't want to preview.

Thoughts?

vboctor:
Code updated per review comments.

@atrol video is now supported.
@syncguru file icons fixed.
@dregad the code now uses mime types instead of depending on extensions to recognize file types.  I didn't remove the old config options.

core/print_api.php
vboctor: done.

core/print_api.php
```diff
@@ -1889,18 +1894,42 @@ function print_bug_attachment( array $p_attachment, $p_security_token ) {
...
 		if( $p_attachment['type'] == 'text' ) {
 			print_bug_attachment_preview_text( $p_attachment );
 		} else if( $p_attachment['type'] === 'image' ) {
 			print_bug_attachment_preview_image( $p_attachment );
+		} else if( $p_attachment['type'] === 'audio' ) {
+			print_bug_attachment_preview_audio( $p_attachment, $p_attachment['file_type'], $p_attachment['preview'] );
+		} else if( $p_attachment['type'] === 'video' ) {
+			print_bug_attachment_preview_video( $p_attachment, $p_attachment['file_type'], $p_attachment['preview'] );
 		}
+
```
dregad: Use a switch statement here

core/print_api.php
```diff
@@ -1889,18 +1894,42 @@ function print_bug_attachment( array $p_attachment, $p_security_token ) {
...
+		# Audio / Video support showing control without preloading.
+		if( $p_attachment['type'] === 'audio' || $p_attachment['type'] === 'video' ) {
+			echo lang_get( 'word_separator' );
+			collapse_icon( $t_collapse_id );
+	
+			if( $p_attachment['type'] === 'audio' ) {
+				print_bug_attachment_preview_audio( $p_attachment, $p_attachment['file_type'], $p_attachment['preview'] );
+			} else if( $p_attachment['type'] === 'video' ) {
+				print_bug_attachment_preview_video( $p_attachment, $p_attachment['file_type'], $p_attachment['preview'] );
+			}
```
dregad: ditto

core/print_api.php
```diff
@@ -2006,6 +2035,46 @@ function print_bug_attachment_preview_image( array $p_attachment ) {
 	echo '</a></div>';
 }
 
+/**
+ * Prints the preview of an audio file attachment.
+ * @param array $p_attachment An attachment array from within the array returned by the file_get_visible_attachments() function.
```
dregad: Break long line

core/print_api.php
```diff
@@ -2006,6 +2035,46 @@ function print_bug_attachment_preview_image( array $p_attachment ) {
...
+	echo '<audio controls="controls"' . $t_preload . '>';
+	echo '<source src="' . string_attribute( $t_file_url ) . '" type="' . string_attribute( $p_file_type ) . '">';
+  	echo lang_get( 'browser_does_not_support_audio' );
+	echo "</audio>";
+	echo "</a></div>";
+}
+
+/**
+ * Prints the preview of an video file attachment.
+ * @param array $p_attachment An attachment array from within the array returned by the file_get_visible_attachments() function.
```
dregad: Break long line

core/print_api.php
```diff
@@ -2006,6 +2035,46 @@ function print_bug_attachment_preview_image( array $p_attachment ) {
...
+ * @param string $p_file_type mime type
+ * @param boolean $p_preload true to preload video, false otherwise.
+ * @return void
+ */
+function print_bug_attachment_preview_video( array $p_attachment, $p_file_type, $p_preload ) {
```
dregad: This is pretty much the same code as *print_bug_attachment_preview_audio()* the only difference is `s/audio/video/`, wouldn't it make sense to have a generic function ?

vboctor:
@syncguru Some of the file formats have a video mime type and some don't.  The one that don't don't end up with a video tag.  The ones that do are not all playable by the browser.  I suspect not all browsers will also support the exact set of video types on all platforms.  Hence, nothing to do here.

@dregad I applied all your comments.  I will not remove the extension confirmation as part of this PR.  In fact, I am not sure that long term we will not end up adding configs to the video/audio scenario.  Hence, I want to put this out and learn how it goes.

PR #1563, opened 2019-09-13
Check that user_pref.redirect_delay is integer and convert it if not

dregad:
Fixes [#26109](https://mantisbt.org/bugs/view.php?id=26109)


dregad:
Without feedback, I assume everyone's OK with this and will proceed with merging this week-end.

PR #1564, opened 2019-09-13
Display MantisGraph pages under their own menu

dregad:
Fixes [#26139](https://mantisbt.org/bugs/view.php?id=26139)

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -602,45 +602,83 @@ function print_menu( array $p_menu_items, $p_current_page = '', $p_event = null 
 }
 
 /**
- * Print the menu for the graph summary section
+ * Print a generic submenu (buttons group).
+ *
+ * @param array  $p_menu_items   List of menu items
+ * @param string $p_current_page Current page's file name to highlight active tab
+ * @param string $p_event        Optional event to signal,
+ */
+function print_submenu( array $p_menu_items, $p_current_page = '', $p_event = null ) {
+	# Plugin / Event added options
+	$t_plugin_menu_items = plugin_menu_items( $p_event );
+
+	if( $p_menu_items || $t_plugin_menu_items ) {
+		echo '<div class="space-10"></div>';
+		echo '<div class="col-md-12 col-xs-12 center">';
+		echo '<div class="btn-group">', "\n";
+
+		$t_btn_template = '<a class="btn btn-sm btn-primary btn-white %s" href="%s">%s%s</a>' . "\n";
+
+		foreach( $p_menu_items as $t_item ) {
+			if( is_array( $t_item ) ) {
+				$t_active = $p_current_page && strpos( $t_item['url'], $p_current_page ) !== false
+					? 'active' : '';
+				$t_icon = array_key_exists( 'icon', $t_item )
+					? '<i class="fa ' . $t_item['icon'] . '"></i>&nbsp;'
+					: '';
+
+				printf( $t_btn_template,
+					$t_active,
+					$t_item['url'],
+					$t_icon,
+					lang_get_defaulted( $t_item['label'] )
+				);
+			} else {
+				# Cooked link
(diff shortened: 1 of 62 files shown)


dregad:
Without feedback, I assume everyone's OK with this and will proceed with merging this week-end.

atrol:
Tested a bit, looks good.
Hope to be able to invest some time also in code review.

I am not sure if it's a good idea to change non-english language files, as I don't know how Translatewiki works.

You missed to remove `$s_plugin_MantisGraph_synthesis_link = 'Sintezo';` from `strings_eo.txt`

core/html_api.php
```diff
@@ -564,35 +564,112 @@ function print_subproject_menu_bar( $p_current_project_id, $p_parent_project_id,
...
-					$t_menu_options[] = $t_callback_menu_options;
-				}
+				# Cooked link
+				echo $t_item;
 			}
 		}
+
+		# Plugins menu items - these are html hyperlinks (<a> tags)
+		foreach( plugin_menu_items( $p_event ) as $t_item ) {
+			$t_active = $p_current_page && strpos( $t_item, $p_current_page ) !== false
```
atrol: `$t_active` is not used

dregad: Fixed

dregad:
> I am not sure if it's a good idea to change non-english language files, as I don't know how Translatewiki works.

@mantisbt/translatewiki-net can you please confirm if this is OK or not ?

> You missed to remove $s_plugin_MantisGraph_synthesis_link = 'Sintezo'; from strings_eo.txt

Fixed

dregad:
@translatewiki your feedback would be appreciated.

dregad:
> I am not sure if it's a good idea to change non-english language files, as I don't know how Translatewiki works.

See [Discussion on translatewiki.net support channel](https://translatewiki.net/wiki/Support#Deleting_unused_strings_from_translated_language_files_56747
).

atrol:
Translatewiki wrote

> If you delete the English strings (the source strings), the next export will take care of deleting the translations as well. 

Seems not to be true.
E.g. when running admin/test_langs.php you get warnings like
```
Testing language file 'strings_afrikaans.txt' (phase 1)...
WARNING: '$s_plugin_MantisGraph_enabled' is not defined in the English language file
WARNING: '$s_plugin_MantisGraph_disabled' is not defined in the English language file
```

siebrand:
| If you delete the English strings (the source strings), the next export will take care of deleting the translations as well.

It is true, and there is a but :). But we only export language files when changes have been made to it recently. So if no one makes changes to Dutch for example, the keys that no longer exist in the English file, will not be cleaned from the Dutch file, simply because we're not (re-)generating it.

My advise is to no worry about superfluous entries in non-English language files if they are not in the way of a correctly functioning application.

dregad:
Thanks for the feedback @siebrand. So if I understand correctly, you recommend not to edit foreign language files (but it does not hurt if we do).

Since you're here, there is still one question that needs answering from @translatewiki (as [posted on translatewiki.net support forum](https://translatewiki.net/wiki/Thread:Support/Deleting_unused_strings_from_translated_language_files/reply_(2))): 

> [renaming a language string](https://github.com/mantisbt/mantisbt/pull/1564/commits/0296fd0f782e7fa258e323471890fba4f1ca33f0) (i.e. not the translation itself, but the string's identifier) ==> this is the one I'm really not sure about. I would think yes based on what the FAQ says, but since it also mentions informing the staff...

Abijeet:
Hi @dregad 

Answered on the support ticket, but here's the summary,

> _renaming a language string_ --> This is also fine. Thanks for the ping regarding this. We will review and do the necessary renaming on our end, which is quite easy.

Regards,

dregad:
@Abijeet great, thanks for the feedback !

PR #1565, opened 2019-09-15
Improve handling of invalid plugins

dregad:
- Add Invalid plugins section on Manage Plugins page
- Also adds admin checks to detect the same issues.

Includes a major refactoring of Manage Plugins page using objects, to simplify the code and avoid  duplication.

Fixes [#26142](https://mantisbt.org/bugs/view.php?id=26142), [#17487](https://mantisbt.org/bugs/view.php?id=17487), [#26143](https://mantisbt.org/bugs/view.php?id=26143)


dregad:
See [#26142](https://mantisbt.org/bugs/view.php?id=26142), for before/after screenshots.

dregad:
Any comments before I merge this ?

cproensa:
@dregad  would like to have a look at this, if i get the time. I hope to do this week.

cproensa:
Regarding new classes: i find our current class directory a growing mess.
Maybe you could find interesting my approach for namespaces+subdirs:

https://github.com/mantisbt/mantisbt/pull/1531/commits/db09198510d92cbaaa4a62e523d1e7c788e5ee11

dregad:
> I would like to have a look at this, if i get the time. I hope to do this week.

I thought you would... :wink: OK I'll wait for your review then.

> i find our current class directory a growing mess.

Agreed. Ideally we should switch to using [PSR-4](https://www.php-fig.org/psr/psr-4/), but this would be no small undertaking...

Will have a look at your commit, I was not aware of it. It's strange, because it seems to be in core repository, yet I don't have it in my local clone :confused:

cproensa:
> It's strange, because it seems to be in core repository, yet I don't have it in my local clone confused

Because its part of a pull request: https://github.com/mantisbt/mantisbt/pull/1531
you can see the structure of the new classes added in that PR.
In your case, the old MantisPlugin class can't be relocated without breaking compatibility, but the new classes could be namespaced, to group them in a common context.

dregad:
> > I would like to have a look at this, if i get the time. I hope to do this week.

@cproensa did you get a chance to review ?

core/plugin_api.php
```diff
@@ -51,14 +51,31 @@
 require_api( 'lang_api.php' );
 require_api( 'logging_api.php' );
 
-# Cache variables #####
-
+/**
+ * Plugins cache
+ * @global MantisPlugin[] $g_plugin_cache
```
cproensa: Now that you are adding better documentation to these variables, i'd find useful to specify the index and value structure
Such as making clear that this array is indexed by the plugin base names.
i don't know the best syntax for that. An example i think i have used before, similar to:
` array( 'plugin_basename' => MantisPlugin, ... )`

core/plugin_api.php
```diff
@@ -739,13 +755,19 @@ function plugin_upgrade( MantisPlugin $p_plugin ) {
...
+ * status.
  * @param MantisPlugin $p_plugin Plugin basename.
+ *
  * @return void
  */
 function plugin_uninstall( MantisPlugin $p_plugin ) {
 	access_ensure_global_level( config_get_global( 'manage_plugin_threshold' ) );
 
-	if( !plugin_is_installed( $p_plugin->basename ) || plugin_protected( $p_plugin->basename ) ) {
+	if( !$p_plugin instanceof MissingClassPlugin &&
```
cproensa: I see how the different subclasses are being used. I'm not saying it's wrong.
I'm thinking, what about a different approach:
- where the validity status of the plugin is contained in the base MantisPlugin class
- those subclasses override that validity status
- The query for validity is used from the base class, don't need to know every subclass.

This would also allow a feature for a valid plugin, to manage it's own validity status, if it needs to perform specific test in order to work properly. Eg: valid configuration, additional resources, etc.

admin/check/check_plugins_inc.php
```diff
@@ -0,0 +1,114 @@
...
+			array(
+				false => "Plugin could not be loaded; "
+					. sprintf( $t_manage_plugins_link, '#invalid', 'remove the Plugin' )
+					. " or reinstall its source code."
+			)
+		);
+	} else {
+		$t_msg = "Contact the Plugin's author.";
+		if( $t_plugin instanceof MissingClassPlugin ) {
+			$t_msg = "Rename the Plugin's directory or " . $t_msg;
```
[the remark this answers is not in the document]
dregad: OK, makes sense

core/plugin_api.php
dregad: OK, will do

core/plugin_api.php
dregad: Just to make sure if I understand properly, you want to have new method(s)s in the base MantisPlugin class, like *isValid()* to further encapsulate the logic within the class instead of doing it within plugin API ?

cproensa: it's not that i want it. It's an idea, just to discuss if it may make sense, or not worth it.

dregad: Sorry, bad wording, I should not have said "want"... would like or mean would have been more appropriate.

Anyway, is my understanding of your idea correct or not ?

cproensa: > , is my understanding of your idea correct or not ?

yes 

> have new method(s)s in the base MantisPlugin class, like isValid() to further encapsulate the logic within the class instead of doing it within plugin API ?

I was thinking that the object provides both a "status" and an optional message, which more or less is what is being derived from each of the invalid plugin subclasses.

> isValid()

i dont kow if it's better a method like that, or just some class properties.

dregad: Fixed.

dregad:
This is now ready to go ! Testing and comments welcome. 
Without feedback by Wednesday, I'll proceed with the merge.

> Regarding new classes: i find our current class directory a growing mess.
> Maybe you could find interesting my approach for namespaces+subdirs:
> 
> [db09198](https://github.com/mantisbt/mantisbt/commit/db09198510d92cbaaa4a62e523d1e7c788e5ee11)

@cproensa I will take care using namespaced classes separately

core/plugin_api.php
```diff
@@ -705,13 +729,13 @@ function plugin_upgrade( MantisPlugin $p_plugin ) {
 		switch( $t_schema[$i][0] ) {
 			case 'InsertData':
 				$t_sqlarray = array(
-					'INSERT INTO ' . $t_schema[$i][1][0] . $t_schema[$i][1][1],
+					'INSERT INTO ' . $t_target . $t_schema[$i][1][1],
```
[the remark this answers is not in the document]
dregad: Not sure where I was going with this, looks like a half-finished code cleanup. I'll revert.

PR #1567, opened 2019-09-22
Relationship graphs improvements and bug fixes

dregad:
Fixes

- [#11365](https://mantisbt.org/bugs/view.php?id=11365) - New Event: EVENT_MENU_ISSUE_RELATIONSHIP
- [#11381](https://mantisbt.org/bugs/view.php?id=11381) - Dependency Graph crash on circular parent child relationships
- [#17594](https://mantisbt.org/bugs/view.php?id=17594) - Output content of title-attributes inside relation graph elements
- [#26163](https://mantisbt.org/bugs/view.php?id=26163) - Relationship Graph page UI lacks MantisBT 2.x layout
- [#26164](https://mantisbt.org/bugs/view.php?id=26164) - Relationship Graph page is missing legend
- [#26165](https://mantisbt.org/bugs/view.php?id=26165) - Relationship Graph - inconsistency between button label and title


bug_view_inc.php
```diff
@@ -946,16 +946,16 @@ function bug_view_relationship_view_box( $p_bug_id, $p_can_update ) {
 	}
 
 	$t_relationship_graph = ON == config_get( 'relationship_graph_enable' );
-	$t_show_top_div = $p_can_update || $t_relationship_graph;
-	?>
+	$t_event_buttons = event_signal( 'EVENT_MENU_ISSUE_RELATIONSHIP', $p_bug_id );
```
vboctor: I would sent the parameter to the event as follows: `issue_id` => $p_bug_id, rather than use order of parameters.  This follows the new model we are trying to adopt.

bug_view_inc.php
```diff
@@ -969,22 +969,41 @@ function bug_view_relationship_view_box( $p_bug_id, $p_can_update ) {
...
+			$t_buttons[lang_get( 'relation_graph' )] =
+				'bug_relationship_graph.php?bug_id=' . $p_bug_id . '&graph=relation';
+			$t_buttons[lang_get( 'dependency_graph' )] =
+				'bug_relationship_graph.php?bug_id=' . $p_bug_id . '&graph=dependency';
+		}
+
+		# Plugin-added buttons
+		foreach( $t_event_buttons as $t_plugin => $t_plugin_buttons ) {
+			foreach( $t_plugin_buttons as $t_callback => $t_callback_buttons ) {
+				if( is_array( $t_callback_buttons ) ) {
```
vboctor: Is there a case where the callback result will be a single element?  If so, should it be added to the array?

docbook/Developers_Guide/en-US/Events_Reference_Output.xml
```diff
@@ -240,6 +240,34 @@ array( '&lt;a href="' . plugin_page( 'mypage' ) . '">' . plugin_lang_get( 'mypag
...
+				<itemizedlist>
+					<title>Parameters</title>
+					<listitem><para>&lt;Integer&gt;: bug ID</para></listitem>
+				</itemizedlist>
+
+				<itemizedlist>
+					<title>Return Value</title>
+					<listitem>
+						<para>&lt;Array&gt;: List of elements to add to the menu.</para>
+						<para>This is an associative array, with the link's label as key
```
vboctor: It would be good to have the event expect the result as an array of elements.  Each element has a `label`, `url`, and possibly more can be added in the future.

bug_view_inc.php
dregad: I did it that way to be consistent with similar event EVENT_MENU_ISSUE, but I'm fine to change it.

bug_view_inc.php
dregad: Good question. That would probably be an error case, as I think we should only accept an array, as documented.

In other, older menu-type events, for legacy reasons we are accepting cooked links from callbacks (i.e. the plugin would provide `<a href=url>label</a>` with the necessary CSS but I don't think this is good practice.

In this context I guess it does not matter that much, but with menu/submenu structures where we need to highlight the currently active item/page, that forces the plugin to handle this logic, which should really be taken care of in core to be consistent. Fixing that would mean a breaking change for plugins though.

Should I just throw an error here ?

docbook/Developers_Guide/en-US/Events_Reference_Output.xml
dregad: That was my initial thought for the implementation, but then I settled on this method, to be consistent with EVENT_MENU_ISSUE.

If we're going the way of associative array, I'm wondering if we should not take it one step further, and define Classes for menu items... This may also help addressing the problem I [mentioned above](https://github.com/mantisbt/mantisbt/pull/1567#discussion_r326889691). 

Thoughts ?

dregad:
@vboctor you have not responded to my comments.

dregad:
@vboctor since you seem to be online, I'm still waiting for a response from you...

bug_view_inc.php
vboctor: I would rather than we have all new events follow the new model.

bug_relationship_graph.php
```diff
@@ -62,20 +62,27 @@
 $f_bug_id		= gpc_get_int( 'bug_id' );
 $f_type			= gpc_get_string( 'graph', 'relation' );
 $f_orientation	= gpc_get_string( 'orientation', config_get( 'relationship_graph_orientation' ) );
+$f_show_summary	= gpc_get_bool( 'summary', false );
```
vboctor: We should have a config option for defaulting the show summary? `relationship_graph_show_summary`.  I would default it to ON.

bug_relationship_graph_img.php
```diff
@@ -55,6 +55,7 @@
 $f_bug_id		= gpc_get_int( 'bug_id' );
 $f_type			= gpc_get_string( 'graph', 'relation' );
 $f_orientation	= gpc_get_string( 'orientation', config_get( 'relationship_graph_orientation' ) );
+$f_show_summary	= gpc_get_bool( 'summary', false );
```
vboctor: Use config option here too.

bug_view_inc.php
vboctor: I would throw an error if the result is not what we expect. GIven that we are handling a new event here, can you elaborate on what will be broken / what existing plugins?

docbook/Developers_Guide/en-US/Events_Reference_Output.xml
vboctor: I agree that classes is the way to go.  It is unfortunate that we will end up with 3 models though (array, assoc array, and classes).  But I think discoverability and functionality will be much better if we have a model where each event has XyzEventArgs for the arguments and potentially functions that can be retrieved by the plugin.  I think we should also consider whether we should have XyzEventResult.

dregad:
@vboctor https://github.com/mantisbt/mantisbt/pull/1567#discussion_r353567145
> GIven that we are handling a new event here, can you elaborate on what will be broken / what existing plugins?

My point was that if we introduce associative arrays or classes for plugin event parameters, then to be consistent, we should also retrofit existing menu-related events (i.e. http://mantisbt.org/docs/master/en-US/Developers_Guide/html-desktop/#dev.eventref.output.menu). That would be a BC-break, affecting all plugins relying on these events, so to be SemVer-compliant we should do this in a major release.

If we don't do this consistency update, then as you pointed out we'd end up with a mess of 3 different ways of handling menu events (cooked links, arrays, classes).

bug_view_inc.php
dregad: Follow-up in https://github.com/mantisbt/mantisbt/pull/1567#issuecomment-561543366

bug_view_inc.php
dregad: Follow-up in https://github.com/mantisbt/mantisbt/pull/1567#issuecomment-561543366

docbook/Developers_Guide/en-US/Events_Reference_Output.xml
dregad: Follow-up in https://github.com/mantisbt/mantisbt/pull/1567#issuecomment-561543366

bug_relationship_graph.php
dregad: Honestly, I don't think it's really worth adding a new config for this, particularly if you'd like to see the feature moved into a plugin.

bug_relationship_graph_img.php
dregad: See https://github.com/mantisbt/mantisbt/pull/1567/files#r373859198

dregad:
Considering the importance of the changes you'd like to make and their backwards-incompatibility impact on existing plugins, while I fully agree that they should be implemented I prefer to handle this as a separate work item. I have opened [#26659](https://mantisbt.org/bugs/view.php?id=26659) to track this.

The new EVENT_MENU_ISSUE_RELATIONSHIP event, which was a long-standing user request, does not add much to the refactoring effort, so I don't think it is a blocking point

As for moving relationship graphs to a plugin, see [#26660](https://mantisbt.org/bugs/view.php?id=26660).

I'll proceed with merging this PR, since it fixes 6 open issues.

PR #1569, opened 2019-09-22
Use max length property of custom field in inputs

cproensa:
When a string, or text, type custom field has a max length defined, we
should use that for the rendered input in the form, to limit the width
of the input.

Fixes: #26141

### DIFF
--- a/core/cfdefs/cfdef_standard.php
+++ b/core/cfdefs/cfdef_standard.php
@@ -448,11 +448,13 @@ function cfdef_input_radio( array $p_field_def, $p_custom_field_value, $p_requir
  * @return void
  */
 function cfdef_input_textbox( array $p_field_def, $p_custom_field_value, $p_required = '' ) {
-	echo '<input ' . helper_get_tab_index() . ' type="text" id="custom_field_' . $p_field_def['id'] . '" name="custom_field_' . $p_field_def['id'] . '" size="80"' . $p_required;
-	if( 0 < $p_field_def['length_max'] ) {
-		echo ' maxlength="' . $p_field_def['length_max'] . '"';
+	echo '<input ', helper_get_tab_index(), ' type="text" id="custom_field_', $p_field_def['id']
+			, '" name="custom_field_', $p_field_def['id'], '" ', $p_required;
+	if( $p_field_def['length_max'] > 0 ) {
+		echo ' maxlength="' . $p_field_def['length_max'] . '"'
+				, ' size="' .  min( 80, $p_field_def['length_max'] ) . '"';
 	} else {
-		echo ' maxlength="255"';
+		echo ' maxlength="255" size="80"';
 	}
 	if( !empty( $p_field_def['valid_regexp'] ) ) {
 		# the custom field regex is evaluated with preg_match and looks for a partial match in the string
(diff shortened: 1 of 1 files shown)


core/cfdefs/cfdef_standard.php
```diff
@@ -448,11 +448,13 @@ function cfdef_input_radio( array $p_field_def, $p_custom_field_value, $p_requir
  * @return void
  */
 function cfdef_input_textbox( array $p_field_def, $p_custom_field_value, $p_required = '' ) {
-	echo '<input ' . helper_get_tab_index() . ' type="text" id="custom_field_' . $p_field_def['id'] . '" name="custom_field_' . $p_field_def['id'] . '" size="80"' . $p_required;
+	echo '<input ' . helper_get_tab_index() . ' type="text" id="custom_field_' . $p_field_def['id'] . '" name="custom_field_' . $p_field_def['id'] . '" ' . $p_required;
 	if( 0 < $p_field_def['length_max'] ) {
 		echo ' maxlength="' . $p_field_def['length_max'] . '"';
+		echo ' size="' .  $p_field_def['length_max'] . '"';
```
vboctor: Shouldn't this be min( max_length, 80 )?

core/cfdefs/cfdef_standard.php
```diff
@@ -448,11 +448,13 @@ function cfdef_input_radio( array $p_field_def, $p_custom_field_value, $p_requir
...
 	} else {
 		echo ' maxlength="255"';
+		echo ' size="80"';
```
vboctor: Why not just have one echo statement in each branch of the if statement.

core/cfdefs/cfdef_standard.php
```diff
@@ -479,6 +481,9 @@ function cfdef_input_textbox( array $p_field_def, $p_custom_field_value, $p_requ
  */
 function cfdef_input_textarea( array $p_field_def, $p_custom_field_value, $p_required = '' ) {
 	echo '<textarea class="form-control" ' . helper_get_tab_index() . ' id="custom_field_' . $p_field_def['id'] . '" name="custom_field_' . $p_field_def['id'] . '"' . $p_required;
+	if( 0 < $p_field_def['length_max'] ) {
```
vboctor: It is much more readable to have length_max > 0.

core/cfdefs/cfdef_standard.php
```diff
@@ -448,11 +448,13 @@ function cfdef_input_radio( array $p_field_def, $p_custom_field_value, $p_requir
  * @return void
  */
 function cfdef_input_textbox( array $p_field_def, $p_custom_field_value, $p_required = '' ) {
-	echo '<input ' . helper_get_tab_index() . ' type="text" id="custom_field_' . $p_field_def['id'] . '" name="custom_field_' . $p_field_def['id'] . '" size="80"' . $p_required;
+	echo '<input ' . helper_get_tab_index() . ' type="text" id="custom_field_' . $p_field_def['id'] . '" name="custom_field_' . $p_field_def['id'] . '" ' . $p_required;
 	if( 0 < $p_field_def['length_max'] ) {
```
vboctor: It is much more readable to have length_max > 0.

core/cfdefs/cfdef_standard.php
cproensa: thanks i overlooked that

core/cfdefs/cfdef_standard.php
cproensa: ok, also break long lines.

core/cfdefs/cfdef_standard.php
cproensa: ok

PR #1576, opened 2019-10-27
Revert "Don’t show completely empty notes"

cproensa:
After implementing the linking of notes and attachments, attachments
linked to a note without text are not displayed to some users becasue of
the time tracking logic, even if that functionality is not enabled.

This reverts commit 1ed44a96567e23da446f2678a717360f678d2d00.

Fixes: #26294

### DIFF
--- a/core/bugnote_api.php
+++ b/core/bugnote_api.php
@@ -490,12 +490,6 @@ function bugnote_get_all_visible_bugnotes( $p_bug_id, $p_user_bugnote_order, $p_
 			# If the access level specified is not enough to see time tracking information
 			# then reset it to 0.
 			if( !$t_time_tracking_visible ) {
-				# If the time tracking is the only data in the note, then skip it.
-				if( is_blank( $t_bugnote->note ) ) {
-					continue;
-				}
-
-				# otherwise, don't return the time tracking information so that it is not visible.
 				$t_bugnote->time_tracking = 0;
 			}
 


cproensa:
@vboctor 
I'm not sure if the logic for not showing empty notes is desirable ([22635](https://www.mantisbt.org/bugs/view.php?id=22635)). The time tracking use case is one that can be controlled within core logic, but we may also have a plugin implement some note related logic that relies on empty notes in similar use cases.
For example, the dedicated timetracking plugin.

Anyway, if the logic for hiding notes must exist, i think the proper place to have it is in `bug_activity_get_all()`, because there you have the attachments info and can make a better decision over what is considered a candidate note to be hidden.

core/bugnote_api.php
```diff
@@ -490,12 +490,6 @@ function bugnote_get_all_visible_bugnotes( $p_bug_id, $p_user_bugnote_order, $p_
 			# If the access level specified is not enough to see time tracking information
 			# then reset it to 0.
 			if( !$t_time_tracking_visible ) {
-				# If the time tracking is the only data in the note, then skip it.
-				if( is_blank( $t_bugnote->note ) ) {
```
vboctor: Shouldn't we just update this check to be empty note and no attached files instead of reverting this check?

cproensa:
> Shouldn't we just update this check to be empty note and no attached files instead of reverting this check?

That's why i suggested:  _if the logic for hiding notes must exist, i think the proper place to have it is in bug_activity_get_all()_
for several reasons:
- In bugnote_get_all_visible_bugnotes(): I think its reasonable to expect the visibility evaluated in this function to be about user permissions, not presentation.
- If implementing that check for attachments in bugnote_get_all_visible_bugnotes(), we need to read all bug linked files, match it with note ids, etc. This is performance cost added, since this is not a cached query, neither there is a function to get files lined to note ids. 

bug_activity_get_all() is closer to the presentation logic, and already have the attachment list to make the decision to show or hide a note with empty text.

>  if we are to support ability for plugins to force notes to be visible, then that would have to be through a plugin event - e.g. is note visible to user?

I would think first: in what scenario do we allow to create empty notes? Currently, only with timetracking enabled and a config option (`time_tracking_without_note`).


To sum up:
- The bug (26294) is already bad.
- This revert is the least change needed to fix up before release, and its side effects have a marginal impact, in my opinion.
- To maintain previous behavior: a subsequent task is needed for reimplementing, in a better way, the current logic for hiding empty notes. I am not going to do that at this moment.
- The stuff about plugins is not critical, no need to dive in to that at this moment.

atrol:
Keep in mind that empty notes must also be displayed if you want to have a look at previous revisions of a note by using the `View revisions` link.

PR #1577, opened 2019-11-02
Pre-fill tags from master bug when cloning

KeyWeeUsr:
When cloning a bug  I expect *everything* to be cloned, even tags, however I don't see the tags as if in an existing issue, so the next reasonable solution was to just pre-fill the input for tags.

With this I can even decide what tags to include before the bug is created so I don't need to waste time with copying the tags on DB level and then remove them one-by-one, instead I can skip the writing completely and proceed as if I'm creating a new bug which writes only the desired tags.

### DIFF
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -605,7 +605,20 @@ if( $t_show_attachments ) {
 			<label for="attach_tag"><?php echo lang_get( 'tag_attach_long' ) ?></label>
 		</th>
 		<td>
-			<?php print_tag_input( '' ); ?>
+			<?php
+				if( $f_master_bug_id > 0 ) {
+					// pre-fill tag string when cloning from master bug
+					$tags = [];
+					foreach(tag_bug_get_attached($f_master_bug_id) as $tag) {
+						array_push($tags, $tag["name"]);
+					}
+					$tag_string = implode(config_get('tag_separator'), $tags);
+					print_tag_input( $f_master_bug_id, $tag_string );
+				} else {
+					// otherwise show just the default empty string
+					print_tag_input( '' );
+				}
+			?>
 		</td>
 	</tr>
 <?php
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -607,15 +607,17 @@ if( $t_show_attachments ) {
 		<td>
 			<?php
 				if( $f_master_bug_id > 0 ) {
-					// pre-fill tag string when cloning from master bug
-					$tags = [];
-					foreach(tag_bug_get_attached($f_master_bug_id) as $tag) {
-						array_push($tags, $tag["name"]);
+					# pre-fill tag string when cloning from master bug
+					$t_tags = [];
+					foreach( tag_bug_get_attached( $f_master_bug_id ) as $t_tag ) {
+						array_push( $t_tags, $t_tag["name"] );
 					}
-					$tag_string = implode(config_get('tag_separator'), $tags);
-					print_tag_input( $f_master_bug_id, $tag_string );
+					$t_tag_string = implode(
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
@dregad the ticket is created [here](https://mantisbt.org/bugs/view.php?id=26326).

KeyWeeUsr:
Any status on this? Do I need to change something?

bug_report_page.php
```diff
@@ -605,7 +605,22 @@
...
+					foreach( tag_bug_get_attached( $f_master_bug_id ) as $t_tag ) {
+						array_push( $t_tags, $t_tag["name"] );
+					}
+					$t_tag_string = implode(
+						config_get( 'tag_separator' ), $t_tags
+					);
+					print_tag_input( $f_master_bug_id, $t_tag_string );
+				} else {
+					# otherwise show just the default empty string
+					print_tag_input( '' );
```
atrol: Didn't test anything, but seems this call was / is wrong as the first parameter is the bug id
```
/**
 * Print the separator comment, input box, and existing tag dropdown menu.
 * @param integer $p_bug_id A bug identifier.
 * @param string  $p_string Default contents of the input box.
 * @return void
 */
function print_tag_input( $p_bug_id = 0, $p_string = '' )
```

dregad: Actually this API call has been like this since it its introduction back in 2015 (see f67ec45972ada6ef3cf3b727b41ce1afb967ff6c).

This has no impact on functionality, as the string is converted to (bug_id) 0, meaning that all available tags are included in the selection list.

dregad:
> Any status on this? Do I need to change something?

Sorry, got sidetracked and forgot about it. Thanks for the reminder

bug_report_page.php
```diff
@@ -605,7 +605,22 @@
...
+				if( $f_master_bug_id > 0 ) {
+					# pre-fill tag string when cloning from master bug
+					$t_tags = [];
+					foreach( tag_bug_get_attached( $f_master_bug_id ) as $t_tag ) {
+						array_push( $t_tags, $t_tag["name"] );
+					}
+					$t_tag_string = implode(
+						config_get( 'tag_separator' ), $t_tags
+					);
+					print_tag_input( $f_master_bug_id, $t_tag_string );
```
dregad: While testing this, I noticed that there is actually a bug in *tag_get_candidates_for_bug()*. The function returns all tags, including those already attached to the specified bug, in contradiction with the function's documentation.

Fixing this issue, the dropdown list of tags no longer contains those associated with the master bug id, and they are added to the text box (as expected).

The problem is that if the user changes the text box's contents (e.g. editing or removing a tag), then they are no longer able to add the original tag back by selecting it from the list.

Therefore I wonder if it would not be better to call `print_tag_input( 0, $t_tag_string );` in this context.

Thoughts ?

KeyWeeUsr: Hm, but wouldn't it mean all of the tags are printed into the text input? If so, then such behavior is undesired because the point of cloning is to have the same tags as the master issue, not all of the tags Mantis has. In small instances that might be even fine, however imagine cloning an issue in mantisbt instance... that might be quite a trouble :)

dregad: > wouldn't it mean all of the tags are printed into the text input

No, the list of tags included in the text box are specified with the 2nd parameter ($t_tag_string).

KeyWeeUsr: Ok, let me patch it then, hopefully it helps with the bug you mentioned.

dregad: Thanks

dregad:
Looks good now. I'll give it one more round of test before merging.

PR #1579, opened 2019-11-14
Do not return tags already attached to the bug

dregad:
The SQL in tag_get_candidates_for_bug() was incorrect, causing the
function to always return all available tags, instead of excluding those
already attached to the bug.

Fixes [#26353](https://mantisbt.org/bugs/view.php?id=26353)

### DIFF
--- a/core/tag_api.php
+++ b/core/tag_api.php
@@ -702,8 +702,6 @@ function tag_get_candidates_for_bug( $p_bug_id ) {
 			$t_where = " WHERE id NOT IN ($t_assoc_tags_query)";
 		}
 		$t_query .= $t_where;
-	} else {
-		$t_query = 'SELECT id, name, description FROM {tag}';
 	}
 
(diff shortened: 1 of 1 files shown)


core/tag_api.php
```diff
@@ -667,46 +667,41 @@ function tag_delete( $p_tag_id ) {
...
 		} else {
-			$t_query = 'SELECT id, name, description FROM {tag} WHERE id IN (
-					SELECT t.id FROM {tag} t
-					LEFT JOIN {bug_tag} b ON t.id=b.tag_id
-					WHERE b.bug_id IS NULL OR b.bug_id != ' . db_param() .
-				')';
+			$t_where = " WHERE id NOT IN ($t_assoc_tags_query)";
 		}
+		$t_query .= $t_where;
 	} else {
```
syncguru: You don't need the else block anymore, right?

dregad: Correct. I'll fix that.

PR #1587, opened 2019-12-07
Use g_normal_date_format to configure datetime picker

syncguru:
Retire g_datetime_picker_format setting
Fixes #22558

---
The modern ui introduced the date time picker widget which is used for `due date` and `product version date` fields. The date picker was also how custom date-time fields were using until the change with reverted to the standard 3 dropdown fields (day, month, year) that we have today.

The picker date-time format was configured its `g_datetime_picker_format` global setting. Having this separate setting was a design mistake since it became so easy to set `g_datetime_picker_format` that is in conflict with the other global date configuration settings `g_normal_date_format` & co. When a user configures custom date format they were almost certain to run into this issue which prevents updating dateline fields when date-time picker is used.

The fix for this issue is to remove the separate setting for date picker configuration and instead use `g_normal_date_format` to auto-generate the format string thus maintain compatibility at all times.


core/date_api.php
```diff
@@ -280,3 +280,62 @@ function print_date_selection_set( $p_name, $p_format, $p_date = 0, $p_default_d
...
+ * Php date time format: https://www.php.net/manual/en/function.date.php
+ * @param string $p_php_format     	Datetime php format
+ * @return string 					Date time format string understood by moment.js based components
+ * @access public
+ */
+function convert_php_to_momentjs_datetime_format( string $p_php_format ) {
+	$t_replacements = [
+		'A' => 'A',      # for the sake of escaping below
+		'a' => 'a',      # for the sake of escaping below
+		'B' => '',       # Swatch internet time (.beats), no equivalent
```
dregad: In the unlikely event that someone would actually use this, shouldn't we trigger an exception or at least a warning, instead of silently dropping the format ?

I guess the same applies also to other PHP formats which do not have a direct corresponding value in momentjs (e.g. `L`, `t` etc)

core/date_api.php
```diff
@@ -280,3 +280,62 @@ function print_date_selection_set( $p_name, $p_format, $p_date = 0, $p_default_d
...
+ * Convert php date time format string to moment.js compatible format
+ * moment.js is a popular date time handling JS library. It is used by datetime picker library
+ * see https://stackoverflow.com/questions/30186611/php-dateformat-to-moment-js-format
+ * Moment.js date time format: https://momentjs.com/docs/#/parsing/string-format/
+ * Php date time format: https://www.php.net/manual/en/function.date.php
+ * @param string $p_php_format     	Datetime php format
+ * @return string 					Date time format string understood by moment.js based components
+ * @access public
+ */
+function convert_php_to_momentjs_datetime_format( string $p_php_format ) {
```
dregad: In terms of naming consistency within date api, maybe the function should be renamed `date_convert_format_from_php_to_momentjs` or something similar

core/obsolete.php
```diff
@@ -226,3 +227,6 @@
 
 # changes in 2.21.0
 config_obsolete( 'my_view_boxes_fixed_position' );
+
+# changes in 2.23.0
+config_obsolete( 'datetime_picker_format' );
```
dregad: Provide suggested replacement:
`config_obsolete( 'datetime_picker_format', 'normal_date_format' );`

docbook/Admin_Guide/en-US/config/date.xml
```diff
@@ -43,31 +43,6 @@
 				</para>
 			</listitem>
 		</varlistentry>
-		<varlistentry>
-			<term>$g_datetime_picker_format</term>
```
dregad: You should update doc for *$g_normal_date_format*, add a note similar to https://github.com/mantisbt/mantisbt/pull/1587/files#diff-dccc8d20b0765881557a770f05f6acb0R1308 explaining that it also covers the datetime picker.

core/date_api.php
atrol: We do still support PHP 5.5.x.
Type declararions / type hints for `string` type are not supported before PHP 7.0
https://www.php.net/manual/en/functions.arguments.php

docbook/Admin_Guide/en-US/config/date.xml
syncguru: Done

syncguru:
@atrol &@dregad
I did a lot more testing to original PR and discover another bug that resulted in make some code changes. I also addressed your comment. Please review/test as your convenience.

core/date_api.php
dregad: It does not look like you've addressed [my earlier note above](https://github.com/mantisbt/mantisbt/pull/1587/files#r355185170)

docbook/Admin_Guide/en-US/config/date.xml
dregad: Thanks

bug_change_status_page.php
```diff
@@ -244,7 +244,7 @@
 		<td>
 			<input type="text" id="due_date" name="due_date" class="datetimepicker input-sm" size="16" maxlength="16"
 				data-picker-locale="<?php lang_get_current_datetime_locale() ?>"
-				data-picker-format="<?php echo config_get( 'datetime_picker_format' ) ?>"
+				data-picker-format="<?php date_get_datetime_picker_format() ?>"
```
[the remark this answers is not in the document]
syncguru: Good catch

core/date_api.php
syncguru: I thought I did reply as part of my last round of comments.
I don't think we should trigger an exception here. It is a valid scenario to have one element in one format that does not have equivalent in another. Mapping routine should not flag that as an error.

dregad: > I don't think we should trigger an exception here. It is a valid scenario to have one element in one format that does not have equivalent in another. Mapping routine should not flag that as an error.

I agree that it is expected not to have equivalence for all formats, but that does not necessarily make it a valid scenario in this specific context.

My concern is that this may lead to inconsistencies or unexpected behavior when processing these dates, and it may be better to consider these formats as invalid and report them as such, to avoid surprises down the line caused by silently ignoring them.

Again, I have not tested this, so it may not be an actual problem.

syncguru: @dregad I looked into your request and tried to satisfy it but it increasingly looks complex and unnecessary. Can you propose a code change that makes sense to you? I am happy to update the PR with it. If not, I don't think we should block this PR any longer.

atrol:
@syncguru I assume the changes in filter_api.php are not intended.

syncguru:
Nope. I think it mistakenly included during the rebase. I will fix.

syncguru:
@atrol I think it is fixed. Can you verify?

atrol:
> @atrol I think it is fixed. Can you verify?

Right, the changes in filter_api.php are no longer part of this PR.

PR #1596, opened 2019-12-14
Add PHP 7.4 to Travis CI, remove PHP 5.5 from Travis CI

atrol:


dregad:
> Don't force Trusty in Travis CI

We can't do without dropping support for PHP 5.5 (see a164d8f9dd26542a28c3b71eb341e95ee53b16b2).

See my work-in-progress branch to update minimum PHP requirement to 7.0: https://github.com/dregad/mantisbt/commits/php7-req. I did not finalize this yet, because doing that also requires migrating to a later PHPUnit version, and I did not have time to finalize and test that (more WIP at https://github.com/dregad/mantisbt/commits/phpunit6).

atrol:
> we can't do without dropping support for PHP 5.5
> See my work-in-progress branch to update minimum PHP requirement to 7.0

Updating runtime minimum PHP requirements to 7.0 impacts our users a lot more than just dropping PHP 5.5 Travis builds.

Using the current PR would at least keep PHP 5.6 for the tests and take the risk that something goes wrong when using PHP 5.5 in Mantis production environment.

I marked the PR as [Draft] as another option could be to ask the Travis team to enable the PHP GD extension (the root cause of failing tests) for PHP 7.4 on Trusty (meanwhile they did for Xenial).
But do we want to go on using Trusty?

Of course, updating runtime minimum PHP requirements to 7.0 would be the easiest thing for us, but until now there was no announcement to our users.
As pointed out before at other places, we shouldn't change minimum PHP version when upgrading from version 2.x to 2.x+1 as this will break existing installations.

Maybe time to create branch 2.x.x as some kind of LTS branch (just security and major bug fixes) and start with version Mantis 3.0.0 on master branch.

dregad:
> Using the current PR would at least keep PHP 5.6 for the tests and take the risk that something goes wrong when using PHP 5.5 in Mantis production environment.

True. I guess that's an acceptable compromise.

> But do we want to go on using Trusty?

I don't.

> Of course, updating runtime minimum PHP requirements to 7.0 would be the easiest thing for us, but until now there was no announcement to our users.

It has been on [the Roadmap](https://mantisbt.org/bugs/roadmap_page.php) since 2.22.0, see [#25956](https://mantisbt.org/bugs/view.php?id=25956)

> As pointed out before at other places, we shouldn't change minimum PHP version when upgrading from version 2.x to 2.x+1 as this will break existing installations.
> 
> Maybe time to create branch 2.x.x as some kind of LTS branch (just security and major bug fixes) and start with version Mantis 3.0.0 on master branch.

OK for me, I don't really care if the next one is called 3.x or 2.x.

atrol:
Changed / simplified PR based on latest feedback from @dregad

atrol:
Merged, based on this note from @dregad 
> True. I guess that's an acceptable compromise.

PR #1597, opened 2019-12-15
enhanced free text filter #26447

being-peace:
See https://mantisbt.org/bugs/view.php?id=26447:

Attached is a patch for improved free text filter function.

Examples:
i:123 OR s:summary c:custom s:"test test" "jo jo"

i:123 searches only for the bug id (useful if you have a list of IDs and create a search link with 'xargs -n 1 printf "i:%s OR "')
c:text searches only the custom fields
s:text searches only the summary (faster and more focussed results).
OR allows mixing AND and OR in one query

### DIFF
--- a/core/classes/BugFilterQuery.class.php
+++ b/core/classes/BugFilterQuery.class.php
@@ -1406,22 +1406,32 @@ class BugFilterQuery extends DbQuery {
 		}
 
 		# break up search terms by spacing or quoting
-		preg_match_all( "/-?([^'\"\s]+|\"[^\"]+\"|'[^']+')/", $this->filter[FILTER_PROPERTY_SEARCH], $t_matches, PREG_SET_ORDER );
+		# optional "s:, i:, c:" for faster summary only requests
+		preg_match_all( "/-?([sic]:)?([^'\"\s]+|\"[^\"]+\"|'[^']+')/", $this->filter[FILTER_PROPERTY_SEARCH], $t_matches, PREG_SET_ORDER );
 
-		# organize terms without quoting, paying attention to negation
-		$t_search_terms = array();
-		foreach( $t_matches as $t_match ) {
-			$t_search_terms[trim( $t_match[1], "\'\"" )] = ( $t_match[0][0] == '-' );
-		}
 
+		$t_bug_table = $this->helper_table_alias_for_bugnote();
 		$t_bugnote_table = $this->helper_table_alias_for_bugnote();
 
 		# build a big where-clause and param list for all search terms, including negations
 		$t_first = true;
-		$t_textsearch_where_clause = '( ';
-		foreach( $t_search_terms as $t_search_term => $t_negate ) {
-			if( !$t_first ) {
-				$t_textsearch_where_clause .= ' AND ';
+		$t_after_or = false; # if we are after an OR 
+		$t_textsearch_where_clause = "( ( "; # second ( for OR
+		foreach( $t_matches as $t_match ) {
+			# analyse search term
+			$t_search_term = trim( $t_match[2], "\'\"" );
+			$t_negate = ( $t_match[0][0] == '-' );
+			$t_summary_only = ( $t_match[1] && $t_match[1][0] == 's' );
+			$t_id_only = ( $t_match[1] && $t_match[1][0] == 'i' );
+			$t_custom_field_only = ( $t_match[1] && $t_match[1][0] == 'c' );
+			if ( !$t_first && !$t_after_or ) {
+				if ( $t_search_term == 'OR' ) {
+					$t_textsearch_where_clause .= ' ) OR ( ';
+					$t_after_or = true;
+					continue;
+				} else {
+					$t_textsearch_where_clause .= ' AND ';
+				}
 			}
 
 			if( $t_negate ) {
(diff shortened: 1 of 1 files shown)


core/classes/BugFilterQuery.class.php
```diff
@@ -1406,48 +1406,86 @@ protected function build_prop_search() {
...
+		# optional "s:, i:, c:" for faster summary only requests
+		preg_match_all( "/-?([sic]:)?([^'\"\s]+|\"[^\"]+\"|'[^']+')/", $this->filter[FILTER_PROPERTY_SEARCH], $t_matches, PREG_SET_ORDER );
 
-		# organize terms without quoting, paying attention to negation
-		$t_search_terms = array();
-		foreach( $t_matches as $t_match ) {
-			$t_search_terms[trim( $t_match[1], "\'\"" )] = ( $t_match[0][0] == '-' );
-		}
 
+		$t_bug_table = $this->helper_table_alias_for_bugnote();
```
atrol: This variable is not used and I assume it's not what you intended with it, as you assigned exactly the same what is assigned to `$t_bugnote_table` on next line.

core/classes/BugFilterQuery.class.php
```diff
@@ -1406,48 +1406,86 @@ protected function build_prop_search() {
...
+					#  id search only in bugs, not in bugnotes
+					if ( !$t_id_only ) {
+						$t_textsearch_where_clause .= ' OR ' . $t_bugnote_table . '.id = ' . $this->param( $c_search_int );
+					}
 				}
 			}
 
+			if ( $t_custom_field_only ) {
+				$t_textsearch_where_clause .=
+					' OR ' . "({bug}.id IN ( SELECT DISTINCT bug_id from " . db_get_table( 'mantis_custom_field_string_table' ) . " where " . $this->sql_like( "value", $c_search )."))";
```
atrol: Should be `db_get_table( 'custom_field_string' )`.
At the moment you get 
```
Deprecated functionality: "db_get_table( 'mantis_custom_field_string_table' )", use "db_get_table( 'custom_field_string' )" instead. (in .../core/classes/BugFilterQuery.class.php line 1472)
```

atrol:
Not sure if you are aware that you don't search for custom fields of type `Text Area`.
The values are stored in column `text` of table `custom_field_string`.

being-peace:
The requested changes were applied, for details see https://mantisbt.org/bugs/view.php?id=26447

core/classes/BugFilterQuery.class.php
```diff
@@ -1406,48 +1439,79 @@ protected function build_prop_search() {
...
-					. ' OR ' . $this->sql_like( '{bug_text}.description', $c_search )
-					. ' OR ' . $this->sql_like( '{bug_text}.steps_to_reproduce', $c_search )
-					. ' OR ' . $this->sql_like( '{bug_text}.additional_information', $c_search )
-					. ' OR ' . $this->sql_like( '{bugnote_text}.note', $c_search );
+			$t_textsearch_where_clause .= '( ';
+			# search in summary only if not id / custom_field_search
+			if ( !$t_id_only && !$t_custom_field_only ) {
+				$t_textsearch_where_clause .= $this->sql_like( '{bug}.summary', $c_search );
+                        } else {
+				$t_textsearch_where_clause .= '0';
```
cproensa: this is not cross-compatible syntax, only works in mysql, but no in other databases: oracle, sql server, postgres

cproensa:
> Only textarea and strings that are marked as "filter by" are searched.

I see a good start to include only those field types in the global search. Other types can be questionable, for example: is really useful to search within enumeration fields? (*1)
But limiting to filter by, probably is not useful.
As i see it, the "filter by" property is needed mainly to not cram the filter dialog, and the current custom field filtering is useful especially on enumerated fields. A free string field is currently a chaotic dropdown with all existing values, so in those cases this new search functionality is better.

Also note, that if at some point we redesign the filter UI, the filter fields will be added on demand (not showing all at once), so the "filter by" may become obsolete... Unless we want to explicitly not allow filtering on some fields.

So my question is: should we limit the search only to those CFs enabled with "filter by"?

core/classes/BugFilterQuery.class.php
being-peace: What is the cross-compatible syntax?

being-peace:
For me, any using "filter_by" or not would be okay. I was only wondering, because the checkbox in the options says: "Add to Filter". Wouldn't it be surprising if they would be used for free text filtering?

core/classes/BugFilterQuery.class.php
cproensa: > What is the cross-compatible syntax?

What is exactly the intention?
Ignore one of the clauses in an ORed expression? 
Seems lke the resulkt would be "0 OR this OR that"
Then i would not add the caluse, nor the first OR.

A dirtier hack is to use a more portable boolean expression like "1=1" (true) or "1=0" (false)
but i'd rather manage properly the building of the conditions.

cproensa:
> For me, any using "filter_by" or not would be okay. I was only wondering, because the checkbox in the options says: "Add to Filter". Wouldn't it be surprising if they would be used for free text filtering?

My vote is to not limit to the "filter by" property, but i'd like to hear other opinions.

core/classes/BugFilterQuery.class.php
being-peace: This would require to rewrite all "t_textsearch_where_clause .=" lines in this function and in the function above, it would require a different handling of the situation where no custom field is visible. Perhaps this is something a person with better php skills can do? It would probably go along the line of adding to a list and then joining the list with " OR ".

I first would like to know if the change is acceptible anyway. For now I will add the 1=0 hack.

cproensa: > adding to a list and then joining the list with " OR "

Yes, that would be the ideal approach. Build an array and an the end, if not emply, implode with " OR ".
Even, if the item count is 1, omit the outer ( ).

MASTER-WM:
Does this code work?
Because I need to be able to search on my custom field, but the codes I added made no difference to me.

PR #1598, opened 2019-12-16
Correct CSS class names for custom fields

atrol:
Not all characters are allowed to be used in CSS class names.
Replace any potentially unsupported character(s) by dashes `-`.

Fixes #26473

### DIFF
--- a/core/custom_field_api.php
+++ b/core/custom_field_api.php
@@ -1433,6 +1433,21 @@ function print_custom_field_input( array $p_field_def, $p_bug_id = null, $p_requ
 	}
 }
 
+/*
+ * Returns a valid CSS identifier for the given custom field.
+ *
+ * The string is built based on the custom field's name, replacing any potentially
+ * unsupported character(s) by dashes `-` to ensure its validity. The resulting
+ * identifier can be used as part of a custom CSS class.
+ *
+ * @param string $p_custom_field_name The custom field's name
+ *
+ * @return string The CSS identifier
+ */
+function custom_field_css_name( $p_custom_field_name ) {
+    return 'custom-' . preg_replace( '/[^a-zA-Z0-9_-]+/', '-', $p_custom_field_name );
+}
+
 /**
  * Constructs the name of a field used as a flag to indicate that a custom field is present on the form
  * @param integer $p_custom_field_id  The custom field id to create the field name for.
--- a/core/custom_function_api.php
+++ b/core/custom_function_api.php
@@ -343,7 +343,7 @@ function custom_function_default_print_column_title( $p_column, $p_columns_targe
 	$t_custom_field = column_get_custom_field_name( $p_column );
 	if( $t_custom_field !== null ) {
 		if( COLUMNS_TARGET_CSV_PAGE != $p_columns_target ) {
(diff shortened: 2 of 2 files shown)


atrol:
> This is just a hack, and it is not sufficient to cover all cases. 

I am aware of that.
No time at the moment, but I will comment later / next days why I think it's not that bad and acceptable for the moment.

atrol:
@dregad @vboctor what about using the custom field id instead of the custom field name to generate the CSS class name?

dregad:
I have used this approach before in another, and I guess it would work. 

The problem is that using the id is not portable, and would force updating the CSS if for any reason the CF record has to be created again (e.g. in another DB).

atrol:
Implemented this approach

>  suggest having a custom_field_css_name() method that converts a custom field name to a css style name. It converts spaces or invalid values by some valid characters (e.g. -) or some encoding. I also suggest using dashes rather than underscores in css style names.

atrol:
@dregad any objections to merge as it is now?
It fixes the issue and allows later enhancements at a central place.

Note to
> The problem is that using the id is not portable, as it could potentially change ...

This could be prevented by using a md5 (or any other) hash of the custom field name.
But I don't like the idea, as the names would be quite cryptic and it would also affect a little bit  performance (not sure if it would be measurable).

dregad:
I didn't have time to review / test yet.

> This could be prevented by using a md5 (or any other) hash of the custom field name.
But I don't like the idea, as the names would be quite cryptic 

Agreed.

core/custom_field_api.php
```diff
@@ -1433,6 +1433,15 @@ function print_custom_field_input( array $p_field_def, $p_bug_id = null, $p_requ
 	}
 }
 
+/**
+ * Constructs a string that can be used as part of a CSS class name based on a custom field name
+ * @param string $p_custom_field_name The custom field name
+ * @return string The CSS string
+ */
```
dregad: Suggested rewording/improvement
```
/*
 * Returns a valid CSS identifier for the given custom field.
 * 
 * The string is built based on the custom field's name, replacing any potentially
 * unsupported character(s) by dashes `-` to ensure its validity. The resulting
 * identifier can be used as part of a custom CSS class.
 * 
 * @param string $p_custom_field_name The custom field's name
 *
 * @return string The CSS identifier
 */
```

core/custom_field_api.php
```diff
@@ -1433,6 +1433,15 @@ function print_custom_field_input( array $p_field_def, $p_bug_id = null, $p_requ
...
+function custom_field_css_name( $p_custom_field_name ) {
+    return 'custom-' . str_replace( ' ', '-', $p_custom_field_name );
```
dregad: To be on the safe side, instead of just dealing with spaces, I would suggest using a regex to also take care of any sequence of potentially invalid chars, and replace them with a dash - something like (offhand, not tested)

```
preg_replace( '/[^A-Za-z0-9_-]+/', '-', $p_custom_field_name );
```

atrol:
Thanks @dregad for the review. So finally you get what you wanted :smile:

PR #1601, opened 2019-12-23
Use custom CSSs in addition to default ones

grv87:
Fixes https://www.mantisbt.org/bugs/view.php?id=12352.

This PR is based on the patch proposed by Kirill in comments, but with RTL fixed too.


core/html_api.php
```diff
@@ -233,6 +233,18 @@ function html_css() {
 	} else {
 		html_css_link( 'dropzone-' . DROPZONE_VERSION . '.min.css' );
 	}
+
+	$t_css_include_file = config_get_global( 'css_include_file' );
+	if( $t_css_include_file !== '' ) {
```
vboctor: `if( $t_css_include_file !== '' && $t_css_include_file !== 'default.css' )`

core/html_api.php
```diff
@@ -233,6 +233,18 @@ function html_css() {
...
+		html_css_link( $t_css_include_file );
+	}
+	# Add right-to-left css if needed
+	if( lang_get( 'directionality' ) == 'rtl' ) {
+		$t_css_rtl_include_file = config_get_global( 'css_rtl_include_file' );
+		if( $t_css_rtl_include_file !== '' ) {
```
vboctor: `if( $t_css_rtl_include_file !== '' && $t_css_rtl_include_file !== 'rtl.css' )`

core/html_api.php
grv87: I agree. Done

PR #1606, opened 2020-01-06
Support adding attachment on bug update

cproensa:
Add attachments section in bug update page and change status page.

Fixes: #21819, #21820

### DIFF
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -342,13 +342,8 @@ layout_page_begin();
 </div>
 
 <div class="space-6"></div>
-<div class="widget-toolbox padding-8 clearfix">
-	<span class="bold">
-	<?php echo lang_get( 'actiongroup_menu_add_note' ) ?>
-	</span>
-</div>
 
-<?php print_bugnote_form_content( $f_bug_id ) ?>
+<?php print_bugnote_form_content_collapse( $f_bug_id ) ?>
 
 <div class="space-6"></div>
 
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -336,91 +336,22 @@ layout_page_begin();
 		printf( '	<input type="hidden" name="resolution" value="%s" />' . "\n", config_get( 'bug_reopen_resolution' ) );
 	}
 ?>
-<?php
-	$t_bugnote_private = $t_default_bugnote_view_status == VS_PRIVATE;
-	$t_bugnote_class = $t_bugnote_private ? 'form-control bugnote-private' : 'form-control';
-
-	if( access_has_bug_level( config_get( 'private_bugnote_threshold' ), $f_bug_id ) ) { ?>
-			<tr>
-				<th class="category">
-					<?php echo lang_get( 'view_status' ) ?>
-				</th>
-				<td>
-<?php
-		if( access_has_bug_level( config_get( 'set_view_status_threshold' ), $f_bug_id ) ) {
-?>
-			<input type="checkbox" id="bugnote_add_view_status" class="ace" name="private"
-				<?php check_checked( $t_default_bugnote_view_status, VS_PRIVATE ); ?> />
-			<label class="lbl padding-6" for="bugnote_add_view_status"><?php echo lang_get( 'private' ) ?></label>
-<?php
-		} else {
-			echo get_enum_element( 'view_state', $t_default_bugnote_view_status );
-		}
-?>
-				</td>
-			</tr>
-<?php } ?>
-			<!-- Bugnote -->
-			<tr id="bug-change-status-note">
-				<th class="category">
-					<?php echo lang_get( 'add_bugnote_title' ) ?>
-				</th>
-				<td>
-					<textarea name="bugnote_text" id="bugnote_text" class="<?php echo $t_bugnote_class ?>" cols="80" rows="7"></textarea>
-				</td>
-			</tr>
-<?php
-	if( config_get( 'time_tracking_enabled' )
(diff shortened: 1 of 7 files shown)


bug_change_status_page.php
```diff
@@ -386,6 +387,33 @@
 <?php
 	}
 
+	if( $t_allow_file_upload ) {
```
atrol: After merging as is, we would have nearly the same code block four times (bug_change_status_page.php, bug_report_page.php, bug_update_page.php, bugnote_add_inc.php)
Maybe you can introduce some kind of `function print_bugnote_attach_zone( ... )`

cproensa:
> If I copy an image, I can't paste it in the new added places to add attachments, but can paste into the existing ones.

Fixed

> If I just add an attachment from existing places, I can hover over their left column later and can make them private, delete, etc. For the ones added via your code, I can only delete them. This is likely being caused by such attachments not having an associated empty note.

Fixed
However, note that currently, attachments added at report time also don't have a related note and can't be made private, only delete.

dregad:
@vboctor @atrol can you please confirm you're OK with the latest fixes following your respetive reviews ?

atrol:
I don't see any change related to my remark concerning code duplication.
Didn't test, but I assume there will also be two more places where we encounter https://www.mantisbt.org/bugs/view.php?id=26627

cproensa:
> I don't see any change related to my remark concerning code duplication.

It's in my to-do list

vboctor:
> @cproensa However, note that currently, attachments added at report time also don't have a related note and can't be made private, only delete.

This is by design since the attachments issued at report time is associated with the issue itself and it derives its private/public from the issue and in the future, it may be displayed within the issue itself.  If the user wants to report a public issue with private attachments, then it is a two step process (report an issue, then attach files to it privately).

bug_update.php
```diff
@@ -412,11 +412,25 @@
...
+	'time_tracking' => array(
+		'duration' => $t_bug_note->time_tracking
+	),
+	'files' => helper_array_transpose( $f_files )
+);
+$t_note_data = array(
+	'query' => $t_note_query,
+	'payload' => $t_note_payload,
+);
+$t_note_command = new IssueNoteAddCommand( $t_note_data );
```
vboctor: The commands should reflect the intention of the user and not be used as re-usable code.  For example, the user intent it to change status or update issue in this case.  Using command here may trigger emails, events, and other semantics that is not expected.  The #1400 PR was attempting at having an update command.

Anyway, whether we have an update command or not, the re-usable code between add note and update issue should be in APIs that are called by both commands.

cproensa: If not using the command, we are at the previous situation: file api don't provide a way to add an attachment that satisfies the expected internal representation.

cproensa:
> I don't see any change related to my remark concerning code duplication.

I have rewritten the form contents, so now it's a separated div:
- visually groups the note specific fields, so they are not confused with the fields of the main action form.
- reusable
- the note section can be expanded/collapsed for a leaner presentation

> Didn't test, but I assume there will also be two more places where we encounter https://www.mantisbt.org/bugs/view.php?id=26627

that's fixed now.

cproensa:
> The commands should reflect the intention of the user and not be used as re-usable code. For example, the user intent it to change status or update issue in this case. Using command here may trigger emails, events, and other semantics that is not expected. The #1400 PR was attempting at having an update command.

I'm not sure that applies for the "add note" command. Adding a note is a differentiated enough action that can be done along other action (eg, update of the issue).
Do you consider that adding a note is part of "updating the isse"? Then, why have a separate note command, when it should be all managed in a "proper" issue update command?
Regarding notifications, in the current implementation for notifications, i'd rather have a notification for issue updated, and another one for the note added. Because, they are formatted differently, and if i expect notes to be notified in the short format, i may overlook a note that was added along an update since there is not specific information that the note was added in that kind of notification.
Additionally, currently its not working very good, eg: https://mantisbt.org/bugs/view.php?id=26701

dregad:
Came back to this following @atrol's note in #1631. 

Not sure what is the status on this, @vboctor @cproensa ? Can it be merged or are you still discussing ?

cproensa:
>  Can it be merged or are you still discussing ?
No merge yet.
My questions are still open, regarding the command, but not blocking though.
There is an UI change that needs some confirmation.
If the UI normalization is valid, it can probably be applied to some other note forms (eg, group update, reminder...)

vboctor:
@cproensa sorry for the delay.  Had another look.

-  I'm OK we the UI update to have notes collapsed by default, but let's see how users find it.
- Can you also update the Edit Note form to support attaching files?
- For the use of `IssueNoteAddCommand` command, we can use it but I wouldn't trigger a different email for the note.  This can be done by adding an "options" block similar to "query" and "payload" with a flag to disable email notification `email_notification`.  The default would be ON, but you can override.  The options block is designed to pass such internal flags that we don't want to be usable via REST API.  You can checkout `IssueAddCommand` which uses options.

vboctor:
@cproensa any chance of finalizing this PR to merge it?

cproensa:
> @cproensa any chance of finalizing this PR to merge it?

Would like to, but at the moment i have little time, and it's complicated with the current situation.I guess in some weeks i can get back to it.
I hope you all are doing well, take care.

dregad:
@cproensa weeks have turned into months, any chance to see this finalized ?

aaribaud:
Hello, I am interested in this PR. @cproensa, do you plan to update it?

vboctor:
Closing stale PR.

PR #1611, opened 2020-01-18
Fix history_default_visible regression in View Issue Details page

atrol:
Fixes #26610


atrol:
> though I question the usefulness of this feature

I don't use it myself, but I think it's not that bad, as most of the time you are not interested in history when viewing at an issue.
In this case, it could be used to reduce a bit the traffic on the server and the time to load the page.

It's been introduced by you https://github.com/mantisbt/mantisbt/commit/2108ff599b1d3e8f75d0a7111b13b4f2702e3ea8

PR #1613, opened 2020-01-20
Guide tweaks

seanm:


seanm:
@atrol @dregad some misc guide fixes after reading through much of it...

seanm:
ticket created: https://mantisbt.org/bugs/view.php?id=26617

dregad:
Another thing, about the removal of IRC and mailing lists sections.

Even though this PR is only aligning the documentation with the current state of these legacy support channels, I do not like that we entirely removed all traces of them and their history. While I agree that they are not actively used nowadays, they are part of the project's legacy and we should still reference the archives somewhere.

@vboctor What about a section on the website, _archives of legacy support channels_ or something like that ?

seanm:
I didn't set out to remove the IRC per se, just that when checking if the links supported https, I found they were just about all dead links. So it seemed no one cared.  Shall I restore the IRC sections here and you guys fix it up as you like?

dregad:
@seanm I know, my comment was not directed at you. What you did is consistent with mantisbt.org's current state, and no action is expected from you at this stage. 

Let's wait for @vboctor and other dev's feedback.

vboctor:
@dregad I like the idea of maintaining links to archives.  We can feature the Gitter chat on the website support page as a pear to other channels.  Then have a statement at the end of the page that refer to archives.

seanm:
So I guess I'll just revert my deletions related to IRC?

dregad:
@seanm yes please, and also address https://github.com/mantisbt/mantisbt/pull/1613#discussion_r368902722

seanm:
OK, I think I addressed all your comments...

vboctor:
@dregad please check and merge or let me know and i'll merge it.

dregad:
@seanm sorry for taking so long to respond, and thanks a lot for your contribution.

PR #1614, opened 2020-01-23
Improve LDAP API performance through caching

dregad:
Note: this only affects actual LDAP server queries, LDAP simulation functions are not cached.

This fixes [#26622](https://mantisbt.org/bugs/view.php?id=26622) and should address [#26600](https://mantisbt.org/bugs/view.php?id=26600) too.

### DIFF
--- a/core/ldap_api.php
+++ b/core/ldap_api.php
@@ -214,7 +214,7 @@ function ldap_cache_user_data( $p_username ) {
 	# Bind
 	$t_ds = @ldap_connect_bind();
 	if( $t_ds === false ) {
-		ldap_log_error( $t_ds );
+		log_event( LOG_LDAP, "ERROR: could not bind to LDAP server" );
 		return false;
 	}
 
@@ -337,10 +337,6 @@ function ldap_authenticate_by_username( $p_username, $p_password ) {
 
 		# Bind
 		$t_ds = ldap_connect_bind();
-		if( $t_ds === false ) {
-			ldap_log_error( $t_ds );
-			trigger_error( ERROR_LDAP_AUTH_FAILED, ERROR );
-		}
 
 		# Search for the user id
 		log_event( LOG_LDAP, 'Searching for ' . $t_search_filter );
--- a/core/ldap_api.php
+++ b/core/ldap_api.php
@@ -212,7 +212,6 @@ function ldap_cache_user_data( $p_username ) {
 	log_event( LOG_LDAP, "Retrieving data for '$p_username' from LDAP server" );
 
 	# Bind
-	log_event( LOG_LDAP, 'Binding to LDAP server' );
 	$t_ds = @ldap_connect_bind();
 	if( $t_ds === false ) {
 		ldap_log_error( $t_ds );
@@ -337,7 +336,6 @@ function ldap_authenticate_by_username( $p_username, $p_password ) {
 		);
 
 		# Bind
-		log_event( LOG_LDAP, 'Binding to LDAP server' );
 		$t_ds = ldap_connect_bind();
 		if( $t_ds === false ) {
 			ldap_log_error( $t_ds );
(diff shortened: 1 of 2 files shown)


dregad:
Note: the PR is currently targeted at master branch, but will probably be merged in 2.23.

core/ldap_api.php
```diff
@@ -207,76 +189,107 @@ function ldap_escape_string( $p_string ) {
...
+	}
+
+	# Store data in the cache
+	$g_cache_ldap_data[$p_username] = $t_data;
+
 	# Free results / unbind
+	# According to documentation, calling ldap_free_result() is not strictly necessary
 	log_event( LOG_LDAP, 'Unbinding from LDAP server' );
-	ldap_free_result( $t_sr );
+	# ldap_free_result( $t_sr );
```
vboctor: Why leave the code commented out?  Either do it anyways (which I would do if we expect something to be called many times in one page) or remove the commented code.  `ldap_free_result()` is also called in other places in `ldap_api.php`.

core/ldap_api.php
```diff
@@ -207,76 +189,107 @@ function ldap_escape_string( $p_string ) {
...
 
 	# Get results
-	$t_info = ldap_get_entries( $t_ds, $t_sr );
-	if( $t_info === false ) {
-		ldap_log_error( $t_ds );
-		log_event( LOG_LDAP, 'ldap_get_entries() returned false.' );
-		return null;
+	$t_entry = ldap_first_entry( $t_ds, $t_sr );
+	if( $t_entry === false ) {
+		log_event( LOG_LDAP, 'No matches found.' );
```
vboctor: It would be useful to include the search filter here as well, since logs tend to interleave across different requests.

core/ldap_api.php
```diff
@@ -207,76 +189,107 @@ function ldap_escape_string( $p_string ) {
...
+ * Values are retrieved from the LDAP cache.
+ * {@see ldap_cache_user_data()} for the list of valid field names.
+ *
+ * @param string $p_username The user name.
+ * @param string $p_field    The LDAP field name.
+ *
+ * @return string The field value or null if not found.
+ */
+function ldap_get_field_from_username( $p_username, $p_field ) {
+	log_event(LOG_LDAP, "Retrieving field '$p_field' for '$p_username'");
```
vboctor: Adjust spaces around `()`.

core/ldap_api.php
dregad: I initially commented the statement out for testing purposes, and also because I was considering a more in-depth refactoring.

The idea was to keep the LDAP connection opened throughout script execution in order to reduce overhead and further improve performance, but I decided not to implement it for now and focus on fixing the immediate performance problem.

I might still do this in a later stage if time allows.

> Either do it anyways (which I would do if we expect something to be called many times in one page

The [documentation](https://www.php.net/manual/en/function.ldap-free-result.php) says:
_In case the script is making successive searches which return large result sets, ldap_free_result() could be called to keep the runtime memory usage by the script low_

We do make multiple ldap searches on a page, but they should normally only return one entry each, so I don't think that qualifies as _large result sets_.

> ldap_free_result() is also called in other places in ldap_api.php.

The other calls are in ldap_authenticate_by_username() which in usual scenarios is only done once. But I guess they should be removed as well for consistency.

core/ldap_api.php
dregad: The search filter _is_ logged already, [just a few lines above](https://github.com/mantisbt/mantisbt/pull/1614/files#diff-aac9367df2352143e7002c2d9fbb1dd3R232).

core/ldap_api.php
dregad: Thanks, will fix.

dregad:
Force-rebased and changed base branch to master-2.23, per https://github.com/mantisbt/mantisbt/pull/1614#issuecomment-577781560

core/ldap_api.php
vboctor: I'm aware of that, I was wondering if it would be helpful where logs from concurrent requests are interleaved.  Not a big deal though.

dregad: Ah OK I see what you mean now. I didn't realize you were talking about concurrent requests. I'll update the log message.

PR #1615, opened 2020-01-24
Fix filter by relationship

cproensa:
Fix scenario when filtering by "none" relation to a given issue id.

Fixes: [26621](https://mantisbt.org/bugs/view.php?id=26621)


dregad:
Should this go in 2.23.1 ?

cproensa:
> Should this go in 2.23.1 ?

I don't think this is a critical issue.

vboctor:
> Should this go in 2.23.1 ?

I would go with master since it is not a recent regression and this area is complex in nature.

PR #1619, opened 2020-01-29
Trim username on login

dregad:
When an LDAP user adds spaces before or after their username when
logging in, they will be authenticated successfully and logged in, but
Mantis will create a new entry in the user table including the spaces.

Fixes [#25097](https://mantisbt.org/bugs/view.php?id=25097)

### DIFF
--- a/login_password_page.php
+++ b/login_password_page.php
@@ -58,7 +58,7 @@ require_css( 'login.css' );
 $f_error                 = gpc_get_bool( 'error' );
 $f_cookie_error          = gpc_get_bool( 'cookie_error' );
 $f_return                = string_sanitize_url( gpc_get_string( 'return', '' ) );
-$f_username              = gpc_get_string( 'username', '' );
+$f_username              = trim( gpc_get_string( 'username', '' ) );
 $f_reauthenticate        = gpc_get_bool( 'reauthenticate', false );
 $f_perm_login            = gpc_get_bool( 'perm_login', false );
 $f_secure_session        = gpc_get_bool( 'secure_session', false );


atrol:
Using whitespaces in usernames is questionable, but such usernames are valid.
We don't ensure at low database level and also not on application level (function user_is_name_valid) that whitespaces are not allowed.

I am not sure if changing input data behind the scenes at a single high level place in code is the right fix for that.

dregad:
> Using whitespaces in usernames is questionable, but such usernames are valid.

True, as per default *$g_user_login_valid_regex* although this is really something that we ought to change IMO.

> I am not sure if changing input data behind the scenes at a single high level place in code is the right fix for that.

This PR is not about removing all spaces from usernames. It is only trimming leading/trailing whitespace from user input, which is quite common if not standard for web applications. Doing this for the username here at login time makes sense, and is consistent with existing MantisBT behavior, e.g.

- [signup](https://github.com/mantisbt/mantisbt/blob/release-2.23.0/signup.php#L59)
- [updating user](https://github.com/mantisbt/mantisbt/blob/release-2.23.0/manage_user_update.php#L85)
- [adding monitors](https://github.com/mantisbt/mantisbt/blob/release-2.23.0/bug_monitor_add.php#L45)

If you disagree with this approach, then for consistency we should also remove the trim() from the above-mentioned locations. Do you have an alternative to suggest ?

atrol:
> this PR is not about removing all spaces from usernames

I am aware of that, my concern is about changing input data in general.

You are right, trimming seems to be quite usual (checked a few other systems).

I don't like that approach, but I don't have time to check if this could introduce other issues, e.g. inconsistent situations when using our API to create users.

Changing `$g_user_login_valid_regex` might be enough to prevent such situations, but will not help  if there are old accounts with invalid usernames.
Low level functions like `user_get_id_by_name()` will also not work in this case.

All in all: We are not clean at the moment, I don't have time to provide a clean solution, "your" trim is not worse than the existing ones, so feel free to merge as is.

PR #1626, opened 2020-02-27
Fix bugnote view_state when sending reminder

nenadalm:
`private` checkbox on the form is ignored. This PR takes view_state from the checkbox rather than ignoring state of the checkbox and using always default value.


syncguru:
Hi @nenadalm - thanks for your contribution.
Is there an issue opened in mantisbt bug tracker for this? If not, would you mind create one with the details about the issue and how to repro?

nenadalm:
I've created the issue: https://mantisbt.org/bugs/view.php?id=26757

dregad:

PR #1630, opened 2020-03-14
Retire `bug_change_status_page_fields` config option

vboctor:
Fixes #26778


atrol:
The option should also be removed from `$g_public_config_names`

vboctor:
@dregad and @atrol feedback addressed.

PR #1632, opened 2020-03-17
Add failed_login_count to user information

atrol:
Fixes #26797

### DIFF
--- a/lang/strings_english.txt
+++ b/lang/strings_english.txt
@@ -720,6 +720,7 @@ $s_p = 'p';
 $s_date_created = 'Date Created';
 $s_last_visit = 'Last Visit';
 $s_last_visit_label = 'Last Visit:';
+$s_failed_login_count = 'Failed Logins';
 $s_edit_user_link = 'Edit User';
 $s_search_user_hint = 'Username, Real Name or E-mail';
 
--- a/manage_user_page.php
+++ b/manage_user_page.php
@@ -357,7 +357,7 @@ $t_user_count = count( $t_users );
 	# Print column headers with sort links
 	$t_columns = array(
 		'username', 'realname', 'email', 'access_level',
-		'enabled', 'protected', 'date_created', 'last_visit'
+		'enabled', 'protected', 'date_created', 'last_visit', 'failed_login_count'
 	);
 
 	foreach( $t_columns as $t_col ) {
@@ -436,6 +436,7 @@ $t_user_count = count( $t_users );
 				</td>
 				<td><?php echo $v_date_created ?></td>
 				<td><?php echo $v_last_visit ?></td>
+				<td><?php echo $v_failed_login_count ?></td>
 			</tr>
 <?php
 	}  # end for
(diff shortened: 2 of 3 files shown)


atrol:
@vboctor I changed that the field is just displayed if the admin deciced to use this functionality, e.g. set `$g_max_failed_login_count` to a value `!= OFF`

vboctor:
@atrol better, but I still think it is clutter.  Why do you think this data is important to the administrator?

atrol:
> Why do you think this data is important to the administrator?

@vboctor not very important (similar to oher columns like `date_created`), but can be useful, e.g.
users contact administrators if they are not able to login.
The reason for it is not clear for the user, as our error message `lang_get( 'login_error' )` is intentionally not clear due to security reasons.
The administrator can ask the user how many times he tried to login.
If the user tried < `$g_max_failed_login_count` but `failed_login_count == $g_max_failed_login_count` the admin knows that someone tried to hack the account.

The administrator can sort the user list by failed_login_count.
If there are a lot of users where `failed_login_count == $g_max_failed_login_count`, the administrator knows that someone tries to hack accounts in general.

vboctor:
@atrol 

1. Relating to the error message, I think if an account is blocked, the owner of the account should be able to reset the password.  Currently we block it.  This way the user can unblock themselves if they or someone else reached the login failure max count.

2. Given that we support model where different users can use different authentication models, even if this feature is enabled, it doesn't apply to all users.

Given the first point, I would rather enable password reset than add this column.

atrol:
Closed after there was no unanimous opinion.

atrol:
Reopened based on discussion on https://www.mantisbt.org/bugs/view.php?id=27572

dregad (approved):

vboctor:
Please hold of on merging.

The number of failed logins is an internal implementation details. The concept that we should surface is:

- Is user account locked? UI, API, etc
- Ability to unlock such account. UI, API, etc

There can be several reasons for an account getting locked, we just happen to implement a single one at the moment.

So for this PR, I would at least move from surfacing failed logins to highlight when an account is locked.

dregad:
> The number of failed logins is an internal implementation details. 

IMO it's not strictly internal, since we let admins decide after how many failed attempts an account gets locked ([$g_max_failed_login_count](https://github.com/mantisbt/mantisbt/blob/release-2.25.5/config_defaults_inc.php#L355)). Therefore it is a legitimate request for an admin to know the actual number of failed logins (in addition to the locked status)

I'll grant you that displaying the *number* of failed logins could be irrelevant and unnecessary in some specific cases (although these remain to be identified - do you have any actual use cases ?) or even overkill in general, but if $g_max_failed_login_count != OFF it definitely makes sense to at least display the account's status (locked or not), both in the user list view and the edit page.

atrol:
> makes sense to at least display the account's status (locked or not)

If you drop the failed logn count, you can't sort any longer based on it.
So you can't distinguish between accounts where someones tries brute force attacks and accounts where the user just forgot the password.

PR #1633, opened 2020-03-19
Filter [none] for enum CF 

Selonka:
Fix the problem discribed in https://mantisbt.org/bugs/view.php?id=4864 .

Enable to filter for [none] for enum CF. This may become necessary if an enum cf is introduced later or contains an empty string as an possible value.


dregad:
I don't think this is an appropriate fix for #4684, because in a given instance there could be several enum CFs, some for which using [none] is desired, and others where it is not. Having a global configuration fails to address this requirement.

Selonka:
Ok..thank you for your repsone, so there is a little more to do to solve this issue...
Might it resaonable to add a coloumn to the "mantis_custom_field_table", like "filter_none", to trigger if a filtering for [none] is desired for the specific cf ?

l2m83:
Seems to me this provides a reasonable solution without breaking anything and it's better to have the possibility to filter on none values on every fields than not being able to.
Perhaps you could use the require report property on the custom field which would mean that the user must set the value from the beginning. Otherwise for all fields, it's a possible value therefore the filter option should always be available.

PR #1634, opened 2020-03-20
Fixes #26811: Increase number of char in DNS extension (username)

realitix:
The regex used to validate username limits the extension dns name to 4 chars.
However, lot of domains exceed this rule.
The following python code give the longest dns suffix:

```
import urllib2

def is_ascii(s):
    return all(ord(c) < 128 for c in s)

data = urllib2.urlopen("https://publicsuffix.org/list/public_suffix_list.dat")
nb_char = 0
longest_domain = None
for line in data:
    if not line.startswith('//') and '.' not in line and is_ascii(line):
        d = line.strip()
        if len(d) > nb_char:
            nb_char = len(d)
            longest_domain = d

print("Longest domain {} with {} chars".format(longest_domain, nb_char))
```

The longest dns name is 18 chars

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -3079,7 +3079,7 @@ $g_bug_count_hyperlink_prefix = 'view_all_set.php?type=' . FILTER_ACTION_PARSE_N
  * http://rubular.com/.
  * @global string $g_user_login_valid_regex
  */
-$g_user_login_valid_regex = '/^([a-z\d\-.+_ ]+(@[a-z\d\-.]+\.[a-z]{2,4})?)$/i';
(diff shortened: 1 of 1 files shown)


dregad:
Thanks for your contribution @realitix. 

This code is quite old (2009), predating the introduction of (most of the) longer gTLDs so it never accounted for them.

Anyway, I'm OK with the change, but before it can be merged, please open an issue in [our tracker](https://mantisbt.org/bugs) and update the PR to reference it the commit message (i.e. `Fixes #xxxx`)

realitix:
Hello, I created an issue in the official bugtracker: https://www.mantisbt.org/bugs/view.php?id=26811

PR #1636, opened 2020-03-23
Fix attachments when notes are private by default

vboctor:
Fixes #26805


atrol:
No need for an additional PR as there is already a fix for it in PR https://github.com/mantisbt/mantisbt/pull/1606

vboctor:
@atrol I think this current PR should be targeted for `2.24.1` and #1606 targeted for `2.25.0`.

vboctor:
@atrol are you ok with this?

atrol:
> @atrol are you ok with this?

I am.
I wanted  to be sure that other devs are aware of the other open PR where we might get merge conflicts after this change.

PR #1637, opened 2020-03-23
Personalize LDAP fields cached

maturbet:
To customize the list of ldap account fields to be cached for a user.


config_defaults_inc.php
```diff
@@ -2153,6 +2153,14 @@
  */
 $g_ldap_simulation_file_path = '';
 
+/**
+ * The LDAP fields of a user to be cached
+ * @global array $g_ldap_cache_fields
+**/
+$g_ldap_cache_fields = array(
+	'mail'
```
vboctor: Having `mail` field here doesn't seem to be needed.

maturbet: Yes, it was just weird to have an empty array

maturbet: https://www.mantisbt.org/bugs/view.php?id=26809

atrol:
> Please update `$g_public_config_names` with the new config option name.

@vboctor why do you want it in `$g_public_config_names`?
Similar options like `$g_ldap_uid_field` are not part of this list.

Independent from that, shouldn't we add all ldap options to `$g_global_settings`?

maturbet:
Is there any hope to view it in MantisBT 2.25 ?

vboctor:
@atrol I'm fine with your recommendations.

vboctor:
@maturbet can you elaborate on scenarios where adding more fields (like the room number) would add value?  I look at the issue and don't see the scenarios where this would be useful.

atrol:
> @atrol I'm fine with your recommendations.

@vboctor I opened https://github.com/mantisbt/mantisbt/pull/1639 for 
> Independent from that, shouldn't we add all ldap options to `$g_global_settings`?

Concerning my other recommendation
> @vboctor why do you want it in `$g_public_config_names`?

To be sure, it means you agree that @maturbet should remove the setting from this array?

maturbet:
> @maturbet can you elaborate on scenarios where adding more fields (like the room number) would add value? I look at the issue and don't see the scenarios where this would be useful.


@vboctor here an exemple how we use this functionnality :
[exemple_ldap.txt](https://github.com/mantisbt/mantisbt/files/4401971/exemple_ldap.txt)
Sélection_001
Sélection_002

maturbet:
@dregad 
It was just an example.
The strength of MantisBT is that it is highly customizable, with the possibility of many parameters, the custom_functions as well as the EVENT_* make this tool very attractive.
It's a sad thing that with such a policy, there is such a static list ($t_search_attrs) with hard-coded values and not customizable.

maturbet:
@vboctor @dregad 
An other exemple (not implemented yet in internal plugin) :
Associated with EVENT_MENU_ACCOUNT we will be able to add to the user's menu, specific links composed with data from LDAP.

maturbet:
an other exemple (not from us)
https://www.mantisbt.org/bugs/view.php?id=23392

docbook/Admin_Guide/en-US/config/auth.xml
```diff
@@ -231,6 +231,16 @@ ldaps://ldap.example.com:3269/
 					the data retrieved from LDAP when ON.</para>
 				</listitem>
 			</varlistentry>
+			<varlistentry>
+				<term>$g_ldap_cache_fields</term>
+				<listitem>
+					<para>Choose witch fields will be add to the cache for
```
vboctor, line 237: typo: `witch`.

docbook/Admin_Guide/en-US/config/auth.xml
```diff
@@ -231,6 +231,16 @@ ldaps://ldap.example.com:3269/
...
+						a user from LDAP. $g_ldap_cache_fields =
+						array( 'mail', 'roomNumer' );
```
vboctor, line 239: It may be worth adding that such extra LDAP fields can be used from plugins.

vboctor: I also suggest remove `mail` from the example since it is built in.

config_defaults_inc.php
```diff
@@ -2153,6 +2153,12 @@
  */
 $g_ldap_simulation_file_path = '';
 
+/**
+ * The LDAP fields of a user to be cached
+ * @global array $g_ldap_cache_fields
+**/
```
vboctor, line 2159: Add extra space before the `**/`.

docbook/Admin_Guide/en-US/config/auth.xml
maturbet: Suggestion added in comment.
`mail` removed but others potentials fields  added to see a real array.

config_defaults_inc.php
maturbet: `*` replaced by space to be conform with other comments

dregad:
> image

:joy:

maturbet:
@dregad seriez-vous francophone ? ;-)

dregad:
Je suis démasqué :angel:

maturbet:
@dregad @vboctor 
Does that make sense to be added ?
https://github.com/Association-Cocktail/mantisbt/pull/1

vboctor:
@maturbet I wonder if what is really needed here is a plugin event that allows adding more fields to the list loaded by LDAP API.  This would enable building a plugin that can fetch more LDAP fields and display them or use them in the appropriate scenarios.  I would generally prefer that model over the approach in this PR and the one you linked to for adding to the user profile page.

maturbet:
@vboctor I agree with you, it would certainly be much better.
However, I'm not a developer, I don't know much about php and I wouldn't know how to create a plugin event. I also don't see how to make these fields appear in the user profile pages via a plugin.
But maybe I'll try ...

maturbet:
I don't succeed... I'm not a dev.
And my management does not seem to want to give me more development time on this issue.

maturbet:
During my personal time :
Do you prefer this way ?
https://github.com/mantisbt/mantisbt/pull/1660

maturbet:
Well, I'm throwing in the towel.

PR #1641, opened 2020-03-30
Fix regression when adding files to project documentation

atrol:
Fixes #26835


atrol:
> Should we target this to 2.24.1?

Right, this was my plan, but I worked on the wrong branch.
Thanks @dregad for pushing to master-2.24

PR #1643, opened 2020-04-01
Fix regression no display of OS Version field on View Issue Details page

atrol:
Fixes #26838


atrol:
> I assume you are targeting `2.24.1`.

Right, done via https://github.com/mantisbt/mantisbt/commit/76fa47a885de2d407f49636d1cd2f016abd9a5f6

PR #1646, opened 2020-04-03
Use correct field name "os_build" instead of "os_version"

atrol:
Fixes #26840

Let me know if I should also 
- rename the corresponding message string (`lang_get( 'os_version' )`, create a new message and remove the old one.
- rename CSS classes like `bug-os-version`


atrol:
> Should we use `os_version` as the main field and from configuration perspective treat `os_build` as an alias in order not to break users that may be using `os_build`?

To avoid confusion I would like to use the same name in whole code without any special handling.
If I go on with the current PR, add a new language string and change CSS names, everything is consistent and backwards compatible without any special handling.
(assuming there is hardly anyone who tweaked CSS of this field)

Getting the same for `os_version` means a lot more, as there is also SOAP, REST, name of filter properties, CSV export, Excel export, XML im- / export , name of the database field in `mantis_user_profile_table`, ERD, ... where `os_build` is used.

@vboctor do you think it's worth to change `os_version` ?
If so, I propose to merge the PR as is to fix the bug and create a new issue for additional work.

dregad:
I don't really care if we use one or the other. While _os_version_ maybe more clear than _os_build_, it may be best to go for _os_build_ if it causes less friction to change.

I would say yes to changing also the language string and CSS.

atrol:
> I would say yes to changing also the language string and CSS.

Done

atrol:
@vboctor do you agree to merge?

vboctor:
Should we show the `os_build` field if the fields configuration reference `os_version`?  Otherwise, the field could disappear from deployments where customers overridden the default fields to show and selected `os_version`. 

Thoughts?

atrol:
> Otherwise, the field could disappear from deployments where customers overridden the default fields to show and selected `os_version`. 

The field disappears already in this case since the regression in 2.23.0 https://mantisbt.org/bugs/view.php?id=26838

admin/check/check_config_inc.php
```diff
@@ -95,6 +95,22 @@
...
 );
 
+$t_field_options = array(
+	'bug_report_page_fields',
+	'bug_view_page_fields',
+	'bug_update_page_fields'
+);
+
+foreach( $t_field_options as $t_field_option ) {
+	$t_fields = config_get_global( $t_field_option );
```
vboctor, line 105: This will not capture user or global config in the database.  Note that this is not a global config option (it can be in the database as well).  Anyways, not a big deal, but you may want to config changing this to `config_get()` instead of `config_get_global()`.

atrol: > This will not capture user or global config in the database

Right, I mentioned that already in my note.

> you may want to config changing this to `config_get()` instead of `config_get_global()`

This is not the solution for it, as the check must not depend on currrent project or current user.

I changed to  `config_get( $t_field_option, null, ALL_USERS, ALL_PROJECTS );`, so the case  where the option is set in config_inc.php and the case where the options is set for all projects and all users in database is covered.

The case where someone sets for a specific project and/or a specific user is still not covered.

PR #1650, opened 2020-04-15
Do not use DISTINCT when it is not needed.

raxoft:
I was recently trying to speed up database queries of our Mantis installation, in particular My View and View Issues. I have noticed that Mantis tends to always use DISTINCT even when it is not necessary. While in theory the SQL optimizer should take care of that, in practice removing the DISTINCT results in noticeable speed improvement. On our database with 100k+ bugs it increased the speed of the slowest query (resolved bugs) from 1.7 second to less than 0.5 second.

The patch I provide thus suppresses the DISTINCT for joins which do not require it. Any joins by default turn on DISTINCT again, but I have found four cases when it is not needed. Thanks for reviewing this and considering this for next release.

### DIFF
--- a/core/classes/BugFilterQuery.class.php
+++ b/core/classes/BugFilterQuery.class.php
@@ -124,6 +124,7 @@ class BugFilterQuery extends DbQuery {
 	protected $rt_table_alias_cf; # keep track of the custom field table joins, to reuse them in order by, or serach matching.
 	protected $rt_table_alias_bugnote = null; # keep track of the bugnote table joins.
 	protected $needs_rebuild; # flag to force a rebuild of the final sql when additions are made after the object is first created.
+	protected $needs_distinct; # flag set when the select needs to use the DISTINCT clause.
 
 	/**
 	 * Constructor.
@@ -185,6 +186,7 @@ class BugFilterQuery extends DbQuery {
 		}
 		$this->query_type = $t_query_type;
 		$this->needs_rebuild = true;
+		$this->needs_distinct = false;
 
 		# The query string must be built here to have a valid DbQuery object ready for use
 		$this->build_main();
@@ -292,9 +294,10 @@ class BugFilterQuery extends DbQuery {
 	 * @param string $p_string
 	 * @return void
 	 */
-	public function add_join( $p_string ) {
+	public function add_join( $p_string, $p_needs_distinct = true ) {
 		$this->parts_join[] = $p_string;
 		$this->needs_rebuild = true;
+		$this->needs_distinct |= $p_needs_distinct;
 	}
 
 	/**
@@ -355,7 +358,8 @@ class BugFilterQuery extends DbQuery {
 			$this->build_order_by();
 			$this->unique_query_parts();
 		}
-		$t_select_string = 'SELECT DISTINCT ' . implode( ', ', $this->parts_select );
+		$t_distinct = $this->needs_distinct ? 'DISTINCT ' : '';
+		$t_select_string = 'SELECT ' . $t_distinct . implode( ', ', $this->parts_select );
 		$t_order_string = ' ORDER BY ' . implode( ', ', $this->parts_order );
 		return $t_select_string . $this->helper_string_query_inner() . $t_order_string;
 	}
@@ -365,7 +369,8 @@ class BugFilterQuery extends DbQuery {
 	 * @return string	The constructed query string
 	 */
 	protected function string_query_count() {
-		$t_select_string = 'SELECT COUNT( DISTINCT {bug}.id )';
+		$t_distinct = $this->needs_distinct ? 'DISTINCT ' : '';
+		$t_select_string = 'SELECT COUNT( ' . $t_distinct . '{bug}.id )';
 		return $t_select_string . $this->helper_string_query_inner();
 	}
 
@@ -509,7 +514,7 @@ class BugFilterQuery extends DbQuery {
 	 * @return void
 	 */
 	protected function build_projects() {
-		$this->add_join( 'JOIN {project} ON {project}.id = {bug}.project_id' );
+		$this->add_join( 'JOIN {project} ON {project}.id = {bug}.project_id', false );
 		$this->add_fixed_where( '{project}.enabled = ' . $this->param( true ) );
 
 		$t_user_id = $this->user_id;
@@ -884,7 +889,7 @@ class BugFilterQuery extends DbQuery {
 		}
 		$t_where = '(' . implode( ' OR ', $t_query_or ) . ')';
 		log_event( LOG_FILTERING, 'category query = ' . $t_where );
-		$this->add_join( $t_join );
+		$this->add_join( $t_join, false );
 		$this->add_where( $t_where );
 	}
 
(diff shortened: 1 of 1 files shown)


cproensa:
The query type `QUERY_TYPE_IDS` returns a list of ids without the DISTINCT applied.
I think, if there is any use case of a filter that does not need a unique row for each bug, then that query type must be used.
For example, for composing a query like :
```
SELECT ... FROM ...
WHERE bug_id in ( [filter subquery] )
```
Having that filter subquery built with QUERY_TYPE_IDS. Because the distinct is not needed there and with a good optimizer it will be ignored, but with a bad optimizer it may degrade performance.

If you find any usage of a filter that dos not rely on the premise of unique bug per row, then that is where the code should be modified.

Otherwise, we need the DISTINCT to avoid duplicates in the resultset, which are caused by any table joined with bug table that has a 1:N cardinality. For example: custom fields. tags, relations, notes... That will happen when multiple matches are successful for the same target table


These are the different query types
```
	 * - 'query_type':	Any of QUERY_TYPE_xxx class constants, meaning:
	 *					QUERY_TYPE_LIST, query listing all fields of matched bugs. This is the default.
	 *					QUERY_TYPE_COUNT, query to return number of matched bugs.
	 *					QUERY_TYPE_IDS, query to return only matched bug ids, which may not
	 *						be unique, but may be faster and convenient for use as a subquery.
	 *					QUERY_TYPE_DISTINCT_IDS, query to return unique matched bug ids .
```

Note that this was a recent introduction, so most of the code does not use it explicitly.

Regarding performance with these join intensive queries, in mysql i found very relevant setting a proper value for `innodb_buffer_pool_size` and other memory settings. Usually those are very low by default, and optimizing them can result in a range of 10x speedup for these queries.

raxoft:
Well, I have seen this query type, but Mantis doesn't seem to use it, does it? In fact, this could be changed now, too, as the joins imply if the DISTINCT is needed or not. No need to pass it from the outside - that looks like awkward design in comparison.

As for the mysql itself, I have already tuned it all for maximum performance. Without those changes, the My View was easily taking around 5-8secs. However, with our database size and the temp tables being per connections, I can't increase the temp table size anymore. So actually fixing this inefficiency was the next major speedup available.

cproensa:
I understand your suggestion, but my concern is that it would add a risk of fragility when:
- Multiple rows exists for a table that is expected to be unique. This should not happen in a consistent database, but if it does, the inconsistency would creep into the application.
- External plugins can provide joins when implementing "plugin filters", so this flag should be available to the plugin integration. Otherwise, a plugin that don't set this flag when possible, would disable the improvements for the whole filter
- Again, with plugin filters: if the plugin code is not well designed, and the flag is not set correctly, can cause visible inconsistencies in the app (duplicated rows, incorrect counts)
- This feature needs to be maintained for each table join, at core and plugins. Adds complexity in development, and to maximize benefit from it, it needs to be accounted in as many places as possible.


I don't have much time, but i would like to think and evaluate a different approach:

- **Count**. Many usages of the filters in "my view" and "view all" execute the query twice: one for counting, and other for the actual resultset. Instead of the `count distinct` query, would it be a better performance avoiding the distinct by using something like:
`select count * from {bug} where id in ( [filter-with-QUERY_TYPE_IDS] )`
?

- **results**. Currently, the result is a direct `select *` from the joined query. For  the standard pages, we only read the relevant fields from bug_table. Not 100% sure, but i can't remember right now if there is any code that uses the returned fields from the joined tables. Usually they are retrieved separately based on the bug id, and even in my own plugins never relied on those.
So probably we could restrict the resultset to only those fields, and rearrange the query as:
`select * from {bug} where id in ( [filter-with-QUERY_TYPE_IDS] )`
This way removing the need for distinct. 

There are COUNT and ORDER operations that needs the whole data, so not as fast as a raw sequential read, but my guess is that even if an implicit "distinct" operation is done in the inside, it will be faster with only the bug ids instead of the full rows.

raxoft:
> I understand your suggestion, but my concern is that it would add a risk of fragility when:

First of all, thanks for such a long reply. I appreciate the time it must have taken to write. However, without any offence, I don't think you understand my suggestion yet, as the following points are quite off. Let's see...

> Multiple rows exists for a table that is expected to be unique. This should not happen in a consistent database, but if it does, the inconsistency would creep into the application.

Table with primary key can't have duplicate rows, regardless of the other columns. This holds even if the value of all rows for each other column was the same. The primary key ensures the uniqueness.

> External plugins can provide joins when implementing "plugin filters", so this flag should be available to the plugin integration. Otherwise, a plugin that don't set this flag when possible, would disable the improvements for the whole filter

Yes, plugins can provide joins. This all goes through the `add_join()` call, and that turns the DISTINCT back on. I am fine with that. I prefer to be safe than sorry. I wanted to improve the cases which matter, not necessarily all possible cases. But if you want to expose this to the plugins, feel free to do so.

> Again, with plugin filters: if the plugin code is not well designed, and the flag is not set correctly, can cause visible inconsistencies in the app (duplicated rows, incorrect counts)

Nope, this can't happen. As it is, the plugins can't set the flag in incorrect way. It's like saying they could set the `$needs_rebuild` flag incorrectly.

> This feature needs to be maintained for each table join, at core and plugins. Adds complexity in development, and to maximize benefit from it, it needs to be accounted in as many places as possible.

Again, not true. This does not *require* any extra effort, unless you want to use the feature. There is no extra work *required*.  But of course, if you want to benefit from it, you would have to say whether your join is 1:1 or 1:N. I guess that's understandable.

> I don't have much time, but i would like to think and evaluate a different approach:
> 
> **Count**. Many usages of the filters in "my view" and "view all" execute the query twice: one for counting, and other for the actual resultset. Instead of the `count distinct` query, would it be a better performance avoiding the distinct by using something like:
>       `select count * from {bug} where id in ( [filter-with-QUERY_TYPE_IDS] )`
>       ?

Of course, querying all rows with one query and than getting the count with another query to show just one page worth of data is completely stupid. It's terrible waste and not even transaction consistent. But that's beyond the scope of this fix, I am afraid.

The much better way is doing either:
  1. just one query for all data. Lot of wasted data transferred, but at least it's one query, not two.
  2. one query for all ids, and then fetch just the data using those ids for the page you are interested in. 
  3. in a transaction, get the count with one query and then just one page of data using offset and limit.

The third is in general the best solution, but would require most changes. 

> **results**. Currently, the result is a direct `select *` from the joined query. For  the standard pages, we only read the relevant fields from bug_table. Not 100% sure, but i can't remember right now if there is any code that uses the returned fields from the joined tables. Usually they are retrieved separately based on the bug id, and even in my own plugins never relied on those.
>       So probably we could restrict the resultset to only those fields, and rearrange the query as:
>       `select * from {bug} where id in ( [filter-with-QUERY_TYPE_IDS] )`
>       This way removing the need for distinct.

No, it's not select from all fields by default. It's select only of the fields from the bug table. I.e., `{bug}.*`, not `*`. The fields from other tables are added to the select case by case only when they are needed.

Besides, converting single query into two queries (which is essentially what happens when you use subselect) is not a good idea here. You would pay an increased price even if the single query would suffice.

> There are COUNT and ORDER operations that needs the whole data, so not as fast as a raw sequential read, but my guess is that even if an implicit "distinct" operation is done in the inside, it will be faster with only the bug ids instead of the full rows.

It's never a sequential read anyway, as by default it checks if the bug projects are enabled even if no project is selected. There is always at least one join. In fact, this was the slowest case this patch improves.

cproensa:
> Of course, querying all rows with one query and than getting the count with another query to show just one page worth of data is completely stupid. It's terrible waste and not even transaction consistent. But that's beyond the scope of this fix, I am afraid.

>    in a transaction, get the count with one query and then just one page of data using offset and limit.

IIRC, that's how it's done




> First of all, thanks for such a long reply. I appreciate the time it must have taken to write. However, without any offence, I don't think you understand my suggestion yet, as the following points are quite off. Let's see...

you're right, 
so i'm keeping out of this until i have the time to properly review this topic

PR #1653, opened 2020-04-15
 Mantis BT is using fix cookies in the DB

kleycardona:
Resolved, session fixation vulnerability, Mantisbt using fixed cookies in the DB.
By updating the contents of MANTIS_string_cookie (random unique hash) at login.

Please refer to:
Project: mantisbt
ID: 0011296
Summary: Mantis BT is using fix cookies in the DB

### DIFF
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -1475,6 +1475,24 @@ function user_get_bug_filter( $p_user_id, $p_project_id = null ) {
 }
 
 /**
+ * Update user string cookie
+ *
+ * @param integer $p_user_id A valid user identifier.
+ * @return boolean always true
+ */
+function user_update_string_cookie( $p_user_id ) {
+    $c_user_id = (int)$p_user_id;
+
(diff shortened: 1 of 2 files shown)


kleycardona:
Hi Damien:

I agree that will be a problem.
Do you have any proposal on how to fix this. I'm willing to work on a
solution if you propose one.
Since the cookie data is stored in the DB and always the same, a malicious
user can use this information to break in into someones account.

Project: mantisbt
ID: 11296

Thanks,
Kley Cardona

kleycardona:
Hi Damien:
Here is (a word doc) a proposal for the session fixation issue I was
referring to.
Can you let me know what you think. Is this a good solution?

Thanks,
Kley

dregad:
> Here is (a word doc) a proposal for the session fixation issue

Sorry for not responding earlier, but where is the document ?

dregad:
Related to #1737, which partially addresses the same issue as this PR.

PR #1654, opened 2020-04-18
Project menu bar fix

dregad:
Fixes the following issues

- [0026887](https://mantisbt.org/bugs/view.php?id=26887): Project Menu Bar does not indent subprojects properly
- [0026888](https://mantisbt.org/bugs/view.php?id=26888): Refactor printing of project selection menus

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -515,7 +515,7 @@ function layout_navbar_projects_menu() {
 	if( !layout_navbar_can_show_projects_menu() ) {
 		return;
 	}
-	echo '<li class="grey" id="dropdown_projects_menu">' . "\n";
+	echo "\n" . '<li class="grey" id="dropdown_projects_menu">' . "\n";
 	echo '<a data-toggle="dropdown" href="#" class="dropdown-toggle">' . "\n";
 
 	$t_current_project_id = helper_get_current_project();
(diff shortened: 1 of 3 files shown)


core/layout_api.php
```diff
@@ -633,6 +635,7 @@ function layout_navbar_subproject_option_list( $p_parent_id, $p_project_id = nul
 	array_push( $p_parents, $p_parent_id );
 	$t_user_id = auth_get_current_user_id();
 	$t_project_ids = user_get_accessible_subprojects( $t_user_id, $p_parent_id );
+	$t_indent = str_repeat( '&nbsp;',4 );
```
vboctor, line 638: Fix spacing after the comma.

core/project_api.php
```diff
@@ -936,3 +936,39 @@ function project_format_id( $p_project_id ) {
...
+ */
+function project_link_for_menu( $p_project_id, $p_active = false, $p_class = '', array $p_parents = array(), $p_indent = '' ) {
+	if( $p_parents ) {
+		$t_full_id = implode( ";", $p_parents ) . ';' . $p_project_id;
+		$t_indent = str_repeat( $p_indent, count( $p_parents ) ) . '&nbsp;';
+	} else {
+		$t_full_id = $p_project_id;
+		$t_indent = '';
+	}
+	$t_url = helper_mantis_url( 'set_project.php?project_id=' . $t_full_id );
```
vboctor, line 965: Consider adding a blank line before this one and deleting the blank line after this one.

core/layout_api.php
dregad: Fixed

core/project_api.php
dregad: Fixed

dregad:
> I would prefer not having more view code in project_api and moving that out to print_api, html_api, or layout_api as appropriate.

I did think about this before putting the new function in project_api... IMO it does not belong

- in print_api, because the new function does not actually *print* anything, it returns a formatted string
- in layout_api because it has nothing to do with page layout

html_api would be the "least bad" choice, but I still think it makes sense to have it in project_api.

Function can still be moved later on if necessary, so this is not blocking this PR's merge.

PR #1599, opened 2019-12-17
add UserResetPasswordCommand as PUT /api/rest/users/:id/reset

tgulacsi:
I need a global password reset ability.

PR #1655, opened 2020-04-18
UserResetPassword Command fixes

dregad:
This is a follow-up on PR #1599

- [0026880](https://mantisbt.org/bugs/view.php?id=26880): Impossible to reset user's password
  => Fixes a regression issue
- [0026885](https://mantisbt.org/bugs/view.php?id=26885): Resetting password for protected user via REST API should fail
  => Improves the *users/:id/reset* endpoint

### DIFF
--- a/api/rest/restcore/users_rest.php
+++ b/api/rest/restcore/users_rest.php
@@ -109,7 +109,5 @@ function rest_user_reset_password( \Slim\Http\Request $p_request, \Slim\Http\Res
 	$t_command = new UserResetPasswordCommand( $t_data );
 	$t_result = $t_command->execute();
 
-	return $p_response
-		->withStatus( HTTP_STATUS_SUCCESS )
-		->withJson( $t_result );
+	return $p_response->withStatus( HTTP_STATUS_NO_CONTENT );
 }
--- a/api/rest/restcore/users_rest.php
+++ b/api/rest/restcore/users_rest.php
@@ -107,7 +107,13 @@ function rest_user_reset_password( \Slim\Http\Request $p_request, \Slim\Http\Res
 	);
 
 	$t_command = new UserResetPasswordCommand( $t_data );
-	$t_command->execute();
+	$t_result = $t_command->execute();
+
+	if( $t_result['result'] == UserResetPasswordCommand::RESULT_FAILURE ) {
+		return $p_response
+			->withStatus( HTTP_STATUS_FORBIDDEN )
+			->withJson( array( 'message' => "Cannot reset a protected user's password") );
+	}
 
 	return $p_response->withStatus( HTTP_STATUS_NO_CONTENT );
 }
--- a/api/rest/restcore/users_rest.php
+++ b/api/rest/restcore/users_rest.php
@@ -109,11 +109,7 @@ function rest_user_reset_password( \Slim\Http\Request $p_request, \Slim\Http\Res
 	$t_command = new UserResetPasswordCommand( $t_data );
 	$t_result = $t_command->execute();
 
-	if( $t_result['result'] == UserResetPasswordCommand::RESULT_FAILURE ) {
-		return $p_response
-			->withStatus( HTTP_STATUS_FORBIDDEN )
-			->withJson( array( 'message' => "Cannot reset a protected user's password") );
-	}
-
-	return $p_response->withStatus( HTTP_STATUS_NO_CONTENT );
+	return $p_response
+		->withStatus( HTTP_STATUS_SUCCESS )
+		->withJson( $t_result );
 }
--- a/core/commands/UserResetPasswordCommand.php
+++ b/core/commands/UserResetPasswordCommand.php
@@ -59,7 +59,10 @@ class UserResetPasswordCommand extends Command {
 	 */
 	function validate() {
 		$this->user_id_reset = (int)$this->query( 'id', null );
-		if( $this->user_id_reset <= 0 || !user_exists( $this->user_id_reset ) ) {
+
+		# Make sure the account exists
+		$t_user = user_get_row( $this->user_id_reset );
+		if( $t_user === false ) {
 			throw new ClientException( 'Invalid user id', ERROR_INVALID_FIELD_VALUE, array( 'id' ) );
 		}
 
@@ -78,12 +81,6 @@ class UserResetPasswordCommand extends Command {
 				'Password reset not allowed for protected accounts',
 				ERROR_PROTECTED_ACCOUNT
 			);
-		}
-
-		# @TODO this seems redundant with the check at beginning of function
-		$t_user = user_get_row( $this->user_id_reset );
-		if( $t_user === false ) { // cannot be
-			throw new ClientException( 'Invalid user id', ERROR_INVALID_FIELD_VALUE, array( 'id' ) );
 		}
 
 		# Ensure that the account to be reset is of equal or lower access than
--- a/core/commands/UserResetPasswordCommand.php
+++ b/core/commands/UserResetPasswordCommand.php
@@ -34,6 +34,13 @@ use Mantis\Exceptions\ClientException;
  */
 class UserResetPasswordCommand extends Command {
 	/**
+	 * Constants for execute() method's return value.
+	 */
+	const RESULT_FAILURE = 0;
+	const RESULT_RESET = 1;
+	const RESULT_UNLOCK = 2;
+
+	/**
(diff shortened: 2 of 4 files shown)


dregad:
@tgulacsi feel free to review & test, your feedback would be appreciated.

api/rest/restcore/users_rest.php
```diff
@@ -107,7 +107,13 @@ function rest_user_reset_password( \Slim\Http\Request $p_request, \Slim\Http\Res
 	);
 
 	$t_command = new UserResetPasswordCommand( $t_data );
-	$t_command->execute();
+	$t_result = $t_command->execute();
+
+	if( $t_result['result'] == UserResetPasswordCommand::RESULT_FAILURE ) {
```
vboctor, line 112: We should have this failure throw an exception and shouldn't need this code.

core/commands/UserResetPasswordCommand.php
```diff
@@ -33,6 +33,13 @@
  * }
  */
 class UserResetPasswordCommand extends Command {
+	/**
+	 * Constants for execute() method's return value.
+	 */
+	const RESULT_FAILURE = 0;
```
vboctor, line 39: Remove this constant.

core/commands/UserResetPasswordCommand.php
```diff
@@ -33,6 +33,13 @@
...
+	const RESULT_RESET = 1;
```
vboctor, line 40: Use meaningful strings that we can return to the REST API client in the response and not magic numbers over the wire.

core/commands/UserResetPasswordCommand.php
```diff
@@ -90,13 +97,15 @@ function validate() {
 	protected function process() {
 		# If the password can be changed, we reset it, otherwise we unlock
 		# the account (i.e. reset failed login count)
-		$t_reset = auth_can_set_password( $this->user_id_reset );
-		if( $t_reset ) {
-			$t_result = user_reset_password( $this->user_id_reset );
+		if( auth_can_set_password( $this->user_id_reset ) ) {
+			$t_result = user_reset_password( $this->user_id_reset )
+				? self::RESULT_RESET
+				: self::RESULT_FAILURE; # Shouldn't we throw an exception in this case ?
```
vboctor, line 103: Yes, we should throw an exception.

core/commands/UserResetPasswordCommand.php
```diff
@@ -90,13 +97,15 @@ function validate() {
...
 		} else {
-			$t_result = user_reset_failed_login_count_to_zero( $this->user_id_reset );
+			user_reset_failed_login_count_to_zero( $this->user_id_reset );
+			$t_result = self::RESULT_UNLOCK;
 		}
-		return array();
+
+		return array( 'result' =>  $t_result );
```
vboctor, line 109: I would use `'action' => 'unlock'` or `'action' => 'reset'`.

dregad:
Thanks for the review, I'll address your comments.

dregad:
Not tested yet... it's 2 am and I need some sleep :wink:

manage_user_reset.php
```diff
@@ -52,33 +52,37 @@
...
-	if( false == $t_result ) {
-		# PROTECTED
-		html_operation_failure( $t_redirect_url, lang_get( 'account_reset_protected_msg' ) );
-	} else {
-		# SUCCESSFUL RESET
-		if( ( ON == config_get( 'send_reset_password' ) ) && ( ON == config_get( 'enable_email_notification' ) ) ) {
-			# send the new random password via email
+switch( $t_result ) {
+	case UserResetPasswordCommand::RESULT_RESET:
+		if(    ( ON == config_get( 'send_reset_password' ) )
```
vboctor, line 70: Why do we need to do config changes here?  We can key the language string from the action returned.

manage_user_reset.php
```diff
@@ -52,33 +52,37 @@
...
 		}
-	}
-} else {
-	# UNLOCK
-	html_operation_successful( $t_redirect_url, lang_get( 'account_unlock_msg' ) );
+		break;
+	case UserResetPasswordCommand::RESULT_UNLOCK:
+		html_operation_successful( $t_redirect_url, lang_get( 'account_unlock_msg' ) );
+		break;
+	case UserResetPasswordCommand::RESULT_FAILURE:
```
vboctor, line 83: This is no longer defined and we should depend on the exceptions for failures.

dregad:
I'm sorry, as mentioned that code was untested. 

In fact the points you reviewed were already identified and fixed locally, but as I did not yet finish testing, I did not push them.

manage_user_reset.php
dregad: Not sure what you mean by _config changes_ here.

> We can key the language string from the action returned.

Are you suggesting that the Command should return the language string to use ? If so it would require special handling in users_rest.php to remove this from the returned JSON, as I don't think this this is something that should be exposed to a client consuming the API.

manage_user_reset.php
```diff
@@ -52,33 +52,37 @@
 );
 
 $t_command = new UserResetPasswordCommand( $t_data );
-$t_command->execute();
+$t_result = $t_command->execute();
+$t_result = $t_result['result'];
```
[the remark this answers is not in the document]
dregad: Actually I removed this completely and used `$t_result['action']` in the switch statement

manage_user_reset.php
dregad: Already fixed.

dregad:
Test cases for resetting password:

Test case | Expected result | Test result GUI | Test result REST API (response code)
-- | -- | :--: | :--:
Non-existing account | Error | :+1: | :+1: (400)
Regular account - send_reset_password = ON | E-mail notification sent | :+1: | :+1: (200) {"action":"reset"}
Regular account - send_reset_password = OFF | Password set to blank | :+1: | :+1: (200) {"action":"reset"}
Regular account, auth_can_set_password == false° | Account unlocked | :+1: | :+1: (200) {"action":"unlock"}
Protected account | Failure | :+1: | :+1: (403)
Protected, locked account | Account unlocked | :+1: (using _Unlock Account_ button) | :+1: (200) {"action":"unlock"}
Protected account, auth_can_set_password == false° | Account unlocked | :+1: | :+1: (200) {"action":"unlock"}

° to test this, set `$g_login_method = LDAP` and make sure user's _failed_login_count_ > 0.

Notes
- The Command does not allow to purposefully attempt to unlock the account. 
- To fix this, we need to have a separate command, or pass a parameter to request unlock or reset.
- Since this is an improvement and not related to the regression targeted by this PR, I'll fix this as a separate work item.

@vboctor this is now ready for your review.

dregad:
@vboctor updated per our discussion on Gitter.

manage_user_reset.php
vboctor: I meant why are we doing config checks here.  You can imagine that the command can return a email_sent = true or false, that is checked here and appropriate string is displayed to the user.  That is not blocking, we can do that as a follow up change as we separate the commands later.

core/commands/UserResetPasswordCommand.php
```diff
@@ -49,30 +55,41 @@ function __construct( array $p_data ) {
...
-		if( $t_user === false ) { // cannot be
+		if( $t_user === false ) {
 			throw new ClientException( 'Invalid user id', ERROR_INVALID_FIELD_VALUE, array( 'id' ) );
 		}
 
-		# Ensure that the account to be reset is of equal or lower access to the
-		# current user.
+		# Mantis can't reset protected accounts' passwords, but if the
+		# account is locked, we allow the operation as Unlock
+		if( auth_can_set_password( $this->user_id_reset )
```
vboctor, line 76: Why are we ANDing these checks?  I would expect an error if `auth_can_set_password()` is false.  I also expect a separate check that errors out if account is protected.  I don't think we should worry about user can sign in or not (i.e. remove user_is_login_request_allowed).

dregad:
Opened [#26929](https://mantisbt.org/bugs/view.php?id=26929) to track creation of a new command + REST API endpoint to unlock a user's account.

core/commands/UserResetPasswordCommand.php
dregad: This is just replicating the logic as it was before the Command was introduced. Without this, there would be no way to unlocked a protected account. 

The user_is_login_request_allowed() calls basically checks whether the account is locked or not. This will disappear when the new User Unlock command is added.

PR #1660, opened 2020-04-21
Add EVENTs to get additionals user attributes from LDAP and show them

maturbet:


maturbet:
Plugin exemple for use : https://github.com/Association-Cocktail/LdapCacheFields

maturbet:
@vboctor 
Is it what you expect here ? https://github.com/mantisbt/mantisbt/pull/1637#issuecomment-612309096

maturbet:
> 1. For attributes to display the plugin event should return logical data that is rendered by core. We have been moving away from have events echo html output. Reason is that it is not API friendly and breaks whenever Mantis UI is changed.

I think I can make that change.

> 2. For the extra LDAP fields, why not just use a config option rather than a plugin event?

I used a plugin event following a [past suggestion](https://github.com/mantisbt/mantisbt/pull/1637#issuecomment-612309096) but I'll use the method you prefer ...

> 3. Should the extra LDAP info be included in the REST API that retrieves user info? eg in ldap_fields element?

Probably, it makes sense.

maturbet:
Hello @vboctor,

> 1. For attributes to display the plugin event should return logical data that is rendered by core. We have been moving away from have events echo html output. Reason is that it is not API friendly and breaks whenever Mantis UI is changed.
I have moved the HTML as requested.

> 2. For the extra LDAP fields, why not just use a config option rather than a plugin event?
Which decision do you make? A plugin or a config option?

> 3. Should the extra LDAP info be included in the REST API that retrieves user info? eg in ldap_fields element?
Is implementation in the API rest essential to validate this PR?

dregad:
Hello @maturbet, 

Please be aware that I merged #2053 (to be included in MantisBT 2.28.0), which conflicts with this PR and will therefore require you to adapt the code to use the new `EVENT_LDAP_CACHE_ATTRS` instead of the one you defined in https://github.com/mantisbt/mantisbt/pull/1660/commits/b1d4e912541b74e5896225e54e1b14ee96fa86ea (`EVENT_LDAP_USER_FIELDS`).

I guess that answers
>> 2. For the extra LDAP fields, why not just use a config option rather than a plugin event?
>
> Which decision do you make? A plugin or a config option?

Your PR only allows adding fields to view_user_page.php and account_page.php; wouldn't it make sense to also allow displaying the extra attributes in manage_user_edit_page.php also ?

As for 
> Is implementation in the API rest essential to validate this PR?

I know @vboctor is quite insistant on being able to do everything with the REST API, so I would say yes. It's not a very complicated change; I would propose the following (takes into account base64 encoding of binary data, as the use case for #2053 was retrieving an image, which would break JSON):

```patch
diff --git a/api/soap/mc_api.php b/api/soap/mc_api.php
index 6a1f25e3d..6c7dcd9f2 100644
--- a/api/soap/mc_api.php
+++ b/api/soap/mc_api.php
@@ -347,6 +347,20 @@ function mci_user_get( $p_user_id, $p_select = null ) {
                        $t_user_data['created_at'] = ApiObjectFactory::datetime( user_get_field( $p_user_id, 'date_created' ) );
                }
 
+               // Extra LDAP attributes, excluding realname and email
+               $t_ldap_fields = ldap_cache_user_data( $t_user_data['name'] );
+               if( $t_ldap_fields !== false ) {
+                       unset( $t_ldap_fields[config_get_global( 'ldap_email_field' )] );
+                       unset( $t_ldap_fields[config_get_global( 'ldap_realname_field' )] );
+                       foreach( $t_ldap_fields as $t_ldap_field => $t_ldap_value ) {
+                               if( !mb_check_encoding($t_ldap_value, 'UTF-8') ) {
+                                       # It's binary data
+                                       $t_ldap_value = base64_encode( $t_ldap_value );
+                               }
+                               $t_user_data['ldap_extra_attr'][$t_ldap_field] = $t_ldap_value;
+                       }
+               }
+
                if( is_null( $p_select ) || in_array( 'projects', $p_select ) ) {
                        $t_project_ids = user_get_accessible_projects( $p_user_id, /* disabled */ false );
                        $t_projects = array();
```

dregad:
Also, this PR is missing a reference to an issue in the bugtracker.

maturbet:
After 4 years, I think I'll stick mostly to @raspopov  's additions (thanks!) and adapt my plugin to match #2053
I'll close this PR if my tests are conclusive.
And I'll leave the user pages to you!

maturbet:
It's OK for me.
There's no UI for viewing cached LDAP items, but at least it's in the official Mantis GIT.

raspopov:
@maturbet Please share your practical idea that you would like to implement in the code. Just displaying user information, i.e. without any additional processing such as filtering in reports? I would also like to add avatar display to user information, аnd I'm thinking about how to do this in the most universal and separate format from the UI rendering...

maturbet:
@raspopov Our first need was to retrieve and cache some user data from LDAP and reuse it in custom attributes in the ticket creation process.

I created this plugin to cache these fields https://github.com/Association-Cocktail/LdapCacheFields (With UI : [TAG 0.4](https://github.com/Association-Cocktail/LdapCacheFields/tree/0.4))

The added rendering in the UI was just to view the data for each of the users.

As I'm not a developer but a sysadmin, I'm not going to spend much more time on it.

PR #1661, opened 2020-04-21
Move release packaging to mantisbt repo

vboctor:
Move release packaging scripts from mantisbt-tools to mantisbt main repository. This is useful in the following ways:

- It makes it easy to co-commit changes to MantisBT with changes to package it.
- The command line to create a release package remains consistent, while the implementation can be different based on the branch.

Fixes #[26903](https://www.mantisbt.org/bugs/view.php?id=26903)


scripts/README.md
```diff
@@ -1,18 +1,55 @@
...
+```
+
+- Update to `MANTIS_VERSION` in `core/constant_inc.php`
+- Update `Revision_History.xml` for Admin Guide
+- Update `Revision_History.xml` for Developer Guide
+- Checkin `Update release to x.y.z`
+
+### Tag the release
+
+These steps typically happen from the machine used for official buids which
```
dregad, line 22:
```suggestion
These steps typically happen from the machine used for official builds which
```

vboctor:
@dregad 

- I have decided it is cleaner to have `scripts` for distribution and `build-scripts` for continuous integration and packaging.
- Moved Travis scripts to `build-scripts`.
- Referenced the commit hash in the commit which brings over the scripts.  I decided not to update each script with the commit hash.
- Brought over more scripts that does some kind of validation on the repository.  We can trim these later if we don't want them.
- I create a `legacy-releases` branch in mantisbt-tools, then delete the scripts from master.  This way we can use the scripts for releases like 1.3.x.
- Starting 2.25.0 release, I would like to use the local scripts.  For older releases, use `mantisbt-tools`
- Fixed minor changes / typos in readme (as applicable after separation into two readme files).

vboctor:
Here is the related `mantisbt-tools` PR:

https://github.com/mantisbt/mantisbt-tools/pull/2

dregad:
@vboctor I think you went a bit too fast with the merge here, considering that **you did not address my review comment**:
> ### Important
> 
> This change will almost certainly break scheduled builds processes on mantisbt.org server. Please make sure to check and adapt them as appropriate. At least take a close look at:
> 
>     * _nightly-builds.sh_ script in mantisbt-tools repo (symlinked from /etc/cron.daily)
>     * _mantis-docbook_ build script is in /etc/cron.hourly (local only, not in mantisbt-tools repo)

Now about your last note https://github.com/mantisbt/mantisbt/pull/1661#issuecomment-619483016
> I have decided it is cleaner to have scripts for distribution and build-scripts for continuous integration and packaging.

OK with the approach, but I would prefer to have the directory simply named `build`, because it is shorter and also we might want to put something other than scripts in there (for example, [CodeSniffer configuration](https://github.com/dregad/mantisbt/tree/codesniffer), etc).

Mind if I rename it ?

> decided not to update each script with the commit hash

OK, good enough.

> scripts that does some kind of validation on the repository. We can trim these later if we don't want them.

OK

> I create a legacy-releases branch in mantisbt-tools, then delete the scripts from master. This way we can use the scripts for releases like 1.3.x.

Unfortunately this is not good enough and, as mentioned before, will break the nightly builds process as it is currently implemented and deployed on the server. [nightly-builds.sh](https://github.com/mantisbt/mantisbt-tools/blob/master/nightly-builds.sh) must be adapted. IMO that's a **blocker for 2.25.0 release**.

To simplify that change, I recommend to stop building 1.3.x nightly, which does not make sense for a branch that is only seeing security fixes nowadays, and assume that the master-2.x stable branch does contains the build scripts (which should be the case after 2.25.0).

PR #1663, opened 2020-04-24
Add filter on versions to hide obsolete ones

maturbet:
Fixes [#26901](https://mantisbt.org/bugs/view.php?id=26901)

### DIFF
--- a/lang/strings_french.txt
+++ b/lang/strings_french.txt
@@ -760,6 +760,7 @@ $s_released = 'Publiée';
 $s_not_released = 'Pas encore publiée';
 $s_scheduled_release = 'Publication planifiée';
 $s_obsolete = 'Périmée';
+$s_hide_obsolete = 'Cacher les versions Périmées';
 $s_version_updated_msg = 'La version a été mise à jour avec succès...';
 $s_account_delete_protected_msg = 'Ce compte est protégé. Vous ne pouvez pas le supprimer.';
 $s_account_deleted_msg = 'Compte supprimé...';
--- a/lang/strings_french.txt
+++ b/lang/strings_french.txt
@@ -760,7 +760,6 @@ $s_released = 'Publiée';
 $s_not_released = 'Pas encore publiée';
 $s_scheduled_release = 'Publication planifiée';
 $s_obsolete = 'Périmée';
-$s_hide_obsolete = 'Cacher les versions Périmées';
 $s_version_updated_msg = 'La version a été mise à jour avec succès...';
 $s_account_delete_protected_msg = 'Ce compte est protégé. Vous ne pouvez pas le supprimer.';
 $s_account_deleted_msg = 'Compte supprimé...';
--- a/manage_proj_edit_page.php
+++ b/manage_proj_edit_page.php
@@ -76,6 +76,7 @@ auth_reauthenticate();
 $f_project_id = gpc_get_int( 'project_id' );
 $f_show_global_users = gpc_get_bool( 'show_global_users' );
 $f_show_obsolete = gpc_get_bool( 'showobsolete' );
+$f_release_type = gpc_get_int( 'release_type', 2 );
 
 # true = show obsolete versions, anything else = hide them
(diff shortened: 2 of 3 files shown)


manage_proj_edit_page.php
```diff
@@ -75,6 +75,11 @@
 
 $f_project_id = gpc_get_int( 'project_id' );
 $f_show_global_users = gpc_get_bool( 'show_global_users' );
+$f_hide_obsolet = gpc_get_bool( 'hideobsolet' );
+
+# OFF = show obsolet versions, anything else = hide them
+$c_hide_obsolet = ( $f_hide_obsolet == OFF ) ? OFF : ON;
+$t_hide_obsolet_filter = '&amp;hideobsolet=' . $c_hide_obsolet;
```
dregad, line 82: use `&` not `&amp;` otherwise there when you call print_link_button() below, the Edit button's URL will have double encoding and the flag will not be passed on correctly:
`http://path.to/mantis/manage_proj_edit_page.php?project_id=132&amp;hideobsolet=1`

manage_proj_edit_page.php
```diff
@@ -492,11 +499,25 @@
 			<input type="submit" name="copy_to" class="btn btn-sm btn-primary btn-white btn-round" value="<?php echo lang_get( 'copy_versions_to' ) ?>" />
 		</fieldset>
 	</form>
-	</div>
+				</div>
+				<div class="pull-right">
+    <form id="manage-project-version-filter" method="post" action="manage_proj_edit_page.php?project_id=<?php echo $f_project_id ?>" class="form-inline">
```
dregad, line 504: It would be more user-friendly if you jumped back to the Versions section of the page after submitting the form.

manage_proj_edit_page.php
```diff
@@ -492,11 +499,25 @@
...
+        <fieldset>
+			<input type="hidden" class="form-control input-sm" name="project_id" value="<?php echo $f_project_id ?>" />
```
dregad, line 506: You're providing the project_id both here and in the form's action property, that's redundant

manage_proj_edit_page.php
```diff
@@ -492,11 +499,25 @@
...
+		</fieldset>
+	</form>
+				</div>
+			</div>
+		</div>
 		<div class="widget-body">
 		<div class="widget-main no-padding">
 	<?php
-	$t_versions = version_get_all_rows( $f_project_id, VERSION_ALL, true );
+	$t_versions = version_get_all_rows( $f_project_id, VERSION_ALL, helper_check_variables_equal( (int)$c_hide_obsolet, OFF, false ) );
```
dregad, line 520: Needlessly complicated - just use the boolean value (negate if needed)

maturbet:
@dregad 
> To avoid a double negation, it's probably better to name the variables `show_obsolete`
> I would also suggest to hide the obsolete versions by default.

OK

> Please avoid long lines, wrap at 80 if possible, and don't make them wider than 120 chars.

I'm gonna do my best.

> Update your commit message to reference the bugtracker issue you're fixing, i.e. `Fixes #26901` so it gets picked up by source integration when committed.

Did you do it for me?
I don't know what else to do?

> Issue #26901 mentions filtering also on _released_ flag, are you planning to integrate this as well ?

Why not. But only if this one is OK for you ...

lang/strings_french.txt
```diff
@@ -768,6 +768,7 @@ $s_released = 'Publiée';
 $s_not_released = 'Pas encore publiée';
 $s_scheduled_release = 'Publication planifiée';
 $s_obsolete = 'Périmée';
+$s_hide_obsolete = 'Cacher les versions Périmées';
```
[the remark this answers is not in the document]
maturbet: removed from

maturbet:
@dregad 
https://github.com/Association-Cocktail/mantisbt/pull/2
Is it the way you expect ?

dregad:
Replaced by #1927

PR #1664, opened 2020-04-25
Added KEYS.md file listing all PGP keys used to sign releases

dregad:
Fixes [#22269](https://mantisbt.org/bugs/view.php?id=22269)


dregad:
See related PR mantisbt/mantisbt-web#5 to link the KEYS file from [mantisbt.org's Downloads page ](https://mantisbt.org/download.php).

PR #1669, opened 2020-05-16
Password managers shouldn’t trigger on monitor box

vboctor:
Fixes #26963

### DIFF
--- a/bug_view_inc.php
+++ b/bug_view_inc.php
@@ -681,8 +681,7 @@ if( $t_flags['monitor_show'] ) {
 			<form method="get" action="bug_monitor_add.php" class="form-inline noprint">
 			<?php echo form_security_field( 'bug_monitor_add' ) ?>
 				<input type="hidden" name="bug_id" value="<?php echo (integer)$f_issue_id; ?>" />
-				<label for="bug_monitor_list_username"><?php echo lang_get( 'username' ) ?></label>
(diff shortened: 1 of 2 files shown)


vboctor:
@mantisbt/core-team ping

bug_view_inc.php
```diff
@@ -681,8 +681,7 @@
 			<form method="get" action="bug_monitor_add.php" class="form-inline noprint">
 			<?php echo form_security_field( 'bug_monitor_add' ) ?>
 				<input type="hidden" name="bug_id" value="<?php echo (integer)$f_issue_id; ?>" />
-				<label for="bug_monitor_list_username"><?php echo lang_get( 'username' ) ?></label>
-				<input type="text" class="input-sm" id="bug_monitor_list_username" name="username" />
+				<input type="text" class="input-sm" id="bug_monitor_list_user_to_add" name="user_to_add" />
```
atrol, line 684: Did you try `autocomplete="off"` ?

Is there some standard how password managers are triggered ?

dregad: I don't know about a _standard_, but for example [Kee](https://addons.mozilla.org/en-GB/firefox/addon/keefox/) relies ona combination of white/black lists based on form or field name and id.

vboctor: @atrol I believe `autocomplete="off"` was the first thing I tried and it didn't work.

PR #1670, opened 2020-05-18
Improved filter : none and not empty are available on custom fields a…

l2m83:
…nd user fields

This pull request aims to provide a patch for these issues. It adds a new option [not empty] to user field filters and custom filters (except for text area).

0004864: Unable to filter on empty custom field value 
0008149: New filter option: [Someone] for Assigned to and other filter fields 
0012798: Add a filter option [Something] as opposite of [None] i.e. field not null 
(and other closed/duplicate issues that are attached)

I have tried to make the smallest possible changes to the existing code except for one function in core/filter_form_api.php which has been factorized. More improvements could have been made to this class but i didn't want to risk having unwanted behaviours.

I hope the patch will be accepted into the official code and i'm expecting your remarks if additional code improvements are needed.

### DIFF
--- a/core/classes/BugFilterQuery.class.php
+++ b/core/classes/BugFilterQuery.class.php
@@ -830,7 +830,8 @@ class BugFilterQuery extends DbQuery {
 	 * @return void
 	 */
 	protected function build_prop_handler() {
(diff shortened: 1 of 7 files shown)


stevecharon:
@cproensa Can you review this or let another reviewer know about this in order to move forward?

PR #1677, opened 2020-05-28
TOO_MANY_REDIRECTS on report issue

maturbet:
https://www.mantisbt.org/bugs/view.php?id=26988

### DIFF
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -138,6 +138,13 @@ if( $f_master_bug_id > 0 ) {
 		$t_project_id = $t_default_project;
 	}
 
+	# Check for bug report threshold
+	if( !access_has_project_level( config_get( 'report_bug_threshold' ) ) ) {
+		# If can't report on current project, show project selector if there is any other allowed project
+		access_ensure_any_project_level( 'report_bug_threshold' );
+		print_header_redirect( 'login_select_proj_page.php?ref=bug_report_page.php' );
+	}
+
 	if( ( ALL_PROJECTS == $t_project_id || project_exists( $t_project_id ) )
 		&& $t_project_id != $t_current_project
 		&& project_enabled( $t_project_id ) ) {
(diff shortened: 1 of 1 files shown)


dregad:
Thanks for your contribution.

I have tested your PR, and it fixes the problem without introducing any regression (as far as I can tell).

That being said, in terms of data integrity I don't think the users' _default_project_ setting should contain an invalid project value, which can happen when a user is removed from the project, or when the project is deleted. 

Maybe we should enforce resetting the default project to ALL_PROJECTS in such case, what do you think ?

dregad:
As a side note, for future contributions
- Commits should actually describe what you changed and why, `Fixes #xxxx` is not sufficient.
- When submitting PR's, you should always **rebase** your feature branch onto mantisbt/mantisbt _master_ (or _master-x.xx_ for a patch release) -- unless it's a very complex, long-running branch that would be too difficult or complex to rebase, in which case merge commits are acceptable.

Unless you're using it for your own purposes, I also recommend to keep your _master_ branch aligned with and identical to ours (i.e. `git fetch upstream && git reset --hard upstream/master`) instead of merging (like you did at c142c21c8f4e6d8edbb9c2c4f5c816c5df213f55). Then you can simply run `git pull upstream` to keep it in sync.

maturbet:
> Maybe we should enforce resetting the default project to ALL_PROJECTS in such case, what do you think ?

This is indeed a feature I had considered but didn't feel like coding myself.

> As a side note, for future contributions [...]

I will try to improve myself, I'm not a developer nor a GIT regular. Thanks for your recommendations.

dregad:
> This is indeed a feature I had considered but didn't feel like coding myself.

I'll take care of it.

> I'm not a developer nor a GIT regular

Feel free to ask if you need help or advice.

maturbet:
Because an administrator can cause this bug, I designed [this feature](https://github.com/mantisbt/mantisbt/pull/1678) to prevent the selection of a default project, inaccessible to a user.

maturbet:
> Feel free to ask if you need help or advice.

Is it better ? [b8bb06338c](https://github.com/Association-Cocktail/mantisbt/commit/b8bb0638c2cf0ab14dfcc9f0cd4bf58412b67f24)
`git pull upstream master`
`git push`

dregad:
So I finally managed to finish testing this :relieved: 
Thanks for your contribution.

I reworded the commit, merged in b162f8fbf500d4bb268a8880a5e30ed3e40a8a57

dregad:
> That being said, in terms of data integrity I don't think the users' default_project setting should contain an invalid project value, which can happen when a user is removed from the project, or when the project is deleted.

This is tracked in https://mantisbt.org/bugs/view.php?id=27144, fix is in #1691

PR #1685, opened 2020-07-27
#26992 Add  jQuery's select2 for more beautiful and usable selects

being-peace:
See https://www.mantisbt.org/bugs/view.php?id=26992


c-schmitz:
@vboctor Why was this never merged? This would really be a huge help and a first step to Ajax based select boxes. (needed if you have a big userbase)

PR #1686, opened 2020-07-28
Fixed parameters borked as soon as code was going through another sql(?)

FrankyB:
Hi!

See this error message when having target versions & trying to resolve an issue via plugins/Source. 

If you like to see, the parameter numbering goes up to 18, then another db action interrupts(uh, thats what I think, at least), and the parameter numbering starts from $1 again (at target_version).

I simply store the condition in a local variable and therefore have no need to restart db_param()

I am not sure if this is a good fix acording to your overall coding. Maybe I can also store db_param() somewhere, do another DB operation and pop it afterwards.

It would be good to have this fixed, it was at least disturbing plugins/Source.

Thank you,
Frank

LINE 11:       target_version=$1,<br />
                              ^<br />
DETAIL:  integer gegen character varying for the query: UPDATE mantis_bug_table<br />
					SET project_id=$1, reporter_id=$2,<br />
						handler_id=$3, duplicate_id=$4,<br />
						priority=$5, severity=$6,<br />
						reproducibility=$7, status=$8,<br />
						resolution=$9, projection=$10,<br />
						category_id=$11, eta=$12,<br />
						os=$13, os_build=$14,<br />
						platform=$15, version=$16,<br />
						build=$17, fixed_in_version=$18,<br />
						target_version=$1,<br />
						view_state=$2,<br />
						summary=$3,<br />
						sponsorship_total=$4,<br />
						sticky=$5,<br />
						due_date=$6<br />
					WHERE id=$7.</p>

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -669,6 +669,7 @@ class BugData {
 		#  as unix timestamps which could confuse the history log and they
 		#  shouldn't get updated like this anyway.  If you really need to change
 		#  them use bug_set_field()
+		$targetversionupdate = access_has_project_level( config_get( 'roadmap_update_threshold' ) );
 		db_param_push();
 		$t_query = 'UPDATE {bug}
 					SET project_id=' . db_param() . ', reporter_id=' . db_param() . ',
@@ -693,7 +694,7 @@ class BugData {
 			$this->build, $this->fixed_in_version,
 		);
 		$t_roadmap_updated = false;
-		if( access_has_project_level( config_get( 'roadmap_update_threshold' ) ) ) {
+		if ( $targetversionupdate ) {
 			$t_query .= '
 						target_version=' . db_param() . ',';
 			$t_fields[] = $this->target_version;


dregad:
@FrankyB Thanks for your contribution. Please note that issues should be reported at https://mantisbt.org/bugs, GitHub PR's are only to submit a patch, which should reference an existing issue.

That being said, I'm afraid that your PR is only addressing the symptoms and not the root cause of the problem; unfortunately you are not giving sufficient information to reproduce the problem, and it is not clear whether it is caused by MantisBT core, by the Source Integration plugin (I assume that's what you mean by _plugins/Source_, and by the way you did not specify which VCS plugin you're using); for the record we have been running that without issues for many years.

Please provide version information, plugins used and detailed steps to reproduce. A stack trace just before the point where the param count gets reset would be helpful as well.

atrol:
@FrankyB please provide also information about the database you use.
I suspect it's not MySQL.

FrankyB:
Thank you. I will create am issue, how to reproduce the bug, and why I am confident its not only happening with me and the source-integration-plugin.

@dregad: Please tell me how do I create a stack trace at that point
@atrol: You are right, I am using postgresql  
PostgreSQL 9.6.17 on x86_64-pc-linux-gnu, compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516, 64-bit
and
PostgreSQL 12.3 (Debian 12.3-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit

dregad:
> how do I create a stack trace at that point

That depends on the tools you're using, e.g. with PHPStorm you can step through the code in debug mode and copy the stack trace. More bare-bones, you could add `trigger_error('stacktrace',ERROR);` (with $g_show_detailed_errors = true) in the right location and reproduce the error; could be at bug_api.php:696 or in the actual location where the query parameter count ($g_db_param->count) gets reset from 18 to 1, if you know it.

dregad:
Follow-up and discuss in https://mantisbt.org/bugs/view.php?id=27113

dregad:
Closing this, as I'm pretty sure now that the problem is with Source Integration plugin

PR #1688, opened 2020-08-03
MantisGraph updates

dregad:
Fixes the following issues

- [27122](https://mantisbt.org/bugs/view.php?id=27122): 3rd-party plugins cannot use chart.js library bundled with MantisGraph
- [27123](https://mantisbt.org/bugs/view.php?id=27123): MantisGraph: stop using chart.js bundled build
- [27124](https://mantisbt.org/bugs/view.php?id=27124): MantisGraph: update Chart.js library to v2.9.3


atrol:
Had a very short look, seems to run in general, but I get one new warning and one new error (compared to current master) when checking the browser console in Firefox.

Sorry, no time to provide English version of the warning, but should be clear

```
Das Skript von "http://localhost:8888/atrol/mastertest/plugin_file.php?file=MantisGraph/Chart.bundle-2.9.3.min.js" wurde geladen, obwohl sein MIME-Typ ("") kein für JavaScript gültiger MIME-Typ ist.

plugin.php
Uncaught SyntaxError: expected expression, got '<'
plugin_file.php:1
```

dregad:
Thanks for testing, good catch

> I suspect you have tested with setting `$g_cdn_enabled = ON;` as the warnig and the error do just occur with standard setting `$g_cdn_enabled = OFF;`

That's exactly it. I forgot to remove the bundle javascript in the else block. Will fix shortly.

PR #1689, opened 2020-08-05
Preserve filters on subprojects

dregad:
Filters of subprojects were not kept when switching projects, the filter
of the highest ancestor was applied instead.

Patch by @polzin

Fixes [#27129](https://mantisbt.org/bugs/view.php?id=27129)

### DIFF
--- a/set_project.php
+++ b/set_project.php
@@ -90,7 +90,7 @@ if( !is_blank( $c_ref ) ) {
 
 		# if view_all_bug_page, pass on filter
 		if( strcasecmp( 'view_all_bug_page.php', $t_referrer_page ) == 0 ) {
-			$t_source_filter_id = filter_db_get_project_current( $f_project_id );
+			$t_source_filter_id = filter_db_get_project_current( $t_bottom );
 			$t_redirect_url = 'view_all_set.php?type=' . FILTER_ACTION_GENERALIZE;
 
 			if( $t_source_filter_id !== null ) {


atrol:
Didn't test, but when looking at the code change I got aware that in some cases  $f_project_id can be a string with a semicolon separated list of project id's in it.
I am wondering if `helper_set_current_project( $f_project_id );` on line 73 works in this case, as the parameter is expected to be a integer project id.
Maybe this has to be `$t_top` instead of `$f_project_id`

dregad:
PHP (int) cast of such a semi-column-separated string de facto keeps the first number, so I don't expect problems if the string is used as integer.

```php
(int)"1;2;3" == 1
```

That being said all *helper_set_current_project()* does not interpret the $p_project_id parameter and just stores it in the Project cookie as-is. *helper_get_current_project()* properly deals with `;`-delimited lists, and in fact expects this format for subprojects; it returns an array of project ids. 

So IMO we should not change this, and I think that it's the PHPDoc block for *helper_set_current_project()* which incorrectly indicates the type of $p_project_id parameter as integer (should be `integer|string`)

atrol:
> PHP (int) cast of such a semi-column-separated string de facto keeps the first number,

Seems I will hardly ever understand PHP 😄 

>  _helper_get_current_project()_ properly deals with `;`-delimited lists, and in fact expects this format for subprojects; it returns an array of project ids.

Confused, please point me to the line where an array of project ids is returned.

dregad:
> Confused, please point me to the line where an array of project ids is returned.

Apologies, I was actually looking at helper_get_current_project_trace() instead of helper_get_current_project() :zany_face: 

Anyway the helper_get_current_project() also deals with the ;-delimited list, and returns the last element.

PR #1690, opened 2020-08-07
Wrap long text strings in tables

syncguru:
Fixes #27114

Before:

Screen Shot 2020-08-07 at 3 39 00 PM

After:

Screen Shot 2020-08-07 at 3 35 44 PM

### DIFF
--- a/css/ace-mantis.css
+++ b/css/ace-mantis.css
@@ -169,6 +169,7 @@
 
 table {
     background-color: #ffffff !important;
+    overflow-wrap: anywhere;
 }
 
 /* Disable printing of link urls */


css/ace-mantis.css
```diff
@@ -169,6 +169,7 @@
 
 table {
     background-color: #ffffff !important;
+    overflow-wrap: anywhere;
```
atrol, line 172: Unfortunately not supported by all browsers
https://caniuse.com/#feat=mdn-css_properties_overflow-wrap_anywhere

syncguru: Yeah... Unfortunately, CSS compliance remains inconsistent.
The goal is to address the issue for most users and these changes does that while not impacting users who happen to use uncommon browsers or versions.

atrol: Got aware of it as PhpStorm CSS inspection treats it as `Mismatched property value` error, but this seems to be caused by outdated validation data.

atrol:
ViewIssues
@syncguru @dregad I don't like the side effects of this. E.g. check the headers on `View Issues` page and the wrapping of ids, user names, dates, ...
What do you think? Revert?

dregad:
@atrol I did not notice this behavior in my tests, but I agree that in this case the cure is worse than the disease. Reverting for now.

@syncguru maybe you can find a way to apply the CSS more selectively ?

dregad:
Reverted in 60d28cce0569d2acd1daccce930e5215a1abda72 and issue [#27144](https://mantisbt.org/bugs/view.php?id=27144) reopened.

PR #1695, opened 2020-09-02
Fix bugnote_clear_cache()

dregad:
bugnote_clear_cache() only removed the note from $g_cache_bugnotes_by_id
if it has also been stored in $g_cache_bugnotes_by_bug_id, but failed in
the scenario where the bug-level cache has not been populated.

Adding a specific unset() call to ensure the bugnote is actually removed
from both caches.

Fixes [#27217](https://www.mantisbt.org/bugs/view.php?id=27217)

### DIFF
--- a/core/bugnote_api.php
+++ b/core/bugnote_api.php
@@ -808,9 +808,12 @@ function bugnote_clear_cache( $p_bugnote_id = null ) {
 		$g_cache_bugnotes_by_id = array();
 		$g_cache_bugnotes_by_bug_id = array();
 	} else {
-		if( isset( $g_cache_bugnotes_by_id[(int)$p_bugnote_id] ) ) {
-			$t_note_obj = $g_cache_bugnotes_by_id[(int)$p_bugnote_id];
-			# current note id will be unset in the following call
+		$p_bugnote_id = (int)$p_bugnote_id;
+		if( isset( $g_cache_bugnotes_by_id[$p_bugnote_id] ) ) {
+			$t_note_obj = $g_cache_bugnotes_by_id[$p_bugnote_id];
+			unset($g_cache_bugnotes_by_id[$p_bugnote_id]);
+
+			# Clear the bug-level cache for the bugnote's parent bug
 			bugnote_clear_bug_cache( $t_note_obj->bug_id );
 		}
 	}


core/bugnote_api.php
```diff
@@ -808,9 +808,12 @@ function bugnote_clear_cache( $p_bugnote_id = null ) {
 		$g_cache_bugnotes_by_id = array();
 		$g_cache_bugnotes_by_bug_id = array();
 	} else {
-		if( isset( $g_cache_bugnotes_by_id[(int)$p_bugnote_id] ) ) {
-			$t_note_obj = $g_cache_bugnotes_by_id[(int)$p_bugnote_id];
-			# current note id will be unset in the following call
+		$p_bugnote_id = (int)$p_bugnote_id;
+		if( isset( $g_cache_bugnotes_by_id[$p_bugnote_id] ) ) {
+			$t_note_obj = $g_cache_bugnotes_by_id[$p_bugnote_id];
+			unset($g_cache_bugnotes_by_id[$p_bugnote_id]);
```
atrol, line 814: I suppose unset is not needed, as this is done later in `bugnote_clear_bug_cache`

Apart from that, this is a pretty confusing caching concept and a confusing function.

The Documentation tells

> Clear a bugnote from the cache

But it does not clear just the single specified bugnote in cache, but all bugnotes which belong to the same bug id like the specified one

Furthermore `bugnote_clear_cache()` and `bugnote_clear_bug_cache()`  if called without parameters do the same.

dregad: > I suppose unset is not needed, as this is done later in bugnote_clear_bug_cache

Actually it *is* needed and that is the whole point of this change. The reason is, if you call bugnote_clear_cache() on an issue for which the bug-level cache has not been set, then it will not be removed from the bugnote-level cache.

> Apart from that, this is a pretty confusing caching concept and a confusing function.

I agree, but didn't want to mess with the plumbing, which was introduced by @cproensa back in 2016, IIRC to address performance issues.

To be clean I guess we should not store bugnote data in the bug-level cache, only pointers to the bugnote-level cache.

> it does not clear just the single specified bugnote in cache, but all bugnotes which belong to the same bug id like the specified one

True, but this is required to guarantee the integrity of the bug-level cache. Without this, if we removed only a single bugnote from it, then subsequent calls to retrieve all bugnotes would return invalid data.

dregad: Without feedback, I assume you're fine with my response to your review comment.

PR #1696, opened 2020-09-05
Improve handling of missing language strings

dregad:
Fixes [#27241](https://mantisbt.org/bugs/view.php?id=27241)

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -4206,13 +4206,12 @@ $g_show_queries_count = OFF;
  * Recommended config_inc.php settings for developers (these are automatically
  * set if the server is localhost):
  * $g_display_errors = array(
- *     E_RECOVERABLE_ERROR => DISPLAY_ERROR_HALT,
  *     E_WARNING           => DISPLAY_ERROR_HALT,
  *     E_ALL               => DISPLAY_ERROR_INLINE,
  * );
  *
- * NOTICE: E_USER_ERROR, E_RECOVERABLE_ERROR and E_ERROR will always be internally
- * set DISPLAY_ERROR_HALT independent of value configured.
+ * NOTICE: E_USER_ERROR, E_RECOVERABLE_ERROR and E_ERROR will always be set
+ * to DISPLAY_ERROR_HALT internally, regardless of the configured value.
  *
  * @global array $g_display_errors
  */
@@ -4222,8 +4221,8 @@ $g_display_errors = array();
 # Note: intentionally not using SERVER_ADDR as it's not guaranteed to exist
 if( isset( $_SERVER['SERVER_NAME'] ) &&
 	( strcasecmp( $_SERVER['SERVER_NAME'], 'localhost' ) == 0 ||
-	  $_SERVER['SERVER_NAME'] == '127.0.0.1' ) ) {
-	$g_display_errors[E_USER_WARNING] = DISPLAY_ERROR_HALT;
+	  $_SERVER['SERVER_NAME'] == '127.0.0.1' )
+) {
 	$g_display_errors[E_WARNING] = DISPLAY_ERROR_HALT;
 	$g_display_errors[E_ALL] = DISPLAY_ERROR_INLINE;
 }
--- a/core/error_api.php
+++ b/core/error_api.php
@@ -117,20 +117,20 @@ function error_stack_trace( $p_exception = null ) {
 }
 
 /**
- * Default error handler
+ * Default error handler.
  *
  * This handler will not receive E_ERROR, E_PARSE, E_CORE_*, or E_COMPILE_*
- *  errors.
+ * errors.
  *
- * E_USER_* are triggered by us and will contain an error constant in $p_error
- * The others, being system errors, will come with a string in $p_error
- *
- * @access private
- * @param integer $p_type    Contains the level of the error raised, as an integer.
- * @param string  $p_error   Contains the error message, as a string.
- * @param string  $p_file    Contains the filename that the error was raised in, as a string.
- * @param integer $p_line    Contains the line number the error was raised at, as an integer.
+ * @internal
+ * @param integer    $p_type  Contains the level of the error raised, as an integer.
+ * @param int|string $p_error For Mantis internal errors (i.e. of type E_USER_*),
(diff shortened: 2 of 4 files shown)


core/error_api.php
```diff
@@ -183,6 +183,15 @@ function error_handler( $p_type, $p_error, $p_file, $p_line ) {
 		$t_method = DISPLAY_ERROR_HALT;
 	}
 
+	# Special handling for missing language strings, as we do not want to
+	# display information about lang_get() call; instead, we need details
+	# about the actual location of the missing string.
+	if( $p_error == ERROR_LANG_STRING_NOT_FOUND ) {
```
atrol, line 189: Does it work?
`$p_error` is the error message as a string, but `ERROR_LANG_STRING_NOT_FOUND` is the integer id of the message.

atrol:
I am fine with `Print lang string if English localization missing`

`Special handling for missing lang string errors` might not work (see my inline comment) and I am not sure if it's a good idea in general to have a special handling at this place.
It shouldn't be too hard for a developer to find the source code where `lang_get` is called, or to set `$g_show_detailed_errors = ON;` to see the call in stack trace.

core/error_api.php
dregad: > Does it work?

Yes

> $p_error is the error message as a string

This is what PHPDoc says for the parameter, but it is not complete; if you look just above, you'll see

```
 * E_USER_* are triggered by us and will contain an error constant in $p_error
 * The others, being system errors, will come with a string in $p_error
```

dregad:
> It shouldn't be too hard for a developer to find the source code where lang_get is called

True it's not very hard, but still extra effort to grep for the missing string, when we actually have this information and can easily provide it to them.

> set $g_show_detailed_errors = ON; to see the call in stack trace.

That does not work with our [recommended dev config for $g_display_errors](https://github.com/mantisbt/mantisbt/blob/release-2.24.2/config_defaults_inc.php#L4206), as there is no stack trace displayed when errors are shown inline. The developer would need to set 
`$g_display_errors[E_USER_WARNING] = DISPLAY_ERROR_HALT;`

core/error_api.php
atrol: I didn't work for me, so I thought at first sight that it's caused by the comparison.
The real root cause is the recursive call if you are using non-english settings.

atrol:
I thought I am using the recommended dev config, as I tried on localhost.
> The developer would need to set `$g_display_errors[E_USER_WARNING] = DISPLAY_ERROR_HALT;` 

When using localhost, [we set](https://github.com/mantisbt/mantisbt/blob/release-2.24.2/config_defaults_inc.php#L4226) `$g_display_errors[E_USER_WARNING] = DISPLAY_ERROR_HALT;`, so I got the trace.

Should we change what is documented as recommended dev config, or should we change the settings for localhost?

dregad:
> Should we change what is documented as recommended dev config, or should we change the settings for localhost

I never realized that we had a discrepancy between documentation and the actual defaulted localhost config. They should definitely be aligned, and they were when I originally introduced this feature (see #215). 

It appears that when @vboctor enforced halting of scripts on errors (as part of #1240, commit c816588d2616225f33896e7f4f4cd5a45447bce6), he changed the defaulted localhost behavior by adding `$g_display_errors[E_USER_WARNING] = DISPLAY_ERROR_HALT;`, not sure where that came from.

IMO, DISPLAY_ERROR_INLINE is better for our internal warnings in development mode, as it is less intrusive; I would suggest to align the localhost settings to the documented values.

core/error_api.php
atrol: @dregad not sure if my last note was clear.
You will not get the right file and line if
- your language is set to `auto` and you are running in a non-english environment
- your language is not set to `English`

dregad: > The real root cause is the recursive call if you are using non-english settings

Can you please describe those settings, and also exactly what you mean by "it didn't work", because I did test with another language and the system behaved as expected.

FTR, the test case is referring to a language string not defined in _strings_english.txt_ (as English is the fall-back language), and obviously also not in the user's current language (English or other).
```php
<?php
include 'core.php';
layout_page_header();
layout_page_begin();
echo lang_get('xxx');
layout_page_end();
```
Master (before)
image

This PR (after)
image

atrol: - In `bug_report_page.php` change `lang_get( 'enter_report_details_title' )` to `lang_get( 'x_enter_report_details_title' )`
- On `My Account` -> `Preferences` change language to `french`
- Click `Rapporter une anomalie`

Result: `Chaîne « x_enter_report_details_title » introuvable. (in '/Applications/MAMP/htdocs/atrol/mastertest/core/lang_api.php' line 304)`

With English settings you get: `String "x_enter_report_details_title" not found. (in '/Applications/MAMP/htdocs/atrol/mastertest/bug_report_page.php' line 253)`

dregad: Duh. I had it under my own eyes, and did not see it... :blush: Thanks

dregad:
I updated Admin Guide, PHPDoc for $g_display_errors and the default developers settings.

This should now be ready to merge.

config_defaults_inc.php
```diff
@@ -4206,13 +4206,12 @@
  * Recommended config_inc.php settings for developers (these are automatically
  * set if the server is localhost):
  * $g_display_errors = array(
- *     E_RECOVERABLE_ERROR => DISPLAY_ERROR_HALT,
```
atrol, line 4209: @dregad I suppose it's not intended to remove this line.

dregad: Yes it is actually. E_RECOVERABLE_ERROR Is forced to _halt_ by our error handler [[1]] so this is redundant 
[1]: https://github.com/dregad/mantisbt/blob/a212dc04c3027d3b0839ab3513a8427004e00cf3/core/error_api.php#L182

atrol: I am confused as you removed E_RECOVERABLE_ERROR from comment but added it in code for localhost.

dregad: Thanks for catching the mistake. I'll fix it.

PR #1700, opened 2020-09-18
Plugin update: validate Priority parameter

dregad:
Plugin Priority must be a number from 1 to 5. Trigger an error if the parameter's value is outside of that range.

Fixes [#27284](https://mantisbt.org/bugs/view.php?id=27284)


atrol:
@dregad you might consider to introduce two constants for the min/max values and use them also in `function print_plugin_priority_list`

dregad:
> consider to introduce two constants for the min/max values

Good idea, will do.

PR #1701, opened 2020-09-19
Don't assign tag to user not allowed to create one

dregad:
Fixes [#27283](https://mantisbt.org/bugs/view.php?id=27283)


atrol:
@dregad didn't have a deeper look.
Are we missing some checks at a lower level (`function tag_update`) ?

vboctor (approved):

PR #1702, opened 2020-09-19
Language strings cleanup

dregad:
There are several duplicated language strings defined for the same labels. We should use a single, common string for a given purpose.

Additionally, there are a few old, unused strings that need to be removed.

Fixes [#27298](https://mantisbt.org/bugs/view.php?id=27298)


bug_view_inc.php
```diff
@@ -682,7 +682,7 @@
 			<?php echo form_security_field( 'bug_monitor_add' ) ?>
 				<input type="hidden" name="bug_id" value="<?php echo (integer)$f_issue_id; ?>" />
 				<input type="text" class="input-sm" id="bug_monitor_list_user_to_add" name="user_to_add" />
-				<input type="submit" class="btn btn-primary btn-sm btn-white btn-round" value="<?php echo lang_get( 'add_user_to_monitor' ) ?>" />
+				<input type="submit" class="btn btn-primary btn-sm btn-white btn-round" value="<?php echo lang_get( 'Add' ) ?>" />
```
atrol, line 685: `add` instead of `Add`

bug_view_inc.php
```diff
@@ -1004,7 +1004,7 @@ function bug_view_relationship_view_box( $p_bug_id, $p_can_update ) {
 		<label class="inline"><?php echo lang_get( 'this_bug' ) ?>&#160;&#160;</label>
 		<?php print_relationship_list_box( config_get( 'default_bug_relationship' ) )?>
 		<input type="text" class="input-sm" name="dest_bug_id" value="" />
-		<input type="submit" class="btn btn-primary btn-sm btn-white btn-round" name="add_relationship" value="<?php echo lang_get( 'add_new_relationship_button' )?>" />
+		<input type="submit" class="btn btn-primary btn-sm btn-white btn-round" name="add_relationship" value="<?php echo lang_get( 'Add' )?>" />
```
atrol, line 1007: `add` instead of `Add`

bug_view_inc.php
dregad: Thanks, good catch

bug_view_inc.php
dregad: fixed

vboctor:
I like the cleanup, but worry about some of these strings may be used by plugins. Thoughts? Should we have a deprecated strings section for ones that are likely to be used - e.g. print, update, etc?

dregad:
> worry about some of these strings may be used by plugins. Thoughts? Should we have a deprecated strings section for ones that are likely to be used

@vboctor I can think of 3 possible approaches here

Nr | Description | Pros | Cons
:---:| --- | --- | ---
1 | Remove the strings without any specific handling (i.e. what I've done in this PR) | No extra work or overhead | May cause breakage or unwanted behavior, depending on _$g_display_errors_ settings - if 3rd-party plugins use a removed string, an error or warning message may be displayed, or nothing at all. **Note:** when PR #1696 is merged, the last case (no display) won't happen anymore, as the missing language string's key will be displayed instead of an empty string.
2 | Leave the strings in the language files, but move them down to an _Obsolete_ section and add a comment that they are deprecated and should not be used | No overhead, simple change| Strings remain in the system, possibly forever; plugins using them are not specifically informed of deprecation, and we can only hope but not enforce that new plugins actually don't use the obsolete strings
3 | Define a list of obsolete language strings and have Lang API trigger a deprecated warning when they are referenced | User-friendly solution | Adds complexity and overhead to the code : global variable to store obsolete strings, extra lookup in lang_get calls and trigger a deprecated warning which may be squelched anyway

I don't think option 2 is a good approach at all, and once #1696 gets merged, option 3 will only add marginal value (displaying the actual deprecated localization instead of the language string key + a warning). IMO it is not worth the effort to implement 3, and I recommend to stick with option 1.

dregad:
FTR I just merged #1696 after fixing the last review comment, so I rebased this branch on top of it.

dregad:
Without further feedback on this, I assume you agree with my proposed approach and will proceed with merging this PR.

PR #1704, opened 2020-09-19
Prevent assignment of categories to non-handler users

dregad:
Fixes [#27268](https://mantisbt.org/bugs/view.php?id=27268)

Also includes a minor fix, adding missing error parameter on EMPTY_FIELD error.

### DIFF
--- a/core/category_api.php
+++ b/core/category_api.php
@@ -173,6 +173,20 @@ function category_update( $p_category_id, $p_name, $p_assigned_to ) {
 	}
 
 	$t_old_category = category_get_row( $p_category_id );
+	$t_project_id = (int)$t_old_category['project_id'];
+
+	# Ensure target user exists and is allowed to handle bugs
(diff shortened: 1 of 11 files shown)


atrol:
@dregad why didn't you prevent assignment on a lower level (`function category_update`) ?

dregad:
> why didn't you prevent assignment on a lower level (`function category_update`) ?

Because at category API level I do not have any context for the update. Without the _project id_, it is not possible to perform the proper access-level check.

dregad:
Actually I take that back, I have it from the category row. 

I'll make the change as suggested, this way it will also be enforced for API calls.

vboctor (approved):
Make sure that you are handling the default handler for a global category correctly in terms of:

1. Are we checking global access level for the user?
2. Are we checking at assignment time based on the specific issue project? That would be a good check to have anyways, just in case a user's access level got reduced after they were assigned as a handler for a project category.

PR #1712, opened 2020-11-13
Fix EVENT_AUTH_USER_FLAGS is not triggered for non-logged in user

elsonwei:
- Set $t_user_id to NO_USER such that the plugins can be triggered.
- The plugins determine what to do if user is not in db and the name is blank.

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -86,17 +86,12 @@ function auth_flags( $p_user_id = null, $p_username = '' ) {
 		# If user id is not provided and user is not authenticated return default flags.
 		# Otherwise, we can get into a loop as in #22740
 		if( !auth_is_user_authenticated() ) {
-			return new AuthFlags();
+			$t_user_id = NO_USER;
+		} else {
+			$t_user_id = auth_get_current_user_id();
 		}
-
-		$t_user_id = auth_get_current_user_id();
 	} else {
 		$t_user_id = (int)$p_user_id;
-	}
-
-	if( !$t_user_id && is_blank( $p_username ) ) {
-		# If user is not in db, must supply the name.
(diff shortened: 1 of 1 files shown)


dregad:
Thanks for your contribution.

Please open an issue at https://mantisbt.org/bugs describing the problem, add a note linking to this PR and update your commit's message with a reference to the issue, i.e. `Fixes #xxxx`.

vboctor:
@elsonwei I'm not sure what you mean by non-logged in user. There is no such concept. It is either the user is logged in using the anonymous user or a regular user, or not logged in at all.

- Case 1: Logged in as regular user.
- Case 2: Logged in as anonymous user.
- Case 3: Not logged in.

Can you explain the current behavior in terms of the above, what you are trying to change and why?

starikov-aa:
I faced the same problem - https://www.mantisbt.org/bugs/view.php?id=27836

dregad:
@vboctor what's the status on this ?

vboctor:
The change introduces a change in behavior in the contract, though it will probably be ok.

starikov-aa:
After applying this patch, by event_auth_user_flags, the plugin starts to be called.
But due to the fact that the new user is missing in the database and in the cache, the login_password_page.php # L84 page will not redirect the plug-in authorization page. The problem is solving such a fix Login_password_page.php # L84: 
```php
$t_user_id = auth_get_user_id_from_login_name( $t_username );
$t_user_id_tmp = $t_user_id === false ? NO_USER : $t_user_id;
if(auth_credential_page( '', $t_user_id_tmp ) != AUTH_PAGE_CREDENTIAL ) {
```

dregad:
Closing in favor of #1792

PR #1714, opened 2020-12-05
Standardize on IEEE 1541 units (KiB, MiB) for file sizes

dregad:
Fixes [#27700](https://mantisbt.org/bugs/view.php?id=27700)


lang/strings_qqq.txt
```diff
@@ -494,3 +494,4 @@ $s_move = '{{Identical|Move}}';
 $s_close = '{{Identical|Close}}';
 $s_add = '{{Identical|Add}}';
 $s_login = '{{Identical|Login}}';
+$s_kib = "IEEE 1541 unit kibibytes (1024 bytes), see https://en.wikipedia.org/wiki/Kibibyte"
```
atrol, line 497: Introduces
`ERROR: Language strings should be single-quoted (line 497)`
when running admin/test_langs.php

dregad: Thanks, fixed.

PR #1719, opened 2020-12-24
Improve timezone setting logic

dregad:
Previously, core.php tried to set the default timezone to the value
stored in global configuration, with error suppression. Due to [[1]],
this does not actually prevent the notice from appearing with PHP 8.

Refactored the code to get the system's default time zone if it is not
defined in configuration. A PHP notice is still thrown if the specified
timezone is not valid, as documented in the code.

Also, since PHP 5.4, date_default_timezone_get() no longer throws a
warning so the '@' operator is not needed anymore.

Fixes [#27796](https://mantisbt.org/bugs/view.php?id=27796)

[1]: https://bugs.php.net/bug.php?id=80548

### DIFF
--- a/admin/install.php
+++ b/admin/install.php
@@ -1249,7 +1249,10 @@ if( 5 == $t_install_state ) {
 	$t_write_failed = true;
 
 	if( !$t_config_exists ) {
-		if( $t_fd = @fopen( $t_config_filename, 'w' ) ) {
+		# Try to create the config file
+		if( is_writable( $g_config_path )
+			&& $t_fd = fopen( $t_config_filename, 'w' )
+		) {
 			fwrite( $t_fd, $t_config );
 			fclose( $t_fd );
 		}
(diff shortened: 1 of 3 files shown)


dregad:
Added permanent fix for `@` operator not suppressing errors in PHP8 [#27799](https://mantisbt.org/bugs/view.php?id=27799).

Also included in this PR, some fixes in install.php :
- Fix error with the same root cause, when generating config_inc.php file
- Fix PHP warnings with null upgrade steps [#27800](https://mantisbt.org/bugs/view.php?id=27800)
- Code cleanup

atrol:
Looks good, but did just some cursory tests, e.g.
- tried old PHP 5.5
- removed write permissions from config folder
- tried print queries

When running again admin/install.php after first installation I get
```
SYSTEM NOTICE: 'date_default_timezone_set(): Timezone ID '1' is invalid' in '.../admin/install.php' line 361
```

dregad:
Thanks for testing. I can reproduce the problem.

atrol (approved):

PR #1721, opened 2020-12-28
Properly check access level when updating issues

dregad:
Commit bf1f02f71a85bd4e05cd2a627b3c25e79f96d5a0 (issues #21393, #22093 - PR #1254)
introduced a bug in the access level check, as it passed the threshold
as a string to access_ensure_bug_level() instead of its actual value.

This seemed to work in PHP < 8.0, because string comparisons behaved
differently [[1]], but in reality there was no actual access check, as
the strings were evaluated to 0 (ANYBODY).

Adding the missing config_get() call to fix the access level check.

Fixes [#27806](https://mantisbt.org/bugs/view.php?id=27806)

[1]: https://www.php.net/releases/8.0/en.php#consistent-type-errors-for-internal-functions

### DIFF
--- a/bug_update.php
+++ b/bug_update.php
@@ -154,24 +154,25 @@ $t_reporter_reopening =
 if ( !$t_reporter_reopening && !$t_reporter_closing ) {
 	switch( $f_update_type ) {
 		case BUG_UPDATE_TYPE_ASSIGN:
-			access_ensure_bug_level( 'update_bug_assign_threshold', $f_bug_id );
+			$t_threshold = 'update_bug_assign_threshold';
 			$t_check_readonly = true;
 			break;
 		case BUG_UPDATE_TYPE_CLOSE:
 		case BUG_UPDATE_TYPE_REOPEN:
-			access_ensure_bug_level( 'update_bug_status_threshold', $f_bug_id );
+			$t_threshold = 'update_bug_status_threshold';
 			$t_check_readonly = false;
 			break;
 		case BUG_UPDATE_TYPE_CHANGE_STATUS:
-			access_ensure_bug_level( 'update_bug_status_threshold', $f_bug_id );
+			$t_threshold = 'update_bug_status_threshold';
 			$t_check_readonly = true;
 			break;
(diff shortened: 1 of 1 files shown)


atrol:
@dregad the PR wants to merge to master branch / version 2.25.0. 
I suspect you want to merge to master-2.24 as you set target version to 2.24.4 on [#27806](https://mantisbt.org/bugs/view.php?id=27806)

atrol (approved):

bug_update.php
```diff
@@ -154,24 +154,25 @@
...
 			$t_check_readonly = false;
 			break;
 		case BUG_UPDATE_TYPE_CHANGE_STATUS:
-			access_ensure_bug_level( 'update_bug_status_threshold', $f_bug_id );
+			$t_threshold = 'update_bug_status_threshold';
 			$t_check_readonly = true;
 			break;
 		case BUG_UPDATE_TYPE_NORMAL:
 		default:
-			access_ensure_bug_level( 'update_bug_threshold', $f_bug_id );
```
vboctor, line 171: Provide the project id in the access check since current project may not be the same as the project associated with issue being edited.

vboctor (changes requested):

bug_update.php
atrol: @vboctor this should not be needed as we have a "nice" way some lines above to deal with it.
```
if( helper_get_current_project() != $t_existing_bug->project_id ) {
	$g_project_override = $t_existing_bug->project_id;
}
```

vboctor: Makes sense.

vboctor (approved):

dregad:
@atrol
> @dregad the PR wants to merge to master branch / version 2.25.0.
> I suspect you want to merge to master-2.24 as you set target version to 2.24.4 on [#27806](https://mantisbt.org/bugs/view.php?id=27806)

Correct, but my 2.24 branch is a mess right now as I'm reorganizing it to take review comments into consideration, which involves descoping some things; I'll cherry-pick the change

Thanks for the review and responding to @vboctor - spot on.

PR #1727, opened 2021-01-04
Review of LDAP code; added StartTLS support

seanm:
- fixed bug #15361: added StartTLS support for LDAP, based on illmnec's patch.
- added new ldap_tls_protocol_min option to specify minimun TLS version.
- changed default g_ldap_protocol_version from 0 to 3.
- changed default g_ldap_follow_referrals from on to off.
- improved some docs in Admin Guide and config_defaults_inc.php comments
- corrected log output for ldap_connect, which, despite its name, doesn't actually perform a network connection, according to its docs.


seanm:
@dregad here's a draft. I'm sure changes are needed...

core/ldap_api.php
```diff
@@ -180,6 +205,8 @@ function ldap_realname_from_username( $p_username ) {
  *
  * @param string $p_string The string to escape.
  * @return string The escaped string.
+ *
+ * @TODO: could we use built-in ldap_escape() instead of this custom code?
```
dregad, line 209: This code exists (since 2009) because ldap_escape() was only introduced in PHP 5.6. Since we still support 5.5.9, I would suggest to update the function with a PHP version check, and use the existing code as fallback if PHP < 5.6, and built-in ldap_escape() if not.

config_defaults_inc.php
```diff
@@ -2038,23 +2038,42 @@
...
 /**
- * Specifies the LDAP or Active Directory server to connect to.
+ * Specifies the LDAP or Active Directory server(s) to connect to.
+ * Multiple servers can be specified as a space-separated list.
  *
  * This must be a full LDAP URI (ldap[s]://hostname:port)
- * - Protocol can be either ldap or ldaps (for SSL encryption). If omitted,
- *   then an unencrypted connection will be established on port 389.
+ * - Protocol must be either: ldap (unencrypted or opportunistic StartTLS) or
+ *   ldaps (for TLS encryption).
```
dregad, line 2046: List the values as an indented, unordered list, e.g.
```
- Protocol must be one of: 
  - ldap (unencrypted or opportunistic StartTLS)
  - ldaps (for TLS encryption).
```

config_defaults_inc.php
```diff
@@ -2038,23 +2038,42 @@
...
  * - Port number is optional, and defaults to 389. If this doesn't work, try
  *   using one of the following standard port numbers: 636 (ldaps); for Active
  *   Directory Global Catalog forest-wide search, use 3268 (ldap) or 3269 (ldaps)
  *
  * Examples of valid URI:
- *   ldap.example.com
```
dregad, line 2051: If this is no longer valid, we need to consider what happens to all installations who are currently using this format - how will the system behave ?

docbook/Admin_Guide/en-US/config/auth.xml
```diff
@@ -91,10 +91,9 @@
 
 					<itemizedlist>
 						<listitem>
-							<para>Protocol can be either
-								<filename>ldap</filename> or
-								<filename>ldaps</filename> (for SSL/TLS encryption).
-								If omitted, then an unencrypted connection will be established on port 389.
+							<para>Protocol must be either:
+								<filename>ldap</filename> (unencrypted or opportunistic StartTLS) or
+								<filename>ldaps</filename> (for TLS encryption).
```
dregad, line 96: I don't think _filename_ tag is appropriate here - suggest `emphasis` instead. 
Suggest formatting as unordered list, as mentioned in my ealier comment on config_defaults_inc.php

dregad (changes requested):
For someone who does not know PHP, I think you did a pretty good job :smile: :+1: 

Kudos for updating the admin guide as well - most people are thrown off by the DocBook XML syntax and give up on that :wink: 

I unfortunately do not have an LDAP setup to actually test this, as I don't use MantisBT in corporate environment anymore. All my review comments are based on looking at the code only.

config_defaults_inc.php
seanm: I removed it only because the docs here https://www.php.net/manual/en/function.ldap-connect.php say "Note that hostname:port is not a supported LDAP URI as the schema is missing."

core/ldap_api.php
```diff
@@ -100,8 +100,33 @@ function ldap_connect_bind( $p_binddn = '', $p_password = '' ) {
...
+			return false;
+		}
+	}
+
+	$t_use_starttls = config_get_global( 'ldap_use_starttls' );
+	if ( $t_use_starttls ) {
+		log_event( LOG_LDAP, 'Attempting StartTLS' );
+		$t_result = @ldap_start_tls( $t_ds );
+		if( !$t_result ) {
+#			ldap_log_error( $t_ds );
```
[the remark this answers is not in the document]
seanm: Ah, was going to ask you about that... not sure how to do the error handling... different bits of code do different things...

config_defaults_inc.php
```diff
@@ -2038,23 +2038,42 @@
...
+ *
+ * @global bool $g_ldap_use_starttls
+ */
+$g_ldap_use_starttls = true;
+
+/**
+ * An integer indicating the minimum version of the TLS protocol to allow.
+ * For example, LDAP_OPT_X_TLS_PROTOCOL_TLS1_2. If 0, then the protocol version
+ * is not set. This maps to the LDAP_OPT_X_TLS_PROTOCOL_MIN ldap library option.
+ * Requires PHP 7.1 or later. For security, a failure aborts the entire connection.
```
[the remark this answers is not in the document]
seanm: Doesn't `OFF` imply a boolean choice?  (this isn't boolean.)

config_defaults_inc.php
```diff
@@ -2070,53 +2089,59 @@
...
- * by LDAP servers or not. This maps to LDAP_OPT_REFERRALS ldap library option.
+ * by LDAP servers or not. This maps to the LDAP_OPT_REFERRALS ldap library option.
  * For Active Directory, this should be set to OFF.
+ * If you have only one LDAP server, setting to this to OFF is advisible to prevent
+ * any man-in-the-middle attacks.
  *
  * @global integer $g_ldap_follow_referrals
  */
-$g_ldap_follow_referrals = ON;
+$g_ldap_follow_referrals = OFF;
```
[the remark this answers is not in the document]
seanm: Yeah, maybe it's not worth it. I'll switch it back to ON. At least the new comments explain that you probably want it off.

config_defaults_inc.php
dregad: Ah, I see what you mean.  In fact, we're relying on the alternate, deprecated signature
*ldap_connect ( string $host = null , int $port = 389 ) : resource|false*

So technically it still works, but we don't know for how long... I'm not sure whether PHP throws any warnings when this is used, so I guess we should inform admins that this is usage is deprecated. 

This could be done in the ldap_connect_bind() function and/or as an Admin check.

config_defaults_inc.php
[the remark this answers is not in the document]
dregad: It usually does, but not always; we sometimes use OFF for such purposes, check for examples e.g. $g_html_make_links

config_defaults_inc.php
```diff
@@ -2038,23 +2038,42 @@
...
  *   ldap://ldap.example.com
  *   ldaps://ldap.example.com:3269/
  *
  * @global string $g_ldap_server
  */
-$g_ldap_server = 'ldaps://ldap.example.com/';
+$g_ldap_server = 'ldap://ldap.example.com/';
+
+/**
+ * A boolean indicating if the connection will attempt opportunistic upgrade to
```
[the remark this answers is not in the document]
seanm: Fixed

config_defaults_inc.php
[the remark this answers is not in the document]
seanm: OK.

config_defaults_inc.php
```diff
@@ -2070,53 +2089,59 @@
...
+ * Defaults to 0 (infinite).
  *
  * @global int $g_ldap_network_timeout
  */
 $g_ldap_network_timeout = 0;
 
 /**
  * Determines whether the LDAP library automatically follows referrals returned
- * by LDAP servers or not. This maps to LDAP_OPT_REFERRALS ldap library option.
+ * by LDAP servers or not. This maps to the LDAP_OPT_REFERRALS ldap library option.
```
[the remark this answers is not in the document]
seanm: Fixed

config_defaults_inc.php
```diff
@@ -2070,53 +2089,59 @@
...
+ * For example, 'cn=ldap,ou=Administrators,dc=example,dc=com'.
  *
  * @global string $g_ldap_bind_dn
  */
 $g_ldap_bind_dn = '';
 
 /**
  * The password for the service account used to establish the connection to
- * the LDAP server.
+ * the LDAP server. For anonymous binding, leave empty.
```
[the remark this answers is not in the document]
seanm: Fixed.

config_defaults_inc.php
seanm: Fixed.

docbook/Admin_Guide/en-US/config/auth.xml
seanm: Fixed.

config_defaults_inc.php
```diff
@@ -2070,53 +2089,59 @@
 $g_ldap_organization = '';
 
 /**
- * The LDAP Protocol Version, if 0, then the protocol version is not set.
- * For Active Directory use version 3.
+ * The LDAP Protocol Version. If 0, then the protocol version is not set,
```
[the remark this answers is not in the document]
seanm: Fixed.

config_defaults_inc.php
seanm: This seems like a confusing PHP thing to me. Both signatures take string as first argument, and the 2nd param has a default value of 389, so does calling `ldap_connect('some string')` call the 1st or 2nd variant? Seem ambiguous.

Anyway, I could log a message from `ldap_connect_bind` if the string doesn't start with `ldap[s]://`...

core/ldap_api.php
seanm: OK, that sounds easy enough.

Now that I look a bit more, I also think the existing code is not escaping everything it should be... will audit that...

dregad: > the existing code is not escaping everything it should be... will audit that

Don't waste too much time on this, I'm planning to increase the minimum PHP requirement to 7.x, in 2.25.0 if I can, or in the next release if not ([#25956](https://mantisbt.org/bugs/view.php?id=25956)).

config_defaults_inc.php
dregad: > This seems like a confusing PHP thing to me.

Unfortunately, PHP is full of such inconsistencies. That may be one of the reasons why this usage is deprecated actually.

> log a message from `ldap_connect_bind` if the string doesn't start with `ldap[s]://`

Sounds good.

core/ldap_api.php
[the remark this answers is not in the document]
dregad: ldap_log_error() will just output the details of the last LDAP error to the log file; this is often useful for troubleshooting purposes, in addition to the log_event() calls that provide move high-level information about the LDAP process.

core/ldap_api.php
```diff
@@ -100,8 +100,33 @@ function ldap_connect_bind( $p_binddn = '', $p_password = '' ) {
 		ldap_log_error( $t_ds );
 	}
 
+	# Set minimum TLS protocol version flag (ex: LDAP_OPT_X_TLS_PROTOCOL_TLS1_2).
+	$t_tls_protocol_min = config_get_global( 'ldap_tls_protocol_min' );
+	if( version_compare( PHP_VERSION, '7.1.0', '>=' ) AND $t_tls_protocol_min > 0 ) {
```
dregad, line 105: By convention, we use `&&` rather than `AND`.

We can avoid an unnecessary config_get_global() call by doing 
```
if( version_compare( PHP_VERSION, '7.1.0', '>=' ) ) {
    $t_tls_protocol_min = config_get_global( 'ldap_tls_protocol_min' );
    if( $t_tls_protocol_min > 0 ) {
```

config_defaults_inc.php
seanm: Actually, `ldap_connect_bind` already checks that the string is a URI, and we already log if that fails.  So I've added an Admin Check instead.  (It's impressive what can be done by grepping and copy-pasting existing code :) )

core/ldap_api.php
seanm: OK done

core/ldap_api.php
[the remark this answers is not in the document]
seanm: OK, I've uncommented it then.

Should I also be calling `trigger_error`?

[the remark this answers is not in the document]
dregad: The following `return false` statement implies that failure to set this is an error, so if that is indeed the case yes you should.

config_defaults_inc.php
dregad: Indeed :wink:

docbook/Admin_Guide/en-US/config/auth.xml
```diff
@@ -171,7 +200,7 @@ ldaps://ldap.example.com:3269/
 					automatically follows referrals returned by LDAP
 					servers or not. This maps to LDAP_OPT_REFERRALS ldap
 					library option. Defaults to
-					<emphasis>ON</emphasis>.</para>
+					<emphasis>OFF</emphasis>.</para>
```
atrol, line 203: The final decision was to keep it as it was before (`ON`)

config_defaults_inc.php
```diff
@@ -2038,60 +2038,89 @@
...
+ * The LDAP Protocol Version.
+ * If 0, then the protocol version is not set, and you get whatever default
+ * the underlying ldap library uses. This maps to the LDAP_OPT_PROTOCOL_VERSION
+ * ldap library option. In almost all cases you should use 3. LDAPv3 was
+ * introduced back in 1997. LDAPv2 was deprecated in 2003 by RFC3494.
  *
  * @global integer $g_ldap_protocol_version
  */
-$g_ldap_protocol_version = 0;
+$g_ldap_protocol_version = 3;
```
atrol, line 2103: Changing the default might introduce regressions.

@dregad you insisted in not changing the default for `ldap_follow_referrals`.
Shouldn't we do the same for `ldap_protocol_version`?

dregad: I think it makes sense to switch to 3, considering the issues we've had with accented characters in the past, e.g. 
- https://mantisbt.org/bugs/view.php?id=20809
- https://mantisbt.org/bugs/view.php?id=22655
- https://mantisbt.org/forums/viewtopic.php?f=3&t=26735

admin/check/check_config_inc.php
```diff
@@ -117,5 +117,15 @@
 	array( false => 'Use "limit_view_unless_threshold" instead.' )
 );
 
+# Check that 'ldap_server' is a proper URI, starting with either ldap:// or ldaps://
+$t_ldap_server = config_get_global( 'ldap_server' );
+$t_starts_with1 = substr( $t_ldap_server, 0, strlen( 'ldap://' ) ) === 'ldap://';
+$t_starts_with2 = substr( $t_ldap_server, 0, strlen( 'ldaps://' ) ) === 'ldaps://';
+check_print_test_row(
+	'"ldap_server" must not be a plain host name, it must be a proper URI',
```
dregad, line 125: Reword to `"ldap_server" must be a valid, full LDAP URI`

admin/check/check_config_inc.php
```diff
@@ -117,5 +117,15 @@
...
+	$t_starts_with1 || $t_starts_with2,
```
dregad, line 126: Probably better to use a regex to test both cases in one go
```
preg_match( '~^ldaps?://~', $t_ldap_server ) == 1
```

docbook/Admin_Guide/en-US/config/auth.xml
seanm: Yeah, I'll leave it alone (still `ON`). I added a comment saying it might be best to set to `OFF`.

config_defaults_inc.php
seanm: And v3 is 24 years old!  Surely vastly more people are using v3 than v2.  Defaulting to v2 is just one more thing new users have to find and change before their LDAP works.

admin/check/check_config_inc.php
seanm: OK

admin/check/check_config_inc.php
seanm: OK

core/ldap_api.php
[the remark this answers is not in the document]
seanm: Yes, I was thinking best to fail hard. If the users want min TLSv1.2, and we can't achieve that, it seems best to fail.

I'll add another error code then.

config_defaults_inc.php
```diff
@@ -2038,60 +2038,89 @@
...
 /**
- * The root distinguished name for LDAP searches
+ * Decides if the connection will attempt an opportunistic upgrade to
+ * a TLS connection (STARTTLS).
+ * Be sure to use the ldap:// scheme and not ldaps://
+ * if setting this to true.  For security, a failure aborts the entire connection.
+ *
+ * @global bool $g_ldap_use_starttls
+ */
+$g_ldap_use_starttls = true;
```
vboctor, line 2068: `ON` instead of `true`.

config_defaults_inc.php
```diff
@@ -2038,60 +2038,89 @@
...
  */
-$g_ldap_server = 'ldaps://ldap.example.com/';
+$g_ldap_server = 'ldap://ldap.example.com/';
 
 /**
- * The root distinguished name for LDAP searches
+ * Decides if the connection will attempt an opportunistic upgrade to
+ * a TLS connection (STARTTLS).
+ * Be sure to use the ldap:// scheme and not ldaps://
+ * if setting this to true.  For security, a failure aborts the entire connection.
```
vboctor, line 2064: You should reference ON here too.

dregad (commented):
This is looking pretty good now. 

I think we need to create a few issues in the tracker, to reference and document significant changes introduced in the PR and make sure they appear on the change log.

Will look into this later.

config_defaults_inc.php
seanm: OK

config_defaults_inc.php
seanm: OK

config_defaults_inc.php
```diff
@@ -2038,60 +2038,89 @@
...
+$g_ldap_server = 'ldap://ldap.example.com/';
 
 /**
- * The root distinguished name for LDAP searches
+ * Decides if the connection will attempt an opportunistic upgrade to
+ * a TLS connection (STARTTLS).
+ * Be sure to use the ldap:// scheme and not ldaps://
+ * if setting this to ON.  For security, a failure aborts the entire connection.
+ *
+ * @global bool $g_ldap_use_starttls
```
atrol, line 2066: After latest changes, change to `@global integer $g_ldap_use_starttls`

docbook/Admin_Guide/en-US/config/auth.xml
```diff
@@ -109,13 +114,36 @@
...
 			</varlistentry>
 
+			<varlistentry>
+				<term>$g_ldap_use_starttls</term>
+
+				<listitem>
+					<para>A boolean indicating if the connection will attempt opportunistic
+					upgrade to a TLS connection (STARTTLS).  Be sure to use the ldap:// scheme
+					and not ldaps:// if setting this to ON.  For security, a failure aborts
+					the entire connection. Defaults to <emphasis>true</emphasis>.</para>
```
atrol, line 130: After latest changes, change to `Defaults to ON`

dregad (changes requested):
Please update the following lines in your commit message:

> - fixed bug #15361: added StartTLS support for LDAP, based on illmnec's patch.

Change to `- fixes #15361: added StartTLS support for LDAP, based on illmnec's patch.`
This allows the Source integration plugin to automatically pick up the commit and link it to the issue

> - changed default g_ldap_protocol_version from 0 to 3.

Append `(fixes #27848)`

> - added an Admin Check to ensure that ldap_server config option is in URI form.

Append `(fixes #27849)`

config_defaults_inc.php
seanm: Why should it be an integer?  A use / don't use choice is by definition a boolean.

atrol: Booleans can be either true or false. https://www.php.net/manual/en/language.types.boolean.php
You set it to constant `ON` which is defined as `1`.

seanm: I set it to `ON` because you told me to :)

As I said, I don't know PHP, I figured ON/OFF where synonymous with true/false.

Why did you suggest using ON instead of true?  Why is integer better than boolean here?

atrol: > I set it to `ON` because you told me to :)

Wasn't me, was @vboctor

> Why is integer better than boolean here?

Not sure what @vboctor intended when requesting the change.

Maybe because you could add more options later on without that much friction.
Maybe because we use integers most of the time for such kind of options (e.g. `$g_enable_email_notification`)

Furthermore integers are easier to handle than booleans at various places, e.g, when persisting them in different database systems and/or different languags.

Of course, we are not complete clean, as there is `$g_reassign_on_feedback` which is currently set to `ON` but also documented as boolean.
And there is `$g_session_save_path` which is currently set to `false`,  documented as boolean, but can also be set to a string.

> As I said, I don't know PHP

so you are certainly confused by the dynamic type system of PHP (still the same for me, even after some years of PHP)

seanm: >Wasn't me, was @vboctor

Sorry, I meant the plural 'you' (darn English).

OK, I defer to 'yous' on this...

docbook/Admin_Guide/en-US/config/auth.xml
seanm: Done

core/ldap_api.php
seanm: OK, I've just removed this TODO for now...

seanm:
BTW I haven't actually tested any of this yet.  Once everyone has reviewed and is happy, I'll actually try it.

dregad (approved):
Looks good to me.
Can't actually test though, as I don't have an LDAP setup.

atrol (approved):
Also fine from my side, but also no environment / time to test.

dregad:
@seanm sorry about the merge conflict, it is caused by me changing the formatting of the $g_global_settings config (see a0291ffd19099e912116544726836237206b8982) - should not be too hard to fix though :wink:

dregad:
And I think you can go ahead and test, @vboctor's review comments should have no impact on the functionality.

seanm:
So I did a git clone on my test server but mantis wouldn't run, I eventually figured out I needed to copy the `vendor` folder from a regular mantis tarball for things to work.

Then, embarrassingly, StartTLS seems to be borked on my server, so now I have to investigate that...

dregad:
> I eventually figured out I needed to copy the vendor folder from a regular mantis tarball for things to work

For the record, the "proper" way to do this, is to run `composer install` after cloning
https://mantisbt.org/docs/master/en-US/Developers_Guide/html-desktop/#dev.contrib.clone.clone

dregad:
@seanm do have any estimate, by when you'll be able to complete your testing ?

I'm not asking to put any kind of pressure on you... I'd just like to know if I can keep this in scope for 2.25, which I'm hoping to release by end of January, or if you'd be more comfortable to move it to next release.

seanm:
I found why TLS was failing on my server: it was using a self-signed cert.

Tomorrow I'll try to finish testing.

seanm:
So I've done testing, and it seems fine to me.  I succeeded with `ldaps://` and with `ldap://` (with and without StartTLS).

On thing to think about: I defaulted the new `g_ldap_use_starttls` to `ON`.  This means that anyone using `ldap://` with a server that doesn't support StartTLS will get a failure.  But it's 2021, such a setup probably deserves an error message thrown in your face. :)

dregad:
> defaulted the new `g_ldap_use_starttls` to `ON`. This means that anyone using `ldap://` with a server that doesn't support StartTLS will get a failure

It seems reasonable to default to the more secure setting, I'm fine with that.
But now that you mention it, we may want to make this more obvious in the PHPDoc / Admin Guide, what do you think ?

seanm:
Not sure how/where to make it more obvious...

dregad:
Merged in 94462f8c88b04370b1c1bc59bb303cd6f4dad163. 

> Not sure how/where to make it more obvious...

I took care of that in 2e952249b06f1277a978bf6c35da58c070bffde9.

Many thanks @seanm for your contribution, and sorry it took me so long to perform the final review and merge it.

seanm:
Thank you for helping me get this in, and all the php help to do so!

PR #1730, opened 2021-01-17
Tag attach group action should trigger event

vboctor:
Use command to attach tags which triggers events.

Fixes #27881


dregad (approved):
LGTM

PR #1731, opened 2021-01-17
Create a Cronjob script and plugin event

vboctor:
This way there is one cronjob script to execute which triggers an event, then all plugins can hook into it.

Fixes #27882

### DIFF
--- a/scripts/cronjob.php
+++ b/scripts/cronjob.php
@@ -17,7 +17,11 @@
 # See the README and LICENSE files for details
 
 /**
- * Cron Script to trigger cronjob work for plugins
+ * Cron script to allow scheduled execution of plugin tasks.
+ *
+ * @package scripts
+ * @copyright Copyright 2021  MantisBT Team - mantisbt-dev@lists.sourceforge.net
+ * @link https://mantisbt.org
  */
 
 /**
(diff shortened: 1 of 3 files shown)


scripts/cronjob.php
```diff
@@ -0,0 +1,38 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+# See the README and LICENSE files for details
+
+/**
+ * Cron Script to trigger cronjob work for plugins
+ */
```
dregad, line 21: _Cron script to trigger cronjob_ is somewhat redundant, suggest replacing by
```
Cron script to allow scheduled execution of plugin tasks.
```

Add 
```
 *
 * @package scripts
 * @copyright Copyright 2021  MantisBT Team - mantisbt-dev@lists.sourceforge.net
 * @link https://mantisbt.org
```

scripts/cronjob.php
```diff
@@ -0,0 +1,38 @@
...
+ * Global Bypass http headers
+ */
+global $g_bypass_headers;
+$g_bypass_headers = 1;
+
+require_once( dirname( dirname( __FILE__ ) ) . '/core.php' );
+
+# Make sure this script doesn't run via the webserver
+if( php_sapi_name() != 'cli' ) {
+	echo "cronjob.php is not allowed to run through the webserver.\n";
```
dregad, line 33:
```
echo basename( __FILE__ ) . " is not [...]
```

dregad (changes requested):

vboctor:
@dregad comments addressed.

dregad (approved):

PR #1732, opened 2021-01-17
DbQuery: push/pop param count before execution

dregad:
This is for backwards compatibility with legacy code still relying on
DB API functions. Without this, parameter count gets reset after query
execution, which causes issues on PostgreSQL.

Fixes [#27113](https://mantisbt.org/bugs/view.php?id=27113)

### DIFF
--- a/core/classes/DbQuery.class.php
+++ b/core/classes/DbQuery.class.php
@@ -241,6 +241,12 @@ class DbQuery {
 	 * @return IteratorAggregate|boolean ADOdb result set or false if the query failed.
 	 */
 	public function execute( array $p_bind_array = null, $p_limit = null, $p_offset = null ) {
+		# For backwards compatibility with legacy code still relying on DB API,
+		# we need to save the parameters count before binding otherwise it will
+		# be reset after query execution, which will cause issues on RDBMS with
+		# numbered params (e.g. PostgreSQL).
+		db_param_push();
+
 		# bind values if provided
 		if( null !== $p_bind_array ) {
 			$this->bind_values( $p_bind_array );
@@ -251,7 +257,9 @@ class DbQuery {
 		$this->process_bind_params();
 		$this->process_sql_syntax();
 
-		return $this->db_execute( $p_limit, $p_offset );
+		$t_result = $this->db_execute($p_limit, $p_offset);
+		db_param_pop();
+		return $t_result;
 	}
 
 	/**


dregad:
Replaces PR #1686

core/classes/DbQuery.class.php
```diff
@@ -241,6 +241,12 @@ public function set_offset( $p_offset = -1 ) {
 	 * @return IteratorAggregate|boolean ADOdb result set or false if the query failed.
 	 */
 	public function execute( array $p_bind_array = null, $p_limit = null, $p_offset = null ) {
+		# For backwards compatibility with legacy code still relying on DB API,
+		# we need to save the parameters count before binding otherwise it will
+		# be reset after query execution, which will cause issues on RDBMS with
+		# numbered params (e.g. PostgreSQL).
+		db_param_push();
```
atrol, line 248: `$g_db_param->push()` is used in `public static function compat_db_query` (same for the `pop` functions).
Not sure which one is the better way, but we should do it the same way in both functions.

Didn't have a deeper look, maybe it's possible to have just a single place where we push/pop, e.g.. in `protected function db_execute`

dregad: > $g_db_param->push() is used [...] instead of db_param_push()
> Not sure which one is the better way, but we should do it the same way in both functions.

They are the same thing. API functions were created to provide abstraction and avoid having to declare the global variable first, but I'm fine with switching to object form if you prefer.

> $g_db_param->push() is used in public static function compat_db_query 

I didn't notice that one - searched for `db_param_push`, didn't think of checking the `->push` variant. 

> maybe it's possible to have just a single place where we push/pop, e.g.. in protected function db_execute

- the push/pop in compat_db_query is here to ensure compatibility with legacy db_query() calls
- the one I added, covers the situation (bug) where we have a legacy query being built, and in the middle of it some other query is executed using the new DbQuery (e.g. through MantisBT core, e.g. User Pref or Project API).

I'm not sure we can have them in a single place, or even if it makes sense.

The problem is with the ADOdb object's *_pnum* parameter, which keeps track of the parameter number and gets increased when binding variables (i.e. when calling $g_db->Param()). That actually happens in DbQuery::process_bind_params(), called from DbQuery::execute() before DbQuery::db_execute(). So that would de facto exclude moving the logic to the latter. 

I'll double-check. Maybe @cproensa can add his 2 cents as well.

dregad:
Without further feedback, I'll proceed with merging.

PR #1733, opened 2021-01-17
Some config options can be set in database, but should be configurable just in config_inc.php

atrol:
There are some configuration options which should just be changed by the Web administrator, but not the MantisBT administrator.
Furthermore these options should not be configurable per project or user, but just for the whole instance.

To prevent it, the options should be listed in `global_settings`.
For better performance, the options should be queried by `config_get_global` instead of `config_get`.

Issue [#27884](https://mantisbt.org/bugs/view.php?id=27884)


vboctor (changes requested):
Some features should be configurable in the database, for example:

- Window Title
- Show Avatar
- Show Avator Threshold
- Sub-projects Enabled

dregad (approved):
Thanks for cleaning this up.

atrol:
> Window Title
 
Thought more about it, I will remove the change.
Not sure if we still want it per project and/or user, but that's not in the scope of this PR.
 
> Show Avatar
 
I thought that similar to `cdn_enable` it should be the Web Admin who decides if content is delivered just by the Mantis server (intranet solution) or also by some external 3rd party server.
The Mantis Admin should not be able to override this decision.
I would like to keep this change, as you are still able to deactivate avatars in database by using option `show_avatar_threshold`.

As a side note, you can set `show_avatar` per project at the moment.
This is confusing when switching between projects with different settings.

> Show Avator Threshold

I didn't touch `show_avatar_threshold`, so you can use it to enable/disable avatars in database if the Web Admin has `show_avatar` set to `ON`.

> Sub-projects Enabled
 
I thought my change ensures what was intended with this option.
The commit message when implementing`subprojects_enabled` was https://github.com/mantisbt/mantisbt/commit/83cfb1da54d0d499889c238c969912b9f1391779

> This will keep the feature enabled by default.
> However, it provides the ability to disable this feature so it won't be used by managers and administrators.

Without my change, managers and administrators can still use this feature by setting the option in database.

As a side note, you can set `subprojects_enabled` per project at the moment.
Not sure if this is wanted / needed.
 
> Some features should be configurable in the database, for example:
> ...
 
@vboctor are there more configuration options where you don't agree with my changes?
Unless you object, I will merge after undoing the changes for `window_title`.
I will also undo the changes for `subprojects_enabled` if you still want it.

atrol:
Ping @vboctor will merge with changes later today as announced in previous note.

vboctor:
Avatars can have multiple providers. Hence, you shouldn't assume that avatars == gravatar. Hence, I would keep this configurable via database. 

There is no need to have the ability to configure avatars by project. Up to you if you make any changes for this. 

As for sub projects enabled.  My point with this config option is to avoid them using the sub projects feature and not disallowing to enable it if desired. The config option also allowed optimizing some queries when option is disabled. 

Note that moving some of these options from database to global only will cause change of behavior for instances that used the database. So unless it makes no sense to have in db at all, we should err on keeping existing behavior.

atrol:
As announced I removed the changes for `window_title` and also for `search_title` as the default setting is `%window_title%`

> Avatars can have multiple providers. Hence, you shouldn't assume that avatars == gravatar. Hence, I would keep this configurable via database.

Agreed, changed

> There is no need to have the ability to configure avatars by project. Up to you if you make any changes for this.

Will not change as part of this PR. There are some more options where configuration by project is questionable.

> As for sub projects enabled.

@vboctor I didn't remove the change for `subprojects_enabled` until now, as I don't understand what you mean with the following sentence

> My point with this config option is to avoid them using the sub projects feature

How can you avoid using the sub projects feature, if you allow to activate it in database?
What do you mean with `them`? The Mantis admins?

atrol:
@vboctor without some more feedback to the `subprojects_enabled` option I will merge the current state of the PR later today.

I have no strong opinion on that, but would like to understand the advantage of using config_get.
So feel free to revert the commits related to this option later on.

As a side note, I am  not sure if it makes sense to go on having have such an option as a per user/project option in database.
E.g. see latest regression that is related to this option https://mantisbt.org/bugs/view.php?id=27902

PR #1736, opened 2021-02-12
REST API PATCH /issue/:id now fails with HTTP 400

dregad:
Previously, PHP warnings would be triggered when the payload is empty.

Fixes [#27973](https://www.mantisbt.org/bugs/view.php?id=27973)

### DIFF
--- a/api/rest/restcore/issues_rest.php
+++ b/api/rest/restcore/issues_rest.php
@@ -156,6 +156,9 @@ function rest_issue_get( \Slim\Http\Request $p_request, \Slim\Http\Response $p_r
  */
 function rest_issue_add( \Slim\Http\Request $p_request, \Slim\Http\Response $p_response, array $p_args ) {
 	$t_issue = $p_request->getParsedBody();
+	if( !$t_issue ) {
+		return $p_response->withStatus( HTTP_STATUS_BAD_REQUEST, "Invalid request body or format");
+	}
 
 	if( isset( $t_issue['files'] ) ) {
 		$t_issue['files'] = files_base64_to_temp( $t_issue['files'] );
--- a/api/rest/restcore/projects_rest.php
+++ b/api/rest/restcore/projects_rest.php
@@ -248,8 +248,8 @@ function rest_project_hierarchy_delete( \Slim\Http\Request $p_request, \Slim\Htt
  */
 function rest_project_add( \Slim\Http\Request $p_request, \Slim\Http\Response $p_response, array $p_args ) {
 	$t_payload = $p_request->getParsedBody();
-	if( $t_payload === null ) {
-		return $p_response->withStatus( HTTP_STATUS_BAD_REQUEST, "Unable to parse body, specify content type" );
+	if( !$t_payload ) {
+		return $p_response->withStatus( HTTP_STATUS_BAD_REQUEST, "Invalid request body or format");
 	}
 
 	$t_project_id = mc_project_add( /* username */ '', /* password */ '', (object) $t_payload );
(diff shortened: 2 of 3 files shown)


api/rest/restcore/issues_rest.php
```diff
@@ -368,6 +368,9 @@ function rest_issue_update( \Slim\Http\Request $p_request, \Slim\Http\Response $
 
 	# Construct full issue from issue from db + patched info
 	$t_issue_patch = $p_request->getParsedBody();
+	if( !$t_issue_patch ) {
+		return $p_response->withStatus( HTTP_STATUS_BAD_REQUEST, "Empty Body");
```
vboctor, line 372: There are several reasons why the getParsedBody() would fail:

1. Content-Type header is not passed, rather than passed to indicate json.
2. The body is set to '{}'
3. The body is set to ''
4. The body is set to invalid json '{'

So empty body may not be the right error to return here, but `Invalid request body or format`.

vboctor (changes requested):

api/rest/restcore/issues_rest.php
dregad: Fixed

vboctor (approved):
REST API looks good.  Do you want to change the string in SOAP API? I haven't looked too much into this code path, so not sure.

dregad:
> Do you want to change the string in SOAP API? I haven't looked too much into this code path, so not sure.

I looked into this some more following your remark, and realized that the fix I proposed for mc_issue_update() was targeted at the tree that was hiding the forest... In fact the root cause of the error I got was incorrect checking of `project` key in IssueData, which obviously could also be triggered with an non-empty payload.

After fixing the root cause, it turns out my original fix may not really be necessary at all, as mc_issue_update() throws a _Mandatory field 'summary' is missing_ error when given an empty payload, which is good enough to inform the user that their input is not correct.

Now that is somewhat outside of this PR's scope, but I'm wondering about the reason for this error. I may intend to update another field, so why should I have to provide _summary_ and _description_ ? I don't see a reason for them to be required in this context. What do you think ?

vboctor:
Here is my comment that I added to the issue you created:

"I haven't checked the code again, but as far as I remember the SOAP API expected the full payload of an issue to update it. This is not ideal, and hence, I designed the REST API to allow patching by supplying only the fields to be overwritten. So I would say this is by design and I wouldn't invest in changing this in SOAP API."

I'm not sure why you would need any changes in `mc_issue_update()`. I would expect this PR to not touch this code since the REST API merges the incremental payload passed by the client with the full definition of the issue before calling the `mc_issue_update()` method.

As an FYI, there are multiple calls to getParsedBody(), multiple of these don't check for null result. if you want to full fix, we can add validation to the other places and using a consistent error messages across all.

dregad:
> I'm not sure why you would need any changes in mc_issue_update(). I would expect this PR to not touch this code since the REST API merges the incremental payload passed by the client with the full definition of the issue before calling the mc_issue_update() method.

When I started working on this, I got problems in mc_issue_update() so I included a fix for it in this PR. 

Technically, the SOAP API fix should probably be handled separately now (that's why I opened #27981), but out of laziness I thought I'd keep this as a single PR. If you want I can split things up.

> there are multiple calls to getParsedBody(), multiple of these don't check for null result. if you want to full fix, we can add validation to the other places and using a consistent error messages across all.

Thanks for the info. I did not look beyond the problem I was facing, but it makes sense to deal with it globally.

dregad:
Moved SOAP API fix to separate PR #1740. Will force-push branch to remove the commit it from here.

The updated branch will also contain fix for additional occurrences of PHP warning caused by empty payload.

PR #1737, opened 2021-02-14
Reset user session cookie string upon logout

dregad:
When a user logs out from Mantis, we clear their session cookie string
(i.e. set mantis_user_table.cookie_string column to an empty string).
This ensures that anyone knowing its value is no longer able to login
with it.

On login, after successfully authenticating the user, when setting
the cookies in auth_set_cookies() we check if the cookie_string is
defined in the DB, and if not a new hash is generated and stored.

While not a complete fix for issue [#11296](https://mantisbt.org/bugs/view.php?id=11296), this does improve the
situation by providing an easy and logical means for users to
effectively invalidate all their previous sessions.

Additionally, using an empty value to indicate an invalidated cookie
string instead of directly generating a new hash makes it easy to:
- identify user records which should be considered as logged out
  (e.g. last_visit older than $g_cookie_time_length)
- invalidate login cookies (set them to '')
Leveraging this is left for future improvements.

Note: an empty string in the session cookie always triggers an anonymous
login (or sends the user back to login page if anonymous login is
disabled).

Fixes [#27976](https://mantisbt.org/bugs/view.php?id=27976)

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -935,17 +935,7 @@ function auth_generate_unique_cookie_string() {
  * @throws ClientException
  */
 function auth_is_cookie_string_unique( $p_cookie_string ) {
-	db_param_push();
-	$t_query = 'SELECT COUNT(*) FROM {user} WHERE cookie_string=' . db_param();
-	$t_result = db_query( $t_query, array( $p_cookie_string ) );
-
-	$t_count = db_result( $t_result );
-
-	if( $t_count > 0 ) {
-		return false;
-	} else {
-		return true;
-	}
+	return false === user_get_id_by_cookie( $p_cookie_string );
 }
 
 /**
@@ -1108,17 +1098,7 @@ function auth_is_cookie_valid( $p_cookie_string ) {
 	}
 
 	# look up cookie in the database to see if it is valid
-	db_param_push();
-	$t_query = 'SELECT * FROM {user} WHERE cookie_string=' . db_param();
-	$t_result = db_query( $t_query, array( $p_cookie_string ) );
-
-	# return true if a matching cookie was found
-	if( 1 == db_num_rows( $t_result ) ) {
-		user_cache_database_result( db_fetch_array( $t_result ) );
-		return true;
-	} else {
-		return false;
-	}
+	return false !== user_get_id_by_cookie( $p_cookie_string );
 }
 
 /**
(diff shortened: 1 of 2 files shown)


vboctor (changes requested):
I would go with an option that is consistent with resetting the cookie when password is changed.  See `user_set_password()`.

Basically, in the logout API, generate a new cookie.  This makes it consistent with reset password and have the change in one place.

dregad:
> in the logout API, generate a new cookie

That was my initial approach actually, but I changed it to set the cookie to `''` instead. 

The reason for this is explained in the commit message:

```
Additionally, using an empty value to indicate an invalidated cookie
string instead of directly generating a new hash makes it easy to:
- identify user records which should be considered as logged out
  (e.g. last_visit older than $g_cookie_time_length)
- invalidate login cookies (set them to '')
Leveraging this is left for future improvements.
```

vboctor:
We have 3 cases where we have a cookie that is never used:

1. The cookie for administrator user after setup.
2. Any user after change password.
3. With this change, any user after logout.

In both 1 and 2, we generate a cookie and place it in the db to be used for next login. It keeps our code consistent and we don't have a worry about a case where we make a code error and validate an empty browser cookie against an empty cookie value in db. Hence, I think case 3 should follow the same model as 1 and 2.

vboctor:
I have updated my comment above to include the user creation/signup case.

dregad:
So how would you propose to identify and possibly invalidate outdated cookies then (short of implementing something like the community-proposed solution https://www.mantisbt.org/bugs/view.php?id=11296#c65103 which is no small undertaking) ?

As for
> worry about a case where we make a code error and validate an empty browser cookie against an empty cookie value in db

Unless I missed something, this just can't happen as auth_is_cookie_valid() always returns _false_ if cookie string is empty (see https://github.com/mantisbt/mantisbt/blob/release-2.24.4/core/authentication_api.php#L1045)

dregad:
@vboctor waiting for your reply on https://github.com/mantisbt/mantisbt/pull/1737#issuecomment-779333627

vboctor:
If we are to do any short term change, then as I suggested in my message, treat logout like change of password, where we generate a new cookie value and update it in the database. This effectively logs the user out of all sessions they have across browsers and machines. Remove the code that sets cookie to empty string and handles such case later.

dregad:
I really don't understand why you're taking such a strong stance on this; from my perspective, the consistency argument is a bit light in comparison to the benefits of my proposed solution. 

But fine, I'll change it per your demand.

And by the way, you have not answered the question.
> how would you propose to identify and possibly invalidate outdated cookies

PR #1738, opened 2021-02-15
Make `mantis_email_table.email` same size as `mantis_user_table.email`

vboctor:
Fixes #27982


dregad (approved):
Missed that one when I applied the change on user table. Thanks.

PR #1744, opened 2021-03-01
Version field is now required

dregad:
When adding a new version from manage_proj_edit_page.php via the
'Add Version' or 'Add and Edit Version', it should not be allowed to
submit an empty field.

The HTML required attribute has been added to the 'version' input, and
the manage_proj_ver_add.php page displays an error message if no version
was created.

Fixes [#27994](https://mantisbt.org/bugs/view.php?id=27994)

### DIFF
--- a/manage_proj_edit_page.php
+++ b/manage_proj_edit_page.php
@@ -559,7 +559,7 @@ if ( config_get_global( 'subprojects_enabled') == ON ) {
 		<fieldset>
 			<?php echo form_security_field( 'manage_proj_ver_add' ) ?>
 			<input type="hidden" name="project_id" value="<?php echo $f_project_id ?>" />
-			<input type="text" class="input-sm" name="version" size="32" maxlength="64" />
+			<input type="text" class="input-sm" required name="version" size="32" maxlength="64" />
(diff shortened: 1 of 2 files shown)


manage_proj_edit_page.php
```diff
@@ -559,7 +559,7 @@
 		<fieldset>
 			<?php echo form_security_field( 'manage_proj_ver_add' ) ?>
 			<input type="hidden" name="project_id" value="<?php echo $f_project_id ?>" />
-			<input type="text" class="input-sm" name="version" size="32" maxlength="64" />
+			<input type="text" class="input-sm" required name="version" size="32" maxlength="64" />
```
atrol, line 562: While you are on it, it would be good to change also the category field some lines above.

Not sure if we should add
`<span class="required">*</span>`
like we have it for the project name and the user name on the same page.

dregad: > While you are on it, it would be good to change also the category field some lines above.

Right, I meant to add the required property here too initially, then I forgot... Will do.
Note that no change is necessary in the "add" script, as the empty string case was already dealt with.

> Not sure if we should add
> `<span class="required">*</span>`
> like we have it for the project name and the user name on the same page.

I would vote against that - it's not a "regular" form, and I believe it should be quite obvious that the field is mandatory, without this additional visual hint, which looks a bit awkward IMHO

image

dregad:
Did some code cleanup too.

PR #1746, opened 2021-03-26
Fix invalid HTML + various improvements in manage_user_proj_delete.php

dregad:
See details including bugtracker references in the commit messages.


atrol:
Clicking  "Remove" on manage_user_edit_page if no project is selected throws 
APPLICATION ERROR # 200
A required parameter to this page (project_id) was not found.

dregad:
As expected...

PR #1748, opened 2021-04-16
Add resolved percentage on Roadmap section title

dregad:
On the Roadmap page, a version's progress is now displayed on the right
of the section's title, with a tooltip showing the number of issues
resolved vs planned.

The resolved_progress language string has been modified to remove the
3rd placeholder (percentage) which is redundant.

Script has been refactored to use a new internal RoadmapProgress class,
simplifying code and removing some duplication.

Fixes [#28182](https://mantisbt.org/bugs/view.php?id=28182)

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -1107,7 +1107,7 @@ function auth_is_cookie_valid( $p_cookie_string ) {
  * @return integer user id
  * @access public
  *
- * @throws ClientException
+ * @noinspection PhpDocMissingThrowsInspection
  */
 function auth_get_current_user_id() {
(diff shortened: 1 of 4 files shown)


dregad:
### Before
image

### After
image

atrol:
> Script has been refactored to use a new internal RoadmapProgress class, simplifying code and removing some duplication.

There is similar code in changelog_page.php
Maybe you can consider this while you are on refactoring.

dregad:
> There is similar code in changelog_page.php
> Maybe you can consider this while you are on refactoring.

I am aware of that, but I'll save this for another day.

PR #1757, opened 2021-06-15
PHP 8: Fix "Bad Request" error on custom field filters

dregad:
Fixes [#28803](https://mantisbt.org/bugs/view.php?id=28803)

### DIFF
--- a/return_dynamic_filters.php
+++ b/return_dynamic_filters.php
@@ -37,6 +37,8 @@
  * @uses helper_api.php
  */
 
+use Mantis\Exceptions\StateException;
+
 # Prevent output of HTML in the content if errors occur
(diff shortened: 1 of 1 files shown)


virtualdreams:
With PHP 7.4 under Debian 11 this problem with custom fields occurs again. Therefore I cannot use the latest version 2.25.2.

dregad:
@virtualdreams I don't get your point. 

This pull request fixed an issue caused by use of @ operator on PHP 8. As far as I know, there was never a problem with PHP 7.4 but you seem to say there is. 

Please open an issue in our bugtracker and provide all required details including steps to reproduce so we can analyze the problem.

virtualdreams:
@dregad Done. :)

PR #1758, opened 2021-06-16
Update MantisMarkdown.php

dirkeinecke:
When < and " are output, the HTML entities are output up to now. This change corrects this.


dregad:
Thanks for your contribution. I have actually tried this before, and I wish the solution were as simple as that, but unfortunately this approach does not fully fix the problem.

HTML entities are blindly converted back to the corresponding character regardless of whether they were part of the original raw string, or added by MantisCoreFormatting plugin. For example, if the raw string is `<script> &lt;`, it will be output as `<script> <`, note that the `&lt;` is unexpectedly converted to `<` (for the record, current Mantis code prints `&lt;script> &lt;`)

I guess one might argue which one is the lesser evil...

PR #1761, opened 2021-06-29
Set overflow-wrap to 'anywhere' for tbody only (#1690)

MaloMn:
This new css rule applies the solution proposed by @syncguru in a more selective way.
The main problem with the proposed solution in #1690 was its application to the whole table.
As css tables are divided in two main parts (header -thead-, and body -tbody-), applying the rule only to tbody does not change the header behavior.

### DIFF
--- a/css/ace-mantis.css
+++ b/css/ace-mantis.css
@@ -171,6 +171,10 @@ table {
     background-color: #ffffff !important;
 }
 
+table tbody {
+	overflow-wrap: anywhere;
+}
+
 /* Disable printing of link urls */
 a:link:after, a:visited:after {
     content: normal !important;
--- a/css/ace-mantis.css
+++ b/css/ace-mantis.css
@@ -171,8 +171,12 @@ table {
     background-color: #ffffff !important;
 }
 
-table tbody {
-	overflow-wrap: anywhere;
+td {
+    overflow-wrap: anywhere;
+}
+
+td a {
+    overflow-wrap: normal;
 }
 
 /* Disable printing of link urls */
--- a/css/ace-mantis.css
+++ b/css/ace-mantis.css
@@ -171,12 +171,8 @@ table {
     background-color: #ffffff !important;
 }
 
-td {
-    overflow-wrap: anywhere;
-}
-
-td a {
-    overflow-wrap: normal;
+[class*="description"] {
+	overflow-wrap: anywhere;
 }
 
 /* Disable printing of link urls */
--- a/css/ace-mantis.css
+++ b/css/ace-mantis.css
@@ -171,8 +171,21 @@ table {
     background-color: #ffffff !important;
 }
 
-[class*="description"] {
-	overflow-wrap: anywhere;
+/* Breaking lines when needed */
+[class*="description"],
+[class*="steps-to-reproduce"],
+[class*="additional-information"],
+[class*="bugnote"],
+[class*="custom-field"] {
+    overflow-wrap: anywhere;
+}
+
+pre {
+    white-space: pre-wrap;       /* Since CSS 2.1 */
+    white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
+    white-space: -pre-wrap;      /* Opera 4-6 */
+    white-space: -o-pre-wrap;    /* Opera 7 */
+    word-wrap: break-word;       /* Internet Explorer 5.5+ */
 }
 
 /* Disable printing of link urls */


MaloMn:
The main problem with the proposed solution in #1690 was its application to the whole table.
As css tables are divided in two main parts (header -thead-, and body -tbody-), applying the rule only to tbody does not change the header behavior.

dregad (approved):
@atrol this looks good to me, and I don't see the behavior you reported in https://github.com/mantisbt/mantisbt/pull/1690#issuecomment-674407921.

I'll wait for your feedback before merging.

atrol:
@dregad this one looks better, no unwanted side effects on the table headers.
So the only disadvantage is that it's still not supported on Safari.

atrol (approved):

dregad:
> So the only disadvantage is that it's still not supported on Safari.

Support for overflow-wrap is tracked in Webkit https://bugs.webkit.org/show_bug.cgi?id=195345 but that does not seem to be going anywhere fast.

Is there any negative side-effect on this browser, or is it just ignoring the CSS attribute ?

MaloMn:
Thanks for your reviews.
I've actually run into cases where the table display gets worse : it's on the bug_update_page (see capture below).
The issue with this table is that some "headers" cells are within the table. To solve this, changing the rule from ~`table tbody`~ to `td` seems to work, as header cells are *th* cells.

One downside of this is the second capture I've included. In `view_all_bug_page`, the number of notes is displayed on two rows, which is not really nice. A solution to this would be to disable the CSS rule on links contained in td cells.

I'll commit the modifications I've made so that you can check them. These new CSS rules are stricter than the previous ones, so it should not cause any more issue.

atrol:
> Is there any negative side-effect on this browser, or is it just ignoring the CSS attribute ?

@dregad I don't see any side effect. Safari seems just to the attribute, but I can't test latest Safari at the moment.

@MaloMn the original issue is no longer fixed after your latest changes.
E.g. try the contents of the "Steps To Reproduce" field from 
https://mantisbt.org/bugs/view.php?id=27114

MaloMn:
After playing around with this css rule, here are (according to me) the three available options :
1. Keep changes proposed initially, bearing in mind that it causes [issues about bugnotes-count column](https://github.com/mantisbt/mantisbt/pull/1761#issuecomment-880427339) (and I've also noticed recently it also happens sometimes with the category-column : the last "n" in Installation goes on a new line)
2. Restict this css rule by using `:not(.class-to-restrict)`. This has the massive drawbacks of having to add manually fields where the css rule breaks the display.
3. Manually select the elements where we want the lines to be broken. As the initial issue was raised on the description field, I've come up with the css rule below that also applies to column-description on the *view_all_bug_page*. However, as this is manual, it might miss areas where this feature might be needed.

```css
[class*="description"] {
    overflow-wrap: anywhere;
}
```

I'm personally not a fan of the first and second options, as there are quite a lot of places where this rule worsens the display, and targeting them one by one is quite a lot of manual labour.
The third is still not ideal in my view, because it targets too few elements. But at least it solves the initial issue and also what my team has witnessed with descriptions on *view_all_bug_page*. This is why we will go with this third option with our team, but I'm not sure if it should be included in the official version of Mantis. I hope I've given you enough information for you to judge on that.

atrol:
All text area fields should behave the same (Description, Steps To Reproduce, Additional Information, Notes, Custom Fields)

MaloMn:
Following your suggestion @atrol, I added class selectors  on the following fields:
```css
[class*="description"],
[class*="steps-to-reproduce"],
[class*="additional-information"],
[class*="bugnote"],
[class*="custom-field"] {
     overflow-wrap: anywhere;
}
```

However, once that was handled, I noticed another issue, also related to long unbreakable lines. It has nothing to do woth text in general : it only concerns text contained in `<pre>` tag. To solve this issue, this solution proposed here worked a treat for me!

```css
pre {
    white-space: pre-wrap;       /* Since CSS 2.1 */
    white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
    white-space: -pre-wrap;      /* Opera 4-6 */
    white-space: -o-pre-wrap;    /* Opera 7 */
    word-wrap: break-word;       /* Internet Explorer 5.5+ */
}
```

One question though: where should I put these pieces of code? It seems the current location is not right to me.

dregad:
@MaloMn sorry for only coming back to this now. 

> the current location is not right to me

I'm not sure what you mean by that. Can you please clarify what you are asking ?

dregad:
@MaloMn you never answered my question... What is the status of this PR ?

dregad:
@MaloMn any chance you could answer https://github.com/mantisbt/mantisbt/pull/1761#issuecomment-955477859 so this PR can be merged ?

MaloMn:
Hello @dregad, sorry for the late reply.

About the location of the piece of code, I meant the location in the CSS file. If it seems right to you, then I think it can be merged.
Otherwise, is there a better place in the css file where this piece of code might fit better?

atrol:
@MaloMn @dregad as mentioned before https://github.com/mantisbt/mantisbt/pull/1761#issuecomment-882823898
The original issue is no longer fixed after the latest changes.
E.g. try the contents of the "Steps To Reproduce" field from
https://mantisbt.org/bugs/view.php?id=27114

PR #1762, opened 2021-06-30
FIX: rest api: call to undefined methods when handling RestFault

marc-dll:
When calling `ApiObjectFactory::throwIfFault()` to check whether an API call returned a fault (`RestFault` or `SoapFault` depending of the API type), if that is indeed the case, a `LegacyApiFaultException` is created by getting the message and code from the fault object.

This is done by calling `getCode()` and `getMessage()` on the fault. Yet, `RestFault` implements none of those methods (contrary to `SoapFault`, which extends PHP's standard `Exception` class).

This causes a `Slim Application Error` page in HTML instead of the exected JSON error response.

### DIFF
--- a/api/soap/mc_api.php
+++ b/api/soap/mc_api.php
@@ -58,6 +58,24 @@ class RestFault {
 		$this->status_code = $p_status_code;
 		$this->fault_string = $p_fault_string === null ? '' : $p_fault_string;
 	}
+
+    /**
+     * Error description getter
+     *
+     * @return string The error description
+     */
+	function getMessage() {
+		return $this->fault_string;
+	}
+
+    /**
+     * Http status code getter
+     *
+     * @return integer The http status code
+     */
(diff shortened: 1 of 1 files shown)


dregad:
@marc-dll could you please file an issue at https://mantisbt.org ?

Also I'm wondering if _RestFault_ should not extend Exception, like _SoapFault_ does... @vboctor you introduced this class, what do you think ?

vboctor:
We never throw instances of this class. Hence, it didn't inherit from Exception. It is just a class who instances are returned from functions.

marc-dll:
@dregard : Done => https://mantisbt.org/bugs/view.php?id=28927

As far as i know, `RestFault` and `SoapFault` use their parameters differently from Exception (the `code` is an HTTP status and is the second parameter)

dregad (approved):

PR #1767, opened 2021-08-04
Attachment to new tab

MaloMn:
I think the behavior when viewing attachment should change: when clicking on an image in a bugnote, it displays it in the same tab. I think it would be better to redirect this full display of the image to a new tab. That would allow you to go back and forth between seeing the image and seeing the bugnote text associated.

This new redirection would be effective for both clicking on the icon or on the name of the attachment (as shown below), and also when clicking on an image preview in a bugnote, or a preview of an audio / video.

image

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2066,6 +2066,13 @@ $g_max_dropdown_length = 40;
  */
 $g_wrap_in_preformatted_text = ON;
 
+/**
+ * Controls whether attachment links redirect users t a new tab or not
+ * If turned off, a click on a link will not redirect to a new tab
+ * @global integer $g_attachment_to_new_tab
+ */
+$g_attachment_to_new_tab = OFF;
+
 #############################################
 # MantisBT Authentication and LDAP Settings #
 #############################################
@@ -4557,6 +4564,7 @@ $g_public_config_names = array(
 	'antispam_max_event_count',
 	'antispam_time_window_in_seconds',
 	'assign_sponsored_bugs_threshold',
+	'attachment_to_new_tab',
 	'auto_set_status_to_assigned',
 	'backward_year_count',
 	'bottom_include_page',
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -1971,7 +1971,7 @@ function print_bug_attachment( array $p_attachment, $p_security_token ) {
 function print_bug_attachment_header( array $p_attachment, $p_security_token ) {
 	if( $p_attachment['exists'] ) {
 		if( $p_attachment['can_download'] ) {
-			echo '<a href="' . string_attribute( $p_attachment['download_url'] ) . '">';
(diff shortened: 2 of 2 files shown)


atrol:
Worth to read in this context https://mantisbt.org/forums/viewtopic.php?f=2&t=25561

MaloMn:
Thanks for the forum link.

From what I understand:
* It's something that should be an option, and not something permanent,
* It's unclear whether it should behave the same for pdf / audio / images / text files.

I don't think setting an option for each file type is a good idea (I agree with your argument that it would confuse users), and I also think that having an option for all file types is a good thing. I don't think it woud be right to open pdf files in a new tab, and not images. In both cases, you have to use the browser "Go back" button if it does not open in a new tab.
I also don't think that the question about whether the browser can handle displaying the attachment is valid, because whether it is in a new tab or not, the browser will behave the same. The only major change in my view is if it can't display the attachment : it will open a new tab and close it immediately, before opening the dialog box to open the attachent in another software.

I'll reply in this forum with a link to this Pull Request.

Following up from what I've said, I added a constant in config_default_inc.php, and a function in print_api.php that returns the `target` tag, filleed with either *_self* or *_blank*. I'm not sure whether it's the right way to do it, but it works on my local Mantis installation.

core/print_api.php
```diff
@@ -2125,6 +2125,17 @@ function get_filesize_info( $p_size, $p_unit ) {
 	return sprintf( lang_get( 'max_file_size_info' ), number_format( $p_size ), $p_unit );
 }
 
+/**
+ * Returns target tag to be added in attachment links
```
dregad, line 2129: *target* is not a tag, it's an attribute

core/print_api.php
```diff
@@ -2125,6 +2125,17 @@ function get_filesize_info( $p_size, $p_unit ) {
...
+ * @return string
+ */
+function get_attachment_link_target() {
```
dregad, line 2132: By convention, we usually prefix functions with the API's name, so this should be called *print_attachment_link_target()*

config_defaults_inc.php
```diff
@@ -2066,6 +2066,13 @@
  */
 $g_wrap_in_preformatted_text = ON;
 
+/**
+ * Controls whether attachment links redirect users t a new tab or not
+ * If turned off, a click on a link will not redirect to a new tab
+ * @global integer $g_attachment_to_new_tab
+ */
+$g_attachment_to_new_tab = OFF;
+
```
dregad, line 2075: I believe it would make more sense to move this to the *Bug Attachments Settings* section

dregad (changes requested):

dregad:
@MaloMn as you have not responded to the review, I'm going to close this PR without merge. Feel free to reopen after making the required changes.

dregad:
Replaced by #1994

PR #1772, opened 2021-08-07
Refactored admin/test_langs.php

dregad:
This is a nearly complete rewrite of this page, which now has 

- a nice "2.x" GUI displaying errors, similar to the that of the Installer
- Output is split in 2 distinct sections: Core strings and plugins. 
- Errors and warnings are displayed as bulleted lists, with red or yellow background depending on severity.

### Before
image
image
image

### After
image
image
image


admin/test_langs.php
```diff
@@ -100,7 +100,7 @@
  * @return void
  */
 function checkplugins( $p_path ) {
-	$t_path = rtrim( $p_path, DIRECTORY_SEPARATOR ) . DIRECTORY_SEPARATOR;
+	$t_path = rtrim( $p_path, DIRECTORY_SEPARATOR ) . '/';
```
vboctor: You still have directory separator used. The trim and concat should use the same construct.

admin/test_langs.php
```diff
@@ -152,7 +152,7 @@ function get_lang_files( $p_path ) {
  * @return void
  */
 function checklangdir( $p_path ) {
-	$t_path = rtrim( $p_path, DIRECTORY_SEPARATOR ) . DIRECTORY_SEPARATOR . 'lang' . DIRECTORY_SEPARATOR;
+	$t_path = rtrim( $p_path, DIRECTORY_SEPARATOR ) . '/lang/';
```
vboctor, line 561: You still have directory separator used. The trim and concat should use the same construct.

vboctor (approved):

admin/test_langs.php
dregad: > You still have directory separator used.

That's intentional and required in this case, because $p_path is coming from FileSystemIterator so may contain `/` or `\` depending on the underlying OS.

> The trim and concat should use the same construct.

This is actually not necessary, as PHP internal functions correctly interpret `/` regardless of OS. So not using DIRECTORY_SEPARATOR saves keystrokes, reduces the need for concatenation (e.g. `'/lang/'` vs `DIRECTORY_SEPARATOR . 'lang' . DIRECTORY_SEPARATOR`, and ultimately makes the code easier to read.

admin/test_langs.php
dregad: ditto.

PR #1773, opened 2021-08-08
Include cache key when including unversioned css

vboctor:
When including our css files that don't have a version in the name,
include cache key based on Mantis version.

Fixes #26148

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -208,11 +208,16 @@ function require_css( $p_stylesheet_path ) {
  */
 function html_css() {
 	global $g_stylesheets_included;
-	html_css_link( config_get_global( 'css_include_file' ) );
+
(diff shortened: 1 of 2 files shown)


vboctor:
I pushed to PR branch to MantisBT repository by mistake, so will leave it there and cleanup after merge.

vboctor:
A similar change is needed for inclusion of unversioned js files. I'm planning to do it once this PR is merged.

dregad (changes requested):
I believe it would be better to move the cache key generation and concatenation logic to the html_css_link() function, and drive it with a bool $p_use_cache_key parameter (defaulted to false).

vboctor:
@dregad I pushed the refactored version.

PR #1774, opened 2021-08-08
Show issue attachments with issue header

vboctor:
- Issue attachments displayed with issue header information.
- Issue notes attachments displayed with notes.

Fixes #[28965](https://mantisbt.org/bugs/view.php?id=28965)


bug_view_inc.php
```diff
@@ -588,6 +590,39 @@
...
+
+		foreach( $t_activity['attachments'] as $t_attachment ) {
+			$t_attachments[(int)$t_attachment['id']] = $t_attachment;
+		}
+
+		ksort( $t_attachments, SORT_NUMERIC );
+		foreach( $t_attachments as $t_attachment ) {
+			print_bug_attachment( $t_attachment, $t_security_token_attachments_delete );
+		}
+	}
```
dregad, line 621: This yields
> SYSTEM WARNING: 'Undefined array key "attachments"' in '/home/dregad/dev/mantisbt/bug_view_inc.php' line 613
> SYSTEM WARNING: 'foreach() argument must be of type array|object, null given' in '/home/dregad/dev/mantisbt/bug_view_inc.php' line 613

Unless I'm mistaken, only activities of type _note_ have an `attachments` key, and these are ignored by the loop (line 604), so you could get rid of the offending inner foreach entirely.

The ksort and subsequent foreach do not make any sense, because each _attachment_ type activity can only have a single attachment.

So the loop boils down to this:
```php
	foreach( $t_activities as $t_activity ) {
		if( $t_activity['type'] === ENTRY_TYPE_ATTACHMENT ) {
			print_bug_attachment( $t_activity['attachment'], $t_security_token_attachments_delete );
		}
	}
```

lang/strings_english.txt
```diff
@@ -1314,6 +1314,7 @@ $s_advanced_filters = 'Advanced Filters';
 $s_simple_filters = 'Simple Filters';
 $s_monitored_by = 'Monitored By';
 $s_attachments = 'attachment(s)';
+$s_files = 'File(s)';
```
dregad, line 1317: Why a new string, and not reuse the existing $s_attachments ?

bugnote_view_inc.php
```diff
@@ -137,14 +141,13 @@
...
 
-	# Tokens for action buttons are created only once, if needed
-	$t_security_token_state = null;
-	$t_security_token_notes_delete = null;
-	$t_security_token_attachments_delete = null;
-
 	for( $i=0; $i < $t_activities_count; $i++ ) {
 		$t_activity = $t_activities[$i];
 
+		if( $t_activity['type'] === ENTRY_TYPE_ATTACHMENT ) {
```
dregad, line 147: I know we currently only have 2 activity types at the moment, but shouldn't this be `!== ENTRY_TYPE_NOTE` ?

bugnote_view_inc.php
```diff
@@ -137,14 +141,13 @@
 	$t_normal_date_format = config_get( 'normal_date_format' );
 	$t_total_time = 0;
 
-	# Tokens for action buttons are created only once, if needed
-	$t_security_token_state = null;
-	$t_security_token_notes_delete = null;
-	$t_security_token_attachments_delete = null;
-
 	for( $i=0; $i < $t_activities_count; $i++ ) {
 		$t_activity = $t_activities[$i];
```
dregad, line 145: Use a foreach loop here
```php
foreach( $t_activites as $t_activity ) {
```

bugnote_view_inc.php
```diff
@@ -329,21 +332,17 @@
...
 		}
 
 		if( isset( $t_activity['attachments'] ) && count( $t_activity['attachments'] ) > 0 ) {
-			if ( !$t_security_token_attachments_delete ) {
-				$t_security_token_attachments_delete = form_security_token( 'bug_file_delete' );
-			}
-
-			foreach( $t_activity['attachments'] as $t_attachment ) {
-				print_bug_attachment( $t_attachment, $t_security_token_attachments_delete );
+			if( $t_activity['type'] === ENTRY_TYPE_NOTE ) {
```
dregad, line 338: Remove the if - condition is always true, since we exclude other activity types at the beginning of the main loop (line 147)

bugnote_view_inc.php
```diff
@@ -329,21 +332,17 @@
...
-		} else {
-			if ( !$t_security_token_attachments_delete ) {
-				$t_security_token_attachments_delete = form_security_token( 'bug_file_delete' );
-			}
-
-			print_bug_attachment( $t_activity['attachment'], $t_security_token_attachments_delete );
 		}
 
 		if( isset( $t_activity['attachments'] ) && count( $t_activity['attachments'] ) > 0 ) {
-			if ( !$t_security_token_attachments_delete ) {
```
dregad, line 341: Activities of type _note_ always have an `attachments` key, so the isset() is not necessary

dregad (changes requested):

lang/strings_english.txt
vboctor: We can't use $s_attachments due to casing. It is used in cases that expect lower case. We will need a new string which can be `$s_files = 'File(s)'` or `$s_attachments_capitalized = `Attachment(s)` depends on the string we want.

bugnote_view_inc.php
vboctor: I have fixed the code, so it always has attachments array and not a mix of attachment and attachments. This was there because of a bug in `bug_activity_api.php`.

vboctor:
@dregad I applied review comments and fixed a bug in bug activity API that caused complexity / bug. I didn't change the language string yet, until we decide.

dregad (changes requested):
> fixed a bug in bug activity API that caused complexity / bug

That does not fix the problem. I still get

SYSTEM WARNING: 'Undefined array key "attachments"' in '/home/dregad/dev/mantisbt/bug_view_inc.php' line 608
SYSTEM WARNING: 'foreach() argument must be of type array|object, null given' in '/home/dregad/dev/mantisbt/bug_view_inc.php' line 608

This occurs in a legacy issue (i.e. created before 2.23.0) having 2 attachments, the 2nd added (without a bugnote), after the initial submission (bugnote_id = 0, date_added > bug.date_submitted + 3s).

I do not understand the reason for applying the fix in bug_activity_combine(), would it not make more sense, not to create the `attachment` entry in the first place in bug_activity_get_all() ?

dregad:
> I didn't change the language string yet, until we decide.

I would suggest to use the existing `$s_attached_files = 'Attached Files';`, which is used in similar context (print_all_bug_page_word.php). This string is duplicated (`$s_attached_files = 'Attached Files'`, used in email_api.php).

vboctor:
@dregad 

1. Re-used existing string `attached_files` rather than introducing `files`.
2. I tested with legacy and new issues and applied a fix.
3. I also did the simplification you requested by always settings attachments array in `bug_activity_get_all()`.

PR #1776, opened 2021-08-16
Fix gpc_set_cookie ignoring HttpOnly parameter

aaribaud:
Function `gpc_set_cookie` ignores parameter `$p_httponly` and sets all cookies to be HttpOnly, which prevents legitimate use cases where Javascript code would need to change the cookie value.


dregad (changes requested):
@aaribaud thanks for your contribution.

Please open an issue in our bugtracker, and reference it in your commit message with `Fixes #xxxx`

dregad (approved):

dregad:
Thanks

PR #1779, opened 2021-09-01
Fixed PHP 7.2 warning on calling count() on null value

fbruneau:
This should fix bug https://mantisbt.org/bugs/view.php?id=29034


dregad:
I removed unnecessary `!= null` and improved the commit message.

Thanks for your contribution @fbruneau !

fbruneau:
Perfect thank you !

PR #1780, opened 2021-09-02
Fix failure to move attachments when there are too many attachments

chozekun:
This issue is caused by a lack of memory because the query results are too large.
It is important not to fail even if execute more queries.

### DIFF
--- a/admin/move_attachments.php
+++ b/admin/move_attachments.php
@@ -157,19 +157,25 @@ function move_attachments_to_disk( $p_type, array $p_projects ) {
 	# Build the SQL query based on attachment type
 	switch( $p_type ) {
 		case 'project':
-			$t_query = 'SELECT f.*
+			$t_query = 'SELECT f.id
 				FROM {project_file} f
 				WHERE content <> \'\'
 				  AND f.project_id = ' . db_param() . '
 				ORDER BY f.filename';
+			$t_query2 = 'SELECT *
+				FROM {project_file}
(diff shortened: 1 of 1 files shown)


admin/move_attachments.php
```diff
@@ -193,55 +199,58 @@ function move_attachments_to_disk( $p_type, array $p_projects ) {
...
-							$t_update_query,
-							array( $t_upload_path, $t_row['id'] )
-						);
-
-						if( !$t_update_result ) {
-							$t_status = 'Database update failed';
-							$t_failures++;
+			while( $t_row2 = db_fetch_array( $t_result ) ) {
+				$t_result2 = db_query( $t_query2, array( $t_row2['id'] ) );
+				while( $t_row = db_fetch_array( $t_result2 ) ) {
```
dregad, line 204: Variable naming is confusing - you're using $t_row to process the results of $t_query2/$t_result2 and $t_row2 for $t_query/$t_result. I suggest to either refactor to use more meaningful names, or at the very least switch usage of $t_row/$t_row2 to match the corresponding query.

dregad (commented):
Thanks for your contribution.

Kindly open an issue in our tracker at https://mantisbt.org/bugs, providing a bit of context about what the issue is and how to reproduce it, then reference the ID in the commit message.

dregad:
@chozekun since you did not respond to the review comment, I'm closing this PR. Feel free to reopen it after providing the requested feedback.

PR #1788, opened 2021-12-24
Custom field id lookup by name now case-insensitive

dregad:
The $g_cache_name_to_id_map now uses lowercase custom field names as key
which prevents APPLICATION ERROR 1300 Custom field not found when using
a case-sensitive database.

Fixes [#29413](https://mantisbt.org/bugs/view.php?id=29413)

### DIFF
--- a/core/custom_field_api.php
+++ b/core/custom_field_api.php
@@ -169,8 +169,9 @@ function custom_field_cache_array_rows( array $p_cf_id_array = null ) {
 	$t_ids_not_found = $c_cf_id_array;
 	while( $t_row = db_fetch_array( $t_result ) ) {
 		$c_id = (int)$t_row['id'];
+		$c_name = mb_strtolower($t_row['name']);
 		$g_cache_custom_field[$c_id] = $t_row;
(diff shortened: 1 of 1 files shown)


core/custom_field_api.php
```diff
@@ -788,24 +792,19 @@ function custom_field_get_id_from_name( $p_field_name ) {
...
+				$t_name = mb_strtolower($t_row['name']);
+				$g_cache_name_to_id_map[$t_name] = $t_row['id'];
+			}
+		}
 	}
 
-	$g_cache_name_to_id_map[$p_field_name] = $t_row['id'];
-
-	return $t_row['id'];
+	return $g_cache_name_to_id_map[$p_field_name] ?? false;
```
atrol, line 807: You want to merge to master-2.25 which is the last version where we support PHP 5.
The null coalescing operator ?? is supported starting from PHP 7.0. https://www.php.net/manual/en/migration70.new-features.php

dregad: Thanks for the review. You're absolutely right, good catch ! I'll fix that.

PR #1791, opened 2022-01-25
Removes WikkaWiki from Wiki Integration.

vendeeglobe:
WikkaWiki development ceased and is no longer maintained.

Fixes [#29497](https://mantisbt.org/bugs/view.php?id=29497)


dregad:
I have no idea how widely the WikkaWiki integration is used today, but if we just remove it like that, it will be a problem for those who rely on it. Just because the wiki is no longer developed, does not mean that everyone has migrated away from it.

So I'd rather start by marking it as a deprecated feature first, and actually remove it in a later version.

In any case, please open an issue at https://mantisbt.org/bugs and reference it in the commit message (e.g. Fixes #xxx)

vboctor:
Closing this request, since there is not much upside to removing this from MantisBT and the PR has been inactive for a while.

PR #1805, opened 2022-05-04
Process bugnote stats in chunks for MSSQL

dregad:
This PR 
- refactors bug_get_bugnote_stats_array() to use a DbQuery object (+ some minor code cleanup and optimization).
- implements @obmsch's [suggested approach](https://mantisbt.org/bugs/view.php?id=24393#c66522) to process the bug Id's in chunks of 2100 to work around SQL Server's limitation.

This hopefully fixes [#24393](https://mantisbt.org/bugs/view.php?id=24393) - feedback from MSSQL users welcome.

### DIFF
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -1648,12 +1648,7 @@ function bug_get_bugnote_stats_array( array $p_bugs_id, $p_user_id = null ) {
 		$t_id_array[$t_id] = (int)$t_id;
 	}
 
-	if ( null === $p_user_id ) {
-		$t_user_id = auth_get_current_user_id();
-	}
-	else {
-		$t_user_id = $p_user_id;
-	}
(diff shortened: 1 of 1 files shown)


cproensa:
Hi Damien, 
I would, for a cleaner code:
- take the DbQuery object out of the loop,
- use a static label, use `$t_query->sql_in( 'n.bug_id', 'ids_array' )`
- reuse the query object, binding the whole array to said label in each loop iteration

and instead of `fetch_all()`, i'd use `while( $t_query_row = $t_query->fetch() )`
for saving some memory 


_* having the static label added via `sql_in` will take care also of oracle 1k limitation_

dregad:
Hi Carlos, many thanks for the review and advice, I will implement the suggested improvements.

And it's nice to see you're still around ! It's been many months without seeing any activity from you :smile:

PR #1806, opened 2022-05-07
Fix access check on "My View"

atrol:
Fixes [#30178](https://www.mantisbt.org/bugs/view.php?id=30178)

### DIFF
--- a/my_view_inc.php
+++ b/my_view_inc.php
@@ -65,7 +65,6 @@ if( $t_filter === false ) {
 $t_sort = $t_filter['sort'];
 $t_dir = $t_filter['dir'];
 
-$t_update_bug_threshold = config_get( 'update_bug_threshold' );
 $t_bug_resolved_status_threshold = config_get( 'bug_resolved_status_threshold' );
 $t_hide_status_default = config_get( 'hide_status_default' );
 $t_default_show_changed = config_get( 'default_show_changed' );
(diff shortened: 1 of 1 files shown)


my_view_inc.php
```diff
@@ -351,7 +350,9 @@
 			print_icon( 'fa-square', 'fa-status-box ' . $t_status_css, $t_status );
 			echo ' ';
 
-			if( !bug_is_readonly( $t_bug->id ) && access_has_bug_level( $t_update_bug_threshold, $t_bug->id ) ) {
+			$t_can_update = !bug_is_readonly( $t_bug->id ) &&
+				access_has_bug_level( config_get( 'update_bug_threshold', null, $t_current_user_id, $t_bug->project_id  ), $t_bug->id );
```
dregad, line 354: To make code more readable, I would keep the *$t_update_bug_threshold* variable
```php
$t_update_bug_threshold = config_get( 'update_bug_threshold', null, $t_current_user_id, $t_bug->project_id  );
$t_can_update = !bug_is_readonly( $t_bug->id ) && access_has_bug_level( $t_update_bug_threshold, $t_bug->id );
```

atrol: I did it this way because of better performance / less server stressing (fewer calls of config_get in some cases)
@dregad what do you think, readability or performance?

dregad: Ah yes, I see what you mean. Let's leave it like that then.

dregad (approved):

PR #1021, opened 2017-02-07
New 'Export issues threshold'

MrBricodage:
Add a threshold to restrict access to export functions (CSV and Excel export displayed on view_all_bugs_page.php)

default value is VIEWER for this new config option.

Fixes [#22224 ](https://mantisbt.org/bugs/view.php?id=22224)

PR #1810, opened 2022-05-12
New configs to restrict access to Export and Print issues

dregad:
Add 2 new config options, *$g_print_reports_threshold* and *$g_export_issues_threshold* (both defaulting to `VIEWER`), to restrict access to the mass export functions on view_all_bugs_page.php:

- CSV and Excel export
- Print reports

Fixes [#22224](https://mantisbt.org/bugs/view.php?id=22224), [#25492](https://mantisbt.org/bugs/view.php?id=25492)

This replaces PR #1021, which I finally got around to reviewing. The code looks good and tested OK with just a minor correction: adding a missing access check in print_all_bug_page_word.php.

Many thanks @MrBricodage for the initial submission and your work on this, with our apologies for taking so long to handle it.

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2783,7 +2783,7 @@ $g_monitor_add_others_bug_threshold = DEVELOPER;
 $g_monitor_delete_others_bug_threshold = DEVELOPER;
 
 /**
- * Access level needed to export filtered bugs
+ * Access level required to export issues
  * Look in the constant_inc.php file if you want to set a different value.
  * @global integer $g_export_issues_threshold
  */
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2751,7 +2751,6 @@ $g_view_bug_threshold = VIEWER;
 
 /**
  * Access level needed to monitor bugs.
- * Look in the constant_inc.php file if you want to set a different value.
  * @global integer $g_monitor_bug_threshold
  */
 $g_monitor_bug_threshold = REPORTER;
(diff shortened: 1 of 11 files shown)


dregad:
On second thoughts, considering that the Print Reports functionality can be leveraged to perform DOS attacks, I'll set *$g_print_reports_threshold* to a safer default of `UPDATER`, so it is not accessible out of the box by anonymous or self-registered users.

dregad:
The last 5 commits are just code cleanup

config_defaults_inc.php
```diff
@@ -2787,7 +2787,7 @@
  *
  * @global integer $g_print_reports_threshold
  */
-$g_print_reports_threshold = VIEWER;
+$g_print_reports_threshold = UPDATER;
```
atrol, line 2790: I agree that it's safer, but finally it's a weakness of the implementation.
Printing reports is just viewing, so it looks a bit strange to need update rights to view something.
Nevertheless we should keep your change, but it should be mentioned in the release announcement that admins have to set `$g_print_reports_threshold = VIEWER;` to get back the old behavior.

atrol (approved):

config_defaults_inc.php
dregad: I thought about putting DEVELOPER, but in the end I picked UPDATER as the lowest possible role giving a safe default behavior. I am fine with changing that if you prefer.

I am fine with adding a comment in the release notes.

vboctor (approved):

config_defaults_inc.php
dregad: > I am fine with adding a comment in the release notes.

Done -> https://mantisbt.org/bugs/roadmap_page.php?version_id=360

PR #1811, opened 2022-05-13
Allow reporters to update their own issues

atrol:
Fixes [#8141](https://mantisbt.org/bugs/view.php?id=8141)


dregad:
Even if they can edit their own issues, a reporter should not be able to update an issue's status if `$g_update_bug_status_threshold = DEVELOPER;` 

In my test setup, it is currently possible to set the status back to NEW (which is the only status allowed to REPORTERs) if the issue's status is > NEW.

atrol:
> In my test setup, it is currently possible to set the status back to NEW

@dregad maybe because of setting `$g_allow_reporter_reopen = ON;`?

dregad:
> maybe because of setting `$g_allow_reporter_reopen = ON;`?

It *is* ON (default value), and that could explain allowing a status change from (>= $g_bug_resolved_status_threshold) to NEW, but not let me, as reporter, change an ASSIGNED issue back to NEW.

image

vboctor:
@atrol few comments:

1. What can the reporter update on the issue?
2. Does the check for each row on View Issues / My Views page impact the query count?
3. Does the Issue Update REST / SOAP API work for this scenario?

atrol:
@vboctor 

> What can the reporter update on the issue?

At he moment they get offered the same options like any user that has `update_bug_threshold` rights.
As @dregad pointed out, the current implemantation allows too much, e.g. as a reporter, you could change an ASSIGNED issue back to NEW.
That's why I converted the PR to a draft PR, as this must be fixed, but I don't know if/when I will be able to work on that.

> Does the check for each row on View Issues / My Views page impact the query count?

No, it's just about offering the additional functionality to update the issue

> Does the Issue Update REST / SOAP API work for this scenario?

Not with the current implementation.
Seems I forgot to change also function `mc_issue_update` .

PR #1812, opened 2022-05-13
Use filter key instead of URL to build permalink

dregad:
Refactor permalink_page.php to accept a temporary filter key and
generate the URL from that, instead of receiving a fully-formed URL.

This prevents issues when the filter criteria contain a `&` (e.g. a
category named "a & b"), causing the value to be interpreted as 2
distinct parameters due to string_sanitize_url() decoding the `%26`
before processing the query string.

Fixes [#30204](https://mantisbt.org/bugs/view.php?id=30204)


dregad:
@cproensa as always when I touch filter API, I would appreciate if you could have a look. Thanks !

PR #1813, opened 2022-05-17
Apply overflow-wrap more selectively

dregad:
Issue [#27114](https://mantisbt.org/bugs/view.php?id=27114) / PR #1761 introduced a regression in the bugnotes count column's
display on View Issues page.

We now only target those fields susceptible to have long, unbreakable
content (i.e. textarea fields), by

- adding markup to Textarea custom fields (`cftype-textarea` class)
- using more precise CSS selectors for description, steps-to-reproduce,
  additional-information and bugnotes, as well as Textarea custom fields
  (using the new cftype-textarea class).

Fixes [#30268](https://mantisbt.org/bugs/view.php?id=30268)


MaloMn:
Just tested it, works as intended and I haven't seen any side-effects yet :) Thanks!

PR #1817, opened 2022-05-26
implementations in the rest api

gilsonjoanelo:
I created some implementations in the rest api, I am using it in a personal project and I missed some functions that have in the SOAP api and I implemented some in the new api, among them the possibility of searching a question through its summary. Another improved point was the ability to remove or add a version of a given project.

### DIFF
--- a/core/commands/VersionRemoveCommand.php
+++ b/core/commands/VersionRemoveCommand.php
@@ -1,0 +1,82 @@
+<?php
+# MantisBT - A PHP based bugtracking system
+
+# MantisBT is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# MantisBT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+require_api( 'constant_inc.php' );
+require_api( 'config_api.php' );
+require_api( 'helper_api.php' );
+
+use Mantis\Exceptions\ClientException;
+
+/**
+ * A command that adds a project version.
+ */
+class VersionRemoveCommand extends Command {
+	/**
+	 * $p_data['query'] is expected to contain:
+	 * - project_id (integer)
+	 *
+	 * $p_data['payload'] is expected to contain:
+	 * - name (string)
+	 * - description (string)
+	 * - released (bool)
+	 * - obsolete (bool)
+	 * - timestamp (timestamp) - e.g. 2018-05-21T15:00:00-08:00
+	 *
+	 * @param array $p_data The command data.
+	 */
+	function __construct( array $p_data ) {
+		parent::__construct( $p_data );
+	}
+
+	/**
+	 * Validate the data.
+	 */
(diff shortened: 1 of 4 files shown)


gilsonjoanelo:
thinking in this way, it is not a good idea, because more than one record can come, it only returns a single record if it is exactly the same. But the steps of changing and removing project versions can exist right? I did the search for the summary due to an integration that we have in our project, but thinking about your vision, I realized that it is not the best practice to adopt, I will review the code.

atrol:
>  But the steps of changing and removing project versions can exist right?

I think so, but would like to get some feedback from @vboctor

I created [Issue 30415](https://www.mantisbt.org/bugs/view.php?id=30415) to track this change.
The issue must be referenced in the commit message via `Fixes #30415`

core/commands/VersionUpdateCommand.php
```diff
@@ -0,0 +1,99 @@
...
+            # viewing, override the current project. This to avoid problems with
+            # categories and handlers lists etc.
+            global $g_project_override;
+            $g_project_override = $t_project_id;
+        }
+        
+        $t_name = trim( $this->payload( 'name' ) );
+        $t_description = trim( $this->payload( 'description', '' ) );
+        $t_released = $this->payload( 'released', false );
+        $t_obsolete = $this->payload( 'obsolete', false );
```
dregad, line 79: Providing a default value will potentially result in unexpectedly changing the target field if it is not provided by the client in the payload.

In my opinion, in an Update operation the command should only update the information provided in the payload, and leave other fields unchanged. @vboctor what do you think ?

core/commands/VersionUpdateCommand.php
```diff
@@ -0,0 +1,99 @@
...
+        
+        $t_timestamp = $this->payload( 'timestamp', '' );
+        if( is_blank( $t_timestamp ) ) {
+            $t_timestamp = null;
+        } else {
+            $t_timestamp = strtotime( $t_timestamp );
+        }
+        
+        $t_version = version_get($t_id);
```
dregad, line 88: Coding guidelines

core/commands/VersionRemoveCommand.php
```diff
@@ -0,0 +1,82 @@
...
+			# viewing, override the current project. This to avoid problems with
+			# categories and handlers lists etc.
+			global $g_project_override;
+			$g_project_override = $t_project_id;
+		}
+
+		$t_id = trim( $this->payload( 'version_id' ) );
+		//$t_idNew = trim( $this->payload( 'idNew' ) );
+		
+		$t_version_id = version_remove($t_id, null);
```
dregad, line 78: Coding guidelines

core/commands/VersionUpdateCommand.php
```diff
@@ -0,0 +1,99 @@
...
+        $t_version->version = $t_name;
+        $t_version->description = $t_description;
+        $t_version->released = $t_released? VERSION_RELEASED : VERSION_FUTURE;
+        $t_version->obsolete = $t_obsolete;
+        
+        $t_version_id = version_update($t_version);
```
dregad, line 94: Coding guidelines

dregad (changes requested):
I second what @atrol said about searching for Issues by Summary; mc_issue_get_id_from_summary() is is flawed, because if there is more than 1 match for the given summary, it will just return the first one found. Therefore I also don't think this should be implemented in REST API as-is, and changes to *issues_rest.php* should be removed from this PR.

Some code review comments follow regarding the Version update/delete endpoints.

NB: I did not actually test the code yet.

dregad:
@gilsonjoanelo are you planning to address the review comments ? If not, I'll close the PR.

vboctor:
See related PR #1883

PR #1821, opened 2022-06-02
Manage project edit page improvements

dregad:
Improves the page layout and markup, resolves static analysis warnings and fixes the following issues,  :

- [#27274](https://mantisbt.org/bugs/view.php?id=27274)
- [#28557](https://mantisbt.org/bugs/view.php?id=28557)
- [#28562](https://mantisbt.org/bugs/view.php?id=28562)
- [#28606](https://mantisbt.org/bugs/view.php?id=28606)
- [#30423](https://mantisbt.org/bugs/view.php?id=30423)
- [#30435](https://mantisbt.org/bugs/view.php?id=30435)
- [#30490](https://mantisbt.org/bugs/view.php?id=30490) and [#30494](https://mantisbt.org/bugs/view.php?id=30494)


dregad:
The commit _Move Delete Custom Field button to form footer_ is not related to Manage project edit page, but as it is part of the fix for #27274, I'm adding it here to avoid creating a separate PR for it.

atrol:
Unfortunately no time for a deeper look, but did some cursory tests.
All in all it looks good, did not encounter a major issue.

In section `Subprojects` there is a button `Create New Subproject`.
Maybe place it after button `Add as Subproject`, so that the drop down to select a project is the first UI element like in `Categories`, `Versions`, ...

@dregad while you are on it you might want to fix a CSP violation on this page.
There are buttons in section `Manage Accounts` for pagination of users.
The buttons work in general, but I am seeing CSP errors in Firefox console when clicking the buttons.

dregad:
> In section `Subprojects` there is a button `Create New Subproject`.
> Maybe place it after button `Add as Subproject`, so that the drop down to select a project is the first UI element like in `Categories`, `Versions`, ...

I agree it would be more consistent, but after implementing the change, the buttons' vertical alignment is off by ~ 1px 
image
I believe this is caused by the ACE theme which vertically centers the _Add as Subproject_ button relative to the select box, which is slightly higher. I'm not good enough with CSS to fix that and TBH I don't really want to mess with it. 

I also tried to move the button within the form and use a _formaction_ attribute, but that has other issues (the project select is required, CSRF protection) so it's not ideal either.

So unless @syncguru has a better idea, the choices are 
- status quo
- consistent ordering of buttons with misalignment

Let me know what you prefer.

dregad:
> Looks like the inline javascript is introduced by the list.js library

I did some research and this appears to be a known issue (see javve/list.js#498), which has been fixed already (javve/list.js@88874a4c011a9b338c5cce95d51f0911feae1d28). The commit is included in release 2.0.0, but we're on 1.5.0 so I need to check if there are any breaking changes that could be affecting us.

dregad:
According to the changelog and a high-level look at git history, the only breaking change in list.js 2.x is end of support for IE 6-8, which we don't support either.

I have opened [#30490](https://mantisbt.org/bugs/view.php?id=30490) to track the CSP issue (fixed by upgrading list.js to 2.3.1), and related [#30494](https://mantisbt.org/bugs/view.php?id=30494) to fix a regression on the pagination buttons.

atrol:
> So unless @syncguru has a better idea, the choices are
> 
>     * status quo
> 
>     * consistent ordering of buttons with misalignment
> 
> 
> Let me know what you prefer.

I prefer consistent ordering of buttons with misalignment

dregad:
OK, thanks for your feedback.

dregad:
> I agree it would be more consistent, but after implementing the change, the buttons' vertical alignment is off by ~ 1px
> image
> I believe this is caused by the ACE theme which vertically centers the _Add as Subproject_ button relative to the select box, which is slightly higher. I'm not good enough with CSS to fix that and TBH I don't really want to mess with it.

Coming back to this after a few days, I just realized there is actually a very simple solution, which is to make the button's height identical to that of the select (30px) - before it was slightly smaller (28.9333px). Opened [#30550](https://mantisbt.org/bugs/view.php?id=30550) to track this.

PR #1822, opened 2022-06-02
Further improvements to admin/test_langs.php

dregad:
This is a follow-up on PR #1772.

It includes a bit of code cleanup, and brings a few additional UI enhancements (related to [#27383](https://mantisbt.org/bugs/view.php?id=27383)):
- add missing error messages
- improved visibility for plugin headers in table
- add a TOC for tested plugins, allowing user to scroll straight to a specific one
- add buttons to navigate between the core and plugin sections

It also implements 2 extra checks for language strings:
- basic syntax check (invalid tags, unexpected end tags, open/end tag mismatch, etc), fixing [#30447](https://mantisbt.org/bugs/view.php?id=30447)
- Use [HTMLPurifier](http://htmlpurifier.org/) to detect if any unauthorized tags are present, which is a partial fix for [#12242](https://mantisbt.org/bugs/view.php?id=12242)

### DIFF
--- a/admin/test_langs.php
+++ b/admin/test_langs.php
@@ -254,12 +254,14 @@ class LangCheckFile {
 						break;
 					case '[':
 						if( $t_last_token != T_VARIABLE ) {
+							$this->logFail( "Unexpected opening square bracket '['", $t_line);
 							$t_pass = false;
 						}
 						$t_variable_array = true;
 						break;
 					case ']':
 						if( !$t_expect_end_array ) {
+							$this->logFail( "Unexpected closing square bracket ']'", $t_line);
 							$t_pass = false;
 						}
 						$t_expect_end_array = false;
@@ -276,7 +278,7 @@ class LangCheckFile {
 						if( $t_last_token == T_CONSTANT_ENCAPSED_STRING ) {
 							$t_two_part_string = true;
 						} else {
-							$this->logFail( "string concatenation found at unexpected location", $t_line );
+							$this->logFail( "String concatenation found at unexpected location", $t_line );
 							$t_pass = false;
 						}
 						break;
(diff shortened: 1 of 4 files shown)


dregad:
@atrol I did not mark the HTMLPurifier commit as _fixing_ #12242, because it is a manual process at this stage, I'll let you review and judge whether the issue can be resolved as it is, or if you want to keep it open until a more automated solution is implemented.

admin/test_langs.php
```diff
@@ -225,6 +225,8 @@ protected function checkTokens() {
 			return false;
 		}
 
+		$t_do_html_check = extension_loaded('dom') && extension_loaded( 'libxml' );
```
atrol, line 250: In case of non-loaded extensions, some addional line with a warning like `Syntax check for HTML tags can't be executed due to missing "dom" or "libxml" PHP extensions` would be good.

dregad: Good idea, I'll add that.

dregad:
Thanks for the review @atrol. Your suggestion has been implemented.

image


Please let me know how you would like me to handle #12242 (resolve or leave open).

atrol:
> Please let me know how you would like me to handle #12242 (resolve or leave open).

I would leave it open to track that the checks are part of the build process.

atrol:
> > Please let me know how you would like me to handle #12242 (resolve or leave open).
> 
> I would leave it open to track that the checks are part of the build process.

@dregad not sure if it was intended, but you did not leave open.

dregad:
> not sure if it was intended, but you did not leave open.

Source integration closed it automatically due to the wording in the merge commit. I reopened it now.

PR #1824, opened 2022-06-11
Use standard table layout for proj_doc_page.php

dregad:
Previously the files were presented with a hybrid layout using floating
<span> blocks within a table cell. The Use of print_form_buttons()
within the span tag generated invalid HTML, a problem that was fixed in
87ec8b45bca75ade53150f1f4b669604b71a67ad but that commit introduced a
regression because the CSRF token was not passed on so it was no longer
possible to delete documents.

We now switch to a fully tabular layout, moving the data within the
spans (project, creation date, action buttons) to their own columns and
adjust the markup to align to standard MantisBT UI.

Fixes [#30541](https://mantisbt.org/bugs/view.php?id=30541)

### DIFF
--- a/proj_doc_page.php
+++ b/proj_doc_page.php
@@ -123,59 +123,70 @@ print_doc_menu( 'proj_doc_page.php' );
 </div>
 
 <div class="widget-body">
-<div class="widget-main">
+<div class="widget-main no-padding">
 <div class="table-responsive">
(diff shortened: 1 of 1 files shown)


dregad:
### Before
image

### After
image

vboctor (approved):

PR #1829, opened 2022-06-23
Allow for a Simple Report screen

CasN:
Allow for a Simple Report screen for specific user levels, this fixes issue #30737


config_defaults_inc.php
```diff
@@ -5117,3 +5117,15 @@
...
+#################################
+# Simple Report Screen settings #
+#################################
+
+/**
+ * Define the max Level Treshold  for users to present the simple entry screen
+ * Define the fileds to be displayed ( initial value is same as current set
+ * example : $g_simple_bug_report_page_fields = array(	'additional_info',	'attachments',	'category_id',);
+ */
+ g_simple_report_level = UPDATER;
```
atrol, line 5130: The build is broken because the missing "$" introduces a syntax error.

CasN: Thanks for correcting the typo

atrol: @CasN I did not correct anything. This was just one of these typical notes you will get during PR / code review in GitHub.
You have to correct it in your branch and push it, so that the PR gets updated.

CasN: I did correct om my side, followed by a push but I do not see the change within the pull request yet. How much time can be in between?

atrol: It typically takes just a few seconds.
I don't see any new change since yesterday on your clone in branch Simple-report-screen https://github.com/CasN/mantisbt/tree/Simple-report-screen

CasN: OK, added a second commit in order to fix the missing character, hope i got it right now

atrol:
The commit messages do not reference the issue in our bugtracker that was entered for it.
They should contain  `Fixes #30737`

The new options
- must be added to $g_public_config_names
- should be documented in Admin Guide

`simple_bug_report_page_fields` should be added to the array `$t_field_options` in admin/checkcheck_config_inc.php

`$g_simple_report_level` could not just be treated as a threshold, but as an array like many other options.
E.g. this way it would be possible to configure that any other access level than developer gets the simple page.

Before going on, I would like to get some feedback from @dregad and @vboctor if 
- they agree in general to introduce these new options (we could also tell to implement based on https://github.com/mantisbt-plugins/EasyEnter)
- simple_report_level should be an array

config_defaults_inc.php
```diff
@@ -5117,3 +5117,15 @@
...
+# Simple Report Screen settings #
+#################################
+
+/**
+ * Define the max Level Treshold  for users to present the simple entry screen
+ * Define the fileds to be displayed ( initial value is same as current set
+ * example : $g_simple_bug_report_page_fields = array(	'additional_info',	'attachments',	'category_id',);
+ */
+$g_simple_report_level = UPDATER;
+$g_simple_bug_report_page_fields = $g_bug_report_page_fields;
```
atrol, line 5131: Should be set to `'%bug_report_page_fields%'`
This will be less confusing if someone changes `$g_bug_report_page_fields` but not `$g_simple_bug_report_page_fields` in his own config/config_inc.php

bug_report_page.php
```diff
@@ -193,7 +193,14 @@
 $f_copy_notes_from_parent         = gpc_get_bool( 'copy_notes_from_parent', false );
 $f_copy_attachments_from_parent   = gpc_get_bool( 'copy_attachments_from_parent', false );
 
-$t_fields = config_get( 'bug_report_page_fields' );
+$t_user_access_level = user_get_access_level( auth_get_current_user_id(), $t_project_id );
+$t_access_levels = config_get( 'simple_report_level' );
+if ( in_array( $t_user_access_level, $t_access_levels ) ){
```
atrol, line 198: The code should be simplified by using existing function `access_has_project_level`.

I didn't try, but the following line should be enough to replace the current line and the two lines above
`if( access_has_project_level( config_get( 'simple_report_level' ), $t_project_id ) ) {`

Furthermore this ensures that `$g_simple_report_level` could be set just to an integer treshold value like UPDATER, but also to an array.

config_defaults_inc.php
```diff
@@ -5117,3 +5119,15 @@
...
+#################################
+# Simple Report Screen settings #
+#################################
+
+/**
+ * Define the access levels for which the simple entry screen should be presented
+ * Define the fileds to be displayed ( initial value is same as current set
+ * example : $g_simple_bug_report_page_fields = array(	'additional_info',	'attachments',	'category_id',);
+ */
+$g_simple_report_level = array( REPORTER, UPDATER, );
```
atrol, line 5132: Can be set to `$g_simple_report_level = UPDATER;` after having addressed my comment concerning usage of `access_has_project_level`

atrol: The options should be moved in file, so that they are near to `$g_bug_report_page_fields`

atrol: There is no PHPDoc comment for `$g_simple_report_level` and `$g_simple_bug_report_page_fields`

bug_report_page.php
CasN: That line indeed works so will change this also

config_defaults_inc.php
CasN: The moving I understand, placed them initially at the bottom for easy finding, will address this in next commit.
As for the PHPDoc comments, I updated 2 XML files (fields & misc) within last commit, what other file needs to be updated?

atrol: See here what PHPDoc is used for https://docs.phpdoc.org/3.0/guide/references/phpdoc/index.html

In this case, use the `@global` tag for the new variables
For example have a look how it is used  for `$g_bug_report_page_fields`

config_defaults_inc.php
```diff
@@ -2620,8 +2685,9 @@
  * This setting can be set on a per-project basis by using the
  * Manage => Manage Configuration administrator page.
  *
- * @global array $g_bug_view_page_fields
+ * @global array $g_simple_bug_view_page_fields
```
atrol, line 2688: Certainly not intended to change this line

config_defaults_inc.php
```diff
@@ -2579,6 +2579,71 @@
...
+ *   - VIEWER
+ *   - REPORTER
+ *   - UPDATER
+ *   - DEVELOPER
+ *   - MANAGER
+ *   - ADMINISTRATOR
+ *
+ * @global array $g_simple_report_level
+ */
+$g_simple_report_level = array( REPORTER, UPDATER, );
```
atrol, line 2595: As mentioned before, this can now be set to `$g_simple_report_level = UPDATER;`

config_defaults_inc.php
```diff
@@ -2579,6 +2579,71 @@
...
+
+/**
+ * An array of optional fields to show on the simplified bug report page.
+ *
+ * The following optional fields are allowed:
```
atrol, line 2600: No need to provide redundant information.
It's better to use `@see $g_bug_report_page_fields` PHPDoc comment

CasN:
I understand that it is possible to change it.
Just was wondering that this way, it is actually more clearer/explicit for which statuses this would run and that it is not simply a threshold.
Certainly if one would like to include the manager level, it would read:
$g_simple_report_level = array( REPORTER, UPDATER, MANAGER, );
Is there a specific reason why we should not use the array statement?

Your other comments I have taken into account (but not yet committed given the above question).

atrol:
@CasN please don't respond via email to single line comments of GitHub PR reviews.
It messes up the conversation and makes it hard to follow up the changes.
I will add some more information to the former inline comments to answer your question.

config_defaults_inc.php
```diff
@@ -2579,6 +2579,71 @@
 	'view_state',
 );
 
+/**
+ * An array of access levels to identify who should be presented a simplified report form.
+ *
+ * The following Statuses are standard available:
```
atrol, line 2585: No need to document it at this place.
We would have to write the same redundant paragraph to **all** options where thresholds for access levels are set.
I think It's enough what we have in Admin Guide https://mantisbt.org/docs/master/en-US/Admin_Guide/html-desktop/#admin.user.access

config_defaults_inc.php
atrol: While more thinking about that, I would change the name of the variable to `$g_simple_report_levels` and set it to `NOBODY` instead of `UPDATER`.

@CasN as response to your question submitted vie email: I asked for the change from array to integer, as it's the way it's done for quite a lot of similar configuration options (`$g_*_threshold`)
Of course, it's not consistent to have "_threshold_" as part of a variable name and to use it not just for integer access level threshold, but also for arrays of access levels.
The reason for that is, that the ability to use arrays in addition to integer values was introduced after naming.
The right clean way would be to rename all these options, deprecate/obsolete them, migrate them, ...
This is certainly nothing that should be done as part of the current PR.
It would have to be implemented  as part of a bigger code cleanup when introducing a new MantisBT major 3.x version, as it would break some 3rd party plugins.

CasN:
@atrol, Initially you advised to change it to an array (https://github.com/mantisbt/mantisbt/pull/1829#issuecomment-1166487115) and provided a one line condition statement (https://github.com/mantisbt/mantisbt/pull/1829#discussion_r906821053).  Now you tell me that I should revert from the array back to a regular statement which also requires the condition line to be adjusted. This also means that I need to adjust all other entries related to this config option. That is not a big issue as such, ( although the array option remains clearer and offers more functionality). Just want to be sure that this then is the final way to go.

atrol:
> Now you tell me that I should revert from the array back to a regular statement which also requires the condition line to be adjusted.

I don't see that any condition line must be adjusted because of that.
Which line are you referring to?
Is it the one with `access_has_project_level`?
If yes, check the source code of this function and finally `access_compare_level`.
It's exactly written to cover the case where access level (thresholds) can be integer but also arrays.

```
function access_compare_level( $p_user_access_level, $p_threshold = NOBODY ) {
	if( is_array( $p_threshold ) ) {
		return( in_array( $p_user_access_level, $p_threshold ) );
	} else {
		return( $p_user_access_level >= $p_threshold );
	}
}
```

CasN:
So here is my sample:
$p_user_access_level = ADMINISTRATOR
$p_threshold =  UPDATER or $p_threshold = array (REPORTER, UPDATER,)
If the user has higher level than UPDATER i want to get a return value of FALSE
Now let's check the function:
1. using the array, return value will be FALSE, which is correct
2. using the integer, return value will be TRUE which is wrong
What am I missing here?

atrol:
Ok, got it.
It's because the option is exactly the other way around like the other "threshold"-ones.
If we want to treat the option like the other ones, it would have to be renamed to something like  `full_report_levels` and initialized to REPORTER.

As mentioned before, I really would like to get some feedback from other devs before investing more time in it.
Ping @dregad @vboctor

dregad:
@atrol 
>    agree in general to introduce these new options (we could also tell to implement based on https://github.com/mantisbt-plugins/EasyEnter)

I am OK with the idea of the new config.
EasyEnter works also, but it's a more fragile approach as it depends on the markup to select the fields to show/hide.

>    simple_report_level should be an array

This should give the same flexibility as our other threshold-like settings, i.e. allow both integer and arrays.

@CasN in reply to https://github.com/mantisbt/mantisbt/pull/1829#issuecomment-1168453465

Access checks are meant to grant additional privileges when user has a given level or higher, but here you're trying to do the opposite, i.e. take away privileges. 

This is easy to do when only dealing with simple thresholds (integer), but if you also consider arrays then there is currently no way to do such a negative check. We would need a new API function _access_does_not_have_level()_ function or something like that.

As an alternative, what about inverting the logic of the simple screen concept, i.e. have the new configs grant access to an extended screen instead ?

dregad:
@atrol we cross-posted, but I think we're saying the same thing.

atrol:
> have the new configs grant access to an extended screen instead ?

As a side note @dregad not sure you remember that we had another approach for that some while ago by providing a separate page https://github.com/mantisbt/mantisbt/blob/release-1.1.9/bug_report_advanced_page.php

IIRC this was dropped after `bug_report_page_fields` was introduced.

CasN:
In my view this access_compare_level function is not serving both options equally. An array does not need to have sequential values, gaps are possible. The second statement is building on that sequential logic. 
Think about this situation when  $p_threshold = array ( REPORTER, UPDATER, MANAGER), it cannot be caught as one integer.
Reverting the logic of course is possible but.... you will encounter the same issue.
I really like the array logic but of course I will stick with the Mantis rules if this would be added to Core, and use :
if ( user_get_access_level( auth_get_current_user_id(), $t_project_id ) <= config_get( 'simple_report_levels' ) ){
or when we prefer to revert the logic to :
if ( user_get_access_level( auth_get_current_user_id(), $t_project_id ) > config_get( 'simple_report_levels' ) ){

dregad:
> As a side note @dregad not sure you remember that we had another approach for that some while ago by providing a separate page https://github.com/mantisbt/mantisbt/blob/release-1.1.9/bug_report_advanced_page.php
> 
> IIRC this was dropped after `bug_report_page_fields` was introduced.

I never knew that, I think it was before I joined the project (or even started using Mantis).

dregad:
> access_compare_level function is not serving both options equally. An array does not need to have sequential values, gaps are possible. 

Which is the whole point. The threshold can be either an integer, OR an array; if array, then individual values are checked, and if integer then it's a >= comparison.

atrol:
@CasN you want to tell that something like that works not for integers and arrays?
```
if( access_has_project_level( config_get( 'extended_report_levels' ), $t_project_id ) ) {
	$t_fields = config_get( 'bug_report_page_fields' );

} else {
	$t_fields = config_get( 'simple_bug_report_page_fields' );
}
```

atrol:
@cas I prefer Option 3)
Do it the way I described in note https://github.com/mantisbt/mantisbt/pull/1829#issuecomment-1168796710
```
if( access_has_project_level( config_get( 'extended_report_levels' ), $t_project_id ) ) {
	$t_fields = config_get( 'bug_report_page_fields' );

} else {
	$t_fields = config_get( 'simple_bug_report_page_fields' );
}
```

This works for integer and arrays, e. g. set
`$g_extended_report_levels = UPDATER;`
REPORTER will get the new simple page based on `simple_bug_report_page_fields`, all other access levels > = UPDATER will get the standard page based on `bug_report_page_fields`
set it to
`$g_extended_report_level = array ( DEVELOPER, ADMINISTRATOR );`
DEVELOPER and ADMINISTRATOR will get the standard page based on `bug_report_page_fields`, any other access level (also MANAGER) will get the new simple page based on `simple_bug_report_page_fields`.

Default setiing in config_defaults_inc.php should be REPORTER to provide full backward compatibility.

The only small inconstistency we would have is the name of the options.
It should be something like `extended_bug_report_page_fields` for the old existing option, but I think we can live with that.

CasN:
I have updated the corresponding files along what we agreed, hope we can proceed (I will be offline till July 11th)

atrol:
Thanks @CasN I will review this weekend or next week.
Enjoy your vacation.

config_defaults_inc.php
```diff
@@ -2579,6 +2579,24 @@
 	'view_state',
 );
 
+/**
+ * An array of access levels or a threshold to identify who should be presented the extended  report form.
+ * Setting it to an array, one can cherry-pick the access levels that should get the extended report screen.
+ * sample: $g_extended_reporter_levels = array( DEVELOPER, ADMINISTRATOR );
+ * Now only the Developers & Admiistrators will get the extended report form, all others will get the simplified report form.
+ *
+ * @global array $g_simple_report_level
```
atrol, line 2588: $g_extended_report_levels

config_defaults_inc.php
```diff
@@ -2579,6 +2579,24 @@
...
+ */
+$g_extended_reporter_levels = REPORTER;
```
atrol, line 2590: $g_extended_report_levels

config_defaults_inc.php
```diff
@@ -4905,6 +4924,8 @@
 	'show_version_dates_threshold',
 	'show_version',
 	'signup_use_captcha',
+	'simple_bug_report_page_fields',
+	'simple_report_level',
```
atrol, line 4928: extended_report_levels

CasN:
Have corrected the wrong entries for config_defaults_inc.php followed by another commit

jrnewton:
Hi All - I'm very much interested in this enhancement.  Any chance this PR will get merged soon?

docbook/Admin_Guide/en-US/config/misc.xml
```diff
@@ -104,6 +104,12 @@
 				</para>
 			</listitem>
 		</varlistentry>
+		<varlistentry>
+			<term>$g_extended_report_levels</term>
+			<listitem>
+				<para>Access levels for which users are provided with the extended report form.	</para>
```
atrol, line 110: Remove trailing blanks.

config_defaults_inc.php
```diff
@@ -4905,6 +4924,8 @@
 	'show_version_dates_threshold',
 	'show_version',
 	'signup_use_captcha',
+	'simple_bug_report_page_fields',
+	'extended_report_levels',
```
atrol, line 4928: Keep the alphabetical order of the options.

atrol:
@CasN  do you plan to solve the review comments in short term?
There is one major open (wrong variable name), one minor (wrong variable name in comment)
I just added some more minor ones.

After solving, please
- squash the commits to a single one 
- reword the commit message
- reference the issue by adding "Fixes #30737" to the commit message body
- rebase to latest master

CasN:
Hope I covered the latest requests in full such that this change is ready for release.

atrol:
> Hope I covered the latest requests in full such that this change is ready for release.

@CasN your PR shows now 812 changed files, 1635 Commits and non-resolved conflicts for 25 files.

CasN:
Perhaps the best thing to do, is to throw this away. I will start from scratch in order to get a clean start. I suspect  that the rebase introduced this (most likely I made the wrong choice there). Do not think that the squashing introduced this. Looked at the commit which only had the changed files.

PR #1831, opened 2022-07-09
Ensure LDAP ext loaded when $g_login_method = LDAP

dregad:
Fixes [#30771](https://mantisbt.org/bugs/view.php?id=30771)

PR Includes a code cleanup commit, moving core.php's helper functions to the end of the file - it has always bothered me to have the main code flow interspersed with function declarations.


atrol:
I would add the check also to `check_php_inc.php`.
Maybe it's even enough to have it that place, instead of checking again and again with each page requst.

dregad:
Thanks for the feedback @atrol 

> I would add the check also to check_php_inc.php.
> Maybe it's even enough to have it that place, instead of checking again and again with each page requst.

Having the test only in admin checks is actually not sufficient. The problem is that in the absence of LDAP extension when $g_login_method = LDAP, the admin check's index page (as well as any other page) does not even load at all. A fatal error is triggered whenever an LDAP function is called, which happens in quite a lot of places (e.g. to display user avatar in navbar, e-mail is required; or whenever auth_flags() is called)

Also, following that logic, we should remove the _mbstring_ check from core.php as well, don't you think ? (rhetorical question, don't answer :wink: )

> remove the check from function ldap_connect_bind()

That makes sense I guess, as it is made redundant by the check in core.php.

> add the check to the other functions where ldap_connect is called (ldap_cache_user_data, ldap_authenticate_by_username), but remove it from core.php.

I assume your concern here is performance; IMO that would be an even worse solution in this case, because the LDAP extension check would then be executed every time an ldap API function is called (which can be quite a lot for some pages, especially if Mantis is configured to show realnames and  $g_use_ldap_realname = ON.

The test I introduced is quite cheap - 1 check for a global variable which is shielding the extension_loaded() call, so I think it's the simplest and most efficient approach to fix the problem.

dregad:
For the record, I tried adding the following to check_php_inc.php, but due to the test in core.php, then it never gets called.

```php
global $g_login_method;
if( $g_login_method == LDAP ) {;
	$t_ldap_loaded = extension_loaded( 'ldap' );

	check_print_test_row(
		'LDAP PHP extension is available to support configured $g_login_method = LDAP',
		$t_ldap_loaded,
		array( false => 'Ensure that the LDAP extension is installed and enabled or switch to a different login method.' )
	);
}
```

atrol:
> Having the test only in admin checks is actually not sufficient. The problem is that in the absence of LDAP extension when $g_login_method = LDAP, the admin check's index page (as well as any other page) does not even load at all.

@dregad the admin/check/index.php page loads without installed ldap extension when $g_login_method = LDAP
You are right concerning the other pages. 

> For the record, I tried adding the following to check_php_inc.php, but due to the test in core.php, then it never gets called.

Right, so `mbstring` could also be removed from `$t_extensions_required` array in check_php_inc.php
I am not sure, I remember some discussion where we decided to keep it. Maybe just to avoid confusion that the required  mbstring extension is not contained in the list of required extensions.

dregad:
> the admin/check/index.php page loads without installed ldap extension when $g_login_method = LDAP

You're right it does load, but with errors in the PHP log:
```
[Sun Jul 10 20:13:12.476048 2022] [php7:notice] [pid 758139] [client 127.0.0.1:33542] Call to undefined function ldap_connect()
/home/dregad/dev/mantisbt/core/ldap_api.php: 248: - - - - ldap_connect_bind()
/home/dregad/dev/mantisbt/core/ldap_api.php: 317: - - - - ldap_cache_user_data()
/home/dregad/dev/mantisbt/core/ldap_api.php: 179: - - - - ldap_get_field_from_username()
/home/dregad/dev/mantisbt/core/ldap_api.php: 167: - - - - ldap_email_from_username()
/home/dregad/dev/mantisbt/core/user_api.php: 961: - - - - ldap_email()
/home/dregad/dev/mantisbt/core/authentication_api.php: 108: - - - - user_get_email()
/home/dregad/dev/mantisbt/core/authentication_api.php: 227: - - - - auth_flags()
/home/dregad/dev/mantisbt/css/status_config.php: 65: - - - - auth_login_page()
, referer: http://localhost/mantis/admin/check/
```

This is caused by loading dynamic CSS status_config.php, which really does not need to be included in this page. I'll track that in [#30773](https://mantisbt.org/bugs/view.php?id=30773)

> Right, so mbstring could also be removed from $t_extensions_required array in check_php_inc.php
> I am not sure, I remember some discussion where we decided to keep it. Maybe just to avoid confusion that the required mbstring extension is not contained in the list of required extensions.

Correct. I'll leave it there, it does not hurt.

PR #1832, opened 2022-07-09
Do not empty email, realname when updating with LDAP

dregad:
Fixes [#24720](https://mantisbt.org/bugs/view.php?id=24720)


atrol:
Looks good, didn't test.
Most of the Travis checks failed (all where PHP version >= 7.2).
I guess this is not caused by the code changes, but I didn't have a deeper look.

dregad:
> Most of the Travis checks failed (all where PHP version >= 7.2).

Yes I noticed that too, but for sure it can't be caused by the change I committed. The logs show a lot of errors like this:

_PHP Deprecated:  Return type of PHPUnit_Framework_TestCase::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/travis/build/mantisbt/mantisbt/vendor/phpunit/phpunit/src/Framework/TestCase.php on line 307_

So maybe it's because I changed the PR's target branch from master to master-2.25 after submitting it.

dregad:
Well I force-pushed the branch and now it's all succeeded. Go figure.

PR #1833, opened 2022-07-10
Only load status_config.php CSS when necessary

dregad:
As mentioned in https://github.com/mantisbt/mantisbt/pull/1831#issuecomment-1179776776

Fixes [#30773](https://mantisbt.org/bugs/view.php?id=30773)

### DIFF
--- a/bug_actiongroup.php
+++ b/bug_actiongroup.php
@@ -348,6 +348,7 @@ foreach( $f_bug_arr as $t_bug_id ) {
 form_security_purge( $t_form_name );
 
 if( count( $t_failed_ids ) > 0 ) {
+	require_css( 'status_config.php' );
 	bug_group_action_print_top();
 	bug_group_action_print_results( $t_failed_ids );
 	bug_group_action_print_bottom();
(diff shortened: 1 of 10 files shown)


lang/strings_english.txt
```diff
@@ -178,6 +178,7 @@ $s_opensearch_text_description = '%s: full-text search';
 $s_select_option = '(select)';
 
 # bug_actiongroup_page.php : mass treatment
+$s_bug_actiongroup_result = 'The following issues could not be processed';
```
atrol, line 181: This change is not related to loading status_config.php CSS

dregad: You're right it belongs to another branch I was working on in parallel, it was committed by mistake.

atrol:
@dregad do we still need this ugly special handling in status_config.php after your change?

```
if( $t_referer_page == auth_login_page() ) {
	# custom status colors not needed.
	http_caching_headers( false );
	exit;
}

switch( $t_referer_page ) {
	case AUTH_PAGE_USERNAME:
	case AUTH_PAGE_CREDENTIAL:
	case 'signup_page.php':
	case 'lost_pwd_page.php':
	case 'account_update.php':
		# We don't need custom status colors on login page, and this is
		# actually causing an error since we're not authenticated yet.
		http_caching_headers( false );
		exit;
}
```

vboctor (approved):

PR #1834, opened 2022-07-24
Allow setting rel=noopener/noreferrer for URL links

dregad:
$g_html_make_links now accepts 2 new contants, LINKS_NOOPENER and
LINKS_NOREFERRER, allowing the admin to set the corresponding type for
the URL links that are generated when the MantisCoreFormatting plugin's
URL Processing is ON.

The default value for $g_html_make_links has been changed from
`LINKS_SAME_WINDOW` to `LINKS_SAME_WINDOW | LINKS_NOOPENER`, for
security reasons.

Fixes [#30791](https://mantisbt.org/bugs/view.php?id=30791)

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2042,15 +2042,25 @@ $g_file_download_xsendfile_header_name = 'X-Sendfile';
 
 /**
  * Convert URLs and e-mail addresses to html links.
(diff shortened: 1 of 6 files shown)


dregad:
Just rebased the PR and added processing of links in Markdown, as well as the fix for [#30918](https://mantisbt.org/bugs/view.php?id=30918)

PR #1838, opened 2022-08-04
Fix tab index

dregad:
Fixes [#30835](https://mantisbt.org/bugs/view.php?id=30835)


bug_update_page.php
```diff
@@ -769,7 +770,7 @@
 if( $t_bottom_buttons_enabled ) {
 ?>
 	<div class="widget-toolbox padding-8 clearfix">
-		<input <?php helper_get_tab_index(); ?>
+		<input <?php echo helper_get_tab_index(); ?>
```
atrol, line 773: The same change is needed at the beginning of the file if `$t_top_buttons_enabled`

PR #1839, opened 2022-08-04
When copying categories, only process "local" ones

dregad:
It does not make sense to copy the global categories, if they are wanted
in the target project, it should inherit them.

Fixes [#30812](https://mantisbt.org/bugs/view.php?id=30812)

### DIFF
--- a/manage_proj_cat_copy.php
+++ b/manage_proj_cat_copy.php
@@ -65,7 +65,7 @@ if( $f_copy_from ) {
 	trigger_error( ERROR_CATEGORY_NO_ACTION, ERROR );
 }
 
-$t_rows = category_get_all_rows( $t_src_project_id );
+$t_rows = category_get_all_rows( $t_src_project_id, false );
 
 foreach ( $t_rows as $t_row ) {
 	$t_name = $t_row['name'];


atrol:
What about still copying global categories if "Inherit Global Categories" of target project is disabled?

dregad:
> What about still copying global categories if "Inherit Global Categories" of target project is disabled?

As mentioned above, if global categories are wanted in the target project, then they should be inherited. I don't see the point of creating duplicates. 

What is your argument ?

atrol:
Consider the global categories as some kind of template / convenience functionality when creating new projects.
You want to copy them, remove some of them, add some new ones, but after that, you want to be independent from any global changes.

dregad:
I didn't see it that way. Makes sense.

dregad:
Thinking about it some more, I failed to consider another scenario: subprojects. Given

```
Hierarchy            Categories
---------            ------------
global                 cat_g
  +- parent project    cat_p
     +- subproject     cat_s
``` 

In its current state, the PR will also **not** process categories inherited from a parent project, i.e. copying _subproject_'s categories will only copy `cat_s` and not `cat_p, cat_g`.

Instead of trying to second-guess the users' intentions, maybe the correct approach would be to give them some flexibility by providing a flag next to the Copy to/from buttons, to indicate whether they want a shallow copy (i.e. just local categories), or a complete one including the parent project(s)'s and global categories.

Thoughts ?

dregad:
> Instead of trying to second-guess the users' intentions, maybe the correct approach would be to give them some flexibility by providing a flag next to the Copy to/from buttons, to indicate whether they want a shallow copy (i.e. just local categories), or a complete one including the parent project(s)'s and global categories.

See #1840 for a proposed implementation.

dregad:
Closed in favor of #1840

PR #1842, opened 2022-08-11
Fix form should be a POST not a GET

chrisgraham:
Browsers/extensions may pre-load any GET URL, including from forms. GET is specified as read-only. A user of mine was having all bugs they viewed auto-monitored.

I have confirmed this fix works. Mantis allows GET or POST for every/most inputs.

### DIFF
--- a/bug_view_inc.php
+++ b/bug_view_inc.php
@@ -723,7 +723,7 @@ if( $t_flags['monitor_show'] ) {
 			if( $t_flags['monitor_can_add'] ) {
 	?>
 			<br /><br />
-			<form method="get" action="bug_monitor_add.php" class="form-inline noprint">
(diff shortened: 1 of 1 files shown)


dregad:
@chrisgraham Thanks for your contribution.

I have no objections with the proposed change, but following our process can you please open an issue at https://mantisbt.org/bugs to track this, and reference its id it in your commit message (i.e. `Fixes #xxxx`)

Thanks.

chrisgraham:
Ok thanks, done.

dregad (approved):

dregad:
Merged in 94520849c8850bf22b0bf2b869057e131cca9830. Thanks again @chrisgraham.

PR #1843, opened 2022-08-21
REST API to fetch project users

vboctor:
- `/api/rest/projects/:id/users` to get all users or all users with a specified access level or above. This functionality is available in SOAP API.
- `/api/rest/projects/:id/handlers` to get all users that can be assigned issues in the project.

The user calling the API must have viewer access to the project. This was not enforced in SOAP API, but now it is.

Will submit a separate PR for Postman collection update, since there is a bunch of unrelated changes that accumulated over time.

Fixes #22791, #30907

### DIFF
--- a/core/commands/ProjectUsersGetCommand.php
+++ b/core/commands/ProjectUsersGetCommand.php
@@ -66,10 +66,10 @@ class ProjectUsersGetCommand extends Command {
 	 */
 	function validate() {
 		$this->project_id = (int)$this->query( 'id' );
-		$this->access_level = (int)$this->option( 'access_level' );
+		$this->access_level = (int)$this->query( 'access_level' );
 		$this->page = (int)$this->query( 'page', 1 );
 		$this->page_size = (int)$this->query( 'page_size', 50 );
-		$this->include_access_levels = (int)$this->option( 'include_access_levels', true );
+		$this->include_access_levels = (int)$this->query( 'include_access_levels', true );
 
 		if( $this->project_id <= ALL_PROJECTS ) {
 			throw new ClientException(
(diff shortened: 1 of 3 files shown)


core/commands/ProjectUsersGetCommand.php
```diff
@@ -0,0 +1,188 @@
...
+		$this->user_id = auth_get_current_user_id();
+		$this->project_id = (int)$this->query( 'id' );
+		$this->access_level = (int)$this->query( 'access_level' );
+		$this->page = (int)$this->query( 'page', 1 );
+		$this->page_size = (int)$this->query( 'page_size', 50 );
+		$this->include_access_levels = (int)$this->query( 'include_access_levels', true );
+
+		if( $this->project_id <= ALL_PROJECTS ) {
+			throw new ClientException(
+				sprintf( "Must specify a specific project id.", $this->project_id ),
```
atrol, line 108: Should be something like
`sprintf( "Must specify a specific project id, %d is invalid", $this->project_id ),`

core/commands/ProjectUsersGetCommand.php
```diff
@@ -0,0 +1,188 @@
...
+				ERROR_PROJECT_NOT_FOUND,
+				array( $this->project_id ) );
+		}
```
atrol, line 111: Is this blocked needed, or is everything covered by the following check to `!project_exists` (didn't try)?
If it's needed, then it should get an own error code, something like ERROR_INAVLID_PROJECT_ID.

dregad: Indeed the check is redundant - project_exists() will return _false_ if given a negative project id.

vboctor: Removed the check.

PR #1846, opened 2022-08-29
Simple Screen

CasN:
Fixes #30737
Allow users to report an issue with a very simple screen based upon their authorization level (Mantis-wide or by project).

This is a replacement for #1829

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2592,6 +2592,24 @@ $g_bug_report_page_fields = array(
 );
 
 /**
+ * An array of access levels or a threshold to identify who should be presented the extended  report form.
+ * Setting it to an array, one can cherry-pick the access levels that should get the extended report screen.
+ * sample: $g_extended_reporter_levels = array( DEVELOPER, ADMINISTRATOR );
+ * Now only the Developers & Administrators will get the extended report form, all others will get the simplified report form.
+ *
+ * @global array $g_extended_report_levels
+ */
+$g_extended_report_levels = REPORTER;
+
+/**
+ *
+ * @see $g_bug_report_page_fields  
+ * 
+ */
+$g_simple_bug_report_page_fields = '%bug_report_page_fields%' ;
+
+
+/**
  * An array of optional fields to show on the bug view page.
  *
  * The following optional fields are allowed:
@@ -4775,6 +4793,7 @@ $g_public_config_names = array(
 	'eta_enum_string',
 	'excel_columns',
 	'export_issues_threshold',
+	'extended_report_levels',
 	'fallback_language',
 	'favicon_image',
 	'file_download_content_type_overrides',
@@ -4917,6 +4936,7 @@ $g_public_config_names = array(
 	'show_version_dates_threshold',
 	'show_version',
 	'signup_use_captcha',
+	'simple_bug_report_page_fields',
 	'sort_by_last_name',
 	'sort_icon_arr',
 	'sponsor_threshold',
--- a/admin/check/check_config_inc.php
+++ b/admin/check/check_config_inc.php
@@ -96,6 +96,7 @@ check_print_test_row( 'Default move category must exists ("default_category_for_
 );
 
 $t_field_options = array(
+	'simple_bug_report_page_fields',
 	'bug_report_page_fields',
 	'bug_view_page_fields',
 	'bug_update_page_fields'
--- a/bug_report_page.php
+++ b/bug_report_page.php
@@ -193,7 +193,12 @@ $f_report_stay			= gpc_get_bool( 'report_stay', false );
 $f_copy_notes_from_parent         = gpc_get_bool( 'copy_notes_from_parent', false );
 $f_copy_attachments_from_parent   = gpc_get_bool( 'copy_attachments_from_parent', false );
 
-$t_fields = config_get( 'bug_report_page_fields' );
+if( access_has_project_level( config_get( 'extended_report_levels' ), $t_project_id ) ) {
+	$t_fields = config_get( 'bug_report_page_fields' );
+} else {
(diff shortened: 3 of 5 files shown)


config_defaults_inc.php
```diff
@@ -2591,6 +2591,24 @@
...
 
+/**
+ * An array of access levels or a threshold to identify who should be presented the extended  report form.
+ * Setting it to an array, one can cherry-pick the access levels that should get the extended report screen.
+ * sample: $g_extended_reporter_levels = array( DEVELOPER, ADMINISTRATOR );
+ * Now only the Developers & Administrators will get the extended report form, all others will get the simplified report form.
+ *
+ * @global array $g_extended_report_levels
+ */
+$g_extended_reporter_levels = REPORTER;
```
atrol, line 2602: Must be 
```
$g_extended_report_levels = REPORTER;
```

CasN: Have corrected the typo and squashed the 2 commits into one(1)

vboctor:
Long time back we used to have a simple and advanced page for report, update, etc. I have done the work to get rid of such concept for reasons like:

1. goldilocks problem - often users will want more on the simple or less on the advanced.
2. duplicate work for developers - often changes need to be applied to both pages.
3. duplicate work for extensions - do extension work with simple and advanced pages?

Hence, I opted to go with having a single set of pages with support for customizability. If we want more customizations, we should add it, but not go with discrete two options for report/update pages.

In short, I'm not in favor of going back to simple/advanced world.

CasN:
This is not going back to the old world with simple (often still too complex) and advanced scripts, no discrete options, this remains one script which allows for tailoring the lay-out based upon the environment it is used in. I have received numerous requests for this from companies and end-users hence this development along Mantis standards. 
1. Goldilocks not applicable, is a decision of the environment
2. Duplicate work is not applicable, just one script
3. Extension will work as designed so no duplicate work hence also not applicable
Clearly it is up to the dev-team to incorporate or not.

atrol:
@vboctor you are coming late to the party
I pinged you in an early state https://github.com/mantisbt/mantisbt/pull/1829#issuecomment-1168628079 to get your opinion to prevent me in investing quite a lot of time in reviewing the PR and guiding the contributor.

> Long time back we used to have a simple and advanced page for report

Right, I mentioned it in discussion of previous PR, see https://github.com/mantisbt/mantisbt/pull/1829#issuecomment-1168743020

>  I have done the work to get rid of such concept for reasons like:
> goldilocks problem - often users will want more on the simple or less on the advanced.

You can have it now by adjusting the options

> duplicate work for developers - often changes need to be applied to both pages.

There is no longer duplicate work, there is just a single page

> duplicate work for extensions - do extension work with simple and advanced pages?

There is just one page, driven by one of the options

vboctor:
I looked at the code, it is different than I thought initially, but I still have concern with the proposal.

If the advanced page is shown to developer+. then a user with access level updated would see a simple report page and a complex update page.

I think we are better served with extending the capabilities of the configuration for what fields to show on a page. For example:

$g_bug_report_page_field can be set to an array of elements. Each element is a field name. This is what we have today.

$g_bug_report_page_field can be set to an array of elements. Each element is either a field name (to always show) or array( "name" => "field_name", access_level => access level threshold or array).

Then at the beginning of each page, we just filter the visible fields based on access level (when applicable). This doesn't introduce more configs and enables this functionality across all pages (report, update, view, etc).

CasN:
@vboctor, for me that indeed was the idea that reporting an issue can be done with just a few fields (as little as required for the organization)  and that the same user can see all data in update/view mode. It is all about simply registering an issue:)

CasN:
@vboctor, so what are we going to do with this feature. We are 2 months down the road now.

CasN:
@vboctor, we are more than a year later now, version 2.26 was released in the mean time without this simple feature. In case you are still reluctant to include in next release, let me at least know. I will then remove the PR.

CasN:
@vboctor any input from your side?

vboctor:
@CasN Thanks for your patience on this. I had another look and here are my thoughts:

If we are to go with the approach in this PR, then here are some thoughts on the naming.

- `extended_report_levels` -> `bug_extended_report_page_threshold` (default NOBODY)
- `bug_report_page_fields` -> use this for the non-extended case.
- use `bug_extended_report_page_fields` for the extended case. (Default `%bug_report_page_fields%`).

As I'm thinking of this approach, I see the following concerns:

- We don't have an equivalent approach for custom fields. Custom fields and native fields currently depend on access level to determine if different user should experience a page differently.
- We don't have this except on the report page. What if we need a simple view page?

So for a field to show:

- Is the feature logically on - e.g. profiles enabled.
- Does the user has the necessary access level to the field.
- Is the field in the list of fields for the page.

We would be adding a fourth criteria and only adding it to report page and native fields. If we believe it is valuable to provide such functionality, I wonder if we are better off using access levels to control visibility of native fields instead. In such case a native field / custom field will leverage an access level to have the field included on any of the pages.

Here are some examples for the configuration:

- Current

```php
$g_bug_report_page_fields = array(
	'additional_info',
        :
	'view_state',
);
```

- New

```php
$g_bug_report_page_fields = array(
	'additional_info',
	'attachments' => DEVELOPER,
	'category_id' => array( DEVELOPER, MANAGER),
	:
	'view_state',
);
```

This can also be applied to `$g_bug_view_page_fields`, `$g_bug_update_page_fields`, etc.

CasN:
@***@***.******@***.***>,
What is currently on the table was discussed with the various dev’s including the naming convention.

Not saying your suggestions are not valid  but is does mean a large overhaul in the area of field access/visibility.
Such a change is not really what I had in mind ( and developed for one of the Mantis users out there), just a simple report screen for reporters.

So I will withdraw this PR from mantisBT and decide if I keep it as a local change or transfer it into a plugin.

PR #1841, opened 2022-08-09
Markdown. Add external css file, remove inline styles from quotes.

grummbeer:
This is an attempt to finish PR #1004 with fewer code changes. This is the first merge request of a small series addressing some Markdown issues.

### Add external css file, clean up quotes

- Reduce lines of codes in MantisMarkdown.php ("decomplexing")
- Fixing https://mantisbt.org/bugs/view.php?id=22190
- tiny change, no impact to core files.
- Remove test. It does not pass and after removing the inline styles, it is just a copy of [parsedown's own test](https://parsedown.org/tests/simple_blockquote).
- Having an external css file in place makes it easy to address styling issues like [this](https://mantisbt.org/bugs/view.php?id=22181) or [this](https://mantisbt.org/bugs/view.php?id=22485) and others in the future.

| before | after |
|---        |---      |
|mantis-quotings-before | mantis-quotings-after      |

If this patch is accepted, i think the [origin PR](https://github.com/mantisbt/mantisbt/pull/1004)  can closed and i will commit a follow up merge request to clean up the "table stuff". [reduce LOC](https://github.com/mantisbt/mantisbt/compare/master...grummbeer:mantisbt:markdown-cleanup-02-tables) and [reduce test](https://github.com/mantisbt/mantisbt/compare/master...grummbeer:mantisbt:markdown-cleanup-03-table-test).

[This note](https://github.com/mantisbt/mantisbt/pull/1004#issuecomment-287067147) is about emails and inline styles. Does Mantis send HTML mail at all or does it no longer send HTML mail? Text mails with the generally accepted `>` are well formatted by the clients, HTML-blockquote should also have a client dependent default styleing that is not quite so terrible. But if this is still relevant, a following pull request should solve this issue. Source code for website is something other than the HTML part of an email.

PR #1848, opened 2022-09-04
Markdown. Simplify table processing.

grummbeer:
After reducing some lines of code in the [first step](https://github.com/mantisbt/mantisbt/pull/1841), this is the second patch of a small series addressing some markdown issues. 

Fix: https://mantisbt.org/bugs/view.php?id=30919 (step 2. and 3.)

### Simplify `<table>` processing

Parsedown provides the method `element` which gives access to all marked elements and makes it easy to change the  properties/attributes of an element just before converted into HTML markup. 

- Reduce lines of codes in MantisMarkdown.php ("decomplexing").
- Reduce the test to its intended purpose.
    - Asserting the existence of the `class` attribute with the value `table table-nonfluid` and not the processing of inline formatting. 
    - The `[link](/)` part made the test unnecessarily dependent on the `process_urls` option.
    - No need to copy an already existing test (https://parsedown.org/tests/table_inline_markdown).

 
| before (master) | after |
|---                       |---     |
| table-before-master | table-after  |

### DIFF
--- a/plugins/MantisCoreFormatting/core/MantisMarkdown.php
+++ b/plugins/MantisCoreFormatting/core/MantisMarkdown.php
@@ -79,6 +79,20 @@ class MantisMarkdown extends Parsedown
 	}
 
 	/**
+	 * @param array $Element Properties of a marked element.
+	 * @return string HTML markup of the element.
+	 */
+	protected function element( $Element )
+	{
+		# Adding CSS classes to tables.
+		if( $Element['name'] === 'table' ) {
+			$Element['attributes']['class'] = $this->table_class;
+		}
+
+		return parent::element( $Element );
+	}
+
+	/**
 	 * Convert a field that supports multiple lines form markdown to html.
 	 * @param string $p_text The text to convert.
 	 * @return string  The html text.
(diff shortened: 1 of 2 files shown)


grummbeer:
@atrol can you have a look at this? No clue why the tests are failing. 

```
SoapFault: SOAP-ERROR: Parsing WSDL: Couldn't load from 'http://localhost:8080/api/soap/mantisconnect.php?wsdl' : Extra content at the end of the document
```
It is not clear to me what is causing this.

dregad:
@grummbeer I apologise for taking so long to follow up on this.

> No clue why the tests are failing.
> 
> ```
> SoapFault: SOAP-ERROR: Parsing WSDL: Couldn't load from 'http://localhost:8080/api/soap/mantisconnect.php?wsdl' : Extra content at the end of the document
> ```
> 
> It is not clear to me what is causing this.

Opening http://localhost/api/soap/mantisconnect.php?wsdl in the browser, displays an error page instead of the expected WSDL:

_This page contains the following errors:
error on line 2 at column 1: Extra content at the end of the document
Below is a rendering of the page up to the first error._

The HTML source of the page reveals the actual root cause:
```html
<br/>
<b>Warning</b>
:  Declaration of MantisMarkdown::element($Element) should be compatible with Parsedown::element(array $Element) in 
<b>/Users/dregad/dev/mantisbt/mantis/plugins/MantisCoreFormatting/core/MantisMarkdown.php</b>
 on line 
<b>42</b>
<br/>
<?xml version="1.0" encoding="ISO-8859-1"?>
<?xml-stylesheet type="text/xsl" href="wsdl-viewer.xsl"?>
```

Changing the method's signature to match Parsedown's fixes the problem.

dregad (approved):

grummbeer:
Thank you, will try to be more accurate in the future.

dregad:
No problem. Thank *you* for your contribution, looking forward for the next. And again, sorry for taking so long to answer.

grummbeer:
> looking forward for the next.

https://github.com/mantisbt/mantisbt/pull/1929

PR #1850, opened 2022-09-12
Exclude adodb-php v5.22.3, it is faulty

starlocke:
See https://github.com/ADOdb/ADOdb/issues/867


dregad:
As mentioned in https://mantisbt.org/bugs/view.php?id=30992#c66987 ADOdb 5.22 is currently not supported with MantisBT. ADOdb 5.22.4 will be released shortly, making this update pointless.

PR #1853, opened 2022-09-24
Implementing the Status field of a category

CasN:
The status field in the Category table is present for quite some time but not being used. This addresses [bug 31017](https://mantisbt.org/bugs/view.php?id=31017). Categories can be de-activated meaning that they will no longer be available when creating or updating an issue. Existing issues will keep the de-activated category.


CasN:
@vboctor, another feature from my side. What is to be expected here?

bug_report_page.php
```diff
@@ -281,7 +281,7 @@
 				echo $t_allow_no_category ? '' : ' required';
 			?>>
 				<?php
-					print_category_option_list( $f_category_id );
+					print_category_option_list( $f_category_id, null, true);
```
vboctor, line 284: Missing space before `)` and I would add hints of parameters that are not clear:

```
print_category_option_list( $f_category_id, /* project_id */ null, /* active_only */ true );
```

bug_update_page.php
```diff
@@ -237,7 +237,7 @@
 		echo '<select ' . helper_get_tab_index()
 			. ( $t_allow_no_category ? '' : ' required' )
 			. ' id="category_id" name="category_id" class="input-sm">';
-		print_category_option_list( $t_bug->category_id, $t_bug->project_id );
+		print_category_option_list( $t_bug->category_id, $t_bug->project_id, true );
```
vboctor, line 240: same here `/* active_only */ true`

core/category_api.php
```diff
@@ -497,7 +497,7 @@ function category_get_filter_list( $p_project_id = null ) {
  * @return array array of categories
  * @access public
  */
-function category_get_all_rows( $p_project_id, $p_inherit = null, $p_sort_by_project = false ) {
+function category_get_all_rows( $p_project_id, $p_inherit = null, $p_sort_by_project = false ,  $p_active_only = false) {
```
vboctor, line 500: Update phpdoc with new parameter added.

core/category_api.php
```diff
@@ -535,6 +535,10 @@ function category_get_all_rows( $p_project_id, $p_inherit = null, $p_sort_by_pro
 		$t_project_where = ' project_id=' . $p_project_id . ' ';
 	}
 
+	if( $p_active_only ) {
+		$t_project_where .= ' and c.status = 0 ';
```
vboctor, line 539: Use constants instead of magic numbers - i.e. replace 0 with a constant.

core/print_api.php
```diff
@@ -752,14 +752,15 @@ function print_profile_option_list_from_profiles( array $p_profiles, $p_select_i
  * @param integer $p_project_id  A project identifier.
  * @return void
  */
-function print_category_option_list( $p_category_id = 0, $p_project_id = null ) {
+function print_category_option_list( $p_category_id = 0, $p_project_id = null, $p_active_only = false ) {
```
vboctor, line 755: Update phpdoc config with new parameter.

core/print_api.php
```diff
@@ -752,14 +752,15 @@ function print_profile_option_list_from_profiles( array $p_profiles, $p_select_i
...
 	if( null === $p_project_id ) {
 		$t_project_id = helper_get_current_project();
 	} else {
 		$t_project_id = $p_project_id;
 	}
 
-	$t_cat_arr = category_get_all_rows( $t_project_id, null, true );
+	$t_active_only = $p_active_only;
```
vboctor, line 762: No need for the temporary variable here.

core/print_api.php
```diff
@@ -752,14 +752,15 @@ function print_profile_option_list_from_profiles( array $p_profiles, $p_select_i
...
+	$t_cat_arr = category_get_all_rows( $t_project_id, null, true, $t_active_only);
```
vboctor, line 763: Missing space before `)`

manage_proj_cat_update.php
```diff
@@ -56,6 +56,7 @@
 $f_category_id		= gpc_get_int( 'category_id' );
 $f_name				= trim( gpc_get_string( 'name' ) );
 $f_assigned_to		= gpc_get_int( 'assigned_to', 0 );
+$f_status		= gpc_get_int( 'status', 0 );
```
vboctor, line 59: Use constant for default. Rename form / php variable to `cartegory_status` and `$f_category_status`.

manage_proj_edit_page.php
```diff
@@ -462,6 +462,7 @@ class="form-inline pull-left padding-right-8">
 		<thead>
 			<tr>
 				<th><?php echo lang_get( 'category' ) ?></th>
+				<td><?php echo lang_get( 'assign_to' ) ?></td>
```
vboctor, line 465: Fix column title.

manage_proj_edit_page.php
```diff
@@ -472,9 +473,15 @@ class="form-inline pull-left padding-right-8">
 		foreach ( $t_categories as $t_category ) {
 			$t_id = $t_category['id'];
 			$t_inherited = ( $t_category['project_id'] != $f_project_id );
+			if( $t_category['status'] == 0){
```
vboctor, line 476: Use constant and add space before `)`

manage_proj_edit_page.php
```diff
@@ -472,9 +473,15 @@ class="form-inline pull-left padding-right-8">
...
+				$cat_status = lang_get( 'category_active' ) ;
```
vboctor, line 477: Rename to `$t_category_status`

manage_proj_page.php
```diff
@@ -244,6 +244,7 @@
 		<thead>
 			<tr>
 				<td><?php echo lang_get( 'category' ) ?></td>
+				<td><?php echo lang_get( 'status' ) ?></td>
```
vboctor, line 247: I would define a new string called `category_status`.

manage_proj_page.php
```diff
@@ -255,9 +256,15 @@
 <?php
 			foreach( $t_categories as $t_category ) {
 				$t_id = $t_category['id'];
+				if ( $t_category['status'] == 0){
```
vboctor, line 259: Use constant and add space before `)`

manage_proj_page.php
```diff
@@ -255,9 +256,15 @@
...
+					$cat_status = lang_get( 'category_active' ) ;
+				} else {
+					$cat_status = lang_get( 'category_not_active' ) ;
+				}
 ?>
 			<tr>
 				<td><?php echo string_display_line( category_full_name( $t_id, false ) )  ?></td>
+				<td><?php echo $cat_status  ?></td>
```
vboctor, line 267: Remove extra space.

manage_proj_cat_edit_page.php
```diff
@@ -94,6 +95,19 @@
 					<input type="text" id="proj-category-name" name="name" class="input-sm" size="32" maxlength="128" value="<?php echo string_attribute( $t_name ) ?>" />
 				</td>
 			</tr>
+			<tr>
+				<td class="category">
+					<?php echo lang_get( 'category_status' ) ?>
+				</td>
+				<td class="center" width="20%">
+					<label><input type="radio" name="status" value="0" <?php echo ( $t_status == 0 ) ? 'checked="checked" ' : ''?>/>
```
vboctor, line 103: Use constant `CATEGORY_STATUS_ENABLED` and `CATEGORY_STATUS_DISABLED`.

vboctor (changes requested):
It would be useful to include screenshots of such changes.

A couple of aspects that come to mind:

1. Should we copy or skip disabled categories in the "copy categories" feature.
2. Should we model category status as an enum? For example, issue view state is modeled as a enum or just two specific options.

I would put these points up for discussion.

I have added inline comments with lower level feedback.

CasN:
[Mantisbt Category activation.docx](https://github.com/mantisbt/mantisbt/files/10049271/Mantisbt.Category.activation.docx)
Attached a document showing how it looks once implemented. Question to use ENUM or the 2 simple options as currently uses requires a decision on the way forward. Both options are present currently. Functional there is no difference at all.
With respect to the copying of categories that is an option if needed, smalll thing to add if we all agree on this.

CasN:
Copying de-activated categories could be a configurable setting like $g_exclude_deactivated_categories which in such case should default to TRUE.

docbook/Admin_Guide/en-US/config/misc.xml
```diff
@@ -287,5 +287,12 @@
 				</warning>
 			</listitem>
 		</varlistentry>
+			<varlistentry>
```
atrol, line 290: Confuses me, what do you want to add in rge Admin Guide?
$p_active_only is not a configuration option.

core/category_api.php
atrol, line 500: Missing blank after `false`

lang/strings_english.txt
```diff
@@ -145,6 +145,9 @@ $s_inherit = 'Inherit Categories';
 $s_inherit_global = 'Inherit Global Categories';
 $s_inherit_parent = 'Inherit Parent Categories';
 $s_update_subproject_inheritance = 'Update Subproject Inheritance';
+$s_category_status = 'Category Status';
+$s_category_active = ' Active';
```
atrol, line 149: Remove leading blank

manage_proj_cat_edit_page.php
```diff
@@ -94,6 +95,19 @@
 					<input type="text" id="proj-category-name" name="name" class="input-sm" size="32" maxlength="128" value="<?php echo string_attribute( $t_name ) ?>" />
 				</td>
 			</tr>
+			<tr>
+				<td class="category">
+					<?php echo lang_get( 'category_status' ) ?>
+				</td>
+				<td class="center" width="20%">
+					<label><input type="radio" name="category_status" value="0" <?php echo ( $t_category_status == OFF ) ? 'checked="checked" ' : ''?>/>
```
atrol, line 103: Wouldn't it be better to use a checkbox?

manage_proj_edit_page.php
```diff
@@ -472,9 +473,15 @@ class="form-inline pull-left padding-right-8">
 		foreach ( $t_categories as $t_category ) {
 			$t_id = $t_category['id'];
 			$t_inherited = ( $t_category['project_id'] != $f_project_id );
+			if( $t_category['status'] == OFF ){
+				$category_status = lang_get( 'category_active' ) ;
```
atrol, line 477: Should be `$t_category_status`

manage_proj_page.php
```diff
@@ -255,9 +256,15 @@
 <?php
 			foreach( $t_categories as $t_category ) {
 				$t_id = $t_category['id'];
+				if ( $t_category['status'] == OFF){
+					$cat_status = lang_get( 'category_active' ) ;
```
atrol, line 260: Should be `$t_category_status` (same name like used in manage_proj_edit_page.php)

CasN: I have addressed all your comments in my latest commit.

atrol: @CasN please test your changes before pushing. There are obvious syntax errors.

atrol:
@CasN again, please test your changes before pushing.
The code does not work, e. g. you can't edit a category , you can't report an issue (getting database error), ...

Also please ensure that you follow our coding guidelines https://www.mantisbt.org/wiki/doku.php/mantisbt:coding_guidelines

core/category_api.php
```diff
@@ -494,10 +494,11 @@ function category_get_filter_list( $p_project_id = null ) {
  * @param integer $p_project_id      A Project identifier.
  * @param boolean $p_inherit         Indicates whether to inherit categories from parent projects, or null to use configuration default.
  * @param boolean $p_sort_by_project Whether to sort by project.
+ * @param boolean $p_active_only	 Wheterto select all categories or just the actie ones
```
atrol, line 497: Typos in comment

core/category_api.php
```diff
@@ -535,6 +536,10 @@ function category_get_all_rows( $p_project_id, $p_inherit = null, $p_sort_by_pro
 		$t_project_where = ' project_id=' . $p_project_id . ' ';
 	}
 
+	if( $p_active_only ) {
+		$t_project_where .= ' and c.status = OFF ';
```
atrol, line 540: Can't work, will raise a database errot

core/print_api.php
atrol: Still missing PHPDoc

lang/strings_english.txt
```diff
@@ -145,6 +145,9 @@ $s_inherit = 'Inherit Categories';
 $s_inherit_global = 'Inherit Global Categories';
 $s_inherit_parent = 'Inherit Parent Categories';
 $s_update_subproject_inheritance = 'Update Subproject Inheritance';
+$s_category_status = 'Category Status';
+$s_category_active = 'Active';
+$s_category_not_active =  'Not Active';
```
atrol, line 150: Remove one space after `=`

manage_proj_cat_edit_page.php
```diff
@@ -94,6 +95,16 @@
 					<input type="text" id="proj-category-name" name="name" class="input-sm" size="32" maxlength="128" value="<?php echo string_attribute( $t_name ) ?>" />
 				</td>
 			</tr>
+			<tr>
+				<td class="category">
+					<?php echo lang_get( 'category_status' ) ?>
+				</td>
+				<td class="center" width="20%">
+					<label><input type="checkbox" name='category_status'  value="<?php echo $t_category_status ?>" />
+					<?php if ( $t_category_status == CATEGORY_STATUS_ENABLED ) { echo "checked='checked'"; } ?>>
```
atrol, line 104: Remove trailing `>`

manage_proj_cat_update.php
```diff
@@ -56,6 +56,7 @@
 $f_category_id		= gpc_get_int( 'category_id' );
 $f_name				= trim( gpc_get_string( 'name' ) );
 $f_assigned_to		= gpc_get_int( 'assigned_to', 0 );
+$f_category_status	= gpc_get_int( 'category_status', OFF );
```
atrol, line 59: Use new constant instead of OFF

manage_proj_page.php
```diff
@@ -255,9 +256,15 @@
 <?php
 			foreach( $t_categories as $t_category ) {
 				$t_id = $t_category['id'];
+				if ( $t_category['status'] == OFF ){
```
atrol, line 259: No space after `if`
https://www.mantisbt.org/wiki/doku.php/mantisbt:coding_guidelines#braces_and_parentheses

phpcheck.cmd
```diff
@@ -0,0 +1,3 @@
+For /R d:\data\Github\mantisbt\ %%G IN (*.php) do (
```
atrol, line 1: File should be removed from PR

results.txt
```diff
@@ -0,0 +1,1433 @@
+
```
atrol, line 1: File should be removed from PR

CasN:
Incorporated all comments in another commit, will squash them into one PR once ok.
Have tested this in a dedicated setup and found no issues.
Did swap the checkbox back to radio button for readability.

CasN:
@atrol, thanks for the patience. Have a nice farewell of 2022 and a great 2023!

core/print_api.php
CasN: Which file needs to be updated with the parameter description?

dregad:
I totally missed that PR for some reason...

Here's a quick review after reading the comments, but without actually testing or closely looking at the code. Note: I have marked as resolved all the outdated review notes, and did not consider them in my review.

Overall, I really like the idea, it makes perfect sense and it's definitely worth implementing. 

Thanks @CasN for your contribution.

From https://github.com/mantisbt/mantisbt/pull/1853#pullrequestreview-1173260635 by @vboctor
>> Should we copy or skip disabled categories in the "copy categories" feature.
>
> Copying de-activated categories could be a configurable setting like $g_exclude_deactivated_categories which in such case should default to TRUE.

To be honest, I don't think this is worth adding yet another config option. 

If a category is disabled then don't copy it.

>> Should we model category status as an enum? For example, issue view state is modeled as a enum or just two specific options
>
> Question to use ENUM or the 2 simple options as currently uses requires a decision on the way forward. Both options are present currently. Functional there is no difference at all.

IMO, adding an Enum for this just unnecessary overhead. At this time, this is just a boolean flag, just like _enabled_ and _protected_ in the User table, and I can't think of any use case requiring other possible values.

I would suggest to treat the category status as a boolean flag for now, and we can always change the implementation to an Enum later on if needed.

From https://github.com/mantisbt/mantisbt/pull/1853#issuecomment-1368199727
> Did swap the checkbox back to radio button for readability.

Why ? Again, no benefit from radio button here, just overhead in code (language strings, etc) - a checkbox is simple to use and implement.  Look at manage_user_page.php and manage_user_edit_page.php.

https://github.com/mantisbt/mantisbt/pull/1853#discussion_r1017408487
> Use constants instead of magic numbers - i.e. replace 0 with a constant

Considering my above comment, I don't think that's necessary. Just make that a query parameter and use `false` instead, like
https://github.com/mantisbt/mantisbt/blob/bf7a3c22eab2f7546b23299ac28cb73b279c3541/manage_user_page.php#L259-L260

CasN:
@dregad Do you want me to reverse the changes made last time, using constants which was requested by Victor?
Same question for the radiobutton.
The other 2 remarks have not been implemented so no action required there.

dregad:
I finally found the time to have a closer look a this.

For starters,I find it somewhat confusing that the **terminology for the category status is not consistent**. In some cases it is *Enabled/Disabled* and others usage refer to *Active/Not active*. 

We need to decide on what terms to use, and stick to that.

Much more problematic is the fact that **the constants' meaning and the actual behavior observed in the application are reversed**, i.e. CATEGORY_STATUS_DISABLED = 0, but a category with status = 0 is actually active.

https://github.com/mantisbt/mantisbt/blob/e65a2015e6d7ec514ec6771dd5b2ef1251cea47d/core/constant_inc.php#L37-L38

```
mysql> select * from mantis_category_table;
+----+------------+---------+---------------+--------+
| id | project_id | user_id | name          | status |
+----+------------+---------+---------------+--------+
|  1 |          0 |       0 | General       |      0 |
| 97 |          0 |       0 | test          |      0 |
| 98 |          0 |       0 | test disabled |      1 |
+----+------------+---------+---------------+--------+
```

image

It makes complete sense to have 0 == DISABLED as it's a falsey value, allowing the field to behave as a boolean. 

Problem is, the field's default value is 0, meaning all existing categories will be considered disabled; so we need a schema update to set the field to 1, and change the column's default value too.

I have a bunch of other, low-level comments and minor code adjustments, but I guess it will be easier and faster if I just fix them myself. @cas Is it OK if I squash all your commits, rebase to latest master and force-push your branch ?

dregad (changes requested):
See above comment

CasN:
I do recognize the issue with Status = 0 = Active. Indeed a schema update will be able to fix this.
Clearly i have no issue if you make the final adjusts.

dregad:
This should fix the issues mentioned in my review, but it's not fully tested yet.

There is one problem that I noticed already, is if someone edits an issue assigned to a disabled category, then it is defaulted to the first available one from the list, and silently updated upon save. This is not good.

dregad:
> if someone edits an issue assigned to a disabled category, then it is defaulted to the first available one from the list, and silently updated upon save

That last commit should address this bug.

Testing and feedback welcome.

dregad (changes requested):
CI tests need fixing

There was 1 error:
1) CategoryTest::testAddRenameDeleteCategory
SoapFault: Too few arguments to function category_update(), 3 passed in /home/travis/build/mantisbt/mantisbt/api/soap/mc_project_api.php on line 332 and exactly 4 expected
/home/travis/build/mantisbt/mantisbt/tests/soap/CategoryTest.php:73

dregad (approved):
Fixed the failing tests, by modifying category_update() signature, $p_status is now optional and defaults to a backwards-compatible behavior of not updating the status.

CasN:
> > if someone edits an issue assigned to a disabled category, then it is defaulted to the first available one from the list, and silently updated upon save
> 
> That last commit should address this bug.
> 
> Testing and feedback welcome.

This is also a neat way to handle this but downside is that the disabled category remains in place. I had another solution (not yet pushed) which adds a few lines of code after "event_signal( 'EVENT_UPDATE_BUG_FORM_TOP', array( $t_bug_id ) );" and this would raise a warning and blanks the category_id. Code is along these lines:
if ( category_get_field ( $t_bug->category_id , 'status') ) {
	echo "<b>". lang_get('category_warning1') . category_get_field ($t_bug->category_id , 'name') . lang_get('category_warning2') . "</b>";
	$t_bug->category_id = 0;
}
$s_category_warning1 = 'Original category (';
$s_category_warning2 = ') no longer available, please select new category';
Nevertheless I can perfectly live with your fix.

dregad:
> downside is that the disabled category remains in place

That is true, but keep in mind this is not a normal situation.
And changing data without conscious user action is not good practice.

From my perspective, the proper way to handle this situation would be after the admin disables a category, they should search for issues assigned to it and update them.

CasN:
As for remark 2, yes that would be a good idea.
Introducing another setting ($g_default_category_for_moves) in my view is over the the top. Solution proposed by @dregad really takes care of the possible issue.

atrol:
>  Introducing another setting ($g_default_category_for_moves) in my view is over the the top. 

This is not about introducing a new option. It's about considering an existing option.

CasN:
@atrol, you are right, forgot about that one (which i came across in the past)

dregad:
@atrol thanks for your feedback

> $g_default_category_for_moves should be considered (don't allow to deactivate this category).

Good point, I'll make necessary adjustments.

> It should be possible to get/set the category status via REST/SOAP API.

I agree this is something that should be done, but considering that at the moment there are no REST API endpoints to specifically deal with categories, IMO developing them should be tracked as a separate work item (i.e. open a new issue in the tracker and fix that with another PR).

I am not so familiar with SOAP and hardly ever used our API except for some tests, so unsure if adding a new parameter to the existing endpoints would cause any regressions to existing clients. If it does, then a new endpoint may be needed, just to enable/disable a category.

atrol:
> > It should be possible to get/set the category status via REST/SOAP API.
> 
> I agree this is something that should be done, but considering that at the moment there are no REST API endpoints to specifically deal with categories, IMO developing them should be tracked as a separate work item (i.e. open a new issue in the tracker and fix that with another PR). EDIT: actually this has already been submitted, see [#32470](https://mantisbt.org/bugs/view.php?id=32470).
>

@dregad this is about Add/Update/Delete

To get the category status as part of the get response for a project is pretty easy.
Add a line for the status to function mci_project_categories

```
		$t_result = array(
			'id' => (int)$t_category['id'],
			'name' => $t_category['name'],
			'status' => $t_category['status'],
			'project' => array( 'id' => $t_project_id, 'name' => $t_category['project_name'] ),
		);
```

dregad:
> > > It should be possible to get/set the category status via REST/SOAP API.
> > 
> > [...] REST API endpoints to specifically deal with categories [...]  should be tracked as a separate work item [...] see [#32470](https://mantisbt.org/bugs/view.php?id=32470).
> 
> @dregad this is about Add/Update/Delete

Yes indeed.
I was responding to your request to be able *to get/**set** the category status*

But if you only want to add a new field in GET project endpoint, indeed it's a minor change that can be implemented easily, and I think the risk of regression for this is quite low.

dregad:
> > $g_default_category_for_moves should be considered (don't allow to deactivate this category).

Done c63cb7720fc452f028b5cb55e38aba087e5e7610

> > It should be possible to get/set the category status via REST/SOAP API.

Done 9112887a37c9e34de95e8a2e2f962fafd8d68e9e

dregad:
Rebased on top of master to resolve conflicts and prepare for merge.

PR #1859, opened 2022-12-09
#31666 Hook for Custom field on bug_change_status_page added

Selonka:
Hook for Custom field on bug_change_status_page added

[#31666](https://www.mantisbt.org/bugs/view.php?id=31666)


dregad (changes requested):
I don't object to introducing a new event, but it would be useful to have a bit of background on why it is needed and how it is expected to be used.

The new event needs to be properly documented in [Event Reference](https://www.mantisbt.org/docs/master/en-US/Developers_Guide/html-desktop/#dev.eventref).

In terms of naming, the *DISPLAY* group of events is more about processing data for display; this new event as I understand it is about showing or hiding a field on a form, so IMO it would be more appropriate to use `EVENT_UPDATE_BUG` prefix.

I am not sure about the logic of signalling the event before the "native" custom field display check. Being able override the default behaviour makes sense, but I'm wondering if letting the core processing take place first, then fire the event with the custom field's display/hide status, allowing a plugin to change it as needed wouldn't be better. Thoughts ?

Selonka:
Hi,

I need the event to show different custom fields on the form when switching between none native states and custom field is not generally set to required. (See https://github.com/Selonka/Mantis_Show_Custom_Name_Plugin as an Example)

The event would probably become unnecessary with ticket #5650. But since I don't have the experience or time to solve the ticket I thought of the event as a workaround until the ticket is eventually resolved. 

I changed the event name and also the logic. Now the core processing takes place followed by the Event.

The Event only allows you to show fields, not hide them because hiding a required field my lead to errors.

I will add the documentation when everything is approved.

core/events_inc.php
```diff
@@ -54,6 +54,7 @@
 	'EVENT_DISPLAY_RSS' => EVENT_TYPE_CHAIN,
 	'EVENT_DISPLAY_EMAIL' => EVENT_TYPE_CHAIN,
 	'EVENT_DISPLAY_EMAIL_BUILD_SUBJECT' => EVENT_TYPE_CHAIN,
+
```
dregad, line 57: Extra newline should be removed

dregad (changes requested):
Thanks for the update. Event name looks OK now, and I think it's better with the reversed logic.

Your code is still not following our guidelines, please fix (in particular, space after `(` and before `)`)

> I will add the documentation when everything is approved.

Documentation is an integral part adding an event, it needs to be submitted as part of the PR before it can be merged.

dregad:
@Selonka just a friendly reminder that we are waiting for you to add event documentation and fix the coding guidelines before we merge your PR.

Selonka:
Hope everything is fine now...sry for the delay.

dregad:
> Hope everything is fine now

Unfortunately you re-introduced space indentation. I'll fix that. 
The rest looks OK at first glance, I'll check more in-depth later.

Selonka:
Thx

dregad:
Sorry for the delay and thanks for your contribution.

PR #1860, opened 2022-12-13
Fix syntax error in Graphviz output

atrol:
Fixes [#31827](https://www.mantisbt.org/bugs/view.php?id=31827)


vboctor (approved):

dregad (approved):
Funny I never actually noticed this warning, despite having used relationship graphs for years at my previous company... good catch.

PR #1863, opened 2023-01-03
PHP 8 - fix for "Unknown named parameter" when params is associative …

sederther:


dregad:
@sederther Thanks for your contribution. Please [open an issue](https://www.mantisbt.org/bugs/bug_report_page.php) at mantisbt.org to track this, and reference it in your commit message (`Fixes #xxx`).

It would be useful to provide some context (which plugin and event hook is being called, including function signature) and details about the error itself i.e. where does it occur and how (stack trace), because I've never faced such error despite running PHP 8 for several months.

core/event_api.php
```diff
@@ -203,13 +203,15 @@ function event_callback( $p_event, $p_callback, $p_plugin, $p_params = null ) {
...
+  			$function_params = array_merge( [ $p_event ], $p_params );
+			$t_value = call_user_func_array( array( $g_plugin_cache[$p_plugin], $p_callback ), array_values($function_params) );
 		}
 
 		plugin_pop_current();
 	} else {
 		if( function_exists( $p_callback ) ) {
-			$t_value = call_user_func_array( $p_callback, array_merge( array( $p_event ), $p_params ) );
+			$function_params = array_merge( [ $p_event ], $p_params );
+			$t_value = call_user_func_array( $p_callback, array_values($function_params) );
```
dregad, line 214: Coding guidelines (space after `(` / before `)`

sederther:
Issue with my plugin.
https://mantisbt.org/bugs/view.php?id=31892

PR #1864, opened 2023-01-06
Improve conversion of date strings to unix timestamps

dregad:
Fixes [#31836](https://www.mantisbt.org/bugs/view.php?id=31836): When editing a project version, if the date format is set for a non-US format, the strtotime will fail on date conversion.

Move conversion logic to date_strtotime() API.

Use modified API function where relevant, i.e. when date string could be user-provided.

@vboctor When this gets merged, the REST API documentation should be updated to indicate that dates can be submitted in $g_normal_date_format in addition to formats recognised by strtotime(), see https://documenter.getpostman.com/view/29959/7Lt6zkP#d9c16d5e-cae9-20e6-2b46-556c68c451cc but I cannot do this myself.

### DIFF
--- a/core/commands/VersionAddCommand.php
+++ b/core/commands/VersionAddCommand.php
@@ -25,6 +25,11 @@ use Mantis\Exceptions\ClientException;
  */
 class VersionAddCommand extends Command {
 	/**
+	 * @var int|null $timestamp
+	 */
+	protected $timestamp;
+
+	/**
 	 * $p_data['query'] is expected to contain:
 	 * - project_id (integer)
 	 *
@@ -43,8 +48,9 @@ class VersionAddCommand extends Command {
 
 	/**
 	 * Validate the data.
+	 * @throws ClientException
 	 */
-	function validate() {		
+	function validate() {
 		$t_project_id = helper_parse_id( $this->query( 'project_id' ), 'project_id' );
 
 		if( !access_has_project_level( config_get( 'manage_project_threshold' ), $t_project_id ) ) {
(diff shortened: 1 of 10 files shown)


core/commands/VersionAddCommand.php
```diff
@@ -55,6 +61,9 @@ function validate() {
 		if( is_blank( $t_name ) ) {
 			throw new ClientException( 'Invalid version name', ERROR_EMPTY_FIELD, array( 'name' ) );
 		}
+
+		$t_timestamp = $this->payload( 'timestamp', '' );
+		$this->timestamp = is_blank( $t_timestamp ) ? null : date_strtotime( $t_timestamp );
```
atrol, line 66: Maybe this is the wrong place to deal with date format.
Does this mean that the payload of a REST client must be prepared depending on  configuration option `normal_date_format`?

dregad: > Does this mean that the payload of a REST client must be prepared depending on configuration option normal_date_format?

No. 

The API will accept dates in _normal_date_format_ **in addition** to the standard formats accepted by strtotime

vboctor:
Let's not rush this in.

I would rather have the API take less formats than more. If designing from scratch, I would use a strongly typed ISO format.

dregad:
I would agree with only accepting ISO format argument, but we're not designing from scratch.

I think it makes perfect sense to allow dates in the MantisBT-defined format in addition to standard ISO dates as well as all the US-centric formats accepted by strtotime, can you explain why you are opposed to this ? It does not hurt and would not cause any compatibility issues.

atrol:
Example with wrong behavior:

$g_normal_date_format is set to 'Y-m-d H:i', payload is sent as 'Y-d-m H:i'.
The conversion will succeed using $g_normal_date_format in cases where the sent day is <= 12.

dregad:
I removed changes to REST/SOAP API. Testing and feedback welcome.

dregad:
Without any feedback on this for over a week, I assume that everyone is fine with the proposed changes, and will proceed with merging the PR.

PR #1865, opened 2023-01-10
Fix unreleased version date not showing in selection lists

dregad:
Fixes [#31889](https://www.mantisbt.org/bugs/view.php?id=31889)

### DIFF
--- a/core/prepare_api.php
+++ b/core/prepare_api.php
@@ -171,11 +171,10 @@ function prepare_version_string( $p_project_id, $p_version_id, $p_show_project =
 	$t_version_text = version_full_name( $p_version_id, $p_show_project, $p_project_id );
 
 	if( access_has_project_level( config_get( 'show_version_dates_threshold' ), $p_project_id ) ) {
-		$t_short_date_format = config_get( 'short_date_format' );
-
-		$t_version = version_cache_row( $p_version_id );
-		if( 1 == $t_version['released'] ) {
-			$t_version_text .= ' (' . date( $t_short_date_format, $t_version['date_order'] ) . ')';
+		$t_date_order = version_get_field( $p_version_id, 'date_order' );
+		if( $t_date_order != date_get_null() ) {
+			$t_short_date_format = config_get( 'short_date_format' );
+			$t_version_text .= ' (' . date( $t_short_date_format, $t_date_order ) . ')';
 		}
 	}
 
--- a/core/prepare_api.php
+++ b/core/prepare_api.php
@@ -170,10 +170,11 @@ function prepare_version_string( $p_project_id, $p_version_id, $p_show_project =
 
 	$t_version_text = version_full_name( $p_version_id, $p_show_project, $p_project_id );
 
-	if( access_has_project_level( config_get( 'show_version_dates_threshold' ), $p_project_id ) ) {
+    $t_show_version_dates = config_get('show_version_dates_threshold', null, null, $p_project_id);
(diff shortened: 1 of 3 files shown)


core/prepare_api.php
```diff
@@ -171,11 +171,10 @@ function prepare_version_string( $p_project_id, $p_version_id, $p_show_project =
...
 
 	if( access_has_project_level( config_get( 'show_version_dates_threshold' ), $p_project_id ) ) {
-		$t_short_date_format = config_get( 'short_date_format' );
-
-		$t_version = version_cache_row( $p_version_id );
-		if( 1 == $t_version['released'] ) {
-			$t_version_text .= ' (' . date( $t_short_date_format, $t_version['date_order'] ) . ')';
+		$t_date_order = version_get_field( $p_version_id, 'date_order' );
+		if( $t_date_order != date_get_null() ) {
+			$t_short_date_format = config_get( 'short_date_format' );
```
vboctor, line 176: While you are here, you may want to provide the `$p_project_id` to config_get() calls in the access check and the fetching of the `short_date_format`.

vboctor (approved):
It makes sense to revert to the original behavior.

I noticed that when date is not provided, the release is pushed to the end of the list, rather than being at the stop as an upcoming release. I also noticed that very old versions of MantisBT doesn't have the date set, so changing the sort behavior will likely surface older release at the top, though it may be a cleanup task to set the dates. This is out of scope of this specific targeted fix, but something for us to think about.

core/prepare_api.php
dregad: OK

dregad:
@vboctor there is another occurrence of this issue on bug_view_page.php, that should also be reverted for consistency.

Here the regression was actually introduced with IssueViewPageCommand (see 2b163046849d3445894537f7a7b4762edd41ff8b), when you [removed the *prepare_version_string()* calls](https://github.com/mantisbt/mantisbt/commit/2b163046849d3445894537f7a7b4762edd41ff8b#diff-f5e42df6e73add5ec18ec7f88f1662ef618b1ab3e9f266771fa9c5a2e0482573L128) and [only displayed the version name](https://github.com/mantisbt/mantisbt/commit/2b163046849d3445894537f7a7b4762edd41ff8b#diff-f5e42df6e73add5ec18ec7f88f1662ef618b1ab3e9f266771fa9c5a2e0482573R467), as returned by the Command.

This data is generated by mci_get_version(), and currently only has id and name, e.g.: `"version": {"id": 1, "name": "1.0"}`.

I'm not sure how to best go about fixing this, here are some options that I considered
- The simplest would be to **update *bug_view_inc.php* to call prepare_version_string()** again, using the issue and project id returned by the Command
- Modify the **command to return a formatted *version (date)* string** in the $t_issue_view array 
- **Include date_order in returned version data**, but then for consistency I guess this should be ISO date format, which would require some post-processing to display in expected Mantis $g_short_date_format

What do you think is best ? Any other, better suggestions ?

dregad:
@vboctor your feedback would be appreciated.

vboctor:
The pattern that is already in place is to include such formatted fields in issue view. Examples include issue_view['created_at'], issue_view['id_formatted'], etc. The same can be done for the version attributes. Then use these new fields in the bug_view_inc.php page.

dregad:
Great, thanks for the information, I'll do that.

dregad:
I finally got around to implementing this...

dregad:
Without any feedback on this for over a week, I assume that everyone is fine with the proposed changes, and will proceed with merging the PR.

PR #1869, opened 2023-03-01
Fix error when displaying Issue without Version

dregad:
Commit d9464fe8132f6cfd694625c8d050575aed95d5ba (PR #1865) introduced a regression when displaying an Issue without Product, Target or Fixed In Version:

APPLICATION ERROR 1601 - Version "" not found
PHP Notice (or Warning on 8.1+): Undefined array key "version"

This is because mci_issue_data_as_array() removes null keys and empty arrays from the Issue data, so the IssueViewPageCommand needs to check for array key existence before calling prepared_version_string().

Fixes [#32086](https://mantisbt.org/bugs/view.php?id=32086), [#31889](https://mantisbt.org/bugs/view.php?id=31889)

### DIFF
--- a/core/commands/IssueViewPageCommand.php
+++ b/core/commands/IssueViewPageCommand.php
@@ -131,18 +131,16 @@ class IssueViewPageCommand extends Command {
 			in_array( 'target_version', $t_fields ) &&
 			access_has_bug_level( config_get( 'roadmap_view_threshold' ), $t_issue_id );
 
-		$t_issue_view['product_version'] = prepare_version_string( $t_project_id,
-			$t_issue['version']['id'],
-			false
-		);
-		$t_issue_view['target_version'] = prepare_version_string( $t_project_id,
-			$t_issue['target_version']['id'],
-			false
-		);
-		$t_issue_view['fixed_in_version'] = prepare_version_string( $t_project_id,
-			$t_issue['fixed_in_version']['id'],
-			false
-		);
+		# Formatted version strings for display
+		$t_issue_view['product_version'] = isset( $t_issue['version'] )
+			? prepare_version_string( $t_project_id, $t_issue['version']['id'], false )
+			: '';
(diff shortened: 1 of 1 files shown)


dregad:
User testing [feedback](https://mantisbt.org/bugs/view.php?id=32086#c67436): 
> I've tested your patch successfully with 8.0.28. Test with php 8.1 will follow as soon as my environment is reconfigured for that again.

ProfCyberNaught:
Just wondering if this fix was going to be passed into the main download build on the website?

The current download website page displays 2.25.6: https://mantisbt.org/download.php

Nightly Builds Downloads: https://mantisbt.org/builds.php

Your requirements page clearly states support from 7.2 or later: https://mantisbt.org/docs/master/en-US/Admin_Guide/html-desktop/#admin.install.requirements.software.versions

Anybody downloading a latest version of PHP either directly or via such packages as XAMPP or LAMPP containing PHP8 for instance, would struggle to get Mantis to work. This could have negative impacts of gaining new users. Many will want to download the latest version of PHP especially when the first option presented on such sites as Apache Friends is PHP8.2 for Windows, Linux, and OS X: https://www.apachefriends.org/index.html

dregad:
@ProfCyberNaught If you see a _fixed_ Issue in _resolved_ state on our tracker, it means not yet released (Status changes to _closed_ when we cut the release). You can also see from the [Changelog](https://mantisbt.org/bugs/changelog_page.php?version_id=369) that 2.25.7 is not yet released. That will happen at some point but I currently have other, more pressing priorities. 

As for PHP 8 support, you make a valid point but again, I do this in my spare time and sadly I do not have the bandwidth to finalize a (long-overdue) 2.26.0 release at the moment. But it will happen eventually. 

MantisBT 2.25 should work fine on 8.0 but not 8.1+; 2.26.0-dev nightly build should work on 8.2, feel free to try and report any incompatibility issues you find on our tracker.

PR #1872, opened 2023-03-23
html_javascript_link function skip default directory if path specified

gatis-ozols:
html_javascript_link function now checks if path is specified and does not append js directory to JavaScript file path, Fixes [#32227](https://mantisbt.org/bugs/view.php?id=32227)


dregad:
Closed without merging, following discussion in [#32227](https://mantisbt.org/bugs/view.php?id=32227#c67583)

PR #1873, opened 2023-03-25
Project Add/Update/Delete commands for REST, SOAP, and Web UI

vboctor:
This PR fixes the following issues:

- [32231](https://www.mantisbt.org/bugs/view.php?id=32231): Create ProjectAddCommand
- [32238](https://www.mantisbt.org/bugs/view.php?id=32238): Create ProjectUpdateCommand
- [32232](https://www.mantisbt.org/bugs/view.php?id=32232): Create ProjectDeleteCommand
- [32237](https://www.mantisbt.org/bugs/view.php?id=32237): REST API Create Project API requires administrator rather than create_project_threshold
- [32236](https://www.mantisbt.org/bugs/view.php?id=32236): REST API Create Project doesn't trigger EVENT_MANAGE_PROJECT_CREATE plugin event
- [32235](https://www.mantisbt.org/bugs/view.php?id=32235): SOAP API Create Project doesn't trigger EVENT_MANAGE_PROJECT_CREATE plugin event
- [32234](https://www.mantisbt.org/bugs/view.php?id=32234): SOAP API Create Project API requires administrator rather than create_project_threshold
- [32248](https://www.mantisbt.org/bugs/view.php?id=32248): Get Project REST API returns html if user doesn't have access
- [32249](https://www.mantisbt.org/bugs/view.php?id=32249): Get Project Issues returns html if user doesn't have access to project

### DIFF
--- a/api/rest/restcore/issues_rest.php
+++ b/api/rest/restcore/issues_rest.php
@@ -104,25 +104,31 @@ function rest_issue_get( \Slim\Http\Request $p_request, \Slim\Http\Response $p_r
 
 		# Get a set of issues
 		$t_project_id = (int)$p_request->getParam( 'project_id', ALL_PROJECTS );
-		if( $t_project_id != ALL_PROJECTS && !project_exists( $t_project_id ) ) {
-			$t_result = null;
+		if( $t_project_id != ALL_PROJECTS ) {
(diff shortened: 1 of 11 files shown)


vboctor:
Removed trailing commas to improve PHP 7 compatibility.

PR #1875, opened 2023-03-26
REST API to create/delete API tokens + revoke API tokens when user is deleted

vboctor:
This PR fixes the following issues:

- [32245](https://www.mantisbt.org/bugs/view.php?id=32245): REST API for creating API tokens for users
- [32247](https://www.mantisbt.org/bugs/view.php?id=32247): REST API for deleting API token
- [32246](https://www.mantisbt.org/bugs/view.php?id=32246): Deleting a user should revoke (delete) all their API tokens

Scenarios:
- Admin that automates user creation + tokens for such users.
- Testing where an admin token can create token for a reporter, update, manager, etc and use these tokens for test cases.
- User wants to create automation to create token, leverage it for a purpose, and then delete it without disclosing their master long living token.


core/commands/UserTokenCreateCommand.php
```diff
@@ -0,0 +1,130 @@
...
+	 */
+	function validate() {
+		// acting user id
+		$t_current_user_id = auth_get_current_user_id();
+
+		// user is to create token for
+		$this->user_id = $this->query( 'user_id' );
+
+		// any user can create a token for themselves, but only an admin can create for other users.
+		if( $t_current_user_id != $this->user_id ) {
```
atrol, line 59: Maybe option `$g_impersonate_user_threshold` has also be considered.
Being able to create a token for another user means being able to act as another user in two ways
- creat a token in the name of the user
- use the token in long term to act as this user

api/rest/restcore/users_rest.php
```diff
@@ -81,6 +91,131 @@ function rest_user_create( \Slim\Http\Request $p_request, \Slim\Http\Response $p
...
+ *
+ * @param \Slim\Http\Request $p_request   The request.
+ * @param \Slim\Http\Response $p_response The response.
+ * @param integer $p_user_id The id of the user to create token for.
+ *
+ * @return \Slim\Http\Response The augmented response.
+ */
+function execute_delete_token_command( \Slim\Http\Request $p_request, \Slim\Http\Response $p_response, $p_user_id, $p_token_id ) {
+	// if body is empty or {} it will fail, this is acceptable for this API.
+	$t_payload = $p_request->getParsedBody();
```
atrol, line 201: `$t_payload` is not used in this function

core/commands/UserTokenCreateCommand.php
vboctor: Good suggestion. I switched over to use impersonation threshold instead. It makes more sense for the scenario. Thanks.

api/rest/restcore/users_rest.php
vboctor: Fixed.

PR #1876, opened 2023-03-27
REST API for Setting Configs

vboctor:
This PR fixes:

- [32258](https://www.mantisbt.org/bugs/view.php?id=32258): Add REST API for setting config options that are settable via database


vboctor:
- Updated all commit messages to reference the associated issue
- Removed trailing commas for better PHP 7 compatibility

PR #1880, opened 2023-04-10
REST testing framework + User test cases

vboctor:


vboctor:
I'll merge this and use it as the basis for User API changes and tests for other APIs.

PR #1881, opened 2023-04-10
REST API for Users Improvements

vboctor:
- Implemented UserGetCommand
- [27128](https://www.mantisbt.org/bugs/view.php?id=27128): Can not get userid from another user with REST API - Added `GET /users/username/:username` to get a user by name and `/users/:id` .
- [27133](https://www.mantisbt.org/bugs/view.php?id=27133) delete user by username - approach: get user info by username, then delete by id.
- [32357](https://www.mantisbt.org/bugs/view.php?id=32357): REST API: Support select for fields to return when getting user info - this applies to `/users/me`, `/users/:id`, `/users/username/:username` and include select query string with field names.

### DIFF
--- a/api/rest/restcore/users_rest.php
+++ b/api/rest/restcore/users_rest.php
@@ -29,7 +29,7 @@
  */
 $g_app->group('/users', function() use ( $g_app ) {
 	$g_app->get( '/me', 'rest_user_get_me' );
-	$g_app->get( '/{user_id}', 'rest_user_get' );
+	$g_app->get( '/{ref}', 'rest_user_get' );
 
 	$g_app->post( '/', 'rest_user_create' );
 	$g_app->post( '', 'rest_user_create' );
@@ -106,7 +106,7 @@ function rest_user_get_me( \Slim\Http\Request $p_request, \Slim\Http\Response $p
  * @return \Slim\Http\Response The augmented response.
  */
 function rest_user_get( \Slim\Http\Request $p_request, \Slim\Http\Response $p_response, array $p_args ) {
-	$t_user_id = $p_args['user_id'];
+	$t_user_ref = $p_args['ref'];
 
 	$t_select = $p_request->getParam( 'select', null );
 	if( is_null( $t_select ) ) {
@@ -126,7 +126,7 @@ function rest_user_get( \Slim\Http\Request $p_request, \Slim\Http\Response $p_re
 
 	$t_data = array(
 		'query' => array(
-			'id' => $t_user_id,
+			'ref' => $t_user_ref,
 			'select' => $t_select
 		),
 		'options' => array(
--- a/api/rest/restcore/users_rest.php
+++ b/api/rest/restcore/users_rest.php
@@ -28,8 +28,16 @@
  * @var \Slim\App $g_app
  */
 $g_app->group('/users', function() use ( $g_app ) {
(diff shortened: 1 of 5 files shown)


vboctor:
I'm not sure why a test is failing on Travis, it is passing on my machine and it is checking for a user that was created by the test case. I'll look into it more, strange!

core/commands/UserGetCommand.php
```diff
@@ -0,0 +1,167 @@
...
+				}
+			}
+		} else {
+			$this->target_user_id = (int)$t_user_id;
+		}
+
+		if( !is_null( $t_user_ref ) && $this->target_user_id == 0 ) {
+			throw new ClientException(
+				"User not found with id, name, realname or email equal to '$t_user_ref'",
+				ERROR_USER_BY_NAME_NOT_FOUND,
```
dregad, line 106: We actually have distinct error codes for the various scenarios (ERROR_USER_BY_ID_NOT_FOUND, ERROR_USER_BY_EMAIL_NOT_FOUND, ERROR_USER_BY_REALNAME_NOT_FOUND), so the Exception should use them and not throw a potentially incorrect ERROR_USER_BY_NAME_NOT_FOUND.

Set a variable with the error code in the if blocks above.

api/rest/restcore/users_rest.php
```diff
@@ -76,20 +106,37 @@ function rest_user_get_me( \Slim\Http\Request $p_request, \Slim\Http\Response $p
...
+		$t_select = array(
+			'id',
+			'name',
+			'real_name',
+			'email',
+			'access_level',
+			'language',
+			'timezone',
+			'created_at'
+		);
```
dregad, line 122: The Command already defines the default fields, is it necessary to duplicate code here ?

core/commands/UserGetCommand.php
```diff
@@ -0,0 +1,167 @@
...
+				$this->select = $t_select;
+			}
+		}
+
+		# if select field is not specified or empty, then use defaults
+		if( is_null( $this->select ) ) {
+			$this->select = array(
+				'id',
+				'name',
+				'real_name',
```
dregad, line 142: What happens if user has no access to real_name ?

dregad (changes requested):
Email and Real Name have no guarantee to be unique. If there are multiple matches, the API will just return the first one (exactly which one is not defined) and caller will never know that there are other occurrences.

Therefore I don't think it's a good idea to allow querying by email or real name using user_get_id_by_XXX as-is. IMO, in this scenario we should bet a list of matching users.

Also consider you introduced user_get_id_by_user_info() some years back, so there is some duplicated logic.

core/commands/UserGetCommand.php
vboctor: I have changed the API to be:

/users/me
/users/:user_id
/users/username/:username

And remove by email and real name.

api/rest/restcore/users_rest.php
vboctor: I have refactored the code to have a method that returns the default fields and used it from everywhere. The /users/me uses default fields and adds `projects` field to the list.

core/commands/UserGetCommand.php
vboctor: If user doesn't have access to real name, then they won't see it. If they are retrieving their own user, then they will see it. That is a standard behavior for getting account ref structure across the board.

PR #1883, opened 2023-04-22
REST API Project Version Get / Update / Delete APIs + Tests

vboctor:
- [32445](https://mantisbt.org/bugs/view.php?id=32445]: REST API: Project Add API to return information about added version
- [30415](https://mantisbt.org/bugs/view.php?id=30415] REST API: Add API to Get / Delete / Update versions

### DIFF
--- a/api/rest/restcore/projects_rest.php
+++ b/api/rest/restcore/projects_rest.php
@@ -44,6 +44,8 @@ $g_app->group('/projects', function() use ( $g_app ) {
 	# Project versions
 	$g_app->post( '/{id}/versions', 'rest_project_version_add' );
 	$g_app->post( '/{id}/versions/', 'rest_project_version_add' );
+	$g_app->delete( '/{id}/versions/{version_id}', 'rest_project_version_remove' );
+	$g_app->delete( '/{id}/versions/{version_id}/', 'rest_project_version_remove' );	
 
 	# Project hierarchy (subprojects)
 	$g_app->post( '/{id}/subprojects', 'rest_project_hierarchy_add' );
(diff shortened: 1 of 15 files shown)


api/rest/restcore/projects_rest.php
```diff
@@ -42,8 +42,16 @@
...
+	$g_app->get( '/{id}/versions', 'rest_project_version_get' );
+	$g_app->get( '/{id}/versions/', 'rest_project_version_get' );
+	$g_app->get( '/{id}/versions/{version_id}', 'rest_project_version_get' );
+	$g_app->get( '/{id}/versions/{version_id}/', 'rest_project_version_get' );
 	$g_app->post( '/{id}/versions', 'rest_project_version_add' );
 	$g_app->post( '/{id}/versions/', 'rest_project_version_add' );
+	$g_app->patch( '/{id}/versions/{version_id}', 'rest_project_version_update' );
+	$g_app->patch( '/{id}/versions/{version_id}/', 'rest_project_version_update' );
+	$g_app->delete( '/{id}/versions/{version_id}', 'rest_project_version_delete' );
+	$g_app->delete( '/{id}/versions/{version_id}/', 'rest_project_version_delete' );
```
dregad, line 54: Shouldn't `/{id}/versions` be defined as a group to avoid duplication?

Also, what about making use of optional parameters so handle the route variants for a given method ? Something like
```php
$g_app->get( ''/{id}/versions[/{version_id}][/]', 'rest_project_version_get' );
```
(I have not actually tested this)

core/commands/VersionDeleteCommand.php
```diff
@@ -0,0 +1,118 @@
...
+			/* default */ null,
+			/* user */ null,
+			$this->project_id );
+		if( !access_has_project_level( $t_access_level, $this->project_id ) ) {
+			throw new ClientException( 'Access denied to delete version', ERROR_ACCESS_DENIED );
+		}
+
+		$this->version_id = helper_parse_id( $this->query( 'version_id' ), 'version_id' );
+		if( !version_exists( $this->version_id ) ) {
+			return;
```
dregad, line 82: @vboctor is it intentional that specifiying a non-existing version is not triggering an ERROR_VERSION_NOT_FOUND ClientException ? This results in the API returning a 204 (version deleted)

dregad: Proposed fix in PR #1905

vboctor: I'll comment on the proposed fix PR.

PR #1885, opened 2023-04-30
REST API: User Update API

vboctor:
- [32464](https://www.mantisbt.org/bugs/view.php?id=32464): Implement UserUpdateCommand
- [27130](https://www.mantisbt.org/bugs/view.php?id=27130): change username via rest api
- [24757](https://www.mantisbt.org/bugs/view.php?id=24757): To move a user to disabled

### DIFF
--- a/api/rest/restcore/users_rest.php
+++ b/api/rest/restcore/users_rest.php
@@ -194,14 +194,16 @@ function rest_user_update( \Slim\Http\Request $p_request, \Slim\Http\Response $p
 
 	$t_data = array(
 		'query' => $t_query,
-		'payload' => $t_payload );
+		'payload' => $t_payload
+	);
 
 	$t_command = new UserUpdateCommand( $t_data );
 	$t_result = $t_command->execute();
 	$t_user_id = $t_result['user']['id'];
 
-	return $p_response->withStatus( HTTP_STATUS_SUCCESS, "User with id $t_user_id updated" )->
-		withJson( $t_result );
+	return $p_response
+		->withStatus( HTTP_STATUS_SUCCESS, "User with id $t_user_id updated" )
+		->withJson( $t_result );
 }
 
 /**
--- a/api/rest/restcore/users_rest.php
+++ b/api/rest/restcore/users_rest.php
@@ -42,6 +42,8 @@ $g_app->group('/users', function() use ( $g_app ) {
 	$g_app->post( '/', 'rest_user_create' );
 	$g_app->post( '', 'rest_user_create' );
 
+	$g_app->patch( '/{id}', 'rest_user_update' );
+
 	$g_app->post( '/me/token/', 'rest_user_create_token_for_current_user' );
 	$g_app->post( '/me/token', 'rest_user_create_token_for_current_user' );
 	$g_app->post( '/{user_id}/token/', 'rest_user_create_token' );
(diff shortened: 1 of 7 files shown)


core/commands/UserUpdateCommand.php
```diff
@@ -0,0 +1,345 @@
...
+		if( is_null( $t_user ) ) {
+			throw new ClientException( 'Missing user data', ERROR_EMPTY_FIELD, array( 'user' ) );
+		}
+
+		# Protected
+		$t_old_protected = user_is_protected( $this->user_id );
+		$t_new_protected = isset( $t_user['protected'] ) ? (int)$t_user['protected'] : null;
+		if( !is_null( $t_new_protected ) && $t_old_protected !== $t_new_protected ) {
+			$this->protected = $t_new_protected ? true : false;
+		}
```
dregad, line 140: Not sure why you're assigning int|null to $t_new_protected, when the finality is to compare it with a bool returned by user_is_protected() (which would btw always fail due to the strict-type compare bool vs int) and finally assign true/false to the class property.

This should be simplified, something like (not tested)
```php
if( isset( $t_user['protected'] ) ) {
    $t_new_protected = (bool)$t_user['protected'];
    if( $t_old_protected != $t_new_protected ) {
        $this->protected = $t_new_protected;
    }
}
```

api/rest/restcore/users_rest.php
```diff
@@ -166,6 +168,42 @@ function rest_user_create( \Slim\Http\Request $p_request, \Slim\Http\Response $p
...
+	$t_data = array(
+		'query' => $t_query,
+		'payload' => $t_payload );
+
+	$t_command = new UserUpdateCommand( $t_data );
+	$t_result = $t_command->execute();
+	$t_user_id = $t_result['user']['id'];
+
+	return $p_response->withStatus( HTTP_STATUS_SUCCESS, "User with id $t_user_id updated" )->
+		withJson( $t_result );
```
dregad, line 204: For better readability, I would suggest to format the code like this:
```php
$p_response
	->withStatus( HTTP_STATUS_SUCCESS, "User with id $t_user_id updated" )
	->withJson( $t_result );
```

api/rest/restcore/users_rest.php
```diff
@@ -166,6 +168,42 @@ function rest_user_create( \Slim\Http\Request $p_request, \Slim\Http\Response $p
...
+	}
+
+	$t_payload = $p_request->getParsedBody();
+	if( !$t_payload ) {
+		return $p_response->withStatus( HTTP_STATUS_BAD_REQUEST, "Invalid request body or format");
+	}
+
+	$t_data = array(
+		'query' => $t_query,
+		'payload' => $t_payload );
```
dregad, line 197: `);` on new line

core/commands/UserUpdateCommand.php
```diff
@@ -0,0 +1,345 @@
...
+use Mantis\Exceptions\ClientException;
+
+require_once( dirname( __FILE__ ) . '/../../api/soap/mc_account_api.php' );
+require_once( dirname( __FILE__ ) . '/../../api/soap/mc_api.php' );
+
+/**
+ * A command that updates a user account.
+ *
+ * Note that only fields to be updated need to be provided. If extra fields
+ * are provided and it matches current state, it will be ignored.
```
dregad, line 32: *they* will be ignored

core/commands/UserUpdateCommand.php
```diff
@@ -0,0 +1,345 @@
...
+				throw new ClientException( 'Access denied to update own enabled status', ERROR_ACCESS_DENIED );
+			}
+		} else {
+			if( !$t_actor_can_manage_users ) {
+				throw new ClientException( 'Access denied to update users', ERROR_ACCESS_DENIED );
+			}
+		}
+
+		# Don't allow updating accounts to access levels higher than that of
+		# the user actor's the account.
```
dregad, line 242: I think there's an extra *the* here. Check meaning and fix

core/commands/UserUpdateCommand.php
```diff
@@ -0,0 +1,345 @@
...
+		if( !access_has_global_level( $t_old_access_level ) ||
+		    ( !is_null( $t_new_access_level ) && !access_has_global_level( $t_new_access_level ) ) ) {
+			throw new ClientException(
+				'Access denied to update users with higher access level',
```
dregad, line 246: with -> to  ?

core/commands/UserUpdateCommand.php
```diff
@@ -0,0 +1,345 @@
...
+
+		# check that we are not downgrading the last administrator
+		if( user_is_administrator( $this->user_id ) ) {
+			$t_admin_threshold = config_get_global( 'admin_site_threshold' );
+			$t_admin_count = user_count_level( $t_admin_threshold, /* enabled */ true );
+
+			if( $t_admin_count <= 1 ) {
+				if( ( !is_null( $this->enabled ) && !$this->enabled ) ||
+				    ( !is_null( $this->access_level ) && $this->access_level < $t_admin_threshold ) ) {
+						throw new ClientException( "Disabling or reducing access level of last admin not allowed.", ERROR_USER_CHANGE_LAST_ADMIN );
```
dregad, line 258: break long line

core/commands/UserUpdateCommand.php
```diff
@@ -0,0 +1,345 @@
...
+		$t_query_params = array(
+			$p_user['username'],
+			$p_user['email'],
+			$p_user['access_level'],
+			$p_user['enabled'],
+			$p_user['protected'],
+			$p_user['real_name'],
+			$p_user['id'] );
+
+		db_query( $t_query, $t_query_params );
```
dregad, line 342: Shouldn't this be a low-level API (user_update() function in user_api.php) ?

I suggest to use a DbQuery object here instead of using legacy db api calls. See project_update() for example: 
https://github.com/mantisbt/mantisbt/blob/8c5b133eb1058c6222f4f26b44649a65a8d213b3/core/project_api.php#L469-L488

dregad (changes requested):

core/commands/UserUpdateCommand.php
vboctor: I have change int to bool and removed strict equality. This style is consistent with the rest of the code.

core/commands/UserUpdateCommand.php
vboctor: We don't have the concept of a low level private/internal API, so will leave here for now. The code is also moved from the action page IIRC. Given that this implementation is tested already, I'll leave as is for now.

PR #1894, opened 2023-06-10
Remove function check_php_version

atrol:
function check_php_version is questionable in some terms
- check for $p_version == PHP_MIN_VERSION will hardly ever evaluate to true in real life (who is running exactly PHP 7.0.0?)
- parameter $p_version is not used in else case, this was hardly what was wanted, but did not introduce any bug, as phpversion() is used instead of the parameter
- check function_exists( 'version_compare' ) is not necessary as
  - according documentation, version_compare() is available starting from PHP 4.1 [1]
  - we use version_compare() at various other places in code without checking if it exists

Thus
- remove the function
- replace the single place where it's called by using version_compare()

Fixes #32714

[1] https://www.php.net/manual/de/function.version-compare.php

### DIFF
--- a/admin/install.php
+++ b/admin/install.php
@@ -292,7 +292,7 @@ if( 0 == $t_install_state ) {
 <?php
 	print_test(
 		'Checking PHP version (your version is ' . phpversion() . ')',
-		check_php_version( phpversion() ),
+		version_compare( phpversion(), PHP_MIN_VERSION, '>=' ),
 		true,
 		'Upgrade to a more recent version of PHP'
 	);
--- a/admin/install.php
+++ b/admin/install.php
@@ -287,17 +287,8 @@ if( $t_config_exists ) {
 
 if( 0 == $t_install_state ) {
 	?>
-
-<!-- Check PHP Version -->
+<!-- Check UTF-8 support -->
 <?php
-	print_test(
-		'Checking PHP version (your version is ' . phpversion() . ')',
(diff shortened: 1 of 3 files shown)


dregad (approved):

admin/install.php
```diff
@@ -292,7 +292,7 @@ function print_test( $p_test_description, $p_result, $p_hard_fail = true, $p_mes
 <?php
 	print_test(
 		'Checking PHP version (your version is ' . phpversion() . ')',
-		check_php_version( phpversion() ),
+		version_compare( phpversion(), PHP_MIN_VERSION, '>=' ),
 		true,
 		'Upgrade to a more recent version of PHP'
```
vboctor, line 297: Maybe change this message to upgrade to a version equal to or newer than min version.

vboctor (approved):

admin/install.php
atrol: This code can't be reached at all, as I introduced a check in an early state in core.php some while ago to prevent using non-supported old PHP versions, see https://github.com/mantisbt/mantisbt/commit/1eb5ff1c7e1d8d99e97eadbd54da087d38b0fa55

What you will get is a good understandable message, e. g.  something like
```
FATAL ERROR: Your version of PHP is too old. MantisBT requires 7.0.0 or newer
Your are running PHP version 5.6.40
```

I pushed one more commit to remove the unreachable code.

PR #1896, opened 2023-06-19
Allow filtering on projection

tgulacsi:
Simply add the "projection' field as a filterable element.

[Fixes #32726](https://mantisbt.org/bugs/view.php?id=32726)

### DIFF
--- a/core/classes/BugFilterQuery.class.php
+++ b/core/classes/BugFilterQuery.class.php
@@ -476,6 +476,9 @@ class BugFilterQuery extends DbQuery {
 				case FILTER_PROPERTY_RELATIONSHIP_TYPE:
 					$this->build_prop_relationship();
 					break;
+				case FILTER_PROPERTY_PROJECTION:
+					$this->build_prop_projection();
+					break;
 			}
 		}
 		# these have several properties that must be built only once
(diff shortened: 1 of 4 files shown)


dregad (changes requested):
Thanks for your contribution. Please

- [open an Issue](https://mantisbt.org/bugs/bug_report_page.php) in our tracker, and 
- reference the Issue number in your commit message (i.e. `Fixes #xxx`)

I'll have a closer look at the code and test later, I don't have time now.

dregad (changes requested):
I don't like where you placed the new filter field - it causes _Target Version_ to move to the next line, but it should remain next to the related _Project Version_ / _Fixed in Version_ fields.

It would be better IMO to put it after _Resolution_ and shift the date filters to the right, as shown in screenshot below

filter_projection

More importantly, **the new filter field does not seem to have any effect** - I get the same results regardless of what I select in the dropdown. 

Looking at the code I don't see any changes to the BugFilterQuery class, which would be expected to handle building the filter's database query. Maybe you forgot to commit/push that ?

tgulacsi:
> I don't like where you placed the new filter field - it causes _Target Version_ to move to the next line, but it should remain next to the related _Project Version_ / _Fixed in Version_ fields.
> 
> It would be better IMO to put it after _Resolution_ and shift the date filters to the right, as shown in screenshot below
> 

It can be done, but those date filters spanning 2 fields, so that insertion pushes the "Filter by Last Updated Date" to the next row.

It fits if I shrunk one/both of them to have a span=1.
See 2023-06-21T17:41:44,464408088+02:00

dregad:
Right, I completely forgot that those were wider... Let me think about this

dregad:
I thought about several variants to make room for the new filter criteria by reorganizing the screen, but none of them are really satisfactory so I'm leaning towards the colspan=1 approach for the date range filter criteria. It shouldn't be an issue when they are expanded, since the browser will resize the column dynamically to fit contents.

dregad:

PR #1897, opened 2023-06-19
[refactor] Use range() function instead of string increment

Girgias:
This was found during the analysis for https://wiki.php.net/rfc/saner-inc-dec-operators

Fixes [#32735](https://www.mantisbt.org/bugs/view.php?id=32735)

### DIFF
--- a/manage_tags_page.php
+++ b/manage_tags_page.php
@@ -61,13 +61,13 @@ $f_page_number = gpc_get_int( 'page_number', 1 );
 # Start Index Menu
 $t_prefix_array = array( 'ALL' );
 
-for( $i = 'A'; $i != 'AA'; $i++ ) {
-	$t_prefix_array[] = $i;
-}
+// Uppercase letters
+$u = range('A', 'Z');
+// Digits
+$d = range('0', '9');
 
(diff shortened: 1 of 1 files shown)


dregad (changes requested):
Thanks for your contribution. I'm OK with the proposed change on principle, but kindly

- update your patch to follow our [Coding Guidelines](https://www.mantisbt.org/wiki/doku.php/mantisbt:coding_guidelines): give variables meaningful names and prefix them with `t_`; space after `(` / before `)`
- [open an Issue](https://mantisbt.org/bugs/bug_report_page.php) in our tracker, and 
- reference the Issue number in your commit message (i.e. `Fixes #xxx`)

dregad:
Fixed in 05a5dd9f4ed0279cd108fb8910aceec5783b08a3

dregad:
For the record, this caused a regression (see #1904 / [#32811](https://www.mantisbt.org/bugs/view.php?id=32811)) because `range( '0', '9' )` actually returns an array of integers instead of strings on PHP < 8.3 (details in https://github.com/mantisbt/mantisbt/pull/1904#issuecomment-1670959543)

PR #1899, opened 2023-07-29
Fix user_is_email_unique() on case-sensitive DB 

dregad:
Fixes [#32451](https://mantisbt.org/bugs/view.php?id=32451)

This PR follows #1898 and [@atrol's remark](https://mantisbt.org/bugs/view.php?id=32451#c67922) questioning the target version for #32451 being 2.25.8. 

This is splitting the original PR, to target the bug fix at _master-2.25_ branch, while the enhancement will be merged into _master_ branch only

### DIFF
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -193,13 +193,16 @@ function user_update_cache( $p_user_id, $p_field, $p_value ) {
  *
  * @param string $p_field The user object field name to search the cache for.
  * @param mixed  $p_value The field value to look for in the cache.
+ * @param bool   $p_case_sensitive False to perform case-insensitive search; defaults to true.
+ *
  * @return integer|boolean
  */
-function user_search_cache( $p_field, $p_value ) {
+function user_search_cache( $p_field, $p_value, $p_case_sensitive = true ) {
 	global $g_cache_user;
 	if( isset( $g_cache_user ) ) {
+		$t_compare = $p_case_sensitive ? 'strcmp' : 'strcasecmp';
 		foreach( $g_cache_user as $t_user ) {
-			if( $t_user && $t_user[$p_field] == $p_value ) {
+			if( $t_user && 0 == $t_compare( $t_user[$p_field], $p_value ) ) {
 				return $t_user;
 			}
 		}
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -725,22 +725,36 @@ function user_get_id_by_name( $p_username, $p_throw = false ) {
 }
 
 /**
- * Get a user id from their email address
+ * Get a user's id from their email address.
+ *
+ * The check is case insensitive.
+ *
+ * This function should not be used when {@see $g_email_ensure_unique} is OFF:
+ * if there are multiple users found matching the given email, the function will
+ * not consider that to be an error, and arbitrarily return one of them (exactly
+ * which one is undetermined).
  *
  * @param string $p_email The email address to retrieve data for.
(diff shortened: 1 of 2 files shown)


core/user_api.php
```diff
@@ -288,18 +291,20 @@ function user_is_email_unique( $p_email, $p_user_id = null ) {
 	}
 
 	$p_email = trim( $p_email );
+	// Escape SQL LIKE pattern chars to ensure exact match
+	$p_email = preg_replace( '/([%_])/', '\\\\$1', $p_email );
```
vboctor, line 295: The escaping for SQL LIKE should be part of the implementation for `$t_query->sql_ilike` function. In this PR it is done before each call.

tests/Mantis/UserTest.php
```diff
@@ -0,0 +1,144 @@
...
+		$t_cookie = user_create(
+			'User' . rand(),
+			'password',
+			self::TEST_EMAIL
+		);
+		/** @noinspection PhpUnhandledExceptionInspection */
+		self::$user_id = user_get_id_by_cookie( $t_cookie );
+	}
+
+	public static function tearDownAfterClass() {
```
vboctor, line 54: Should you call `parent::tearDownAfterClass()` similar to calling `parent::setUpBeforeClass()` above?

core/user_api.php
```diff
@@ -720,22 +725,36 @@ function user_get_id_by_name( $p_username, $p_throw = false ) {
...
 
-	db_param_push();
-	$t_query = 'SELECT * FROM {user} WHERE email=' . db_param();
-	$t_result = db_query( $t_query, array( $p_email ) );
+	// Escape SQL LIKE pattern chars to ensure exact match
+	$p_email = preg_replace( '/([%_])/', '\\\\$1', $p_email );
+
+	$t_query = new DbQuery;
+	$t_query->sql( 'SELECT * FROM {user} WHERE '
+		// Case-insensitive like
```
vboctor, line 753: Is there a reason the fetching from cache is configurable for case insensitive vs. not, and here it is always case insensitive?

vboctor (changes requested):

core/user_api.php
dregad: > The escaping for SQL LIKE should be part of the implementation for $t_query->sql_ilike function

No we don't want that, the replacement has to be performed by the caller. 

If we did it in DbQuery::sql_ilike(), then the function would no longer interpret LIKE wildcards `%` or `_`, effectively turning it into a case-insensitive `=` operator (which basically what I'm going after here).

tests/Mantis/UserTest.php
dregad: Technically yes I guess, but the method is not defined in the parent, so it would just be unnecessary overhead

core/user_api.php
dregad: I'm not sure I understand what you're asking.

When searching a user Id by e-mail, we **always** want the search to be case insensitive, this is what this PR is changing.

The user cache contains other data, which is looked-up in a case-sensitive manner. So user_search_cache()'s new $p_case_sensitive parameter defaults to true ensuring unchanged behavior for existing usages, while allowing insensitive searches for email.

Does this answer your question ?

dregad:
@vboctor without further feedback from you, I assume you're satisfied with my replies to your review comments and plan to merge this tonight or tomorrow.

PR #1901, opened 2023-07-31
0010603: Automatic compression of uploads for download

CasN:
Allow zipping of attachments while uploading them to mantis (DISK or DB) with option to exclude certain file types

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -1956,26 +1956,26 @@ $g_allow_file_upload = ON;
 $g_file_upload_method = DATABASE;
 
 /**
- * Use Zip_attachments: zip attachments in order to optimize  
+ * Zip_attachments in order to optimize  
  * network performance and save disk/db space.
  * @global integer $g_zip_attachments
  */
 $g_zip_attachments		=	OFF;
 
 /**
- * Use Zip_excludes: define the file types to omit from the zipping process 
+ * Zip_excludes defines the file types to omit from the zipping process 
  * because the files are already compressed and/or one does not want to
  * compress a specific file type.
  * Values should be provided in lowercase.
  * @global integer $g_zip_excludes
  */
-$g_zip_excludes			= array('pdf','zip','docx','xlsx','pptx');
+$g_zip_attachments_excluded_extensions		= array('pdf','zip','docx','xlsx','pptx');
 
 /**
  * Use Zip_threshold: define minimal size of the attachment for the zipping process.
  * @global integer $g_zip_excludes
  */
-$g_zip_threshold		= 1024;
+$g_zip_attachments_minimum_size		= 1024;
 
 /**
  * Use File dropzone: enable drag and drop into a drop zone functionality for
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -1963,17 +1963,17 @@ $g_file_upload_method = DATABASE;
 $g_zip_attachments		=	OFF;
 
 /**
- * Zip_excludes defines the file types to omit from the zipping process 
+ * zip_attachments_excluded_extensions defines the file types to omit from the zipping process 
  * because the files are already compressed and/or one does not want to
  * compress a specific file type.
  * Values should be provided in lowercase.
- * @global integer $g_zip_excludes
+ * @global array $g_zip__attachments_excluded_extensions
  */
 $g_zip_attachments_excluded_extensions		= array('pdf','zip','docx','xlsx','pptx');
 
 /**
- * Use Zip_threshold: define minimal size of the attachment for the zipping process.
- * @global integer $g_zip_excludes
+ * zip_attachments_minimum_size defines minimal size of the attachment for the zipping process.
+ * @global integer $g_zip__attachments_minimum_size
  */
 $g_zip_attachments_minimum_size		= 1024;
 
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -1956,18 +1956,26 @@ $g_allow_file_upload = ON;
 $g_file_upload_method = DATABASE;
 
 /**
- * Use Zip_attachments: zip selected file types in order to optimize performance 
- * and save disk/db space
+ * Use Zip_attachments: zip attachments in order to optimize  
+ * network performance and save disk/db space.
  * @global integer $g_zip_attachments
  */
-$g_zip_attachments		=	ON;
+$g_zip_attachments		=	OFF;
 
 /**
- * Use Zip_excludes: define the file types to ommit from the zipping process 
- * because the files are already compressed and/or one does not want to compress a specific file type
+ * Use Zip_excludes: define the file types to omit from the zipping process 
+ * because the files are already compressed and/or one does not want to
+ * compress a specific file type.
+ * Values should be provided in lowercase.
  * @global integer $g_zip_excludes
  */
-$g_zip_excludes			=	"pdf,zip,jpg,docx,xlsx,pptx,x-zip-compressed";
+$g_zip_excludes			= array('pdf','zip','docx','xlsx','pptx');
+
+/**
+ * Use Zip_threshold: define minimal size of the attachment for the zipping process.
+ * @global integer $g_zip_excludes
+ */
+$g_zip_threshold		= 1024;
 
 /**
  * Use File dropzone: enable drag and drop into a drop zone functionality for
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -1956,6 +1956,20 @@ $g_allow_file_upload = ON;
 $g_file_upload_method = DATABASE;
 
 /**
+ * Use Zip_attachments: zip selected file types in order to optimize performance 
+ * and save disk/db space
+ * @global integer $g_zip_attachments
+ */
+$g_zip_attachments		=	ON;
+
+/**
+ * Use Zip_excludes: define the file types to ommit from the zipping process 
+ * because the files are already compressed and/or one does not want to compress a specific file type
+ * @global integer $g_zip_excludes
+ */
+$g_zip_excludes			=	"pdf,zip,jpg,docx,xlsx,pptx,x-zip-compressed";
+
+/**
  * Use File dropzone: enable drag and drop into a drop zone functionality for
  * file upload fields
  * @global integer $g_dropzone_enabled
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -69,8 +69,50 @@ function file_attach_files( $p_bug_id, $p_files, $p_bugnote_id = 0 ) {
 	}
 
 	$t_file_infos = array();
+	$t_excluded_files = config_get( 'zip_excludes' );
+	$t_excluded_arr = explode( ',', $t_excluded_files );
+	$t_zipfiles = config_get( 'zip_attachments' );
 	foreach( $p_files as $t_file ) {
 		if( !empty( $t_file['name'] ) ) {
+			# check if this file is excluded from the zipping excercise
+			$t_zip = true ;
+
+			# grab extension
+			$t_extension = $t_file['type'];
+			if ( ( $pos = strpos($t_extension, "/" ) ) !== FALSE) { 
+				$t_extension = substr( $t_extension, $pos+1 );  ; 
+			}
+			# check against excluded files
+			if( !is_blank( $t_excluded_files ) ) {
+				foreach( $t_excluded_arr as $t_val ) {
+					if( 0 == strcasecmp($t_val, $t_extension ) ) {
+						$t_zip = false ;
+					}
+				}
+			}
+			
+			## zip attached file 
+			if( ON == $t_zipfiles and $t_zip == true ) {
+				if ( extension_loaded( 'zip' ) ) {
+					// Creating a ZIP compressed archive of uploaded file
+					$t_new_filename  = $t_file['tmp_name'];
+					$t_new_filename .= ".zip";
+					// make zip archive
+					$zip = new ZipArchive();
+					$zip->open( $t_new_filename, ZIPARCHIVE::CREATE );
+					//add file
+					$t_datafile =  $t_file['tmp_name'];
+					$zip->addFile( "$t_datafile" );
+					// close archive
+					$zip->close();
+					// file to be used for mantis
+					$t_name = $t_file['name'];
+					$t_name .=".zip";
+					$t_file['name'] = $t_name;
+					$t_file['tmp_name'] = $t_new_filename;
+				}
+			} 			
+				
 			# $p_bug_id, array $p_file, $p_table = 'bug', $p_title = '', $p_desc = '', $p_user_id = null, $p_date_added = 0, $p_skip_bug_update = false, $p_bugnote_id = 0
 			$t_file_infos[] = file_add(
 				$p_bug_id,
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -72,7 +72,7 @@ function file_attach_files( $p_bug_id, $p_files, $p_bugnote_id = 0 ) {
 	$t_zip_attachments = config_get_global( 'zip_attachments' );
 	if ( ON == $t_zip_attachments && extension_loaded( 'zip' ) ) {
 		$t_zip_attachments_excluded_extensions = array_map( 'strtolower', config_get_global( 'zip_attachments_excluded_extensions' ) );
-		$t_zip__attachments_minimum_size = config_get_global( 'zip_attachments_minimum_size' ) ;
+		$t_zip_attachments_minimum_size = config_get_global( 'zip_attachments_minimum_size' ) ;
 	} else {
 		$t_zip_attachments = OFF;
 	}
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -70,26 +70,19 @@ function file_attach_files( $p_bug_id, $p_files, $p_bugnote_id = 0 ) {
 
 	$t_file_infos = array();
 	$t_zip_attachments = config_get( 'zip_attachments' );
-	$t_zip_go = false;
 	if ( ON == $t_zip_attachments && extension_loaded( 'zip' ) ) {
-		$t_excluded_files = array_map( 'strtolower', config_get( 'zip_excludes' ) );
-		$t_zip_go = true;
+		$t_zip_attachments_excluded_extensions = array_map( 'strtolower', config_get( 'zip_attachments_excluded_extensions' ) );
+		$t_zip__attachments_minimum_size = config_get( '$g_zip_attachments_minimum_size' ) ;
+	} else {
+		$t_zip_attachments = OFF;
 	}
 	foreach( $p_files as $t_file ) {
 		if( !empty( $t_file['name'] ) ) {
-			if ( $t_file['size'] > config_get( 'zip_threshold' ) ) {
-				if ( $t_zip_go ) {
-
-					# check if this file is excluded from the zipping excercise
-					$t_zip = true ;
-
-					# grab extension
-					$t_extension = strtolower( pathinfo( $t_file['name'] , PATHINFO_EXTENSION ) );
-					# check against excluded files
-					if ( in_array( $t_extension , $t_excluded_files ) ) {
-						$t_zip = false ;
-					}
-
+			if ( $t_zip_attachments  && $t_file['size'] > $t_zip__attachments_minimum_size ) {
+				# grab extension
+				$t_extension = strtolower( pathinfo( $t_file['name'] , PATHINFO_EXTENSION ) );
+				# check against excluded files
+				if ( !in_array( $t_extension , $t_zip_attachments_excluded_extensions ) ) {
 					## zip attached file 
 					if( $t_zip ) {
 						# Creating a ZIP compressed archive of uploaded file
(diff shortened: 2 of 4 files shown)


config_defaults_inc.php
```diff
@@ -1955,6 +1955,20 @@
...
+ * Use Zip_attachments: zip selected file types in order to optimize performance 
+ * and save disk/db space
+ * @global integer $g_zip_attachments
+ */
+$g_zip_attachments		=	ON;
+
+/**
+ * Use Zip_excludes: define the file types to ommit from the zipping process 
+ * because the files are already compressed and/or one does not want to compress a specific file type
+ * @global integer $g_zip_excludes
```
dregad, line 1968: Same comment as above. 

Avoid long lines (try to wrap around 80, no more than 100)

`omit` takes only one M

I think some clarification on what is meant by *file types* is needed here - elsewhere in config_defauts, we speak about extensions, but here I see `x-zip_compressed` which is more a mime type, so it's a bit confusing.

config_defaults_inc.php
```diff
@@ -1955,6 +1955,20 @@
...
+ */
+$g_zip_excludes			=	"pdf,zip,jpg,docx,xlsx,pptx,x-zip-compressed";
```
dregad, line 1970: Single-quoted strings are preferred.

And it maybe better to define the config as an array of strings, for consistency with $g_preview_text_extensions and others.

Indicate that values should be provided in lower case.

docbook/Admin_Guide/en-US/config/misc.xml
```diff
@@ -287,5 +287,25 @@
...
+			</listitem>
+		</varlistentry>
+		<varlistentry>
+			<term>$g_zip_excludes</term>
+			<listitem>
+				<para>A list of filetype extensions (separated by comma) which should be excluded from the zipping process.
+					The default value is "pdf,zip,jpg,docx,xlsx,pptx,x-zip-compressed".
+				</para>
+			</listitem>
+		</varlistentry>
```
dregad, line 309: Update per changes in config_defaults_inc.php (see my above comments).

config_defaults_inc.php
```diff
@@ -1955,6 +1955,20 @@
  */
 $g_file_upload_method = DATABASE;
 
+/**
+ * Use Zip_attachments: zip selected file types in order to optimize performance 
+ * and save disk/db space
+ * @global integer $g_zip_attachments
+ */
+$g_zip_attachments		=	ON;
```
dregad, line 1963: Default must be **OFF**

dregad (changes requested):
Thanks for your contribution.

I must say I'm personally not a big fan of this feature, sharing the same opinion that was expressed in [dhx's comment](https://mantisbt.org/bugs/view.php?id=10603#c22237), but I can see the benefits in terms of storage for trackers storing lots of large text file or uncompressed image attachments. 

If this is implemented only because of performance when downloading large attachments, then I think it's the wrong solution, gzip compression of HTTP traffic would be much more appropriate.

Additionally, I think it's not so user-friendly that the preview is not available anymore. This forces  users to download and unzip the file to view the contents.

I did not actually test the code, but I'm concerned about the system's behavior when changing these config's values, causing inconsistency.

Not sure if it's worth the effort, but maybe some admin tool should be implemented, to allow selectively compressing / decompressing historical attachments to have a consistent behavior.

Also, did you consider a size threshold below which attachments are NOT zipped ? It makes no sense to compress a small text file, as the resulting ZIP would be bigger than the original...

An alternative approach to alleviate storage issues (maybe more suited for an archival process, like in [cbradney's note](**url**)), could be to have a plugin that would compress attachments for old bugs based on some criteria (i.e. status, last updated date).

CasN:
It is done for a few reasons:

  1.  Uploading (big) attachments in light of network performance
  2.  Downloading (big) attachments in light of network performance
  3.  Reducing DB/DISK storage
That indeed is the question to be answered, hence it can be configured or not.
ZIP-files do not have a preview so I do not think there will be an issue with existing code around previewing some file types or not.

An admin-tool can be created but is not yet there. If it is decided not to include, it may become part of a plugin😊.

Size threshold makes sense, I will include that too.
Reactions on the other comments added in the section itself.

config_defaults_inc.php
```diff
@@ -1955,6 +1955,28 @@
...
 
+/**
+ * Use Zip_attachments: zip attachments in order to optimize  
+ * network performance and save disk/db space.
+ * @global integer $g_zip_attachments
+ */
+$g_zip_attachments		=	OFF;
+
+/**
+ * Use Zip_excludes: define the file types to omit from the zipping process
```
dregad, line 1966: Same comment as above

core/file_api.php
```diff
@@ -69,8 +69,47 @@ function file_attach_files( $p_bug_id, $p_files, $p_bugnote_id = 0 ) {
 	}
 
 	$t_file_infos = array();
+	$t_zip_attachments = config_get( 'zip_attachments' );
+	$t_zip_go = false;
```
dregad, line 73: This extra variable is not necessary. Just use $t_zip_attachments, as I suggested in my 1st review

core/file_api.php
```diff
@@ -69,8 +69,47 @@ function file_attach_files( $p_bug_id, $p_files, $p_bugnote_id = 0 ) {
...
+	if ( ON == $t_zip_attachments && extension_loaded( 'zip' ) ) {
+		$t_excluded_files = array_map( 'strtolower', config_get( 'zip_excludes' ) );
+		$t_zip_go = true;
+	}
 	foreach( $p_files as $t_file ) {
 		if( !empty( $t_file['name'] ) ) {
+			if ( $t_file['size'] > config_get( 'zip_threshold' ) ) {
```
dregad, line 80: Move config_get() out of loop, initialize variable if $t_zip_attachments == true

core/file_api.php
```diff
@@ -69,8 +69,47 @@ function file_attach_files( $p_bug_id, $p_files, $p_bugnote_id = 0 ) {
...
+				if ( $t_zip_go ) {
```
dregad, line 81: Check this before the size
```
if( $t_zip_attachments && $t_file['size] > $t_zip_threshold ) {
```

config_defaults_inc.php
```diff
@@ -1955,6 +1955,28 @@
...
+$g_zip_attachments		=	OFF;
+
+/**
+ * Use Zip_excludes: define the file types to omit from the zipping process 
+ * because the files are already compressed and/or one does not want to
+ * compress a specific file type.
+ * Values should be provided in lowercase.
+ * @global integer $g_zip_excludes
+ */
+$g_zip_excludes			= array('pdf','zip','docx','xlsx','pptx');
```
dregad, line 1972: Rename to $g_zip_attachments_excluded_extensions

config_defaults_inc.php
```diff
@@ -1955,6 +1955,28 @@
...
+
+/**
+ * Use Zip_threshold: define minimal size of the attachment for the zipping process.
+ * @global integer $g_zip_excludes
+ */
+$g_zip_threshold		= 1024;
```
dregad, line 1978: Rename to $g_zip_attachments_minimum_size

Indicate that it is to be specified in bytes

dregad (changes requested):

config_defaults_inc.php
CasN: Adjusted accordingly

core/file_api.php
```diff
@@ -69,8 +69,47 @@ function file_attach_files( $p_bug_id, $p_files, $p_bugnote_id = 0 ) {
...
+					## zip attached file 
+					if( $t_zip ) {
+						# Creating a ZIP compressed archive of uploaded file
+						$t_new_filename  = $t_file['tmp_name'].".zip";
+						# make zip archive
+						$zip = new ZipArchive();
+						$zip->open( $t_new_filename, ZIPARCHIVE::CREATE );
+						#add file
+						$t_datafile =  $t_file['tmp_name'];
+						$zip->addFile( "$t_datafile" );
```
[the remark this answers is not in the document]
CasN: Adjusted accordingly

core/file_api.php
CasN: Adjusted accordingly

core/file_api.php
CasN: Adjusted accordingly

core/file_api.php
CasN: Adjusted accordingly

config_defaults_inc.php
CasN: Adjusted accordingly

config_defaults_inc.php
```diff
@@ -1955,6 +1955,28 @@
  */
 $g_file_upload_method = DATABASE;
 
+/**
+ * Use Zip_attachments: zip attachments in order to optimize
```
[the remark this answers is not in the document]
CasN: Adjusted accordingly

docbook/Admin_Guide/en-US/config/misc.xml
```diff
@@ -287,5 +287,33 @@
 				</warning>
 			</listitem>
 		</varlistentry>
+		<varlistentry>
+			<term>$g_zip_attachments</term>
+			<listitem>
+				<para>Indicator to zip all attachments with the exception of the filetypes defined with $g_zip_excludes
```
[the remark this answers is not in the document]
CasN: Adjusted accordingly

config_defaults_inc.php
CasN: Adjusted accordingly

config_defaults_inc.php
```diff
@@ -1955,6 +1955,28 @@
...
+$g_zip_attachments		=	OFF;
+
+/**
+ * Zip_excludes defines the file types to omit from the zipping process 
+ * because the files are already compressed and/or one does not want to
+ * compress a specific file type.
+ * Values should be provided in lowercase.
+ * @global integer $g_zip_excludes
+ */
+$g_zip_attachments_excluded_extensions		= array('pdf','zip','docx','xlsx','pptx');
```
atrol, line 1972: What about dropping `$g_zip_attachments` and have just option `$g_zip_attachments_included_extensions` without any predefined extensions as default?
This would allow a more precise definition of what's wanted.
If we should stay on the blacklist approach, there should be some more extensions added (investigation needed) , e. g. 7z, jpg, jpeg, mov, mpeg, .... one more reason why I prefer the whitelist approach.

config_defaults_inc.php
```diff
@@ -1955,6 +1955,28 @@
...
+
+/**
+ * Use Zip_threshold: define minimal size of the attachment for the zipping process.
+ * @global integer $g_zip_excludes
```
atrol, line 1976: $g_zip_attachments_minimum_size

config_defaults_inc.php
```diff
@@ -1955,6 +1955,28 @@
...
+ * @global integer $g_zip_attachments
+ */
+$g_zip_attachments		=	OFF;
+
+/**
+ * Zip_excludes defines the file types to omit from the zipping process 
+ * because the files are already compressed and/or one does not want to
+ * compress a specific file type.
+ * Values should be provided in lowercase.
+ * @global integer $g_zip_excludes
```
atrol, line 1970: Wrong name, wrong type

core/file_api.php
```diff
@@ -69,8 +69,40 @@ function file_attach_files( $p_bug_id, $p_files, $p_bugnote_id = 0 ) {
 	}
 
 	$t_file_infos = array();
+	$t_zip_attachments = config_get( 'zip_attachments' );
```
atrol, line 72: @dregad I think we should not allow this setting to be a project / user specific one, so config_get_global should be used. Do you agree?

core/file_api.php
```diff
@@ -69,8 +69,40 @@ function file_attach_files( $p_bug_id, $p_files, $p_bugnote_id = 0 ) {
...
+	if ( ON == $t_zip_attachments && extension_loaded( 'zip' ) ) {
+		$t_zip_attachments_excluded_extensions = array_map( 'strtolower', config_get( 'zip_attachments_excluded_extensions' ) );
+		$t_zip__attachments_minimum_size = config_get( '$g_zip_attachments_minimum_size' ) ;
```
atrol, line 75: Can't work with `$g_zip_attachments_minimum_size`, must be `zip_attachments_minimum_size`

config_defaults_inc.php
CasN: Adjusted accordingly

config_defaults_inc.php
CasN: Adjusted accordingly

core/file_api.php
CasN: Adjusted accordingly

core/file_api.php
CasN: Can of course change it but depends on your first comment blacklist versus whitelist)  if we still need this setting

config_defaults_inc.php
CasN: I could change it to whitelisting. Before making any changes we should have consensus on the way forward.

dregad:
In reply to @CasN https://github.com/mantisbt/mantisbt/pull/1901#issuecomment-1658474550
> It is done for a few reasons: 
>1. Uploading (big) attachments in light of network performance 
>2. Downloading (big) attachments in light of network performance

I still believe that zipping attachments is the wrong fix for this problem - optimizing network performance should be done globally, and in fact is normally taken care of automatically by Compress API when `$g_compress_html = ON` (default).

Regarding point 2, please note that *file_download.php* currently disables compression explicitly. 
https://github.com/mantisbt/mantisbt/blob/32d58871e8231d8b79d671edaf837d590e4c5f0a/file_download.php#L42
https://github.com/mantisbt/mantisbt/blob/32d58871e8231d8b79d671edaf837d590e4c5f0a/file_download.php#L145-L147

This goes back to 2009, and according to Git blame (6b96804ac7b2cdb09e5a001d7c1a2573c2a71e59) this was done to fix downloading of large files hitting memory limit. Not sure that's still an issue in 2023... Maybe compression should be re-enabled. 


In reply to @atrol's review comment](https://github.com/mantisbt/mantisbt/pull/1901/files#r1279777529)
> @dregad I think we should not allow this setting and the other ones to be a project / user specific one, so config_get_global should be used. Do you agree?

Absolutely. It should definitely be a global config.


Regarding [whitelist vs blacklist](https://github.com/mantisbt/mantisbt/pull/1901/files#r1279771139), I don't have any real preference, both would work IMO. Which one to choose depends on the usage scenario(s), which I don't know. 

Blacklisting and @CasN's original implementation was [suggested by @vboctor](https://mantisbt.org/bugs/view.php?id=10603#c22201). 

The whilelisting approach proposed by @atrol has the advantage of requiring one less config option; we could even have it allow 3 states if you think that's useful
- OFF
- array listing file extensions to zip
- ON (= zip everything)

I guess offering both white and black list is probably overkill.

CasN:
Perhaps the better option here would be to transform it into a plugin given it is not considered as the correct solution.
If we agree on that, suggest the pull request is deleted and I will follow the other track.

CasN:
In that case I suggest to stick with Victor's suggestion using a blacklist. I will commit the next update.

core/file_api.php
```diff
@@ -69,8 +69,38 @@ function file_attach_files( $p_bug_id, $p_files, $p_bugnote_id = 0 ) {
 	}
 
 	$t_file_infos = array();
+	$t_zip_attachments = config_get_global( 'zip_attachments' );
+	if ( ON == $t_zip_attachments && extension_loaded( 'zip' ) ) {
+		$t_zip_attachments_excluded_extensions = array_map( 'strtolower', config_get_global( 'zip_attachments_excluded_extensions' ) );
+		$t_zip__attachments_minimum_size = config_get_global( 'zip_attachments_minimum_size' ) ;
```
atrol, line 75: Can't work, variable name is `$t_zip__attachments_minimum_size` but must be `$t_zip_attachments_minimum_size`

dregad:
> > @dregad I think we should not allow this setting and the other ones to be a project / user specific one, so config_get_global should be used. Do you agree?
> 
> Absolutely. It should definitely be a global config.

Following up on this, the new configs should be added to *$g_global_settings*, and *config_get_global()* used instead of *config_get()* to retrieve values.

atrol:
@CasN  I recommend to stop pushing more changes as long as we don't have an agreement how to proceed.

If we should stay with the current approach in general, I prefer using a whitelist instead of the blacklist.

- less options needed
- setting the default out of the box value is easy, just set it to OFF
- behavior is complete predictable, you have to think about / you know which types are compressed
- when using a blacklist, you will get surprised by files that will be compressed, but should not be compressed, e.g. the current default will compress .jpg files (and a lot more file types), were compression makes not that much sense

There is a side effect of the current approach, that I don't like at all.
Especially for the types where inline-preview is possible, it's annoying to download the zip, unzip the download, open the unzipped file.

Also to have just one setting `$g_zip_attachments_minimum_size = 1024;` for all file types is not optimal.
Compression ratio of different file type can be pretty different.
E.g comression of 100KB .xml files makes sense, whereas compression of 100KB .chm or .7z makes hardly any sense.
This introduces CPU stress on server and client without having a benefit in terms of amount of transferred data.
 
So you might like the following approach.

To reduce needed storage on disk / database on Server:

If attachment compression is enabled (default OFF), when uploading try to compress on server.
If compression ratio is higher than a configurable minimum compression ratio, store compressed file and mark it as compressed in an additional column of file table
When downloading, uncompress if it was compressed.

To optimize data transfer between client and server:

If delivering of the zipped file should needed like it's done in the current approach, we could offer an option for that with default OFF.
Setting to ON is still questionable, but better than the current implementation, as we would not uncompress zip-files with low compression ratio. 

Delivering of the zipped files might not be needed at all, as
- starting from 2.25.0, transfer of some attachments is already opzimized (lazy loading of inline attachments https://www.mantisbt.org/bugs/view.php?id=27150)
- like dregad mentioned it, allowing low level compression (negotiation between client and server) https://developer.mozilla.org/en-US/docs/Web/HTTP/Compression might work
 
Maybe there is also a way to calculate compression ratio already via JS on client side before uploading, but that's another story.

@CasN @dregad @vboctor comments welcome

CasN:
@atrol, I see the point of using a whitelist (already stated that I have that code on the shelf). This way users can decide themselves what they want to compress. I followed the other approach based upon the input of Victor in the corresponding ticket. And his input in general seems to be leading with such changes.
As for the second part of your suggestions, to me seems to lead to a lot of overhead. If I understand it correctly you suggest to decompress it while downloading. Personally I would not like that, would prefer to do that myself. Clearly this can be controlled by another setting but that is far from optimal in my view. Downloading a zip in explorer offers quite often the option to open the zip-file almost on the fly which for me is enough. My initial intention was and still is to keep it simple.

vboctor:
This discussion started in 2009, my comment refers to BMPs and now we have 4k videos 😄 

Here are my thoughts:

- We shouldn't compromise previews, there should be no behavior changes.
- If I upload a text file, I expect to download a txt file.
- We should leverage (enable) network compression (gzip) to optimize performance and client-server bandwidth.

Hence, beyond enabling gzip, here are a couple of possible ideas:

1. Optimize for Network: For large image previews, we could consider downsizing for the purpose of previews - think about it as creating thumbnails. This would mean two copies -- hence negative on disk space.
2. Optimize for Storage (Disk/Database): For saving space on disk/database, this is an implementation details of files API. System must be aware of which attachment is compressed vs. not. Ideally there is a script can compress/uncompress existing attachments based on configuration. In the meantime, changes apply going forward without breaking existing attachments.

At one point we need to refactor attachment handles to enable plugins to store in different stores (e.g. S3) as well as behaviors (e.g. compression or thumbnails).

So looks like we are all on agreement on gzip enablement. I think download as zip action from client can also be useful (maybe can be turned ON/OFF). Compression at rest (storage) is OK as long as it does violate the above constrains that I mentioned.

As for what to compress - given that attachments are one average downloaded or previewed more often that uploaded, we could consider zipping for files not in denyList, but only retaining the zip if it is x% smaller than original file.

CasN:
Let me comment on the views from Victor

There is absolutely a huge difference between technology of 2009 and 2023, most of the times to our advantage.

  *   We shouldn't compromise previews, there should be no behavior changes.
     *   That is up to the management of the installation to decide, possible downside is listed in the manual. It is just another feature.
  *   If I upload a text file, I expect to download a txt file.
     *   A lot of browsers on Windows and MacOS can automatically unzip compressed files. Perhaps not known to every end-user but it is available.
     *   All file explorers support opening compressed files (from the most common versions)
  *   Optimize for Network: For large image previews, we could consider downsizing for the purpose of previews - think about it as creating thumbnails. This would mean two copies -- hence negative on disk space.
     *   One could also offer the option to only keep the small version overcoming the negative impact.
  *   Optimize for Storage (Disk/Database)
     *   Certainly when storing in the cloud, size does matter. Given that is still somewhere on the horizon, we do not need to worry too much here.
  *   As for what to compress - given that attachments are one average downloaded or previewed more often that uploaded, we could consider zipping for files not in denyList, but only retaining the zip if it is x% smaller than original file.
     *   So still working with a blacklist and adding a check on relative size loss (where the percentage should be configurable). This can be easily implemented.
     *   This also can be used with a white-list which actually would have my preference now (denylist might become very large)..

dregad:
I'm going to step out of this discussion for now. I'll gladly perform a technical review once an agreement has been reached as to what to implement and how.

For that, I think what you need is feedback from Victor on your last comment.

vboctor:
I have stated my view on this @CasN and you response didn't change it. I think the principles I stated can be implemented, but it adds more complexity than value.

CasN:
Withdrawn for now

PR #1904, opened 2023-08-08
Fix tag filter display when using PHP 7.x

atrol:
Fixes [#32811](https://www.mantisbt.org/bugs/view.php?id=32811)

### DIFF
--- a/manage_tags_page.php
+++ b/manage_tags_page.php
@@ -112,7 +112,7 @@ print_manage_menu( 'manage_tags_page.php' );
 	<?php
 	foreach ( $t_prefix_array as $t_prefix ) {
 		$t_caption = ( $t_prefix === 'ALL' ? lang_get( 'show_all_tags' ) : $t_prefix );
-		$t_active = $t_prefix == $f_filter ? 'active' : '';
+		$t_active = (string)$t_prefix == (string)$f_filter ? 'active' : '';
 		echo '<a class="btn btn-xs btn-white btn-primary ' . $t_active .
 		'" href="manage_tags_page.php?filter=' . $t_prefix .'">' . $t_caption . '</a>' ."\n";
 	} ?>


vboctor (approved):

dregad (approved):

dregad:
In find it kind of weird that `range( '0', '9' )` generates an array of integer values and not strings... Good catch.

dregad:
This was bugging me so I checked further... And it looks like this behavior of *range()* will [change in PHP 8.3](https://wiki.php.net/rfc/proper-range-semantics), `range( '0', '2' )` **will** generate an array of strings, as I would expect.

https://onlinephp.io?s=s7EvyCjg5SpLLIpPKc0t0FAoSsxLT9VQN1DXUVA3UtdU0LQGAA%2C%2C&v=8.2.9%2C8.1.22%2C8.0.29%2C8.3.0beta1%2C7.4.33

Result for 8.2.9, 8.1.22, 8.0.29, 7.4.33:
``` 
array(3) { [0]=> int(0) [1]=> int(1) [2]=> int(2) }
```
Result for 8.3.0beta1:
```
array(3) { [0]=> string(1) "0" [1]=> string(1) "1" [2]=> string(1) "2" }
```

Anyway an alternative fix for the regression would be to ensure the generated array of digits is made of strings and not integers
```patch
diff --git a/manage_tags_page.php b/manage_tags_page.php
index 8cfea8035..e6728fd05 100644
--- a/manage_tags_page.php
+++ b/manage_tags_page.php
@@ -59,10 +59,14 @@ $f_filter = mb_strtoupper( gpc_get_string( 'filter', config_get( 'default_manage
 $f_page_number = gpc_get_int( 'page_number', 1 );
 
 # Start Index Menu
+$t_digits = range( '0', '9' );
+if( version_compare( PHP_VERSION, '8.3', '<' ) ) {
+       $t_digits = array_map( 'strval', $t_digits );
+}
 $t_prefix_array = array_merge(
        array( 'ALL' ),
        range( 'A', 'Z' ),
-       range( '0', '9' )
+       $t_digits
 );
 
 if( $f_filter === 'ALL' ) {
```

Girgias:
Yeah... range() is a bit scuffed at the moment, also discovered this while working on the ``++/--`` operator RFC which lead to the initial PR.

atrol:
Would like to be sure that you are aware that this gives (bool)true with all PHP 7.x versions, but gives (bool)false starting from PHP 8.o
```
<?php
var_dump("ALL" == 0);
```

dregad:
> Would like to be sure that you are aware that this gives (bool)true with all PHP 7.x versions, but gives (bool)false starting from PHP 8.o

@atrol not sure if you're talking to me or @Girgias... 
Answering for myself, yes I'm aware of that, it is a consequence of [this RFC](https://wiki.php.net/rfc/string_to_number_comparison) and was announced in the [PHP 8.0 release page](https://www.php.net/releases/8.0/en.php#saner-string-to-number-comparisons).

What I missed (and should have tested more carefully before merging #1897) is that range() created an integer instead of string array as I expected (which will be fixed starting with PHP 8.3).

PR #1906, opened 2023-08-19
Rest API issue relationship add/delete fixes

dregad:
- fixing errors
- code cleanup
- add unit tests

Fixes [#32835](https://mantisbt.org/bugs/view.php?id=32835)

Planning to back port the fixes to 2.25.x

### DIFF
--- a/api/soap/mc_issue_api.php
+++ b/api/soap/mc_issue_api.php
@@ -1056,7 +1056,20 @@ function mc_issue_update( $p_username, $p_password, $p_issue_id, stdClass $p_iss
 		$t_bug_data->severity = mci_get_severity_id( $p_issue['severity'] );
 	}
 	if( isset( $p_issue['status'] ) ) {
-		$t_bug_data->status = mci_get_status_id( $p_issue['status'] );
+		$t_old_status = $t_bug_data->status;
+		$t_new_status = mci_get_status_id( $p_issue['status'] );
+
+		# Check if we are resolving (or closing) the issue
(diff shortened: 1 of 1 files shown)


vboctor (approved):
Looks good. Nice cleanup!

You could add two more tests if you like:
- The reversing of relationship logic, not sure if this exercised or not.
- Attempting to resolve an issue that has non-resolved children.

dregad:
Thanks for the review.

> You could add two more tests if you like:
> 
> * The reversing of relationship logic, not sure if this exercised or not.

This is probably covered by the existing test, as it basically the same as replacing an existing relationship, but I'll have a closer look to make sure.

> * Attempting to resolve an issue that has non-resolved children.

Good one, didn't think about it. Will add.

dregad:
> > * Attempting to resolve an issue that has non-resolved children.
> 
> Good one, didn't think about it. Will add.

As it turns out, **the test should fail but does not**: the API will happily set the issue to *resolved*. So we have a discrepancy between _mc_issue_update()_ and _bug_update.php_ - and quite possibly more than one actually.

I can implement a quick fix (i.e. patch mc_issue_update() to call relationship_can_resolve_bug() and return a faultBadRequest if not) for now, but to be clean we really need a new **IssueUpdateCommand**... That is a complex beast however, and I don't really have the time to work on this right now.

dregad:
> to be clean we really need a new **IssueUpdateCommand**

Opened [#32861](https://mantisbt.org/bugs/view.php?id=32861) to track this.

PR #1907, opened 2023-08-26
Return consistent status codes from REST API delete operations

dregad:
Fixes [#32858](https://mantisbt.org/bugs/view.php?id=32858) + code cleanup

Follows up on discussion in #1905

### DIFF
--- a/api/rest/restcore/projects_rest.php
+++ b/api/rest/restcore/projects_rest.php
@@ -171,9 +171,9 @@ function rest_project_user_add( \Slim\Http\Request $p_request, \Slim\Http\Respon
 /**
  * A method to remove user access to a project.
  *
- * @param \Slim\Http\Request $p_request   The request.
+ * @param \Slim\Http\Request  $p_request  The request.
  * @param \Slim\Http\Response $p_response The response.
- * @param array $p_args Arguments
+ * @param array               $p_args     Arguments
  * @return \Slim\Http\Response The augmented response.
  */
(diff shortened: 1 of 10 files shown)


dregad:
Note: there are currently no test cases for REST API /projects endpoint

vboctor (changes requested):
Thanks @dregad

The API tests shouldn't call internal APIs like tag_exists(). The need to do so is a symptom of missing APIs and couples the API tests with the codebase. Please remove such calls and identify gaps that we need with todos or issues in bug tracker - e.g. if we are missing a Get Tags API or equivalent.

dregad:
> The API tests shouldn't call internal APIs like tag_exists(). The need to do so is a symptom of missing APIs and couples the API tests with the codebase.

I can hear the argument about such usage being an indicator that the API has missing endpoints and needs improvement. On the other hand, I strongly disagree that the internal calls should be removed before the new endpoints are available. 

Doing so means reducing the tests's coverage and quality, which does not make sense. I'm OK to flag them with a TODO, but getting rid of them is not a good idea.

Some of these usages, e.g. the loop to find a non-existing Id or ensuring a tag name's uniqueness, make absolutely no sense to implement with repeated REST calls; it adds no value, and is less efficient.

dregad:
Opened [#32863](https://mantisbt.org/bugs/view.php?id=32863) to track adding CRUD endpoints for Tags management.

vboctor (approved):

PR #1909, opened 2023-08-27
REST API compatibility on PHP 8.1+ with Slim 3.x

dregad:
Fixes [#32866](https://mantisbt.org/bugs/view.php?id=32866)


dregad:
It actually works ! 🚀 😃

PR #1912, opened 2023-08-31
Remove "Operation Successful" from action pages

vboctor:
Removed operation successful from all pages except:

- Ones that show information in addition to "operation successful".
- Pages that don't reflect the changes or redirect to another page.

Fixes #[5189](https://www.mantisbt.org/bugs/view.php?id=5189)

### DIFF
--- a/account_prefs_update.php
+++ b/account_prefs_update.php
@@ -139,10 +139,5 @@ user_pref_set( $f_user_id, $t_prefs, ALL_PROJECTS );
 
 form_security_purge( 'account_prefs_update' );
 
-layout_page_header( null, $f_redirect_url );
+print_header_redirect( $f_redirect_url );
 
-layout_page_begin();
-
-html_operation_successful( $f_redirect_url );
-
-layout_page_end();
--- a/account_prefs_update.php
+++ b/account_prefs_update.php
@@ -139,5 +139,5 @@ user_pref_set( $f_user_id, $t_prefs, ALL_PROJECTS );
 
 form_security_purge( 'account_prefs_update' );
 
-print_header_redirect( $f_redirect_url );
+print_successful_redirect( $f_redirect_url, /* force_show */ true );
 
--- a/account_prefs_update.php
+++ b/account_prefs_update.php
@@ -139,5 +139,5 @@ user_pref_set( $f_user_id, $t_prefs, ALL_PROJECTS );
 
 form_security_purge( 'account_prefs_update' );
 
-print_successful_redirect( $f_redirect_url, /* force_show */ true );
+print_header_redirect( $f_redirect_url, /* force_show */ true );
 
--- a/account_prefs_update.php
+++ b/account_prefs_update.php
@@ -139,5 +139,5 @@ user_pref_set( $f_user_id, $t_prefs, ALL_PROJECTS );
 
 form_security_purge( 'account_prefs_update' );
 
-print_header_redirect( $f_redirect_url, /* force_show */ true );
+print_header_redirect( $f_redirect_url );
 
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -151,9 +151,12 @@ function print_header_redirect_view( $p_bug_id ) {
  *
  * @param integer $p_bug_id A bug identifier.
  * @return void
- * @deprecated Use print_header_redirect() instead.
+ * @deprecated 2.26.0 Use print_header_redirect() instead.
  */
 function print_successful_redirect_to_bug( $p_bug_id ) {
+	error_parameters( __FUNCTION__ . '()', 'print_header_redirect()' );
+	trigger_error( ERROR_DEPRECATED_SUPERSEDED, DEPRECATED );
+
 	$t_url = string_get_bug_view_url( $p_bug_id );
 	print_header_redirect( $t_url );
 }
@@ -165,9 +168,12 @@ function print_successful_redirect_to_bug( $p_bug_id ) {
  * @param string $p_redirect_to URI to redirect to.
  * @param bool $p_force_show Force showing operation successful
  * @return void
- * @deprecated Use print_header_redirect() instead.
+ * @deprecated 2.26.0 Use print_header_redirect() instead.
  */
 function print_successful_redirect( $p_redirect_to, $p_force_show = false ) {
+	error_parameters( __FUNCTION__ . '()', 'print_header_redirect()' );
+	trigger_error( ERROR_DEPRECATED_SUPERSEDED, DEPRECATED );
+
 	print_header_redirect( $p_redirect_to );
 }
 
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -163,10 +163,11 @@ function print_successful_redirect_to_bug( $p_bug_id ) {
  * If the show query count is OFF, redirect right away.
  *
  * @param string $p_redirect_to URI to redirect to.
+ * @param bool $p_force_show Force showing operation successful
  * @return void
  */
-function print_successful_redirect( $p_redirect_to ) {
-	if( helper_log_to_page() ) {
+function print_successful_redirect( $p_redirect_to, $p_force_show = false ) {
+	if( $p_force_show || helper_log_to_page() ) {
 		layout_page_header( null, $p_redirect_to );
 		layout_page_begin();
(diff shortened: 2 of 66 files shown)


core/print_api.php
```diff
@@ -163,10 +163,11 @@ function print_successful_redirect_to_bug( $p_bug_id ) {
...
+ * @param bool $p_force_show Force showing operation successful
  * @return void
  */
-function print_successful_redirect( $p_redirect_to ) {
-	if( helper_log_to_page() ) {
+function print_successful_redirect( $p_redirect_to, $p_force_show = false ) {
+	if( $p_force_show || helper_log_to_page() ) {
 		layout_page_header( null, $p_redirect_to );
 		layout_page_begin();
 		echo '<br /><div class="center">';
```
dregad, line 173: This code block should be replaced by a call to html_operation_successful().

https://github.com/mantisbt/mantisbt/blob/4a20beb2d08ba7655c1db2fef5c2a5c6963e0eff/core/print_api.php#L173-L176

dregad (changes requested):
I have not fully tested this yet.

Noticed that you have not changed bug_report.php, is that intentional ?

2nd Commit message has long lines, see https://mantisbt.org/wiki/doku.php/mantisbt:git_commit_messages

Will do more tests later

vboctor:
Before this PR and now, we have some action pages that always showed operational successful and some pages that only showed operation successful when page logging is enabled. Here are the possible options:

1. Standardize on `print_successful_redirect()` which uses - pro - debugging is easier, con - bad naming for the main product case.
2. Standardize on `print_header_redirect()` and use file logging when debugging performance of an action page - pro - naming consistent with the product behavior, con - no on page logging for action pages. Note that most of the profiling is typically done on view pages.
3. Deprecate print_successful_redirect() and make it just call print_header_redirect, then handle the debug case as a special case there. Maybe have an override flag for behavior, default/always/never show operational successful.

I like the simplicity and consistency of naming with product behavior of 2.

dregad:
To be honest I don't think I've ever actually used `$g_log_destination = 'page';` so I'm pretty sure we can live without *print_successful_redirect()*.

> use file logging when debugging performance of an action page [...] Note that most of the profiling is typically done on view pages.

Not sure what you mean / intend to do, can you clarify ?

core/print_api.php
vboctor: I have deprecated print_successful_redirect*(), so this code no longer exists.

vboctor:
> To be honest I don't think I've ever actually used `$g_log_destination = 'page';` so I'm pretty sure we can live without _print_successful_redirect()_.

I have deprecated `print_successful_redirect*()` and used `print_header_redirect*()` instead.

> > use file logging when debugging performance of an action page [...] Note that most of the profiling is typically done on view pages.
> 
> Not sure what you mean / intend to do, can you clarify ?

I was saying the same thing you are saying, we live without the page logging feature on action pages. I was saying that other logging options can be utilized as needed.

vboctor:
> Noticed that you have not changed bug_report.php, is that intentional ?

It wasn't intentional. Fixed now.

core/print_api.php
```diff
@@ -165,19 +165,10 @@ function print_successful_redirect_to_bug( $p_bug_id ) {
...
-		layout_page_begin();
-		echo '<br /><div class="center">';
-		echo lang_get( 'operation_successful' ) . '<br />';
-		print_link_button( $p_redirect_to, lang_get( 'proceed' ) );
-		echo '</div>';
-		layout_page_end();
-	} else {
-		print_header_redirect( $p_redirect_to );
-	}
+	print_header_redirect( $p_redirect_to );
```
dregad, line 171: Add
```php
	error_parameters( __FUNCTION__ . '()', 'print_header_redirect()' );
```

core/print_api.php
```diff
@@ -151,11 +151,11 @@ function print_header_redirect_view( $p_bug_id ) {
  *
  * @param integer $p_bug_id A bug identifier.
  * @return void
+ * @deprecated Use print_header_redirect() instead.
  */
 function print_successful_redirect_to_bug( $p_bug_id ) {
 	$t_url = string_get_bug_view_url( $p_bug_id );
-
-	print_successful_redirect( $t_url );
+	print_header_redirect( $t_url );
```
dregad, line 158: Add
```php
	error_parameters( __FUNCTION__ . '()', 'print_header_redirect()' );
```

dregad:
Maybe that's just the force of habit, but I must admit that it does feel really weird, not to have some kind of visual confirmation that an operation completed successfully. 

I made this remark on several occasions in the related tracker issues, but @vboctor I don't think you ever commented on that.

dregad (changes requested):
A quick grep shows 2 remaining occurrences of print_successful_redirect() that should probably be updated as well.

```
manage_plugin_upgrade.php:60:print_successful_redirect( 'manage_plugin_page.php' );
manage_proj_custom_field_update.php:67:print_successful_redirect( $t_redirect_url, /* force_show */ true );
```

core/print_api.php
```diff
@@ -151,32 +151,24 @@ function print_header_redirect_view( $p_bug_id ) {
  *
  * @param integer $p_bug_id A bug identifier.
  * @return void
+ * @deprecated Use print_header_redirect() instead.
```
dregad, line 154: PHPoc tag should indicate the version in which the functionality was deprecated
```
@deprecated 2.26.0 Use print_header_redirect() instead.
```

core/print_api.php
```diff
@@ -151,32 +151,24 @@ function print_header_redirect_view( $p_bug_id ) {
...
 }
 
 /**
  * If the show query count is ON, print success and redirect after the configured system wait time.
  * If the show query count is OFF, redirect right away.
  *
  * @param string $p_redirect_to URI to redirect to.
+ * @param bool $p_force_show Force showing operation successful
  * @return void
+ * @deprecated Use print_header_redirect() instead.
```
dregad, line 168: PHPoc tag should indicate the version in which the functionality was deprecated
```
@deprecated 2.26.0 Use print_header_redirect() instead.
```

dregad (changes requested):

account_prefs_update.php
```diff
@@ -139,10 +139,5 @@
 
 form_security_purge( 'account_prefs_update' );
 
-layout_page_header( null, $f_redirect_url );
+print_header_redirect( $f_redirect_url, /* force_show */ true );
```
dregad, line 142: I think this is wrong - print_header_redirect()'s second parameter is not *force_show*... if true, it stops the script after redirecting.

The 2nd parameter should be dropped.

dregad (changes requested):

dregad:
> A quick grep shows 2 remaining occurrences of print_successful_redirect() that should probably be updated as well.
> 
> ```
> manage_plugin_upgrade.php:60:print_successful_redirect( 'manage_plugin_page.php' );
> manage_proj_custom_field_update.php:67:print_successful_redirect( $t_redirect_url, /* force_show */ true );
> ```

Fixed.

dregad:
@vboctor I have addressed my own review comments so this should be good to go.

But before it's merged, I'd like to have your feedback on https://github.com/mantisbt/mantisbt/pull/1912#issuecomment-1709744343
> Maybe that's just the force of habit, but I must admit that it does feel really weird, not to have some kind of visual confirmation that an operation completed successfully.
> 
> I made this remark on several occasions in the related tracker issues, but @vboctor I don't think you ever commented on that.

vboctor:
> Maybe that's just the force of habit, but I must admit that it does feel really weird,
> not to have some kind of visual confirmation that an operation completed successfully.

@dregad I think we started off with a mix, some pages did switch directly and some did show operational successful page. I think it is partially that we are used to the old behavior. Typically what happens after the action is the page is updated or user is directed to a different page, so it is obvious that everything worked as expected. There were a case or two, where it was not obvious and I left them as is. Don't recall the exact cases, but it was a form where values are applied but the form looks exactly the same. In such case a feedback is useful.

For all cases, we can consider in the future a js based feedback bubble that self-dismiss itself after a couple of seconds. But that will be a bigger change to consider in the future. I would start with this change and see if there is feedback.

PR #1913, opened 2023-09-03
Remove usage of function mcrypt_create_iv

atrol:
Starting with MantisBT version 2.26.0 our PHP minimum version is 7.2.5. Function mcrypt_create_iv was deprectated in PHP 7.1.0, and removed in PHP 7.2.0

Fixes [#32900](https://www.mantisbt.org/bugs/view.php?id=32900)

### DIFF
--- a/core/crypto_api.php
+++ b/core/crypto_api.php
@@ -75,17 +75,6 @@ function crypto_generate_random_string( $p_bytes, $p_require_strong_generator = 
 		}
 	}
 
-	# Attempt to use mcrypt_create_iv - this is built into newer versions of php on windows
-	# if the mcrypt extension is enabled on Linux, it takes random data from /dev/urandom
-	if( !isset( $t_random_string ) ) {
-		if( function_exists( 'mcrypt_create_iv' ) ) {
-			$t_random_bytes = mcrypt_create_iv( $p_bytes, MCRYPT_DEV_URANDOM );
-			if( $t_random_bytes !== false && strlen( $t_random_bytes ) === $p_bytes ) {
-				$t_random_string = $t_random_bytes;
-			}
-		}
-	}
-
 	# Next we try to use the /dev/urandom PRNG provided on Linux systems. This
 	# is nowhere near as secure as /dev/random but it is still satisfactory for
 	# the needs of MantisBT, especially given the fact that we don't want this


dregad (changes requested):
Thanks for drawing my attention to this.

I believe that *crypto_generate_random_string()* requires much more cleanup than that. It was written by @davidhicks nearly 15 years ago, when there was no good, cross-platform way of generating cryptographically strong random data.

Since then, PHP 7 introduced the [random_bytes()](https://www.php.net/manual/en/function.random-bytes.php) function, which is generally available on all platforms and will use an appropriate CSPRNG for the operating system it is running on.

It's also worth mentioning that *openssl_random_pseudo_bytes()*, which is currently the "best" PRNG implemented in our code is **not considered cryptographically secure** anymore, according to [this article](https://paragonie.com/blog/2015/07/how-safely-generate-random-strings-and-integers-in-php).

Based on the above, *random_bytes()* should be used in replacement of the whole crypto_generate_random_string() function; that means the *$p_require_strong_generator* parameter becomes useless.

We should probably even consider getting rid of crypto_generate_random_string() completely (i.e. mark it as deprecated and have it throw a warning for now, then remove it in a future release of Mantis), and just call random_bytes() instead in the few places where it's used 
```
$ git grep crypto_generate_random_string
admin/install.php:1234: $t_crypto_master_salt = crypto_generate_random_string( 32 );
core/crypto_api.php:65:function crypto_generate_random_string( $p_bytes, $p_require_strong_generator = true ) {
core/crypto_api.php:153:        $t_random_string = crypto_generate_random_string( $p_bytes, true );
core/crypto_api.php:182:                $t_random_bytes = crypto_generate_random_string( $t_raw_bytes_required, false );
core/file_api.php:764:          $t_string = md5( crypto_generate_random_string( 32, false ) );
```

The following code block is also obsolete
https://github.com/mantisbt/mantisbt/blob/1254f0ddcbb7fd4eedb64325b2303097bf64ced5/core/crypto_api.php#L176-L183

Likewise, *crypto_generate_strong_random_string()* will be pretty much useless after this, as well as the ERROR_CRYPTO_CAN_NOT_GENERATE_STRONG_RANDOMNESS constant and related language strings
https://github.com/mantisbt/mantisbt/blob/1254f0ddcbb7fd4eedb64325b2303097bf64ced5/core/crypto_api.php#L152-L158

atrol:
@dregad I fully agree, but I didn't want to risk late regressions or introduce even more delay in getting out 2.26.0.
We didn't use random_bytes() unti now, so some testing in multiple platforms is needed.
My proposal is to merge as-is for 2.26.0 and to follow up on more cleanup in 2.27.0.

dregad:
> We didn't use random_bytes() until now, so some testing on multiple platforms is needed.

Honestly I don't think it's worth it. Keep in mind that it's a _standard PHP core function_ that has been out there for many years (PHP 7.0 came out in 2015), so any issues with it would definitely have come up by now. 

I also believe that the risk is negligible, the effort to implement very low and it improves security overall. This is why I'd rather put this in 2.26 scope.

If you don't have the time, I'm perfectly fine to do it myself.

atrol:
Using just random_bytes() in whole code would be nice, but I'm still a little worried as Random\RandomException can be thrown if no appropriate source of randomness can be found.
Does it mean that we would still need a less secure fallback?

davidhicks:
@dregad your assessment is good! I can't believe it has been 15 years since that code was written.

@atrol I'd be surprised if that exception would trigger on Linux systems, as https://www.php.net/manual/en/function.random-bytes.php states `/dev/urandom` is the source, and this source is non-blocking. Ask for 100PB of "random data" and you'll get it, just with reducing entropy because a little bit of actually random data will get stretched a lot by the CPRNG in the kernel. The exception to this would be if someone had an obscure platform or kernel configuration and PHP is monitoring the level of entropy in the pool (human equivalent: `watch -n .1 cat /proc/sys/kernel/random/entropy_avail`) and only reading if a certain level of entropy is available.

atrol:
@davidhicks nice to hear from you after such a long time.

Created  new PR https://github.com/mantisbt/mantisbt/pull/1917

PR #1917, opened 2023-09-05
Remove usage of function mcrypt_create_iv()

atrol:
Fixes [#32900](https://www.mantisbt.org/bugs/view.php?id=32900)

Replaces #1913


vboctor (approved):

dregad:
@atrol https://github.com/mantisbt/mantisbt/pull/1913#issuecomment-1705658533
> I'm still a little worried as Random\RandomException can be thrown if no appropriate source of randomness can be found.

Following up on that, I've done some research on the behavior of CSPRNG routines used by random_bytes() on [Linux](https://man7.org/linux/man-pages/man2/getrandom.2.html), [Windows](https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom) and [MacOS](https://keith.github.io/xcode-man-pages/arc4random.3.html). As far as I can tell, it is highly unlikely that any of these algorithms would fail to return data and cause an exception to be thrown on those platforms as under normal system operations they all produce an unlimited amount of random data (with potentially reduced randomness in the absence of sufficient entropy). 

That leaves us with the corner case of some exotic OS or system configuration not supported by PHP, or relying on a blocking PRNG. I think we can agree to ignore this. 

In any case, should that exception be thrown anyway, it would still be captured by our error handler as an *Internal application error* and halt execution which, IMO is the desired behavior and similar to what was happening before with crypto_generate_strong_random_string().

The only difference compared to the the previous implementation is that we would **now also get failure when a strong random string is not truly required** (e.g. file_api.php generating a unique file name). I believe the risk of that ever happening is very low, but if you think that's a problem, then we can easily catch the exception and implement a fallback, like @davidhicks original implementation https://github.com/mantisbt/mantisbt/blob/1254f0ddcbb7fd4eedb64325b2303097bf64ced5/core/crypto_api.php#L114-L136

Some of our usages of crypto_generate_uri_safe_nonce(), which was previously allowing fall-back to non-secure PRNG was debatable (generating cookie string or random password), so for these use cases we actually benefit from enforcing strong randomness.

We could decide to abstract the random_bytes() call in our own api function and always catch the exception to throw our own, more explicit error (ERROR_CRYPTO_CAN_NOT_GENERATE_STRONG_RANDOMNESS) instead if you think that's better in terms of user-friendliness and worth the effort. 

That being said, I'm fine with the proposed change as it is.

dregad (approved):

dregad:
_utility_api_ is no longer needed.

davidhicks:
This looks wrong: https://github.com/mantisbt/mantisbt/blob/66317ad3a3574f7335243eace295054f5cf20f46/core/session_api.php#L118

Shouldn't the session key be using `random_bytes` e.g. `$this->key = bin2hex(random_bytes(32))` rather than hashing two constant values together, one being the now-deprecated `$g_crypto_master_salt`? The documentation of the session key generation function states the key is expected to be the unique, which is not the case currently.

core/file_api.php
```diff
@@ -761,7 +761,7 @@ function file_clean_name( $p_filename ) {
  */
 function file_generate_unique_name( $p_filepath ) {
 	do {
-		$t_string = md5( crypto_generate_random_string( 32, false ) );
+		$t_string = md5( random_bytes( 32 ) );
```
davidhicks, line 764: I'd suggest `bin2hex(random_bytes(16))` instead, but really I doubt this matters :)

vboctor: I think there is minor benefit to having the newly generated filenames be consistent to existing ones in terms of length and format.

dregad:
That's weird, I got this notification by mail but can't see the
corresponding post in the Github interface... Anyway, 2 comments

On Thu, 7 Sept 2023 at 04:24, 'David Hicks' wrote:
> This looks wrong:
https://github.com/mantisbt/mantisbt/blob/66317ad3a3574f7335243eace295054f5cf20f46/core/session_api.php#L118
>
> Shouldn't the session key be using random_bytes e.g. $this->key =
bin2hex(random_bytes(32)) rather than hashing two constant values together,
one being the now-deprecated $g_crypto_master_salt? The documentation of
the session key generation function states the key is expected to be the
unique, which is not the case currently.

First, this key is used to compartiment the session store, and while it
does need to be unique, it also must be persistent otherwise it would not
be possible to retrieve stored session data, hence the use of constants. We
can't use random_bytes() here.

Second, $g_crypto_master_salt is **not** deprecated as part of this
PR, only the crypto_generate_random_string() and
crypto_generate_strong_random_string() functions are.

dregad:
Thanks for the clarification David, I thought that was the reason, but would have expected Github to show a "message deleted" or something in the thread.

atrol:
> _utility_api.php_ is no longer needed.

Removed it.
@dregad do you hace some tool to find unneeded require_api calls?

dregad:
> do you hace some tool to find unneeded require_api calls?

No, I just knew that is_windows_server() was part of it, and I checked manually if other functions were referenced.

PR #1919, opened 2023-09-16
Improving manage_user_page.php performance

dregad:
Follow-up on #1898, Fixes [#32787](https://mantisbt.org/bugs/view.php?id=32787)

Also includes some code cleanup and the fix for [32940](https://mantisbt.org/bugs/view.php?id=32940) (bundled in this PR for convenience)

### DIFF
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -306,7 +306,65 @@ function user_ensure_name_unique( $p_username, $p_user_id = null ) {
 }
 
 /**
+ * Returns a list of duplicate email addresses (case-insensitive).
+ *
+ * Each array entry's key is the e-mail address, and the value is an array
+ * listing the associated user accounts, with key=id and value=username.
+ *
+ * Note that the complex query used by this function is quite expensive (average
+ * execution time of about 0.3 seconds for 50K users with 1000 duplicates).
+ *
+ * @return array
+ */
+function user_get_duplicate_emails() {
+	static $s_duplicate_emails;
+
+	if( $s_duplicate_emails === null ) {
+		$t_user_table = db_get_table( 'user' );
+
+		# Using a sub-query within the IN clause's SELECT statement, as a workaround for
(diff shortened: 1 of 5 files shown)


manage_user_page.php
```diff
@@ -409,7 +416,8 @@
 				</td>
 				<td><?php echo string_display_line( $v_realname ) ?></td>
 				<td><?php
-					if( $t_unique_emails && !user_is_email_unique( $v_email, $v_id ) ) {
+					# Display warning icon if emails should be unique and a duplicate exists
+					if( array_key_exists( $v_email, $t_duplicate_emails ) ) {
```
atrol, line 420: Will not work with email addresses with uppercase characters in it,  as user_get_duplicate_emails() delivers lowercase email addresses.

core/user_api.php
```diff
@@ -305,9 +305,67 @@ function user_ensure_name_unique( $p_username, $p_user_id = null ) {
...
+	static $s_duplicate_emails;
+
+	if( $s_duplicate_emails === null ) {
+		$t_user_table = db_get_table( 'user' );
+
+		# Using a sub-query within the IN clause's SELECT statement, as a workaround for
+		# MySQL >= 5.7 with sql_mode=only_full_group_by throwing ERROR 1055 (42000):
+		# Expression #1 of HAVING clause is not in GROUP BY clause
+		$t_sql = <<< ENDSQL
+SELECT lower(email) email, id, username
```
atrol, line 329: Suspect that lower(email) must be removed to fix https://github.com/mantisbt/mantisbt/pull/1919/files#r1327981213

atrol:
I don't get any longer the list of duplicate emails on admin/check page.

manage_user_page.php
dregad: Good catch, thanks ! Will fix

PR #1923, opened 2023-09-19
No longer required to install a plugin for gravatar

g0tmi1k:


g0tmi1k:
I got it wrong

g0tmi1k:
> Support for avatar plugins – We still ship Gravatar out of the box, but it is now very easy to implement other avatar providers that can be based on file system or LDAP.


https://mantisbt.org/blog/archives/date/2016/06

PR #1925, opened 2023-09-20
Consistency with default config file

g0tmi1k:
Very mirror commits to address formatting item for each variable

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -3854,7 +3854,8 @@ $g_file_type_icons = array(
 	'xlt'	=> 'fa-file-excel-o',
 	'xml'	=> 'fa-file-code-o',
 	'zip'	=> 'fa-file-archive-o',
-	'?'	=> 'fa-file-o' );
+	'?'	=> 'fa-file-o'
+);
 
 /**
  *
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -3701,18 +3701,19 @@ $g_custom_headers = array();
  * information is displayed. Note that this will be bypassed (and caching is
  * allowed) for the bug report pages.
  *
+ * @todo The code that references this config considers it ON if it is set (doesn't use config_get_global() / $g_global_settings)
  * @global integer $g_allow_browser_cache
  */
-## TODO - the code that references this config considers it ON if it is set (doesn't use config_get_global() / $g_global_settings)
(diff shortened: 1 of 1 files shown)


config_defaults_inc.php
```diff
@@ -3698,14 +3703,15 @@
...
+$g_allow_browser_cache = OFF;
+
 /**
  * File caching - This will allow the browser to cache downloaded files.
  * Without this set, there may be issues with IE receiving files, and launching
  * support programs.
  * @global integer $g_allow_file_cache
  */
-# $g_allow_file_cache = ON;
+$g_allow_file_cache = OFF;
```
vboctor, line 3714: This will enable file caching. The reason is that the code that references this config considers it ON if it is set. This is wrong behavior. Hence, I would leave this config and the `$g_allow_browser_cache` disabled until:

- Update the code fetching them to use the standard config_get_global() method.
- Consider them ON only when their value is set to ON.
- Add these two config options to `$g_global_settings` array.

This can be done as part of this PR or as a separate PR.

vboctor (changes requested):
Thanks @g0tmi1k for your contribution. Looks good with the exception of uncommenting of two configs options relating to caching.

config_defaults_inc.php
g0tmi1k: Thanks for getting back to me @vboctor and the feedback.

Gotcha - good to know.
I don't have the time right now to learn where the changes need to go to make this happen :(.
I've altered the commit (and done a force push), to revert the commit explaining why they are not consistent like the others.

g0tmi1k:
Thanks @vboctor your welcome!
Please do say if there is anything else I can do with the pull request.

g0tmi1k:
Thanks for getting back to me @dregad.
Just updated the MR with your suggestion.****

dregad:
Merged in 6277ccb9bd77d27af416015da9fe5c80cd91a1ff

g0tmi1k:
Thanks @dregad !

g0tmi1k:
Was there anything I could of done better @dregad ?

dregad:
> Was there anything I could of done better @dregad ?

No it's fine, I just rebased the branch before merge to avoid spaghetti-plate git history - I try to keep it as linear as possible.

Thanks for your contribution.

g0tmi1k:
all good - make sense
(and thanks for keeping the code history clean!)

PR #1929, opened 2023-10-15
Markdown. Cleanup and fix tests

grummbeer:
The Tests `testHashLetters`, `testHashNumberAny` and `testHashLettersAny` failing.

Fix: https://mantisbt.org/bugs/view.php?id=30919 (step 4.)

### DIFF
--- a/plugins/MantisCoreFormatting/tests/MarkdownTest.php
+++ b/plugins/MantisCoreFormatting/tests/MarkdownTest.php
@@ -24,7 +24,7 @@
  */
 
 # Includes
-require_once( dirname( __FILE__, 2 ) . '../../../tests/TestConfig.php' );
+require_once( dirname( __FILE__, 2 ) . '/../../tests/TestConfig.php' );
 require_once( dirname( __FILE__, 2 ) . '/core/MantisMarkdown.php' );
 
 # MantisBT Core API
@@ -46,7 +46,7 @@ class MantisMarkdownTest extends PHPUnit\Framework\TestCase {
 	 */
 	public function testHashLetters() {
 		$this->assertEquals( '<h1>hello</h1>', MantisMarkdown::convert_text( '# hello' ) );
-		$this->assertEquals( '<h1>hello</h1>', MantisMarkdown::convert_text( '#hello' ) );
+		$this->assertEquals( '<p>#hello</p>', MantisMarkdown::convert_text( '#hello' ) );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


vboctor:
Just commenting on the PR description at this point.

I tested headers without space after the hash in github and [Markdown Lint](https://dlaa.me/markdownlint/) and in both cases, they didn't detect it as a header. I suggest that we do the same and not support "#h1" as a header. I also wonder if one day we would want to link "#abc" to tag "abc".

When testing the detection of bugs or comment references, we should be sure to factor in case where the mention is followed by a separator like `.`, `,`, or `;`.  For example, "This is fixed in issue #123456."

dregad:
> Valid Markdown header #h1 is not detected

This is actually NOT a valid header as per the [CommonMark spec](https://spec.commonmark.org/0.30/#atx-headings), that states:
*At least one space or tab is required between the # characters and the heading’s contents, unless the heading is empty*

So there can't be any conflict with bug links, which are defined as *$g_bug_link_tag* (`#` by default), **directly** followed by a continuous sequence of digits until the next word boundary (which correctly handles the `:`, `,`, `;`, etc. mentioned by @vboctor above).


Note that bug link matching is currently implemented like this:
https://github.com/mantisbt/mantisbt/blob/f996974cdbd659c2f0b7cc8ab353041e19b51e5b/core/string_api.php#L377-L381

grummbeer:
Thanks for the feedback and the to the CommonMark specs.

I guess than only the tests are wrong.

https://github.com/mantisbt/mantisbt/blob/f996974cdbd659c2f0b7cc8ab353041e19b51e5b/plugins/MantisCoreFormatting/tests/MarkdownTest.php#L49

Not `convert_text` returns the wrong value, the expected `<h1>hello</h1>` is the wrong part. Must be `<p>#hello</p>` then.

Should the regex be tested or the returned value from the `convert_text` method?

grummbeer:
> Should the regex be tested or the returned value from the `convert_text` method?

forget this.

grummbeer:
Thanks for your feedback, was on the wrong track. Tests are fixed and do pass again. Title and description updated accordingly.

At first i thought looking for the wanted (a bug mention at the beginning of a line) is the best practice, but now i know that this line ensures also the commonmark specs for headers.

This line also, because it _looks_ wrong and the ide complains.

https://github.com/mantisbt/mantisbt/blob/f996974cdbd659c2f0b7cc8ab353041e19b51e5b/plugins/MantisCoreFormatting/tests/MarkdownTest.php#L27

vboctor (approved):
Assuming the tests pass - looks good to me. Thanks.

dregad:
Thanks @grummbeer I have not yet had time to review this, will do so as soon as I can - kind of busy these days...

grummbeer:
Hi Damian, I'm still motivated, but honestly, I'm not sure if it's worth going ahead or not. There were already some nice attempts that addresses the Markdown stuff, but all failed for some reason … afraid of spending my love for the same result.

dregad:
Well it is your decision and I respect it; I think you already did most of the work for step 5 in https://github.com/mantisbt/mantisbt/compare/master...grummbeer:mantisbt:markdown-cleanup-05-formatting but that is now conflicting with master branch, so maybe it's just a question of resolving the merge conflict and submit a PR (I did not actually look a the code yet).

If you decide not to do it in the end that's fine... Just post a note in [#34040](https://mantisbt.org/bugs/view.php?id=34040) so we know and don't leave things hanging if nothing is going to happen.

grummbeer:
Yes, you're right, the base is already there and I've been running it for quite a while. So no reason to give up at this point. Will do it.

PR #1930, opened 2023-10-16
Remove generated history events for unchanged custom fields

vboctor:
Fixes #22408


dregad:
Have not looked at the code yet, but did you consider the earlier work and discussion in #1035 ?

vboctor:
@dregad I have hit the bug, implemented a fix, and then found the issue related to it, so I didn't take into consideration the previous work. I just had a vague memory that this bug was reported before.  The bug I hit was that multi-select custom field was always adding history events after editing an issue. That is called because "A|B" is not equals to "A|B|". Hence, I did the trimming calls in history events for such custom field type. I then added the check in the higher level business logic to avoid the write of custom field to the db if there is no change.

Feel free to combine the best of the two PRs or close this one.

@atrol +1 for getting 2.26.0 out as is, and not get more features in. I wasn't assuming this would be part of the release, since @dregad is cutting it soon as per my understanding.

PR #1931, opened 2023-10-16
Fix permalink creation when filtering based on projection

atrol:
Issue #32726


vboctor (approved):

dregad (approved):

dregad:
@atrol For this kind of so-obvious fixes there's really no need to open a PR, you can just commit directly...

PR #1933, opened 2023-10-29
Fix Unknown named parameter $files error on PHP 8+

dregad:
- Remove the array key when calling event_signal() in bugnote_add() and IssueNoteAddCommand::process().
- Update documentation

Fixes [#33058](https://www.mantisbt.org/bugs/view.php?id=33058)

### DIFF
--- a/core/bugnote_api.php
+++ b/core/bugnote_api.php
@@ -312,7 +312,7 @@ function bugnote_add( $p_bug_id, $p_bugnote_text, $p_time_tracking = '0:00', $p_
 
 	# Event integration
 	if( $p_trigger_event ) {
-		event_signal( 'EVENT_BUGNOTE_ADD', array( $p_bug_id, $t_bugnote_id, 'files' => array() ) );
+		event_signal( 'EVENT_BUGNOTE_ADD', array( $p_bug_id, $t_bugnote_id, array() ) );
 	}
 
 	# only send email if the text is not blank, otherwise, it is just recording of time without a comment.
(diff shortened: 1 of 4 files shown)


vboctor (approved):
Unless there are plugins that reference this argument via name somehow this should work. The reason I used a named array element was to move away from index based reference of event arguments, but I guess that approach won't work.

I would like to in the future move to a class based model:

```
class AddNoteEventArgs {
  private $argX;
  private $argY;
  private $argZ;

  constructor (X, Y, Z) { ... }

  function getX() {}

  function getY() {}

  function getZ() {}

  function getArgsForMantisVersion( $p_version ) {
     // used by eventing infrastructure to get the arguments array to pass to the event based
     // on the plugins MantisCore dependency version. This will pass an array for older plugins
     // and $this for newer plugins.
  }
}

// Here is how the signaling will look like:
event_signal ( ADD_NOTE_EVENT, new AddNoteEventArgs(x, y, z) );

// Still support old signaling if we want to support older plugins triggering events, but we
// should eventually retire the old signaling approach.
```

dregad:
> I would like to in the future move to a class based model

So I guess there should be a *EventArgs* parent class / interface, inherited by a bunch of event-specific classes ?

Smart. But is it really needed and useful? The only advantage is allowing named parameters. The plugin infrastructure does nothing with them except passing them on to the callbacks, which can currently give them any name they want.

On the down side it would be quite verbose (one class per event -> 89 as of today) so quite a lot of maintenance so overall I'm not sure about the added value.

PR #1944, opened 2023-11-28
Create strings_thai.txt

kongoon:


dregad:
Hello

Thanks for your contribution, but unfortunately we can't accept it as translation  updates ara managed via https://translatewiki.net/

PR #1945, opened 2023-11-30
Fix custom field display on change status page

atrol:
Fixes [#33358](https://www.mantisbt.org/bugs/view.php?id=33358)

### DIFF
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -281,13 +281,9 @@ layout_page_begin();
 		$t_show_custom_field = $t_require;
 
 		if( !$t_show_custom_field ) {
-			if( 'update' != $t_custom_status_label ) {
-				$t_show_custom_field = true;
-			}
 			if( in_array( $t_custom_status_label, array( 'resolved', 'closed' ) ) && $t_display ) {
 				$t_show_custom_field = true;
-			}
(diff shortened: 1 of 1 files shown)


dregad (approved):
I didn't analyze the change in logic but I trust that you did.

Just a suggestion for optimization: with the removal of the `if( 'update` check, the remaining logic is no longer necessary and the whole block 283..296 could be simplified:

```php
if( !$t_show_custom_field && in_array( $t_custom_status_label, array( 'resolved', 'closed' ) ) && $t_display ) {
	$t_show_custom_field = true;
} else {
	$t_plugin_Event_result = event_signal( 'EVENT_UPDATE_BUG_SHOW_CUSTOM_FIELD', array( $t_bug, $t_id ) );
	foreach( $t_plugin_Event_result as $t_event_result ) {
		if( in_array( true, $t_event_result ) ) {
			$t_show_custom_field = true;
			break;
		}
	}
}
```

vboctor (approved):

atrol:
> I didn't analyze the change in logic but I trust that you did

You shouldn't ;-)
 
> Just a suggestion for optimization: with the removal of the `if( 'update` check, the remaining logic is no longer necessary and the whole block 283..296 could be simplified:

Done

PR #1947, opened 2023-12-01
Fix SQL error opening Manage Users page with MSSQL

dregad:
[Microsoft][ODBC Driver 17 for SQL Server][SQL Server]No column name was specified for column 2 of 'tmp'.

Adding aliases to subquery's columns.

Fixes [#33372](https://www.mantisbt.org/bugs/view.php?id=33372)

### DIFF
--- a/core/user_api.php
+++ b/core/user_api.php
@@ -331,7 +331,7 @@ FROM {$t_user_table}
 WHERE lower(email) IN (
 	SELECT email 
 	FROM (
-		SELECT lower(email) email, COUNT(*)
+		SELECT lower(email) AS email, COUNT(*) AS mycount
 		FROM {$t_user_table}
 		GROUP BY lower(email) HAVING COUNT(*) > 1
 		) tmp


vboctor:
Should the having statement use `mycount` rather than `count(*)`?

dregad:
> Should the having statement use `mycount` rather than `count(*)`?

That won't work actually...

### MySQL
```
mysql> SELECT lower(email) email, id, username FROM mantis_user_table 
WHERE lower(email) IN (
  SELECT email  FROM ( SELECT lower(email) email, COUNT(*) FROM mantis_user_table  
  GROUP BY lower(email) HAVING mycount > 1 ) tmp 
) ORDER BY lower(email), username;
ERROR 1054 (42S22): Unknown column 'mycount' in 'having clause'
```

### PostgreSQL
```
bugtracker=# SELECT lower(email) email, id, username FROM mantis_user_table 
bugtracker-# WHERE lower(email) IN (
bugtracker(#   SELECT email  FROM ( SELECT lower(email) email, COUNT(*) FROM mantis_user_table  
bugtracker(#   GROUP BY lower(email) HAVING mycount > 1 ) tmp 
bugtracker(# ) ORDER BY lower(email), username;
ERROR:  column "mycount" does not exist
LINE 4:   GROUP BY lower(email) HAVING mycount > 1 ) tmp 
```

vboctor:
Maybe because you don't have AS mycount in the query?

dregad:
Sorry I copy/pasted the SQL from the wrong branch 😊 let me try again

dregad:
So MySQL is OK with the *AS* but PostgreSQL still chokes

```
bugtracker=# SELECT lower(email) email, id, username FROM mantis_user_table WHERE lower(email) IN ( SELECT email  FROM ( SELECT lower(email) AS email, COUNT(*) AS mycount FROM mantis_user_table GROUP BY lower(email) HAVING mycount > 1 ) tmp ) ORDER BY lower(email), username;
ERROR:  column "mycount" does not exist
LINE 1: ...OM mantis_user_table GROUP BY lower(email) HAVING mycount > ...
```

dregad:
I can't actually test it, but most likely this wouldn't work with MSSQL either according to some quick research I did.

Further reading https://dba.stackexchange.com/questions/21965/why-are-queries-parsed-in-such-a-way-that-disallows-the-use-of-column-aliases-in/

vboctor (approved):

PR #1952, opened 2023-12-04
33402 fix - Updating an Issue through the API sets all comments last edit timestamp

Ephaltes:
Fix problem [33402](https://www.mantisbt.org/bugs/view.php?id=33402)

problem was, that it always compares the first bugnote to all other bugnotes


dregad (approved):

dregad:
Merged in e03d0f63dc57e710bc4380aea02e086b98d06309

Thanks for your contribution.

PR #1953, opened 2023-12-05
[WIP] Fix add monitor regressions

atrol:
Fixes #33404
Fixes #33348

### DIFF
--- a/core/commands/MonitorAddCommand.php
+++ b/core/commands/MonitorAddCommand.php
@@ -97,7 +97,7 @@ class MonitorAddCommand extends Command {
 				/* user */ null,
 				$this->projectId );
 
-			if( !access_has_bug_level( $t_access_level, $t_issue_id, $t_user_id ) ) {
+			if( !access_has_project_level( $t_access_level, $this->projectId, $t_user_id ) ) {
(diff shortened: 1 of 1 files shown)


dregad:
@atrol considering the discussion in https://mantisbt.org/bugs/view.php?id=33404, particularly @vboctor's last note, I thought I should merge this PR but then I noticed you set it as _Draft_, not sure what is the reason for that. Let me know if this is OK to merge or if you are still working on it.

atrol:
As mentioned at https://mantisbt.org/bugs/view.php?id=33404#c68390
After the change proposed by Victor (currently implemented in this WIT PR), we are back to where we have been in 2.25.8.
This means you are able to add users without access to private issues to the monitoring list, but after that, the added user is not able to access the issue.
@dregad @vboctor isn't the current 2.26.0 behaviour better than what we had in 2.25.8?

vboctor:
I think we should merge the fix. The fix addresses case where users being added would have access to the issue, but were not being able to be added. If there are other corner cases we need to address, we should open an issue with exact repro steps and discuss if we can address that without breaking valid functionality. I think the fix should be part of 2.26.1.

atrol:
> The fix addresses case where users being added would have access to the issue, but were not being able to be added. 

@vboctor seems I have not the right configuration for such a case. Which settings are needed to get it?

PR #1956, opened 2024-01-07
Use config API to access allow_browser_cache

dregad:
Until now $g_allow_browser_cache config was handled in a non-standard way in in http_caching_headers(), i.e. by referencing it as a global variable instead of retrieving its value with config_get_global().

Moreover, the setting was commented out in config_default_inc.php (i.e. not defined) instead of being set to OFF.

This follows up on discussion in PR #1925.

Fixes [#33482](https://mantisbt.org/bugs/view.php?id=33482)

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -4281,18 +4281,14 @@ $g_custom_headers = array();
 /**
  * Browser Caching Control.
  *
- * By default, we try to prevent the browser from caching anything.
- * $g_allow_browser_cache will defeat this for some cases.
+ * By default, we try to prevent the browser from caching anything. This setting
+ * will allow the browser to cache all pages. The upside will be better
+ * performance overall, but there may be cases where outdated information is
+ * displayed.
  *
- * Browser Page caching - This will allow the browser to cache all pages. The
- * upside will be better performance, but there may be cases where obsolete
- * information is displayed. Note that this will be bypassed (and caching is
- * allowed) for the bug report pages.
- *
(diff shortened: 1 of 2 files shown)


dregad:
Please read https://mantisbt.org/bugs/view.php?id=33482#c68427 as part of review

atrol (approved):
I am fine with the changes.

As a side note: Using config_get_global is about 7 times slower than direct access of the variable.

dregad:
Thanks for the feedback @atrol 

> Using config_get_global is about 7 times (xdebug.mode = debug) / 11 times (without xdebug) slower than direct access of the variable.

Which is to be expected of course, although I'm surprised it's actually faster with Xdebug than without...

I just followed @vboctor's note in https://github.com/mantisbt/mantisbt/pull/1925#discussion_r1332400620
> Update the code fetching them to use the standard config_get_global() method.

atrol:
> I'm surprised it's actually faster with Xdebug than without...

@dregad It's not faster with Xdebug.
Seems my formulation was not clear.

> As a side note: Using config_get_global is about 7 times (xdebug.mode = debug) / 11 times (without xdebug) slower than direct access of the variable.

I compared two times config_get_global calls with direct access of the variable.
With Xdebug enabled, config_get_global is about 7 times slower than direct access of the variable, (e.g. 7 seconds vs. 1 second)
With Xdebug disabled, config_get_global is about 11 times slower than direct access of the variable. (e.g. 5.5 seconds vs.  0.5 seconds)

dregad:
Thanks for the clarification.

vboctor (approved):

PR #1957, opened 2024-01-18
Adding Issues by project to MantisGraph

CasN:
This addresses issue [33521](https://mantisbt.org/bugs/view.php?id=33521)


plugins/MantisGraph/MantisGraph.php
```diff
@@ -64,7 +64,7 @@ function register() {
 
 		$this->author = 'MantisBT Team';
 		$this->contact = 'mantisbt-dev@lists.sourceforge.net';
-		$this->url = 'https://mantisbt.org';
+		$this->url = 'https://www.mantisbt.org';
```
dregad, line 67: Why are you changing this, the URL is correct

plugins/MantisGraph/MantisGraph.php
```diff
@@ -227,7 +227,7 @@ private function get_url_with_filter( $p_page ) {
 	 */
 	function summary_menu() {
 		$t_menu_items[] = '<a href="'
-			. $this->get_url_with_filter( 'developer_graph.php' )
+			. $this->get_url_with_filter( 'project_graph.php' )
```
dregad, line 230: Changing the default graph is fine, but it needs to be reflected in all graph scripts
```php
print_summary_menu( 'project_graph.php', $t_filter );
```
Otherwise the *Graphs* tab will no longer be marked as active.

plugins/MantisGraph/lang/strings_english.txt
```diff
@@ -88,3 +88,6 @@ $s_plugin_MantisGraph_graph_topdev = 'Top Developers by Fixed Issues';
 $s_plugin_MantisGraph_graph_opendev = 'Developers by Open Issues';
 $s_plugin_MantisGraph_graph_topreporter_fixed = 'Top Reporters by Fixed Issues';
 $s_plugin_MantisGraph_other_categories = 'other categories (%d)';
+
+$s_plugin_MantisGraph_other_projects = 'other projecst (%d)';
```
dregad, line 92: Typo

plugins/MantisGraph/lang/strings_english.txt
```diff
@@ -88,3 +88,6 @@ $s_plugin_MantisGraph_graph_topdev = 'Top Developers by Fixed Issues';
...
+$s_plugin_MantisGraph_graph_imp_project_title = 'By Projects Graphs';
```
dregad, line 93: Not sure why you're tagging these 2 strings at the end, it would be more consistent to put them together with related strings. Also the `imp` prefix is probably the result of a copy/paste, not needed IMO (TBH, I'm not even what its meaning is/was).

dregad (approved):
@CasN many thanks for your contribution. This looks good (mostly) and tested OK.

I'll make the necessary adjustments to fix my review comments and merge.

dregad:
Also, to keep in mind for future contributions: 

When referring to Mantis issues in your commit message, you should always follow the regex from source integration plugin (`(?:fixe?d?s?|resolved?s?)+\s*:?\s+(?:#(?:\d+)[,\.\s]*)+`), to ensure that when it is merged
- the commit is automatically linked to the issue(s)
- the issue(s) is/are marked as resolved

In other words
- DON'T: `This addresses issue 33521`
- **DO**: `Fix #33521`, `Fixes #33521` or `Resolves #33521`, etc.

You can reference multiple issues if needed, e.g. `Fix #1, #2...` and if you just want to link an issue without affecting the resolved status, you can use `(?:bugs?|issues?|reports?)+\s*:?\s+(?:#(?:\d+)[,\.\s]*)+`, i.e. `Bug #1`, `Issues #2, #3` etc.

PR #1958, opened 2024-01-18
Allowing plugins to send array to function print_submenu

CasN:
This addresses issue [33522](https://mantisbt.org/bugs/view.php?id=33522)


dregad (changes requested):
This is introducing code duplication. Please introduce a new API function, e.g. *print_submenu_item()* and call it from within the foreach.

Also see my note https://mantisbt.org/bugs/view.php?id=33522#c69171

I think PHPDoc for plugin_menu_items() should be updated.

dregad:
For consistency, I believe we should do the same for print_menu() as well.

https://github.com/mantisbt/mantisbt/blob/70f2c4a98e87147009c8644bb0f051c3d1de6b57/core/html_api.php#L646-L651

PR #1965, opened 2024-02-13
Update class.ObjectIterator.inc.php

tm8544:
Prevent exception in creating RSS when there are no issues to publish


dregad:
After review and test, I believe it would be a better approach to initialize ObjectList::$objects property to an empty array, this way it will never be null and we won't hit the TypeError in ObjectIterator::size().

dregad:
Fixed in 9a8f1277c1b0a2577ffee9c110e56d73a2ae7f10

PR #1966, opened 2024-02-13
Update class.ObjectIterator.inc.php

tm8544:
Prevent exception in creating RSS when there are no issues to publish. This is more general approach than previous commit.


dregad:
> This is more general approach than previous commit.

Since this is a bugfix, it will be targeted at 2.26 branch, where the minimum PHP version is 7.2; is_countable() was added in 7.3 so I'd rather stick to the original approach in #1965.

PR #1967, opened 2024-02-14
Fix overflowing text in collapsed sidebar menu

dregad:
- CSS now truncates text and appends an ellipsis if it is too wide
- A title attribute is added to those menu items, so the truncated text is visible in full when hovering over it

Fixes [#33651](https://mantisbt.org/bugs/view.php?id=33651)

New behavior:

image


dregad:
Cancelled in favor of #1968 following discussion in [#33651](https://mantisbt.org/bugs/view.php?id=33651).

PR #1971, opened 2024-02-19
Revert unrelated changes to prevent a regression

atrol:
Changes for issue [#33521](https://www.mantisbt.org/bugs/view.php?id=33521) introduce a regression.
Without undoing, we will get again [#32459](https://www.mantisbt.org/bugs/view.php?id=32459).


dregad (approved):

dregad:
Thanks @atrol

PR #1972, opened 2024-02-20
Travis builds timing out since merge of GHSA-mcqj-7p29-9528

dregad:


atrol:
@dregad suspect you have to add now also `path`  to the install.hpp URL.

dregad:
> @dregad suspect you have to add now also `path` to the install.php URL.

That was my first guess too, but I just tried that and it's still not working.

I do not understand why the curl call is not returning any output from the installer.

dregad:
And it's definitely the change to the installer that's breaking it, I just reverted and the build succeeded.

dregad:
Even more strange, I ran the curl command locally and everything worked as expected 😕

dregad:
When setting `path=http://$HOSTNAME:$PORT/` it does not work, maybe some url encoding issue ?

Anyway it works OK with an empty path so let's use that.

PR #1332, opened 2018-04-13
WIP: Fix Markdown rendering issues

dregad:
I started working on a branch targeted at fixing the various rendering issues following the enabling of Parsedown safe mode ([#24186](https://www.mantisbt.org/bugs/view.php?id=24186)).

The work item covers refactoring of the following MantisCoreFormatting plugin features:
- [x] user mentions
- [x] URL and e-mail addresses
- [ ] Bug and Bugnote linking

This is still a work-in-progress at this time, but in the interest of getting some early testing and feedback, hopefully also by the people currently affected by the issue who are facing the difficult choice between disabling Markdown or leaving their MantisBT instance vulnerable to XSS, I thought I'd open a PR now.

Warning: this feature branch is subject to change (including rebasing) without notice.

Note: The PR currently includes a Composer update to a beta version of Parsedown, but it should also be possible to fix the issue (in a less elegant / complete way) with current 1.7.1 version. I'll decide which version to use later on (depending on when Parsedown 1.8.0 is released).

Fixes [#24628](https://www.mantisbt.org/bugs/view.php?id=24628), [#24241](https://www.mantisbt.org/bugs/view.php?id=24241)

PR #1976, opened 2024-03-14
Markdown processing. Cleanup

grummbeer:
This is the 5th step in cleaning up the Markdown processing and a quite nice reduction of ~code and~ complexity. It’s a huge improvment of the status quo.

#### Fixes 

- [#34040](https://mantisbt.org/bugs/view.php?id=34040) 5. cleanup formatting process 
- [#22315](https://mantisbt.org/bugs/view.php?id=22315) Markdown converts " to `&quot;` within code blocks and inline code
- [#22320](https://mantisbt.org/bugs/view.php?id=22320) Don't expand issue ids into URLs within code blocks
- [#24241](https://www.mantisbt.org/bugs/view.php?id=24241) $g_html_valid_tags are not rendered if Markdown is enabled
- [#24628](https://www.mantisbt.org/bugs/view.php?id=24628) Double quotes " and lesser than sign < are shown as HTML entity within Markdown code blocks
- [#24810](https://mantisbt.org/bugs/view.php?id=24810) Markdown links/code always show HTML entities for Ampersand and Less-than sign
- [#22231](https://mantisbt.org/bugs/view.php?id=22231) Fix unit tests for markdown
- [#23738](https://mantisbt.org/bugs/view.php?id=23738) Mantis issue links displayed as raw HTML in code block

#### Candidates 

- [#23106](https://mantisbt.org/bugs/view.php?id=23106) Markdown delete link target _blank by default
- [#28138](https://mantisbt.org/bugs/view.php?id=28138) notes feature: allowable html tags are post fixed with a closed br tag (see last comment, code block is recommended)
- [#14043](https://mantisbt.org/bugs/view.php?id=14043) HTML List Tags break because of newlines (Its about HTML … but suggested by reporter)

#### Replaces 

- #1332
 

### How, in short 

- Process input via Markdownparser and return the result, no further text processing.
- During parsing, catch all `<code>`-blocks and replace them with a hash value.
- After the markup is returned from Parsedown, apply mentions and links
- Restore the untouched `<code>`-Blocks back in place.

<details>
<summary>screenshot</summary>

2024-03-14 19 16 04 mantisbt local 7820f57ed959

</details>

### possible follow ups

Will add a tickets for them. 

- harmonize look for inline code and code blocks [#22181](https://mantisbt.org/bugs/view.php?id=22181) [#22485](https://mantisbt.org/bugs/view.php?id=22485)
- Add prism, with ability to select the languages and plugins
- Modernize plugin class / PHP language level
- Think about splitting or restructuring the formatting process
- evaluate a CS fixer. https://github.com/dregad/mantisbt/tree/codesniffer.
- Add something to show workarounds/tricks for quirks like this [#22344](https://mantisbt.org/bugs/view.php?id=22344)
- evaluate HTMLPurifier / see input of HTML link ($g_html_valid_tags, $g_html_valid_tags_single_line, …) see https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2018532773
 
<details>
<summary>screenshot</summary>

2024-03-14 19 02 52 mantisbt local 386712bd3fd7

</details>


grummbeer:
Add prism

Please review. 

- How matching the unmatched emails adddresses?
- What parts or internals have I missed?

grummbeer:
> How matching the unmatched emails adddresses?

Solved, was already in a previous state, but forgotten.

grummbeer:
@atrol is this what you expected?

2024-03-14 19 20 03 mantisbt local aaaa5449f8cf

grummbeer:
Handline of tags and `process_urls = OFF`
---

```
<info@example.com>
<https://exmaple.com>
[example.com](https://example.com)
```

This inline tags are not covered by `$this->setUrlsLinked( false );`, so this is useless for the requirement of no links when `process_urls = OFF`.

### Two approaches

1. Strip links from generated markup (what i currently did here)

```php
if( self::$mantis_markdown->config_process_urls == OFF ) {
   $t_markup = string_strip_hrefs( $t_markup );
}
```

But this means grepping the whole stuff.

2. Convert anchors to span elements during runtime

```php
protected function element( array $Element )
{
    if( $Element['name'] === 'a' && $this->config_process_urls == OFF ) {
        $Element = [
            'name' => 'span',
  	    'text' => $Element['text']
        ];
    }
}
```

Opinions on this?

grummbeer:
Further consideration.

Now, since the process has a clear single last step `finalizeMarkup()`, this could moved to the Plugin. The text processing logic would then only be at one place and not scattered.

- Add config parameter via caller (no need to get them twice, plugin and immediately again in Markdown class)
- Parser returns an array with markup and codeblocks


```php
// MantisCoreFormatting
if( ON == $s_markdown ) {
    if( $p_multiline ) {
        $t_result = MantisMarkdown::convert_text( $t_string, $s_urls, $s_buglinks );
    } else {
        $t_result = MantisMarkdown::convert_line( $t_string, $s_urls, $s_buglinks );
    }

    // $_result = [string: 'markup', array: 'codeblocks'];
        
    $t_markup = $t_result['markup'];

    if( $s_urls == OFF ) {
	$t_markup = string_strip_hrefs( $t_markup );
    }

    if( $s_buglinks == ON ) {
	$t_markup = string_process_bugnote_link( $t_markup );
	$t_markup = string_process_bug_link( $t_markup );
    }

    $t_markup = mention_format_text( $t_markup );

    foreach( $t_result['codeblocks'] as $t_hash => $t_code ) {
        $t_markup = str_replace( $t_hash, htmlspecialchars($t_code), $t_markup );
    }

    return $t_markup;
}
```

dregad:
Many thanks @grummbeer for this excellent work ! It is much appreciated and indeed a great improvement over the current code.

I have done some limited, informal testing and so far found no issues with your PR. I have not yet had time to fully review the code itself, but here are some general remarks to start with.

1. Replacing code blocks with a hash value to prevent them from being processed twice is a great idea, leading to code simplification. It seems to work perfectly as far as I can tell.

2. Adding Prism for syntax highlighting of code blocks is a nice touch, I really like it. That should be tracked as a separate issue (feature) at https://mantisbt.org/bugs though, so please open one and reference it in your commit message (i.e. `Fixes #xxx). Also, make sure all the relevant code is contained in a single commit (at the moment the update to *MantisCoreFormatting::resources()* is in c15a86fc instead of 628b83a9)

3. Based on the screenshots you posted, you have obviously built some nice and fairly comprehensive test cases for this. Since I unfortunately lost the ones I had prepared when working on #1332 when I changed to a new computer, it would be useful if you could post yours here for reference. They could be used to improve the PHPUnit tests, which are quite limited at the moment. 

4. Regarding the skipped *testHashNumbers()* failing test case, I don't understand what you mean by *unclear state, fail on auth* (what does auth have to do with this ?). I'm not really sure about the intention of the original author either. All I can say is that `#1` should be interpreted as a bug link reference (assuming that A. `1` is an existing issue and B. *$g_bug_link_tag* is set to its default value of `'#'`), which is apparently exactly what `MantisMarkdown::convert_text( '#1' )` is doing. So maybe instead of skipping, we should just fix the expected value ?

> Handling of tags and process_urls = OFF (https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-1999214200)
[...]
requirement of no links when process_urls = OFF 

My understanding of *process_urls*'s purpose, is that when ON it should process "plain" URL and e-mail addresses (in addition to Markdown links), e.g. 
```
URL http://example.com and EMAIL user@example.com
```
should be converted to 
```
URL <a href="http://example.com">http://example.com</a> and EMAIL <a href="mailto:user@example.com">user@example.com</a>
```
which AFAIK CommonMark does not do. In other words, `<xxx>` (autolinks) and other Markdown links like `[text](url)`.  should IMO not be affected by *process_urls* setting.

> Further consideration (https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-1999324037)

I'm not sure I'm following your thoughts here. Can you please clarify what feedback you expect ?

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -52,9 +52,19 @@ class MantisMarkdown extends Parsedown
...
-	 * @var string inline style
+	 * @var int plugin configuration
 	 */
-	private $inline_style = null;
+	private $config_process_buglinks = OFF;
+
+	/**
+	 * @var int plugin configuration
+	 */
+	private $config_process_urls = OFF;
```
dregad, line 62: The properties are initialized in constructor, so it's pointless to assign a default value.

The same goes for *$table_class*, null is the default anyway.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -205,57 +175,38 @@ protected function inlineUrl( $Excerpt ) {
...
-	 *
-	 * MantisBT text processing replaces '&' signs by their entity name. Within
-	 * code blocks or backticks, Parsedown applies the same transformation again,
-	 * so they ultimately become '&amp;amp;'. This reverts the initial conversion
-	 * so ampersands are displayed correctly.
+	 * Customize the blockHeader markdown
 	 *
-	 * @param string $p_text Text block to process
-	 * @return void
+	 * @param string $line The Markdown syntax to parse
```
dregad, line 185: it's not a string, it's an array

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -205,57 +175,38 @@ protected function inlineUrl( $Excerpt ) {
...
+	 * @access protected
+	 * @return string|null HTML representation generated from markdown or null
```
dregad, line 187: Should be `array[]|null` or even `array[]|void|null` since the parent Parsedown method calls `return` without a value.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -205,57 +175,38 @@ protected function inlineUrl( $Excerpt ) {
...
-		$p_text = str_replace( '&amp;', '&', $p_text );
+	protected function blockHeader( $line ) {
+		# Header detection logic
+		# - the opening # may be indented 0-3 spaces
+		# - a sequence of 1–6 '#' characters
+		# - The #'s must be followed by a space or a newline
+		if ( preg_match( '/^ {0,3}#{1,6}(?: |$)/', $line['text'] ) ) {
+			return parent::blockHeader($line);
+		}
 	}
```
dregad, line 198: Missing `return null;` here I suppose

dregad (changes requested):

grummbeer:
Wow! Thank you very much for this nice review!

> Prism

I've added an issue to the tracker [#34124](https://mantisbt.org/bugs/view.php?id=34124) and rebased the commits to reflect this. It already was, but it got messed up at one point during the rebase. Thanks for finding it!

> Markdown test

How to share as plain text? Created my first gist (in hope you can access it)https://gist.github.com/grummbeer/aceb2e76fac60e93df5fae9474154175


> process_urls

`process_urls == OFF`

```
URL http://example.com and EMAIL user@example.com
```

This both are ignored but `<https://example.com>` not? Sounds not logically to me. Why, what's the difference? If `process_urls` is not wanted, then no urls/emails should be clickable at all.

> Test

Of course, it should be fixed. Let me investigate, will came back later, now breakfast with family :-)

the trace is this.
```
 /mantisbt/core/print_api.php:135
 /mantisbt/core/access_api.php:83
 /mantisbt/core/authentication_api.php:1137
 /mantisbt/core/access_api.php:508
 /mantisbt/core/string_api.php:352
 /mantisbt/core/string_api.php:377
 /mantisbt/plugins/MantisCoreFormatting/core/MantisMarkdown.php:230 // $t_markup = string_process_bug_link( $t_markup );
 /mantisbt/plugins/MantisCoreFormatting/core/MantisMarkdown.php:107
 /mantisbt/plugins/MantisCoreFormatting/tests/MarkdownTest.php:78
 
```

atrol:
> @atrol is this what you expected?

Thirst of all, thanks a lot @grummbeer for working on that!

I had a very short look at it until now. The results look good.

atrol:
@dregad concerning test data:  I started with https://github.com/atrol/testdata/blob/master/MantisCoreFormatting.txt some years ago.

grummbeer:
> Further consideration.

The idea is moving the text processing logic from `MantisMarkdown::finalizeMarkup()` to `MantisCoreFormattingPlugin::formatted()`. The tes processing is then an only one place. 

I am considering doing this in a further step to keep this one smaller.

grummbeer:
> process_urls

Ah! You mean when it is intended to display urls that should not be clickable while in the same text there should be links which should be clickable?

```
This url https://example.com is dangerous please see <https://example.com/why-is-this-link-dangerous>
```

`procss_urls = OFF`

```
This url https://example.com is dangerous please see 

<a href="https://example.com/why-is-this-link-dangerous">
https://example.com/why-is-this-link-dangerous
</a>
```

grummbeer:
fixed some doc-block and rebase

grummbeer:
@atrol what is your opinion on `process_urls`?

when `process_urls = OFF`

- `<https://example.com> / [link](https://example.com)` became a link but `https://example.com` not?
- `<user@example.com>` became a mailto link but `user@example.com` not?
 

I am not sure, will implement it as you (maintainers) want.

plugins/MantisCoreFormatting/MantisCoreFormatting.php
```diff
@@ -158,6 +160,21 @@ function formatted( $p_event, $p_string, $p_multiline = true ) {
...
 		$t_string = $p_string;
 
+		if( null === $s_markdown ) {
+			$s_markdown = plugin_config_get( 'process_markdown' );
+		}
+
+		# Process Markdown and return content, no further text processing.
+		if( ON == $s_markdown ) {
+			if( $p_multiline ) {
+				$t_string = MantisMarkdown::convert_text( $t_string );
```
atrol, line 170: As there is no further text processing, you can return the value without assigning to a variable.

plugins/MantisCoreFormatting/MantisCoreFormatting.php
```diff
@@ -158,6 +160,21 @@ function formatted( $p_event, $p_string, $p_multiline = true ) {
...
+			} else {
+				$t_string = MantisMarkdown::convert_line( $t_string );
```
atrol, line 172: Same as above.

grummbeer: fixed and rebased

```php
if( $p_multiline ) {
   return MantisMarkdown::convert_text( $t_string );
} else {
   return MantisMarkdown::convert_line( $t_string );
}
```

Would a ternary here be OK for you?

```php
return $p_multiline
  : MantisMarkdown::convert_text( $t_string )
  ? MantisMarkdown::convert_line( $t_string )
;
```

atrol: Ternary operator was introduced with PHP 5.3, so @grummbeer that's ok, as starting from version 2.26.0 we support just PHP >= 7.2.5 see https://www.mantisbt.org/bugs/view.php?id=27840
Starting from 2.27.0 it will be PHP >= 7.4,0 see https://www.mantisbt.org/bugs/view.php?id=32808

vboctor:
Thanks @grummbeer for your contribution and thorough work.

Any chance we can capture the tests you shared via a screenshot as test cases in MarkdownTest test fixture?

grummbeer:
> Tests

Yes, but …

- bugs `#1`, bug note `~1` and user`@user` are dependent on existing Database entries.
- bugs `#1` and bug notes `~1` lead to `authentication_api.php::access_denied()`. 
- Anchors, especially the attributes, are dependent on settings which are not changeable/configurable by the test. I guess making this testable means making  `helper_api::helper_get_link_attributes` testable/configurable first.

Not sure, but have a strong feeling that the resolving of mentions is somthing what should covered by testing the core functions

- mention_format_text
- string_process_bug_link
- string_process_bugnote_link
- …

The formatting process is using them, not providing them. 


What i think at the moment is:

### Testing the parser

Processing text as Parsedown does. This all is [already testet](https://parsedown.org/tests/)

**MantisBT specific, what is to be tested**

The parser mainly do three thnings.

### 1. Detecting issue mentions

There is already a test for this.

```php
public function testHashNumbers() {
    $this->assertEquals( '<p>#1</p>', MantisMarkdown::convert_text( '#1' ) );
}
``` 

The purpose of "parser" is not resolving (mantis) tokens to somewhat. The `<p>#1</p>` is what expected. It's not a header element. Resolving the bug mention is part of the "formatted()" function. So what i mentioned as "[further consideration](https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-1999324037)" is the path to go.

### 2. Replacing `<code>` elements with a hash

Perfectly testable

```php
// do the same for multiline "```" codeblocks
public function testInlineCode() {
    $token = '@foobar';
    $hash = md5($token); // or better MantisMarkdown::hash($token);

    $this->assertSame(
        '<p>lorem <code>'.$hash.'</code> @user</p>',
	MantisMarkdown::convert_text( 'lorem `'.$token.'` @user' )
    );

    // something like
    // $this->assertCount(1, MantisMarkdown::codeblocks);
    // $this->assertTrue(array_key_exists($hash, MantisMarkdown::codeblocks));
    // $this->assertSame($token, MantisMarkdown::codeblocks[$hash]);
}
```

Again, the retoring of the untouched `<code>` block happens in `formatted()`. Also the resolving of the `@user` mention.


### 3. Adding attributes to Markdown anchors

`<https://example.com>` `<user@exmaple.com>` `[link](https://example.com)`

The interessing here are the attributes, but i don't know how to test in a consist way …

Will give this a try, with own commits, so they can easily removed.

plugins/MantisCoreFormatting/MantisCoreFormatting.php
grummbeer: Thanks

dregad:
Warning: long, detailed answer below 🤓 
I tried to keep it as structured as possible to facilitate further discussion if needed.

## Markdown test data

Re: https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2002377176 and https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2002402762
> - https://gist.github.com/grummbeer/aceb2e76fac60e93df5fae9474154175
> - https://github.com/atrol/testdata/blob/master/MantisCoreFormatting.txt

Thanks, that's helpful. I'll try to consolidate this and maybe store it for future reference as a wiki page.


## Prism

Re: https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2002377176
> I've added an issue to the tracker [#34124](https://mantisbt.org/bugs/view.php?id=34124) and rebased the commits

Thanks.

One more question, since Prism offers a myriad of options and features when downloading it, what did you select to build the packaged js/css files ? I believe this is quite important to document, so we don't break any functionality when the library has to be upgraded.

I also think it would be a good idea to [allow use of a CDN](https://prismjs.com/index.html#basic-usage-cdn) when MantisBT is configured with `$g_cdn_enabled = ON;`, possibly with https://prismjs.com/plugins/autoloader, but this can be done as a separate work item.

Some ideas for future improvement (not needed as part of this PR unless you're feeling fancy and got time on your hands :wink:)
- allow the admin to pick additional languages
- enable selected plugins (via config options), e.g. [Line Numbers](https://prismjs.com/plugins/line-numbers), [Show Language](https://prismjs.com/plugins/show-language), [Copy to Clipboard Button](https://prismjs.com/plugins/copy-to-clipboard)


## process_urls

Re: https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2002377176
> ```
> URL http://example.com and EMAIL user@example.com
> ```
>
> This both are ignored but <https://example.com> not? Sounds not logically to me. Why, what's the difference? If process_urls is not wanted, then no urls/emails should be clickable at all.

Think of process_urls more as an option to *add* an additional, easier way of turning text into clickable links, without the need for any Markdown syntax. When it's OFF, you only get actual links if the appropriate syntax is used  (i.e. `<URL>` or `[link text](URL)`). 

In other words, OFF does **not** mean *I don't want any links, remove even those that were Markdown-ed as such*.

Re: https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2002411022
> You mean when it is intended to display urls that should not be clickable while in the same text there should be links which should be clickable?

Well it could be used for this purpose I guess, but that is not really what I meant. See above comment.

Re: https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2002415282
> when `process_urls = OFF`
>     * `<https://example.com> / [link](https://example.com)` become a link but `https://example.com` not?
>     * `<user@example.com>` become a mailto link but `user@example.com` not?

That's it.


## skipped testHashNumbers() failing test case

Re: https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2002377176
> the trace is this.
>
> /core/authentication_api.php:1137
> /core/access_api.php:508
> /core/string_api.php:352

OK, so the problem may be caused by the user running the unit tests not having access to the issues being linked. 

You may need to define a [bootstrap file](https://mantisbt.org/docs/master/en-US/Developers_Guide/html-desktop/#dev.contrib.test.setup) to run the tests.

It may also be needed to authenticate the user (see [MantisCoreBase::login()](https://github.com/mantisbt/mantisbt/blob/aeeb60ecc8b2c4aaf349967897d4ab2b68b29b46/tests/Mantis/MantisCoreBase.php#L67) for an example).


## Further consideration (https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-1999324037)

Re: https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2002403386

>> I'm not sure I'm following your thoughts here. Can you please clarify what feedback you expect ?
>
> The idea is moving the text processing logic from `MantisMarkdown::finalizeMarkup()` to `MantisCoreFormattingPlugin::formatted()`. The processing is then only at one place.
> 
> I am considering doing this in a further step to keep this one smaller.

OK, it makes sense. 
I'm fine with both approaches (in this PR or a separate work item), I leave it up to you.


## Improving unit tests

Re: https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2005377626

As a preliminary remark, I fully agree with you that there is no point in re-testing what is already covered by other fixtures.

The goal is just to make sure that we have the best coverage possible. 

This also means that we need to develop a way to execute Plugin test suites as part of our continuous integration process, because at the moment *MarkdownTest* suite is not executed by our TravisCI jobs.

> bugs #1, bug note ~1 and user@user are dependent on existing Database entries.

This is normally handled by having the test case generating its own test data (users, issues...) as appropriate. And deleting it at the end of processing, usually via tearDown() method.

> bugs #1 and bug notes ~1 lead to authentication_api.php::access_denied().

See comment under *skipped testHashNumbers() failing test case* section above.

> resolving of mentions is somthing what should covered by testing the core functions

Indeed we already have [MentionTest ](https://github.com/mantisbt/mantisbt/blob/master/tests/soap/MentionTest.php) that (should) cover this.

dregad:
Hello @grummbeer any feedback following my review ? I would really like to move forward with this and get these Markdown issues resolved.

grummbeer:
Thank you Damien, very helpful.

Not pushed anything yet. 

`process_urls`
---

OK. The desired behaviour has been restored.

Further consideration (https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-1999324037)
---

At the moment i am not totally sure, but I have a feeling that the format processing should be split into two parts, preparation (current parser) and finalization (applying the link processing). Maybe with a new class that brings this both parts together. All parts should be testable separately and not mixed or doubled in some way. Not sure, something this way …


Tests
---

No need for login or database. I see the parser more like the "preparer" of the given input. So what to be tested is:

- Implementation of the missing CommonMark specifications for headers.
- Tables have the wanted CSS class.
- Codeblocks are catched and replaced by their hash values.

This both are handled by the "preparer", but i think they should not.

- Attributes are added to links according to `$g_html_make_links`
- Not markdown tagged URLs and emails become a link if `process_url == ON`.


prismjs
---

Yeah, of course, I also considered it. I am willing to do it. But not in this step, this one has already grown enough.

The current collection is: 
https://prismjs.com/download.html#themes=prism&languages=markup+css+clike+javascript+javadoclike+markup-templating+php+phpdoc+php-extras

Just the default with "PHP" + "PHPDoc" + "PHP extra"


For now
---

I'm not sure what exactly the coding style is. I'm adopting what I see in the existing code, but on the other hand, this code is based on a very old PHP version and the refactored code could be modernized?

#### type hinting

What is the target for this merge request? Is it php7.4?

If yes, can i use type hinting for properties (7.4), parameters and return values? (Where possible, without breaking the compatibility with the base class) 

#### Doc block

We have a very inconsistent usage. No doc block, only description, some with visibilty `@access` or only parameters. Are the doc blocks used to build any documenation?

Thanks

dregad:
> `process_urls` The desired behaviour has been restored.

Thanks.

> ## Further consideration ([#1976 (comment)](https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-1999324037))
> 
> At the moment i am not totally sure, but I have a feeling that the format processing should be split into two parts, preparation (current parser) and finalization (applying the link processing). Maybe with a new class that brings this both parts together. All parts should be testable separately and not mixed or doubled in some way. Not sure, something this way …

I agree it would be cleaner to have separated parser and renderer classes.

It sounds like a separate work item though, and probably not a small task either. Unless this is really needed to achieve the immediate goal which is to fix the current, annoying bugs we have with Markdown rendering, I would suggest to treat this later.

> ## prismjs
> 
> Yeah, of course, I also considered it. I am willing to do it. But not in this step, this one has already grown enough.
> [...]
> Just the default with "PHP" + "PHPDoc" + "PHP extra"

Thanks for the clarification. I agree that what you did already will be good enough for now.
I was only outlining future ideas.

> ## For now
> 
> I'm not sure what exactly the coding style is. 

Please read our official [Coding Guidelines](https://mantisbt.org/wiki/doku.php/mantisbt:coding_guidelines) on the project wiki. They have not been updated for the latest PHP releases, but should still be good enough. And we are not so strict about them either, and willing to make them evolve too if needed.

> What is the target for this merge request? Is it php7.4?

Yes.

> can i use type hinting for properties (7.4), parameters and return values? (Where possible, without breaking the compatibility with the base class)

Yes please do. But as you said, make sure it does not introduce regressions, e.g. due to TypeError exceptions in PHP 8+

> #### Doc block
> 
> We have a very inconsistent usage. No doc block, only description, some with visibilty `@access` or only parameters. Are the doc blocks used to build any documenation?

Not presently. Inconsistent usage is our legacy, coming from years of code contributions written by various developers (`@access` is a typical example of this).

Just stick to the basics and apply common sense, to make code self-documenting and readable; mainly, functions should have Short summary / Description / Parameters / Return value / Thrown exceptions if applicable; variables/properties a type and description. 

Anyway if I see something missing or not right, I'll let you know as I review the code.

grummbeer:
> Coding Guidelines

Damn … how embarrassing :-) the whole test class has to be reformatted. brackets spaces, variable names …

Will do this after your other change requests.

dregad:
> No chance of getting a fixer in?

Well, that's another of those things that I started playing with years ago, and never got around to finishing... See https://github.com/dregad/mantisbt/tree/codesniffer. I have absolutely no idea in what state I left that in, give it a try if you're feeling lucky :wink:

> Please review

Don't have time now. Will try my best to do it later tonight.

> What should happen with an input of a HTML-link? The output is ugly and broken.
> 
> ```
> <a href="https://example.com">Link</a>
> ```

Opening a brand new can of worms here...

Commonmark spec says that [raw HTML should be rendered as-is without escaping](https://spec.commonmark.org/0.31.2/#raw-html) i.e. <a href="https://example.com">Link</a> but there are security implications here and we must definitely NOT blindly allow this due to the risk of XSS (`<a href="https://example.com" onclick="alert(1)">test</a>` and many other options) 

MantisBT has config settings to define acceptable HTML tags ($g_html_valid_tags, $g_html_valid_tags_single_line), which I think need to be honored (meaning, don't process the link if `a` tag is not whitelisted). 

So (open to discussion here), I would say given original text `<a href="https://example.com">Link</a>` that the link within href should become clickable (i.e. *&lt;a href="<https://example.com>">Link&lt;/a>*) if process_url = ON, and nothing should happen if OFF (ie. show the raw, not-rendered HTML). 

Of course , the proper solution would be to use HTMLPurifier to ensure we don't allow any harmful HTML to be rendered, but that's outside the scope of this PR.

> some next steps

These should be treated separately, after this PR is merged.

grummbeer:
add basic test for convert_line and convert_text

mantisbt-markdown-coverage

grummbeer:
> Don't have time now. Will try my best to do it later tonight.

Pleae take the time you need.

> Opening a brand new can of worms here...

Sorry, for that. I know that you are very busy. Thanks for your detailed answers, patience and care. 


> These should be treated separately, after this PR is merged.

YES!!

grummbeer:
Off topic, just my thoughts. No need to answer.

> Fixer 

I just run the [ecs](https://github.com/easy-coding-standard/easy-coding-standard) with a few default fixers against plugins/MantisCoreFormatting and it looks great … just a single 

```shell
ecs --fix
```

ecs supports also codeSniffer, so your own rules could adopted.

> https://github.com/dregad/mantisbt/tree/codesniffer. 

I am trying to figure out what exactly you want to solve with your own rules, but (just naive thoughts, not to be taken seriously) saying goodbye to all those prehistoric code styles (probably born while coding with vim) would not be the worst choice … saying now, but in 1h it might sound different. :-)

grummbeer:
Experimental. Added plugin tests to the testsuites.

grummbeer:
I have done something wrong. I added the ability to change the settings at runtime, but the only use case is for testing, what is not very clever. I will revert this with a new commit now.

@dregad should i open a new PR with the very minimum, which has much lesser code and is easier to review? We can then come back to this PR afterward?

grummbeer:
>  I will revert this with a new commit now.

done. ready. No more force pushes :-)

Wish you a happy Easter.

dregad:
> [grummbeer](https://github.com/grummbeer) added 3 commits [2 days ago](https://github.com/mantisbt/mantisbt/pull/1976#commits-pushed-daa04fc)

Didn't you say you were done ? 😉 
Will review later today.

grummbeer:
As longer it takes … more time to find missed things. Will not commit more until your review … really :-)

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -42,220 +42,212 @@
...
-
-		$block = parent::blockCodeComplete( $block );
-
-		if( isset( $block['element']['text']['text'] )) {
-			$this->processAmpersand( $block['element']['text']['text'] );
+	protected function blockHeader( $Line ) {
+		# - the opening # may be indented 0-3 spaces
+		# - a sequence of 1–6 '#' characters
+		# - The #'s must be followed by a space or a newline
+		if ( preg_match( '/^ {0,3}#{1,6}(?: |$)/', $Line['text'] ) ) {
```
dregad, line 194: Commonmark says *The opening sequence of # characters must be followed by spaces or tabs, or by the end of line.* So this needs a little adjustment to be compliant:

```php
# - The #'s must be followed by a space, a tab or a newline
if ( preg_match( '/^ {0,3}#{1,6}(?:[ \t]|$)/', $Line['text'] ) ) {
```

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -42,220 +42,212 @@
...
+	 *
+	 * @var array<string, string>
+	 */
+	private array $codeblocks = [];
 
-		# set the table class
-		$this->table_class = 'table table-nonfluid';
+	public function __construct( ?int $p_process_urls = OFF, ?int $p_process_buglinks = OFF ) {
+		# Plugin configuration
+		if( in_array( $p_process_urls, [ OFF, ON ], true ) ) {
```
dregad, line 75: The constructor is called from MantisCoreFormatting::formatted(), and parameters are coming from plugin_config_get() which does not check or enforce the return type.

While I'm generally in favor of strict type checking, considering that config options are basically user input, I'm concerned about the non-user-friendly behavior if we get non-integer data here (could actually be boolean, null, empty string...). This would trigger a TypeError exception, which is not user friendly.

Maybe cast plugin_config_get() return value to int when assigning $s_url / $s_buglinks.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -42,220 +42,212 @@
...
+			$this->config_process_urls = $p_process_urls;
+		}
 
+		if( in_array( $p_process_buglinks, [ OFF, ON ], true ) ) {
```
dregad, line 79: ditto

plugins/MantisCoreFormatting/tests/MarkdownTest.php
```diff
@@ -14,79 +14,395 @@
...
+	}
+
+	public function provideHeaders(): Generator {
+		# valid headers
+		yield  'valid: # foo' => [ '# foo', '<h1>foo</h1>' ];
+		yield  'valid: ## foo' => [ '## foo', '<h2>foo</h2>' ];
+		yield  'valid: ### foo' => [ '### foo', '<h3>foo</h3>' ];
+		yield  'valid: #### foo' => [ '#### foo', '<h4>foo</h4>' ];
+		yield  'valid: ##### foo' => [ '##### foo', '<h5>foo</h5>' ];
+		yield  'valid: ###### foo' => [ '###### foo', '<h6>foo</h6>' ];
```
dregad, line 294: Aren't we missing some test cases here, for header with leading whitespace, and tab instead of space between # and foo

plugins/MantisCoreFormatting/tests/MarkdownTest.php
```diff
@@ -14,79 +14,395 @@
...
+		yield 'process_urls = OFF; lorem https://exmaple.com ipsum' => [
+			'lorem https://exmaple.com ipsum',
+			OFF,
+			'<a href="https://exmaple.com"',
+			false
+		];
+	}
+
+	public function provideAttributes(): Generator {
+		yield 'LINKS_NEW_WINDOW' => [
```
dregad, line 373: Isn't that redundant, already covered by the changes I recently merged (#1983) ?

dregad (changes requested):
Here is for the formal review of code; looking pretty good, we're almost there I think 🤩 

I still need to run some actual tests, but it's getting late and need some sleep... Will try to do it tomorrow morning if I have time, otherwise in the evening.

Thanks for your work !

plugins/MantisCoreFormatting/core/MantisMarkdown.php
grummbeer: Is it user input? I think the values are already casted to an Integer at this point. The plugin handle the value and stores "1" or "0" with an type 1 into the database.

When `config_get` retrive the value it goes to this line

https://github.com/mantisbt/mantisbt/blob/aa93edbf9d45da731a24aa23875261d8cb78dfb7/core/config_api.php#L154C7-L156

```php
case CONFIG_TYPE_INT:
    return (int)$t_raw_value;
``` 

I want rely on this. If the values any other than the expected `OFF` or `ON`, something is broken and the exception is clear response to that.

Casting the values is some kind of silent failssafe that undermine the strictness. I understand your concerns, but think this situation is safe to be more strict.

grummbeer: Or do i miss something?

grummbeer:
Additionally

- Again some formatting. array style and spacing. PHPStorm "helps" with displaying this nice named arguments. But makes it hard to find the missing spaces after brackets. :-)
 
phpstorm-named-arguments

- Changed "heredoc" parts from named like `<<<Markdown` to `<<<EOD` see https://mlocati.github.io/php-cs-fixer-configurator/#version:3.52|fixer:heredoc_closing_marker

- Adding comments and new samples to emails and URLs without surrounding text. The surrounding text is for more real world usage, while the basics should not be forgotten

plugins/MantisCoreFormatting/core/MantisMarkdown.php
grummbeer: I reorganized the commits, not sure if I should open an issue for the commit "Add tests for plugin MantisCoreFormatting" to link it. I think I should, but very tired now.

No time today, so if you want the config values casted, please just do it. From my side i am fine to merge. Was a pleasure with you, thanks a lot.

atrol:
@grummbeer @dregad 

Some general thoughts about Prism usage.

MantisBT is not used just for software bug tracking, but can also be used for other purposes.
In such a scenario syntax highlighting is not needed, but introduces some overhead.
Maybe we should have it as an optional functionaity, similar to URL and MantisBT internal links processing.

I didn't try, just a guess, could especially be important on mobile devices:
Syntax highlighting might consume noteworthy additional CPU / battery when rendering issues with large code blocks.

As a side note, the current Prism 1.x version is pretty old and it looks like the roadmap for the new major version is delayed.
https://github.com/PrismJS/prism/?tab=readme-ov-file#important-notice
https://github.com/orgs/PrismJS/discussions/3531
Hope we don't introduce something new in MantisBT that is not maintained in mid-term.
In best case Prism 2 is available pretty soon and version 1 can be replaced with small effort.

grummbeer:
@atrol Good point. I'll open a follow-up PR right after this one that adds a switch to turn syntax highlighting on or off.

If you know an alternative I can replace prismjs completely.

atrol:
> If you know an alternative I can replace prismjs completely.

Searched a bit, I have no own experience.
Looks like highlightjs coult be a replacement just in case we should get problems with PrismJS
https://github.com/highlightjs/highlight.js

grummbeer:
hightlightjs would also be fine for me.

Next round … i change the commit "Add prism for markdown", but keep the reference to the issue, it must then changed accordingly.

grummbeer:
Nope, not yet. highlight.js needs an extra portion of attention that I can't provide for now.

I vote to merge this and decide afterwards. I'm willing to take care of it and also replace prismjs if necessary.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
dregad: > Is it user input?

Well technically yes. It is submitted as  an HTML form, and furthermore can be edited manually from adm_config_report.php page, which makes it fairly easy to introduce an incorrect data type e.g. string (intentionally or not).

Of course if you only edit settings using the plugin's config page, then there can be no problem, and furthermore the config_edit.php page captures the submitted form using gpc_get_int() which ensures we are indeed getting a number. "Normal" usage of adm_config_report.php page also results in typecasted to int data.

So we're mostly safe unless the admin does something silly, in which case they should expect to get an error (and hopefully know how to analyze and fix it).

> If the values any other than the expected OFF or ON, something is broken and the exception is the clear response to that.

True, but my concern was that the error message is highly generic and misleading (INTERNAL APPLICATION ERROR) providing zero indication on the actual root cause unless you have $g_show_detailed_errors = ON, which is usually not the case in production systems.

Considering the low probability of this error occurring, let's leave it like that.

dregad:
> PHPStorm "helps" with displaying this nice named arguments. But makes it hard to find the missing spaces after brackets. :-)

Yes I know... I have it configured to show whitespace, which makes it (a little) easier to spot these things thanks to the little dot.

image

dregad:
Re: prism.js vs highlight.js

I have no personal opinion on one or the other; according to https://js.libhunt.com/compare-highlight-js-vs-prism, highlight.js seems both more popular and better maintained; prism.js v2 branch is (somewhat) active but there zero visibility on when it will land.

I do agree that, regardless of which library we end up picking, we need an option in the MantisCoreFormatting plugin's config page to toggle the feature on/off. Additionally, as I [mentioned earlier](https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2007495291):

> I also think it would be a good idea to [allow use of a CDN](https://prismjs.com/index.html#basic-usage-cdn) when MantisBT is configured with `$g_cdn_enabled = ON;`, possibly with https://prismjs.com/plugins/autoloader, but this can be done as a separate work item.
> 
> Some ideas for future improvement (not needed as part of this PR unless you're feeling fancy and got time on your hands 😉)
> 
> * allow the admin to pick additional languages
> * enable selected plugins (via config options), e.g. [Line Numbers](https://prismjs.com/plugins/line-numbers), [Show Language](https://prismjs.com/plugins/show-language), [Copy to Clipboard Button](https://prismjs.com/plugins/copy-to-clipboard)

I really don't want to delay the Markdown fixes any further though. After all, syntax highlighting is a new feature, and there is no urgency to have it land, so I propose to 

1. merge this PR ASAP, without the syntax highlighting, so we can benefit from the long-awaited fixes to Markdown
2. take this discussion to https://mantisbt.org/bugs/view.php?id=34124, a separate PR can be submitted for the syntax highlighting feature

grummbeer:
Kick the commit "[Add prism for markdown](https://github.com/mantisbt/mantisbt/pull/1976/commits/591716813d2aec2f7d9b341724d2014b5289f38d)" … and finally merge! :-)

dregad:
After testing, everything looks good, except `<a href...` links.

Given raw text: `<a href="http://example.com">link linky</a>`, with process_url = OFF system behaves as expected, but when the setting is ON, we get an incorrect link:

- master <br> image
- this PR <br> image

I actually think the new behavior is correct overall, except that the link is wrong (should stop at the href's closing `"`). 

We actually discussed this before, see https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2018532773:

> > What should happen with an input of a HTML-link? The output is ugly and broken.
> > ```
> > <a href="https://example.com">Link</a>
> > ```
> 
> ...
> 
> Commonmark spec says that [raw HTML should be rendered as-is without escaping](https://spec.commonmark.org/0.31.2/#raw-html) i.e. [Link](https://example.com) but there are security implications here and we must definitely NOT blindly allow this due to the risk of XSS (`<a href="https://example.com" onclick="alert(1)">test</a>` and many other variants)
> 
> ...
> 
> So (open to discussion here), I would say given original text `<a href="https://example.com">Link</a>` that the link within href should become clickable (i.e. _<a href="https://example.com">Link</a>_) if process_url = ON, and nothing should happen if OFF (ie. show the raw, not-rendered HTML).
> 
> Of course , the proper solution would be to use HTMLPurifier to ensure we don't allow any harmful HTML to be rendered, but that's outside the scope of this PR.

I just tested Parsedown's behavior, and it actually does exactly what I expect (as stated above) natively when safeMode is ON (which is how we set it in MantisBT):

```php
include 'vendor/erusev/parsedown/Parsedown.php';
$p = new Parsedown;
$p->setsafemode(true);
echo $p->text('<a href="http://example.com" onclick="alert(1);">link</a>');
```

Output: `<p>&lt;a href=&quot;<a href="http://example.com">http://example.com</a>&quot; onclick=&quot;alert(1);&quot;&gt;link&lt;/a&gt;</p>`
image

So something we're doing is breaking this and causing incorrect parsing of the link in the href attribute. 

@grummbeer This needs to be addressed, but it's not a blocking issue so I would be OK with merging the PR in its current state and fix it later, unless you prefer to put the fix in now. Let me know.

grummbeer:
Yes, I noticed that very early and that’s why I mentioned it in the very first description of the PR (screenshot at the bottom).

The reason why the master behaves differently is that `processText` is called and the link is extracted from the `a` with `string_strip_hrefs`. `processText` is omitted now.

That’s why i asked for "Should we strip them with `string_strip_hrefs()` before any other parsing?" https://github.com/mantisbt/mantisbt/pull/1976#issuecomment-2018349482

We could restore this behavior (with `string_strip_hrefs()` very early) or try to fix this stuff now and forever with the HTMLPurifier.

grummbeer:
please merge. We'll fix this later.

I have tried this
```php
	public function convert( string $p_string, bool $p_multiline = false ): string {
		# @todo remove this is HTMLPurifier is in place
		$p_string = string_strip_hrefs( $p_string );
	       // ....
	}
```

But this strips the anchor also from code blocks. Not wanted.

dregad:
I don't think restoring the old behavior of stripping the tags is the way to go. And AFAICT HTMLPurifier is not immediately needed for this use case.

As I said, Parsedown works exactly as I expect natively, so the question is why is it not so with MantisMarkdown class ?

```php
include 'core.php';
$text = '<a href="http://example.com" onclick="alert(1);">link</a>';

$p = new Parsedown;
$p->setSafeMode(true);
echo $p->text($text);
// <p>&lt;a href=&quot;<a href="http://example.com">http://example.com</a>&quot; onclick=&quot;alert(1);&quot;&gt;link&lt;/a&gt;</p>

$p2 = new MantisMarkdown;
echo $p2->text($text);
// <p>&lt;a href=&quot;http://example.com&quot; onclick=&quot;alert(1);&quot;&gt;link&lt;/a&gt;</p>

```

dregad:
> please merge. We'll fix this later.

OK, I'll do it tomorrow.

grummbeer:
your right, good question, looking for the anwser now.

grummbeer:
MantisMarkdown set `$this->setUrlsLinked();` from first constructor parameter (process_urls) default is OFF.


```php
include 'core.php';
$text = '<a href="http://example.com" onclick="alert(1);">link</a>';
$p = new Parsedown;
$p->setSafeMode(true);

config_set('html_make_links', 0);
$p2 = new MantisMarkdown(1);

var_dump($p->text($text) === $p2->text($text)); // bool(true)
```

Both are broken the same way.

The outcome you want: 


```php
$text = '<a href="http://example.com" >http://example.com</a>';
//                                   ^ space here
$p = new Parsedown;
$p->setSafeMode(true);
config_set('html_make_links', 0);
$p2 = new MantisMarkdown(1);

echo $p->text($text).PHP_EOL;
echo $p2->text($text).PHP_EOL;
```

```
<p>&lt;a href=&quot;<a href="http://example.com">http://example.com</a>&quot; &gt;<a href="http://example.com">http://example.com</a>&lt;/a&gt;</p>
<p>&lt;a href=&quot;<a href="http://example.com">http://example.com</a>&quot; &gt;<a href="http://example.com">http://example.com</a>&lt;/a&gt;</p>
```

Try this.

```
<a href="http://example.com" onclick="alert(1);">link</a>
<a href="http://example.com" >link</a>
<a href="http://example.com" >http://example.com</a>

broken
<a href="http://example.com">link</a>
<a href="http://example.com">http://example.com</a>
```

dregad:
Thanks for investigating. So we're hitting a bug in Parsedown. Maybe be a variant of https://github.com/erusev/parsedown/issues/505, which I reported there many years ago.

grummbeer:
So the last action is

- Remove commit "Add prism"
- Useing the whole method inlineUrl with fixed regex in MantisMarkdown?

grummbeer:
I suggest deleting the commit "Add Prism" and merge.

The input of HTML anchors can be handled later with a new merge request.

- take $g_html_valid_tags, $g_html_valid_tags_single_line into account
    - parsdown's safemode must snyc the tags
    - evaluate HTMLPurifier for this, can it do the job?
- eventually adopt the full inlineUrl method and change the regex
- add tests for various input

dregad:
The main changes have been merged in 3aec6e68c537d3a72b6e92ce34d57231c54ddc07 and 8b23f27243a6458334d899eb0ffb07c6bfce4ed9.

As discussed, the following work items were not included as part of this PR, and will be handled separately

- Adding syntax highlighting to code blocks (PrismJS or similar) - tracked in [#34124](https://mantisbt.org/bugs/view.php?id=34124)
- Fixing the links in HTML anchors - tracked in [#34389](https://mantisbt.org/bugs/view.php?id=34389)

dregad:
Again, many thanks @grummbeer for the awesome work you have done with this PR (and the others), resulting in greatly improved Markdown handling in MantisBT 🚀

PR #1977, opened 2024-03-15
Enabling HTML emails based upon templates

CasN:
These changes will allow users of the software to define their own templates for mails being send by MantisBT. Simple templates have been provided, these will be maintained by the local admin in order to reflect company lay-out.

This addresses https://mantisbt.org/bugs/view.php?id=12830


dregad:
Replaced by #2075

PR #1982, opened 2024-03-27
Don't call string_display() on already-escaped data

dregad:
This causes display of `<br />` tags on Advanced Filter form when multiple values for Assigned To and Monitored by when `br` is not allowed in $g_valid_html_tags.

Fixes [#34018](https://mantisbt.org/bugs/view.php?id=34018)


atrol (approved):

dregad:
Merged in bcf62d6e1a1fd59c72d09427e366cf116eec5718

PR #1986, opened 2024-04-04
Syntax highlighting for code blocks

grummbeer:
This is the attempt to make syntax highlighing available for the code blocks created from Markdown formatted input

Fixes [#34124](https://mantisbt.org/bugs/view.php?id=34124)

This implementation is based on [Prism](https://github.com/PrismJS/prism). The state of Prism is currently unclear, which is the reason for some concerns about using an unmaintained third party tool.

The reasons why I decide for prism over some alternatives like [highlightjs](https://github.com/highlightjs/highlight.js) are that Prism has a small file size, offers a bunch of plugins and the important autoloader, whereas hljs has a larger file size and doesn’t provide own plugins such as [copy-to-clipboard](https://github.com/arronhunt/highlightjs-copy) or [line-numbers](https://github.com/wcoder/highlightjs-line-numbers.js) or an [autoloader](https://github.com/s9e/hljs-loader). These are only available as third party components. Having a single tool from one third party, even if it might become unmaintained, is better than having one modern tool but many (as many as plugins wanted) different third parties.

### Features

- Only less and third-party agnosic changes in the "MantisCoreFormattingPlugin.php"
- The implentation is all done in a single js file, which makes it easy to replacing it with an other solution.
- You can activate or deactivate syntax highlighting completely
- You can select a theme
- You can activate one or more plugins
- Languages are autoloaded. `\```dart` loads automatically the file for the Dart language
- with `$g_cdn_enabled = ON;` the resources are loaded from "https://cdnjs.cloudflare.com/ajax/libs/prism/1.29.0/prism.min.js"
- Doesn't load anything if there is no `<code>` block at all

 
<details>
<summary>screenshot of settings</summary>

mantis-sh-confg

</details>

### Themes

Available are only the themes provided by the [main repository](https://github.com/PrismJS/prism/tree/master/themes) because these are deliverd by the cdn (https://cdnjs.com/libraries/prism/1.29.0) under the main Prism URL.

The themes from the [Prism themes repository](https://github.com/PrismJS/prism-themes) are also available [via the cdn](https://cdnjs.com/libraries/prism-themes/1.29.0), but at a different URL, so I am not sure if it is worth to make them available here. Means dealing with two urls for the same type of data (a theme).

### Plugins

- copy-to-clipboard: https://prismjs.com/plugins/copy-to-clipboard/ 
- show-language: https://prismjs.com/plugins/show-language/
- show-invisibles: https://prismjs.com/plugins/show-invisibles/
- line-numbers: https://prismjs.com/plugins/line-numbers/
- normalize-whitespace (default): https://prismjs.com/plugins/normalize-whitespace/
- match-braces (with "rainbow braces"): https://prismjs.com/plugins/match-braces/
- diff-highlight: https://prismjs.com/plugins/diff-highlight/
- inline-color: https://prismjs.com/plugins/inline-color/
- previewers: https://prismjs.com/plugins/previewers/

<details>
<summary>Screehot with plugins</summary>

Source, empty lines after and before opening and closing "```"

```css



span.foo {
	background-color: navy;
	color: #BFD;
}

span.bar {
	background: rgba(105, 0, 12, .38);
	color: hsl(30, 100%, 50%);
	border-color: transparent;
}



```

mantis-sh-css

</details>

### Test content

Some content to test https://gist.github.com/grummbeer/215a8bec5cd794629ac99badb0cbc7e9

TODO
----
- [x] Translations. Add all necessary strings for English and their German translation. ~Add the strings to all language files. I noticed, that not all languages files contains all strings. I thinks is should fixed also.~
- [x] Be sure how to handle the dependencies and do it, remove the others
- [x] clean up code
 

Difficulties
---

### CSS classes are needed by some plugins

<details>

<summary>solved</summary>

Some of the plugins needs special css classes to the `<code>` elements. I don't want to add them during parsing (at `if( $Element['name'] === 'code' )`) to keep the server-side code clean from any specific third party parts as long there is no "Highlighting" abstraction layer.

To archive this, a very early `DOMContentLoaded` should fired, that add the wanted classes to the elements before the prism core script is loaded.

```php
// function resources() { 
http_csp_add('script-src', "'nonce-8IBTHwOdqNKAWeKl7plt8g=='");
// need a nonce but can't add from here. What is perfectly good :-)
echo <<<EOD
<script nonce="8IBTHwOdqNKAWeKl7plt8g==">
    const _plugins = '$t_plugins_string'.split(',')
    const plugins = {
        'line-numbers': ['line-numbers'],
        'match-braces': ['rainbow-braces', 'match-braces'],
        'diff-highlight': ['diff-highlight'],
    }
    if (_plugins.length > 0) {
        document.addEventListener('DOMContentLoaded', () => {
            const codeBlocks = document.querySelectorAll('pre > code');
            _plugins.forEach((_plugin) => {
                if (plugins[_plugin]) {
                    codeBlocks.forEach((code) => code.classList.add(...plugins[_plugin]))
                }
            });
        });
    }
</script>
EOD;
```

At the moment I've created a script file which do the job instead. But the request for the file is not that ideally.

</details>

### Loading of dependencies

<details>

<summary>solved</summary>

Some of the plugins have dependencies to languages or other plugins, but do not autoload them. If a plugin requires a dependencies that is also required by another plugin, I can't load it twice because if the first request is not loaded, the second one wins, but is the first already loaded, the second one throws an exception. It’s not a good style also. So some sort of loading handler must be implemented.

Thas a little weird.

I've tried to solve this with javascript

```
loadScript('dependency', () => { // onload
    loadScript('plugin')
})
``` 

But this fixes nothing. The `<script>` tags are created and injected to `<head>`. Not bad for pagespeed (a little lazyload) but prism don't want it. So this is in PHP with direct echo thie `<script>` tags. The issue is: Sometimes some of the dependencies are not loaded fast enough.

You can observe this with developer tools open and disabled cache. With not disabled cache the files are cached after the first page request and both approaches work well as expected.

So I am not 100% sure how to go. JavaScript or server-side. Just needs a little bit of more fiddeling.

</details>

### CSS font-family / monospace

<details>
<summary>solved</summary>

Under "Settings | User > Font Family" its possible to select a user specific font, that is printed as line style.

```css
* { font-family: [user chosen font] }
```

But this rule applies to all the `<span>` elements which wraps the language tokens, within the `<code>` element. But should not as it breaks the monospaced font-family. Currently, this is overridden by a own rule that is loaded with syntax highlighting.

```css
pre,
pre * {
    font-family: Menlo,Monaco,Consolas,"Courier New",monospace !important;
}
```
</details>


plugins/MantisCoreFormatting/files/prism/fix-code-font.css
```diff
@@ -0,0 +1,23 @@
...
+pre,
+pre * {
+    font-family: Consolas, Menlo, Monaco, "Andale Mono WT", "Andale Mono", "Lucida Console", "Lucida Sans Typewriter", "DejaVu Sans Mono", "Bitstream Vera Sans Mono", "Liberation Mono", "Nimbus Mono L", "Courier New", Courier, monospace !important;
+}
+
+/* Monospace for Textareas
+ * @todo dedicated config "Use monospac type for textarea input"
+ *                                     ^ "e" is missing
+ */
+textarea {
```
grummbeer, line 20: this is just for dev, not meant to be part of this PR.

dregad:
> Translations. Add the strings to all language files. I noticed, that not all languages files contains all strings. I thinks is should fixed also

Translations for bundled plugins are managed via [TranslateWiki](https://translatewiki.net). 

It is generally not necessary to update the language string files directly, unless something is broken. MantisBT should handle Missing strings gracefully with a fall-back to English; new translations are added regularly to master branch as they become available.

grummbeer:
> Translations

I realized that also. The todo item was then "Translations. Add all necessary strings for English and their German translation" but forgot to change it.


### Know issues

The plugin "show-invisibles" sometimes fails for some languages(?). For example for "lisp" or "dart" language. A exception is thrown, but only affects a spcific block.

grummbeer:
@dregad the plugin "show-invisibles" behave not as expected (at least in some situations), should i just remove (comment out) it? If someone misses it and is willing to fix it, we can reactivate it. If this point is clear, this PR ready for a review.

dregad:
@grummbeer it looks like a bug in the show-invisibles plugin, doesn't it ? I [checked the PrismJS issues](https://github.com/PrismJS/prism/issues?q=is%3Aissue+is%3Aopen+invisibles) but did not find anything similar so I'll file an issue there.

> should i just remove (comment out) it?

Dart and Lisp are not so commonly used, and I think it's nice to have the option to show tabs/spaces/LFs, so I would be tempted to leave the plugin enabled, knowing that these 2 languages won't be highlighted; and users have anyway the option to disable show-invisibles.

dregad (changes requested):
image

The new config values need to be initialized in MantisCoreFormattingPlugin::config()

- syntax_highlighting
- syntax_highlighting_plugins
- syntax_highlighting_theme

plugins/MantisCoreFormatting/lang/strings_english.txt
```diff
@@ -56,3 +56,16 @@ $s_plugin_format_process_markdown = 'Markdown Processing';
 $s_plugin_format_process_markdown_info = 'Styling with Markdown support, text formatting syntax designed so that it can be converted to HTML.';
 $s_plugin_format_enabled = 'On';
 $s_plugin_format_disabled = 'Off';
+$s_plugin_format_syntax_highlighting = 'Syntax highlighting for code blocks';
+$s_plugin_format_syntax_highlighting_info = 'Syntax highlighting works only if "Markdown processing" is active.';
+$s_plugin_format_syntax_highlighting_theme = 'Appearance/Theme of the code blocks';
+$s_plugin_format_syntax_highlighting_plugins = 'Extensions for the code blocks';
```
dregad, line 62: Suggest changing above 2 strings to
- `Code blocks theme`
- `Code blocks options`

grummbeer:
### show-invisibles

Thanks for filing the bug. There is another observation, that leads me to the question of removing the plugin. But i am also fine with keep it as a "nice to have".

<details>

Origin, HTML code from obove

mantis-sh-indention-origin

1 + 2: Tab/space indent is displayed correctly
3 + 4: The first occurence of a broken code block prevents the following from being hightlighted also.

Within MantisBT

mantis-sh-indention-mantis

- 1 + 2: A broken code block has no impact on others
- 3: Tab indention is alway show as spaces

</details>


### commits

Have made two commits.

- Initialize default values 
- Adjust the names of the config options

But it's so embarrassing, when I push, it will result in a force push … The reason why i did this horrible annoying force pushes was that something in my workflow must have changed. Maybe my graph is messed up or something. Every rebase on master results in a force push, but should not, it was not how it works until a few weeks ago.

- Go to my fork on github use "sync Branch"
- locally swich to master, pull (fast-forward)
- locally switch to the feature branch, rebase onto master

Normally the feature branch is now sitting on top and I can work further on the rebased base without a force push (honestly, can't remember exactly. At least when working on my gitlab with others it behaves this way). No idea what happened or what I am doing wrong.

This is just for explanation, please don't waste time trying to help, it might be useless :-) I am sure the one and only problem is me.


### Sublines/decriptions of each plugin

I've tried to decripe in own words what they do, please proofread them. But would also fine if you prefer prism’s own short description from their plugin page. e.g. https://prismjs.com/plugins/inline-color/

grummbeer:
> I failed to notice that when I tested last night. Is it the same Javascript TypeError exception in the browser console?

Only for the failing blocks/languages

> rebased / force push

Hopefully the last force push for this (and the next).

Thanks for your advice and the link. Will have to think about/test it. 

If the base is fast forwarded and the feature branch has no conflicting changes, then the feature branch has no changes in his history, right? It just sits with its own history on top of the last change to the base. That's how I've understood and experienced (I belive) it so far … will find the answer.

dregad:
> Hopefully the last force push for this (and the next).

Hopefully, indeed... FYI this is what I get when I open the PR from the notification email 😉 
image

I'll have a closer look and to some tests later (hopefully tonight).

grummbeer:
lesson learned (hopefully)

grummbeer:
@atrol can you please proofread the german translations so that they are already been reviewed when dregad has finished the code review? Tannks.


- $s_plugin_format_syntax_highlighting

   Syntaxhervorhebung für Code-Blöcke

- $s_plugin_format_syntax_highlighting_info

   Diese Einstellung wird nur berücksichtigt, wenn die "Markdown-Verarbeitung" aktiviert wurde.

- $s_plugin_format_syntax_highlighting_theme

   Darstellung der Code-Blöcke

- $s_plugin_format_syntax_highlighting_plugins

   Erweiterungen für die Code-Blöcke

- $s_plugin_format_syntax_highlighting_plugin_clipboard_desc

    Inhalt des Code-Blocks in die Zwischenablage kopieren.

- $s_plugin_format_syntax_highlighting_plugin_language_desc 
 
    Die im code block dargestellte Sprache anzeigen.

- $s_plugin_format_syntax_highlighting_plugin_numbers_desc

    Zeilennummern anzeigen

- $s_plugin_format_syntax_highlighting_plugin_invisible_desc 

    Verborgene Zeichen werden angezeigt (z.B. bei Einrückungen).

- $s_plugin_format_syntax_highlighting_plugin_trim_desc 

    Entfernt leere Zeilen vor und nach dem eigentlichen Code.

- $s_plugin_format_syntax_highlighting_plugin_braces_desc

    Zeigt bei Klammern die Paarung an, wenn man mit der Maus darüberfährt. Inklusive "rainbow braces" (Paare werden in jeweils gleicher Farbe dargestellt).

- $s_plugin_format_syntax_highlighting_plugin_diff_desc

     Hinzugefügte und entfernte Zeilen werden farblich hervorgehoben.

- $s_plugin_format_syntax_highlighting_plugin_colors_desc
    
    Ein kleines farbiges Quadrat wird vor Farbdefinitionen angezeigt (z.B. #000).

- $s_plugin_format_syntax_highlighting_plugin_preview_desc

   Beim überfahren mit der Maus wird eine Vorschau für "angle", "color", "gradient", "easing" oder "time" Definitionen angezeigt.

plugins/MantisCoreFormatting/lang/strings_german.txt
```diff
@@ -55,3 +55,16 @@ $s_plugin_format_process_markdown = 'Markdown-Verarbeitung';
 $s_plugin_format_process_markdown_info = 'Gestaltung mit Markdown-Unterstützung und einer Text-Formatierungs-Syntax, die so gestaltet wurde, dass sie in HTML konvertiert werden kann.';
 $s_plugin_format_enabled = 'An';
 $s_plugin_format_disabled = 'Aus';
+$s_plugin_format_syntax_highlighting = 'Syntaxhervorhebung für Code-Blöcke';
+$s_plugin_format_syntax_highlighting_info = 'Diese Einstellung wird nur berücksichtigt, wenn die "Markdown-Verarbeitung" aktiviert wurde.';
+$s_plugin_format_syntax_highlighting_theme = 'Darstellung der Code-Blöcke';
+$s_plugin_format_syntax_highlighting_plugins = 'Erweiterungen für die Code-Blöcke';
+$s_plugin_format_syntax_highlighting_plugin_clipboard_desc = 'Inhalt des Code-Blocks in die Zwischenablage kopieren.';
+$s_plugin_format_syntax_highlighting_plugin_language_desc = 'Die im code block dargestellte Sprache anzeigen.';
```
atrol, line 63: `'Die im Code-Block dargestellte Sprache anzeigen.'`

plugins/MantisCoreFormatting/lang/strings_german.txt
```diff
@@ -55,3 +55,16 @@ $s_plugin_format_process_markdown = 'Markdown-Verarbeitung';
...
+$s_plugin_format_syntax_highlighting_plugin_numbers_desc = 'Zeilennummern anzeigen.';
+$s_plugin_format_syntax_highlighting_plugin_invisible_desc = 'Verborgene Zeichen werden angezeigt (z.B. bei Einrückungen).';
+$s_plugin_format_syntax_highlighting_plugin_trim_desc = 'Entfernt leere Zeilen vor und nach dem eigentlichen Code.';
+$s_plugin_format_syntax_highlighting_plugin_braces_desc = 'Zeigt bei Klammern die Paarung an, wenn man mit der Maus darüberfährt. Inklusive "rainbow braces" (Paare werden in jeweils gleicher Farbe dargestellt).';
+$s_plugin_format_syntax_highlighting_plugin_diff_desc = 'Hinzugefügte und entfernte Zeilen werden farblich hervorgehoben.';
+$s_plugin_format_syntax_highlighting_plugin_colors_desc = 'Ein kleines farbiges Quadrat wird vor Farbdefinitionen angezeigt (z.B. #000).';
+$s_plugin_format_syntax_highlighting_plugin_preview_desc = 'Beim überfahren mit der Maus wird eine Vorschau für "angle", "color", "gradient", "easing" oder "time" Definitionen angezeigt.';
```
atrol, line 70: `'Beim Überfahren mit der Maus wird eine Vorschau für "angle", "color", "gradient", "easing" oder "time" Definitionen angezeigt.'`

atrol:
> @atrol ~can~ Would you please proofread the german translations so that they are already been reviewed when dregad has finished the code review? 

I reviewed the messages.

Strings like "Copy to clipboard", ... are not translated.
grafik

atrol:
> Do you want them translated?

Better translate now instead of getting users wondering about mixed language UI.

grummbeer:
> Better translate now instead of getting users wondering about mixed language UI.

done.

grummbeer:
Where to place a link to the supported languages? 

https://prismjs.com/index.html#supported-languages

If this is MR is merged a note for this goes to https://mantisbt.org/bugs/view.php?id=22189.

plugins/MantisCoreFormatting/files/syntax-highlighting/README.md
```diff
@@ -1,31 +1,41 @@
...
+- components to  ./components
+- themes to ./themes
 
 
-Files
+Plugins and Components
 ---
 
-Files taken from the repository.
+Some of the plugins has dependencies to other plugins or components which are not loaded automatically by default.
```
dregad, line 25: plugins **have** dependencies **on** other plugins

plugins/MantisCoreFormatting/files/syntax-highlighting/README.md
```diff
@@ -1,31 +1,41 @@
...
+To solve this, the plugins and their dependencies are "registered" in the [init.js](init.js) that handles the proper loading in the right order.
+
+If a plugin is added or removed, the `availableResources` array must be changed accordingly.
+
+
+CDN
+---
 
-Copy only the minimized versions.
+If the options `$g_cdn_enabled` is set to `ON`, the files are loaded from this CDN.
```
dregad, line 35: **option** (no "s")

dregad (commented):
Maybe add a note to the README, regarding the limitations / known issues with show-invisibles including link to upstream issue ?

dregad:
> Where to place a link to the supported languages?

I can't think of any appropriate place for that at the moment.

> If this is MR is merged a note for this goes to https://mantisbt.org/bugs/view.php?id=22189.

Sounds like the best option at the moment.

grummbeer:
> show-invisibles

wait ... As a repoducible observation, the second code block on the same page/bug-note (within MantisBT) is not broken. This could have something to do with the order of execution. Maybe the plugin should wait for something in the first that is present in the second.
 
However, I have added a note about this in the README.

grummbeer:
oh dear, found another one … "previewers" breaks the language specific diff highlighting "e.g. diff-javascript" in some combinations with other languages/plugins. Not sure which.


Sorry, please stop reviewing, have to investigate.

grummbeer:
I have learned that Prism is a bit fragile and some combinations of certain languages/plugins interfere with each other due to Prism's internal processes and the sharing of some globals (I assume). These are all nits and maybe also some edge cases.

Overall, it looks good, is useful and works in the most cases. I will place a note to the plugin description in the backend settings, to inform the users that in some cases it could possible to experience a strange behaviour.

I have patched the show-invisibles plugin to log to console instead of crashing, this is also to commit.

<details>
<summary>All the splendour</summary>

show-invisibles
---

### 1. Tab indent

The reason for why we don't see this

mantis-prism-tabs-parsedown

is dead simple ... at this point, they have already been replaced with whitespaces by parsdown.

https://github.com/erusev/parsedown/blob/cb17b6477dfff935958ba01325f2e8a2bfa6dab3/Parsedown.php#L160

### 2. Failing languages

Some languages such as lisp or dart fail on this line.

https://github.com/PrismJS/prism/blob/59e5a3471377057de1f401ba38337aca27b80e03/plugins/show-invisibles/prism-show-invisibles.js#L23-L24

Because `tokens[name]` is null or undefined.

**dart**: `token['string']` = undefined

https://github.com/PrismJS/prism/blob/59e5a3471377057de1f401ba38337aca27b80e03/components/prism-dart.js#L58

**lisp** `tokens['arguments']` = null, `language.arguments` not present at all

https://github.com/PrismJS/prism/blob/59e5a3471377057de1f401ba38337aca27b80e03/components/prism-lisp.js#L36

I haven't figured out why the missing tokens are magically present in the second block (globals, i guess) … and i won't do, but I have patched the plugin to just log this to the console instead of crashing in whole. At least the locally loaded version is then working as expected.

mantis-prism-show-invisibles-patch

previewers + diff-highlight
---
mantis-prism-previewers-diff-highlight

They can live together also with all plugins enabled … but … doh, the whitespace are not shown in this combination "show-invisibles+diff-highlight" :-)

The culprit was another … instead the "diff-highlight" is broken when it coexists with a `markup` block … jeez :-)

mantis-prism-markup-diff-highlight



</details>

dregad (commented):
Thanks for the thorough analysis and report @grummbeer. Let me know when I can give it another round of testing.

plugins/MantisCoreFormatting/pages/config.php
```diff
@@ -117,6 +117,91 @@
...
+<tr>
+	<th class="category width-40">
+		<?php echo lang_get( 'plugin_format_syntax_highlighting' )?>
+		<br />
+		<span class="small">
+			<?php echo lang_get( 'plugin_format_syntax_highlighting_info' )?>
+		</span>
+	</th>
+	<td class="center">
+		<label><input type="radio" class="ace" name="syntax_highlighting" value="1" <?php echo( ON == plugin_config_get( 'syntax_highlighting' ) ) ? 'checked="checked" ' : ''?>/>
```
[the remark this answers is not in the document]
grummbeer: Should I also break the long lines for process_text, process_links, etc.?

plugins/MantisCoreFormatting/pages/config.php
```diff
@@ -117,6 +117,91 @@
...
+		<?php echo lang_get( 'plugin_format_syntax_highlighting_plugins' ) ?>
+	</th>
+	<td colspan="2">
+		<?php
+			$t_current_plugins = plugin_config_get( 'syntax_highlighting_plugins' );
+			if ( !is_array( $t_current_plugins ) ) {
+				$t_current_plugins = [];
+			}
+		?>
+		<?php foreach( [
```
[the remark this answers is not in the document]
grummbeer: I moved the declaration of the vars to the top to keep the HTML (template) part more clean. I have created a variable for the list, but think that such situations do not need extra variables.

plugins/MantisCoreFormatting/pages/config.php
```diff
@@ -117,6 +117,91 @@
...
+			'show-language',
+			'line-numbers',
+			'show-invisibles',
+			'normalize-whitespace',
+			'match-braces',
+			'diff-highlight',
+			'inline-color',
+			'previewers',
+		] as $t_plugin_id ): ?>
+			<label style="cursor:pointer; margin-bottom: .8rem">
```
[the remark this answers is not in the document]
grummbeer: > CSS

To which file?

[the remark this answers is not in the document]
grummbeer: This is a plugin, so we need a plugin specific `*.css`?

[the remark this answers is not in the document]
grummbeer: the `.block` class from ace.css helps 

Only one `margin-left: 2.3rem` for the plugin description. Acceptable?

mantis-prism-plugins-settings

```php
<?php foreach( $t_available_plugins as $t_plugin_id ): ?>
	<label class="lbl block">
		<input
			name="syntax_highlighting_plugins[]"
			type="checkbox"
			class="ace input-sm"
			value="<?php echo $t_plugin_id ?>"
			<?php echo in_array( $t_plugin_id, $t_current_plugins, true )
				? ' checked'
				: ''
			 ?>
		/>
		<span class="lbl">
			<?php echo lang_get( sprintf(
				'plugin_format_syntax_highlighting_plugin_%s_name',
				str_replace('-', '_', $t_plugin_id)
			) ) ?>
			<small class="block" style="margin-left:2.3rem">
				<?php echo lang_get( sprintf(
					'plugin_format_syntax_highlighting_plugin_%s_desc',
					str_replace('-', '_', $t_plugin_id)
				) ) ?>
			</small>
		</span>
	</label>
<?php endforeach; ?>
```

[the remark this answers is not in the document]
grummbeer: > unless strictly necessary.

its is not …

plugins/MantisCoreFormatting/pages/config.php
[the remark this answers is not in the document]
dregad: > Should I also break the long lines for process_text, process_links, etc.?

If they are > 120 chars, yes please

plugins/MantisCoreFormatting/pages/config.php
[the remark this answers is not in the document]
dregad: > This is a plugin, so we need a plugin specific `*.css`?

Well it's a plugin, but I think having a CSS class to display a checkbox with label and small descriptive text  is something reusable elsewhere, so it makes sense to have it in the MantisBT core CSS files

> Only one margin-left: 2.3rem for the plugin description. Acceptable?

I was thinking of a class that would be set on the parent item (i.e. the label), made of one or more style rules applying to child elements as appropriate, something like this (not tested)

```
.checkbox-with-small-text {
    display: block;
    margin-bottom: 9px;
}
.checkbox-with-small-text > span { ... }
.checkbox-with-small-text > small {
    margin-left: 2.3rem; 
    display: block;
}
```

Does that make sense ?

> To which file? 

`ace-mantis.css` seems most appropriate.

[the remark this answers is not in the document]
grummbeer: > Does that make sense?

Yes, it does.

Sounds like a css component ("checkbox-list-item" or similar) and/or the use case for a [BEM](https://getbem.com/).

But instead introducing something "new", I’ve added another spacing class for the left margin and use a simple `<br>` for the line break. So this looks more like what already exists.

grummbeer:
Again ready for review.

There are a few points where I have doubts. Keep the old style (for consistency) or modernise it (for this plugin only)?

- Removing [closing tag](https://html.spec.whatwg.org/multipage/syntax.html#start-tags) on [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements). We’re using HTML5 doctype
- Try to remove the obsolte `width="20%"` for `td`. Three data cells with width of 40%+20%+20% … where are the remaining 20%?
- Removing value for the `checked` attribute. [Empty attribute syntax](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2) 
- Use a function for displaying the `checked` attribute. From repeating inline check 
    ```php 
    <?php echo( OFF == plugin_config_get( 'process_markdown' ) ) ? 'checked="checked" ' : '' ?>
    ```
    to function
    ```php 
    <?php echo checked( 'process_markdown', OFF ); // or similar ?>
    ```

grummbeer:
> A few points

I think, we can ignore them for now.

dregad:
Thanks for the update. I have not had time for tests or a close look at code changes since my last review yet. Will do as soon as I can. 

In reply to https://github.com/mantisbt/mantisbt/pull/1986#issuecomment-2124395875

> Keep the old style (for consistency)

Please :slightly_smiling_face:

> * Removing [closing tag](https://html.spec.whatwg.org/multipage/syntax.html#start-tags) on [void elements](https://html.spec.whatwg.org/multipage/syntax.html#void-elements). We’re using HTML5 doctype

Those are leftovers from an old attempt to use XHTML from the early 1.3 days, that was finally killed around 2013 (see [#16025](https://mantisbt.org/bugs/view.php?id=16025)). Feel free to clean them up.

> * Try to remove the obsolte `width="20%"` for `td`. Three data cells with width of 40%+20%+20% … where are the remaining 20%?

Again, that's legacy code; you can probably replace these with one of the `width-*` classes as appropriate - or even change the markup 

> * Removing value for the `checked` attribute. [Empty attribute syntax](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2)

That's a leftover from the HTML 4 / XHTML days. I'm OK to use empty attribute syntax.

> * Use a function for displaying the `checked` attribute. 

We already have this, see check_checked() function in helper_api.php.


In reply to https://github.com/mantisbt/mantisbt/pull/1986#discussion_r1609061188

> But instead introducing something "new", I’ve added another spacing class for the left margin and use a simple `<br>` for the line break. So this looks more like what already exists.

This definitely does not look as nice as it did before due to the unindented help text breaking the flow, making it more difficult to read and distinguish the options.

image

plugins/MantisCoreFormatting/files/syntax-highlighting/README.md
```diff
@@ -0,0 +1,75 @@
+Syntax highlighting
+===
```
dregad, line 2: Better use only the `#` syntax for markdown headers to avoid mixing `===`, `---` and `###`

plugins/MantisCoreFormatting/files/syntax-highlighting/monospace.css
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * The result is an inline style that applies this font to all elements.
+ * Including the <span> elements which wraps the language tokens, within
+ * the <code> element.
+ *
+ * <style>
+ *  * { font-family: "Open Sans" }
+ * </style>
+ *
+ * This rule reapplies/forces the monospace font-family to them.
```
dregad, line 13: I think it would be useful to reference the source of this reapply, i.e. where the *font-family* rule was copied from.

dregad (changes requested):
> Indentation: It should look like this. Maybe your css comes from the cache?

Indeed it was a cache problem, sorry about that. I thought I had disabled it with dev tools, but apparently I didn't. Looks OK now.

Indented help text seems slightly off compared to checkbox main label (by 1-2 pixels I'd say). I guess that's acceptable, but considering you introduced a new *margin-left-20* rule for this I was wondering why you did not go for a perfect alignment.

> I’ve added the monospace.css

I'm somewhat confused by the multiple edits and the reference to text areas in your last note. And not sure what you're asking me...

Is it worth creating an extra CSS file for just a single rule, wouldn't it be better to move it into markdown.css ?

### Localization

The * copy-to-clipboard* button is not localized (always displays `Copy` / `Copied` regardless of selected MantisBT language. Maybe it can be changed to some icon (e.g. https://fontawesome.com/v4/icon/clone, https://fontawesome.com/v4/icon/check) so we don't need to worry about managing additional language strings ?

plugins/MantisCoreFormatting/files/syntax-highlighting/monospace.css
grummbeer: The font family for `<code>` elements are originally defined in the bootstrap.css and is overwritten by the inline-style.

https://github.com/mantisbt/mantisbt/blob/a7c9660fd0b5cfaea8bdeef7ba4b2e83e28947f2/css/bootstrap-3.4.1.css#L1529-L1534

grummbeer: Sorry for the confusion. I was completely wrong, forget it. 

> Is it worth creating an extra CSS file for just a single rule, wouldn't it be better to move it into markdown.css?

It is only relevant if syntax highlighting is active. No syntax highlighting, no unused rule in markdown.css

grummbeer:
> Indented help text seems slightly off compared to checkbox main label (by 1-2 pixels I'd say). I guess that's acceptable, but considering you introduced a new margin-left-20 rule for this I was wondering why you did not go for a perfect alignment.

You are right, I also want the perfect alignment, and I am not happy with it either. It is a compromise … "margin-left-20" is in line with the exsisting "margin-left-*" classes. If you want it, I'll do it the way you suggested

### Localization

#### Text 

Translation is possible via data-* attributes 

- `data-prismjs-copy="Copier"`
- `data-prismjs-copy-error="Press Ctrl+C to copy"`
- `data-prismjs-copy-success="Copied!"`

The translations would be added to the `<script>` tag then.

#### Icon

Sounds good too, will investigate how to do.

grummbeer:
First try pure css (hacky).

https://github.com/mantisbt/mantisbt/assets/67791701/f4054735-05b3-4d38-a6ac-6559242cc120

What do you think, going ahead or text translation?

<details>
<summary>css</summary>

looks a little bit hacky. 

```css
.copy-to-clipboard-button span {
	display: inline-block;
	font: normal normal normal 14px/1 FontAwesome;
	text-rendering: auto;
	-webkit-font-smoothing: antialiased;
	-moz-osx-font-smoothing: grayscale;

	/* Hack: Hide the text */
	/* font-size: inherit; */
	font-size: 0;
}

.copy-to-clipboard-button span:before {
	font-size: 1rem;
	padding: 0.2rem;
	content: "\f24d"
}

.copy-to-clipboard-button[data-copy-state="copy-success"] span:before {
	content: "\f00c"
}
```

</details>

grummbeer:
currently

- The color of the buttons and text does not match for dark themes
- The button "copy to clipboard" is very tiny and not very good visible
- The order of "show language" and "copy to clipboard" is not good "copy to clipboard" should be the last button on the right.
- The background color and box-shadow looks outdated

It could look this

https://github.com/mantisbt/mantisbt/assets/67791701/4c92940a-2dd7-4a1c-8217-33daac3b95e1

https://github.com/mantisbt/mantisbt/assets/67791701/d824914c-1080-4ad1-9f3b-c27d787d8adc

You are right, the icon style is seen very often, much more common.

The question is: Is this wanted at all and is this amount of code acceptable to achieve it?

<details>

<summary>first attempt, will be optimized, if is at all acceptable</summary>

```javascript

 // noinspection JSUnresolvedReference
 // Prism.highlightAll();
 // addThemeClass(); // manipulate DOM
 // loadCss(`${localUrl}/monospace.css`);
 //                     ^^^^^^^^^^^^^^ will renamed to "syntax-highlighting.css"

	/*
	 * If the toolbar is used and the background color is "dark",
	 * add the CSS class "dark-theme" to it.
	 */
	const addThemeClass = () => {
		// Toolbar is in use if "ready = true"
		const toolbar = availableResources.find((p) => 'toolbar' === p.name && p.ready);
		if (toolbar) {
			const dark = isDark();
			// Icon for the "copy-to-clipboard" button.
			const icon = document.createElement('i');
			icon.classList.add('fa', 'fa-clone');
			// Find the toolbar containers
			document
				.querySelectorAll('.code-toolbar')
				.forEach((tb) => {
					/*
					 * Add the CSS class "dark-theme"
					 * when the background-color is dark
					 */
					if (dark) {
						tb.classList.add('dark-theme');
					}

					/*
					 * If the "copy-to-clipboard" exists, replace the text
					 * with the icon.
					 */
					const ctcButton = tb.querySelector('.copy-to-clipboard-button');
	 				if (ctcButton) {
						ctcButton.firstChild.remove();
						ctcButton.appendChild(icon.cloneNode()); // todo: test for cloning the icon is cheaper than creating a new one?
					}
			});
		}
	}

	/**
	 * Determines if the color scheme is dark
	 *
	 * @return boolean
	 */
	const isDark = () => {
		// getComputedStyle returns always the RGB value of a color.
		const rgb = window.getComputedStyle(codeBlocks[0].parentElement).backgroundColor;
		const color = rgb.match(/^rgba?\((\d+),\s*(\d+),\s*(\d+)(?:,\s*(\d+(?:\.\d+)?))?\)$/);
		const r = color[1];
		const g = color[2];
		const b = color[3];
		// HSP http://alienryderflex.com/hsp.html
		const hsp = Math.sqrt(0.299 * (r * r) + 0.587 * (g * g) + 0.114 * (b * b));
		return hsp < 127.5;
	}
```

```css

/*
 * Adjust position and buttons of the toolbar.
 *
 * - More space around
 * - display flex for the alignment of the different sizes for
 *   the "show languages" text and the "copy to clipboard" icon.
 *
 */
div.code-toolbar > .toolbar {
	display:flex;
        align-items: center; /* not working as expected? */
	top: 1.5rem;
	right: 1.5rem;
}

/*
 * Nicer look for the toolbar items.
 *
 * - Remove background colors and shadows
 * - Default color for light theme
 */
div.code-toolbar > .toolbar > .toolbar-item > a,
div.code-toolbar > .toolbar > .toolbar-item > span,
div.code-toolbar > .toolbar > .toolbar-item > button {
	color: #333;
 	background: none;
	box-shadow: none;
}

/*
 * Color for dark theme.
 */
div.code-toolbar.dark-theme > .toolbar > .toolbar-item > a,
div.code-toolbar.dark-theme > .toolbar > .toolbar-item > span,
div.code-toolbar.dark-theme > .toolbar > .toolbar-item > button {
	color: #ddd;
}

/*
 * Copy to clipboard. Text has been removed with an icon. Resize it to be more visible.
 */
.copy-to-clipboard-button i:before {
	font-size: 2rem; /* make it a little bit smaller */
}

/*
 * Copy to clipboard. Change this icon for the state "copy-success"
 */
.copy-to-clipboard-button[data-copy-state="copy-success"] i:before {
	content: "\f00c";
	color: green;
}

```


</details>

dregad:
Hello @grummbeer,

I'm sorry for the delayed response, I have been quite busy in the past few weeks... I finally found some time to look at this again today.

> > Indented help text seems slightly off compared to checkbox main label (by 1-2 pixels I'd say). I guess that's acceptable, but considering you introduced a new margin-left-20 rule for this I was wondering why you did not go for a perfect alignment.
> 
> You are right, I also want the perfect alignment, and I am not happy with it either. It is a compromise … "margin-left-20" is in line with the exsisting "margin-left-*" classes. If you want it, I'll do it the way you suggested

Well it is acceptable as it is (meaning I'm OK to merge even if it's not perfect).

Of course if you are willing to make it perfect then of course it's better, but I don't want to ask too much from you, for such a minor cosmetic issue. So I let you decide if it's worth your time and effort to improve the alignment.

> Translation is possible via data-* attributes

In that case, from my perspective it's perfectly fine for me. 

The icon version you propose looks nice, but indeed it seems like a lot of javascript and CSS for little benefit. 

So I propose to go with the simple, localized text approach. Maybe in a future version Prism.js Copy to Clipboard Button extension will have an option to use an icon, and we can revisit this.

>> Is it worth creating an extra CSS file for just a single rule, wouldn't it be better to move it into markdown.css?
>
> It is only relevant if syntax highlighting is active. No syntax highlighting, no unused rule in markdown.css

OK I understand your point, but I think there is more overhead making an additional connection/request to the web server, vs  having an extra, unused rule in the existing CSS file.

grummbeer:
Hi Damien, no problem, same situation here.

- monospace.css removed, rule moved to markdown.css
- added translations
- "copy" button is the last one
 
> CSS

I don't **feel** comfortable with it. I fully agree with having a component like "checkbox-with-decription", but in the current state nobody knows what it does, where it is used (example of what it looks like), it looks kind of lost in the middle of ace-mantis.css, no context, no structure, no organisation, no other components (besides ace owns `.tooltip`, etc), deal with the ace formats (`.ace`, `.lbl`, etc.). The current state of the CSS scares me. :-)

So another simple `margin-left-*` class, where everyone knows what it does or how to use it, is the compromise I mentioned earlier. Thats all, just feelings.

dregad:
> The current state of the CSS scares me. :-)

Yeah. The Ace template legacy is horrible. And preventing us from upgrading to more recent version of Bootstrap, too... But that's a different story 😟 

> So another simple `margin-left-*` class, where everyone knows what it does or how to use it, is the compromise I mentioned earlier.

OK, let's leave it at that for now.

I'll review the last commits later, and then this can (finally) be merged 🚀 🥳

grummbeer:
Sounds great! I'd squashing the commits to a singel one "Add syntax highlighting for code blocks".

dregad:
@grummbeer finally got around to finalising this.

I would like to sincerely thank you for your efforts in making this excellent contribution happen ! It was great to collaborate with you on this PR.

grummbeer:
Thank you!

grummbeer:
Let me know if you have any issues, I will take care for this component.

PR #1987, opened 2024-04-04
Equalize the appearance of inline code and code blocks

grummbeer:
Equalize the appearance of inline code and code blocks in color and adjust the distance from a code block to its previous line in list items.

Fixes

- [#22181](https://mantisbt.org/bugs/view.php?id=22181) Markdown different rendering between inline code (single backtick) and ``` blocks
- [#22485](https://mantisbt.org/bugs/view.php?id=22485)  Increase spacing before ``` blocks

|before|after|
|-- |-- |
|mantis-code-before|mantis-code-after|

### DIFF
--- a/plugins/MantisCoreFormatting/files/markdown.css
+++ b/plugins/MantisCoreFormatting/files/markdown.css
@@ -1,3 +1,17 @@
+/* Adjust color of inline code */
+code {
+    color: #333;
+    background-color: #f5f5f5;
+}
+
+/*
+ * Increases the distance between a code block
+ * and its previous line within a list.
+ */
+li > pre {
+    margin-top: .4rem;
+}
+
 /* quotes */
 blockquote {
     padding: 0.13em 1em;


plugins/MantisCoreFormatting/files/markdown.css
```diff
@@ -1,3 +1,24 @@
+/*
+ * Rules taken from bootstrap.
+ *
+ * The right way is creating boostrap css from scss sources instead
+ * of override the rules here.
+ *
+ * @todo fix this directly in bootstrap and delete this rules.
```
dregad, line 7: I'm not sure what you mean by fixing directly in Bootstrap... Opening an issue to have it fix upstream ? Considering that we are using a fairly old version of that library, I don' think it's worth bothering as it would certainly not be be accepted.

I'd rather keep the bundled bootstrap files untouched, an override of specific CSS rules like here is acceptable to me.

Grammar fixes: 
- instead of **overriding**
- delete **these** rules

plugins/MantisCoreFormatting/files/markdown.css
```diff
@@ -1,3 +1,24 @@
...
+ */
+code {
+    color: #333;
+    white-space: pre;
+    padding: 0.2rem .4rem 0.1rem .4rem;
+    margin: 0;
+    font-size: 90%;
+    background-color: #f5f5f5;
+    border-radius: 4px;
+}
```
dregad, line 17: Is it necessary to redefine the whole rule ? Wouldn't it be enough to only override the necessary properties (I would guess, *color* and *background-color*.

plugins/MantisCoreFormatting/files/markdown.css
```diff
@@ -1,3 +1,24 @@
...
+ * Rules taken from bootstrap.
+ *
+ * The right way is creating boostrap css from scss sources instead
+ * of override the rules here.
+ *
+ * @todo fix this directly in bootstrap and delete this rules.
+ */
+code {
+    color: #333;
+    white-space: pre;
```
dregad, line 11: I don't think that's right; why are you adding that ?

Inline quotes are expected to wrap with text and with `pre` there would no line breaks introduced, so the \<code> block would extend the parent \<td> horizontally which is not wanted.

For me it's fine if whitespace is collapsed in this context

dregad: Github is using `white-space: break-spaces;` in this context

plugins/MantisCoreFormatting/files/markdown.css
grummbeer: > I'd rather keep the bundled bootstrap files untouched,

Of course.

That was misleading, I meant that it is the normal way to customize the bootstrap css via the variables or to remove parts from the scss sources that you don't need. 
The idea was to do it this way, i.e. set the color via the variables and rebuild the css with it. But now I realized that bootstrap is just way too old for that. 

The comment is removed.

plugins/MantisCoreFormatting/files/markdown.css
grummbeer: > I don't think that's right; why are you adding that ?

I can't give you a proper anwser. I can't remember what I was thinking … probably nothing. Maybe it was all too fast, driven by the desire to get the Markdown theme done quickly.

Thanks for the correction.

> For the record

:-) they do it better.

plugins/MantisCoreFormatting/files/markdown.css
grummbeer: Not very precise :-( I can do better.

dregad (approved):
Looks good. Thanks !

PR #1989, opened 2024-04-07
Fix internal error on view_user_page

atrol:
Fixes [#34399](https://mantisbt.org/bugs/view.php?id=34399)

PR #1991, opened 2024-04-08
WIP: Introduce CS Fixer

grummbeer:
This PR is a continuation of the work started by @dregad to [add a CS fixer](https://github.com/mantisbt/mantisbt/compare/master...dregad:mantisbt:codesniffer). This is the first step with adding the "[easy coding standard](https://github.com/easy-coding-standard/easy-coding-standard)" to the CI. No fixers are added yet, these will be added in small steps to keep the changes small an reviewable.

Issue [#34400](https://mantisbt.org/bugs/view.php?id=34400)

First attempt: https://github.com/mantisbt/mantisbt/compare/master...dregad:mantisbt:codesniffer

Some sniffs are outdated, e.g. "Generic.Arrays.DisallowShortArraySyntax", others may no longer be relevant in other ways.

But there is a minimum of basic fixer that can help a lot. Better having just a few basic fixers that work than nothing at all.

For example [spaces_inside_parentheses](https://cs.symfony.com/doc/rules/whitespace/spaces_inside_parentheses.html) > `func( $arg, $arg2 )` to reduce situations [like this ](https://github.com/mantisbt/mantisbt/pull/1989#pullrequestreview-1985115982).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,16 @@ env:
 
 jobs:
   include:
+    - name: Coding Standards
+      # stage: cs
+      php:
+        - 8.3
+      env: []
+      services: []
+      before_script: []
+      script:
+        - echo "Executing Mantis coding style checks"
+        - vendor/bin/ecs check
     # Add a specific build for Documentation
     - name: Documentation
       env: DOCBOOK=1
--- a/build/buildrelease.py
+++ b/build/buildrelease.py
@@ -38,6 +38,7 @@ exclude_list = (
     # Directories
     "docbook/",
     "tests/"
+    "ecs.php"
     )
 
 # Checksum types to includ in digest filese
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,8 @@
     },
     "require-dev": {
         "ext-zip": "*",
-        "phpunit/phpunit": "^9.6"
+        "phpunit/phpunit": "^9.6",
+        "symplify/easy-coding-standard": "^12.1"
     },
     "repositories": [
         {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "fc8c59d7789077470a010a5b6d0a092b",
+    "content-hash": "def99b400e78ac217d6849b379e13d07",
     "packages": [
         {
             "name": "adodb/adodb-php",
@@ -2883,6 +2883,64 @@
             "time": "2020-09-28T06:39:44+00:00"
         },
         {
+            "name": "symplify/easy-coding-standard",
+            "version": "12.1.14",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/easy-coding-standard/easy-coding-standard.git",
+                "reference": "e3c4a241ee36704f7cf920d5931f39693e64afd5"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/easy-coding-standard/easy-coding-standard/zipball/e3c4a241ee36704f7cf920d5931f39693e64afd5",
+                "reference": "e3c4a241ee36704f7cf920d5931f39693e64afd5",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">=7.2"
+            },
+            "conflict": {
+                "friendsofphp/php-cs-fixer": "<3.46",
+                "phpcsstandards/php_codesniffer": "<3.8",
+                "symplify/coding-standard": "<12.1"
+            },
+            "bin": [
+                "bin/ecs"
+            ],
+            "type": "library",
+            "autoload": {
+                "files": [
+                    "bootstrap.php"
+                ]
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "MIT"
+            ],
+            "description": "Use Coding Standard with 0-knowledge of PHP-CS-Fixer and PHP_CodeSniffer",
+            "keywords": [
+                "Code style",
+                "automation",
+                "fixer",
+                "static analysis"
+            ],
+            "support": {
+                "issues": "https://github.com/easy-coding-standard/easy-coding-standard/issues",
+                "source": "https://github.com/easy-coding-standard/easy-coding-standard/tree/12.1.14"
+            },
+            "funding": [
+                {
+                    "url": "https://www.paypal.me/rectorphp",
+                    "type": "custom"
+                },
+                {
+                    "url": "https://github.com/tomasvotruba",
+                    "type": "github"
+                }
+            ],
+            "time": "2024-02-23T13:10:40+00:00"
+        },
+        {
             "name": "theseer/tokenizer",
             "version": "1.2.3",
             "source": {
@@ -2951,5 +3009,5 @@
     "platform-overrides": {
         "php": "7.4.33"
     },
-    "plugin-api-version": "2.6.0"
+    "plugin-api-version": "2.3.0"
 }
--- a/ecs.php
+++ b/ecs.php
@@ -1,0 +1,31 @@
+<?php
+# MantisBT - A PHP based bugtracking system
+#
+# MantisBT is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# MantisBT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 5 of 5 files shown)


.travis.yml
```diff
@@ -51,6 +51,16 @@ env:
 
 jobs:
   include:
+    - name: Coding Standards
+      # stage: cs
+      php:
+        - 8.3
```
vboctor, line 57: What does this php version represent?

grummbeer: The fixers run with this version. They run in a separate job without affecting the other jobs, e.g. Unittest.

grummbeer: I think the CI script should splitted into stages.

- cs
- test
- docs
- build

In this way, the "docs" have their own setup and do not have to read an env value. Each stage can be handled separately ... and can fail separately

grummbeer:
Added some basics to see the results of `ecs check`

Last run: https://app.travis-ci.com/github/mantisbt/mantisbt/jobs/620277330

- OK EncodingFixer
- OK LineEndingFixer
- NoTrailingWhitespaceFixer
- SingleBlankLineAtEofFixer
- NoWhitespaceInBlankLineFixer
- SingleQuoteFixer
- OK FullOpeningTagFixer
- OK EchoTagSyntaxFixer
- OK NoClosingTagFixer
- 1x ConstantCaseFixer
- 2x LowercaseKeywordsFixer
- SpacesInsideParenthesesFixer::class, ('space' => 'single')


SpacesInsideParenthesesFixer looks not good for multiline 
```diff
@@ -29,7 +29,7 @@
 /**
  * @var \Slim\App $g_app
  */
-$g_app->group('/issues', function() use ( $g_app ) {
+$g_app->group( '/issues', function() use ( $g_app ) {
 	$g_app->get( '', 'rest_issue_get' );
 	$g_app->get( '/', 'rest_issue_get' );
 	$g_app->get( '/{id}', 'rest_issue_get' );
@@ -74,7 +74,7 @@
 	$g_app->get( '/{id}/files', 'rest_issue_files_get' );
 	$g_app->get( '/{id}/files/{file_id}/', 'rest_issue_files_get' );
 	$g_app->get( '/{id}/files/{file_id}', 'rest_issue_files_get' );
-});
+} );
```
^ `})` looks better … but I think it's more important that they all look the same, rather than *how exactly* they look.


- Add some basic fixer (e.g. those that don't complain) with this PR?
- Add one rule at a time, check results, then do `ecs --fix` for a couple of files, repeat this,  until the added fixer pass?
- Add more fixers and select the ones with the fewest hits and fix them first?

grummbeer:
Dear maintainers, please leave some hints, suggestions or advice.

Having limits on time for reviews in mind, what do you think is a viable approach?

dregad:
Thanks a lot @grummbeer for picking up and improving on this half-baked attempt I started with PHP_CodeSniffer so many years ago.

I started reading [#34400](https://mantisbt.org/bugs/view.php?id=34400) but it's a long post and I don't have enough time at the moment to go in details. Will give feedback later.

> ```diff
> -});
> +} );
> ```
> ^ }) looks better … but I think it's more important that they all look the same, rather than how exactly they look.

`} );` is actually correct according to the guidelines, IMO. And indeed it's better to be consistent.

For as long as you're developing and testing this, please **disable all unnecessary build jobs** by commenting them out in .travis.yml, so it does not unnecessarily consume build credits - as a nice side effect, it will make the CI execution faster for you, too. Thanks for your understanding.

As for moving forward, this kind of changes will likely introduce a lot of churn, that will cause merge conflicts with all other unmerged PRs and local development. Of course those should be easily fixable by running the fixer in the feature branch, but to reduce the impact I would rather have an incremental approach within a single PR that can be merged in one go, rather than a series of individual PRs.

What I mean by this:
- iterative approach (i.e add one or more rules, check, adjust rules if needed, apply fix and commit updated source code)
- number of rules to add per iteration does not matter, but I would suggest keeping change sets small to facilitate checks and review
- order in which rules are added I also leave up to you, whatever makes the job easier
- **No force-rebase / squash commits !** like you've done in your recent contributions. I don't like it so much in general because it rewrites history and causes loss of traceability when multiple reviews are needed; it is much better to add incremental fix commits, and rebase/force-push just before merge.
- Here I would expect a one or more initial commits with the necessary setup (composer, ecs config), followed by N commits with code style fixes, i.e. one per (batch of) added rule(s).
- When ready to merge, the feature branch can be rearranged as and if needed

dregad:
Just an idea for optimization, not sure if it's worth the effort or even feasible, but maybe the CI job could run only against the files modified in the submitted PR, instead of the whole code base ?

.travis.yml
dregad: Config file triggers a warning on Travis https://app.travis-ci.com/github/mantisbt/mantisbt/builds/269882010/config

```
[warn] on jobs.include.php: unexpected sequence, using the first value (8.3)
```

I think it should be just
```
      php: 8.3
```
But I'm wondering if it is even needed to specify the PHP version... Can't we simply rely on the worker's system default (I think it's 7.4 on focal, which we currently use)

grummbeer: This is my first time with travis, but the config and the keys looks very similar to the gitlab ci config. So what I have done here is based on my experience with Gitlab.

> [warn] on jobs.include.php: unexpected sequence, using the first value (8.3)

Can not find this message on the linked page. I have tested the config with the travis config tool, no warings were shown.

Anyway `php: 8.3` looks good also … like the gitlab notation.


> But I'm wondering if it is even needed to specify the PHP version... Can't we simply rely on the worker's system default (I think it's 7.4 on focal, which we currently use) ? This way we don't have to worry about changing that in the future if and when PHP 8.3 becomes obsolete.

Not sure, will test it with the config tool. We have defined some PHP versions as global or shared versions for all jobs. Adding a key "php" to the job overrides the global/shared "php". Similar to neutralize the services `services: []` oder the `before_script: []` for this specific job. 

Simpler: Will the CS job run with all global defined PHP-Versions or not?
 
Please do not waste time on this, i will find the best working solution (without invisible :-) warnings)

grummbeer:
> To get it right. You mean this PR?

Sorry … reading helps, will do it this way.

.travis.yml
dregad: > > [warn] on jobs.include.php: unexpected sequence, using the first value (8.3)
> 
> Can not find this message on the linked page. I have tested the config with the travis config tool, no warings were shown.

Weird. Maybe an access rights issue caused by you not being member of mantisbt org ? I just checked and I it's still there
image

dregad: Anyway it's gone from more recent builds, so the point is moot.
https://app.travis-ci.com/github/mantisbt/mantisbt/jobs/620323284/config

grummbeer:
@dregad 

> Just an idea for optimization, not sure if it's worth the effort or even feasible, but maybe the CI job could run only against the files modified in the submitted PR, instead of the whole code base?

Sounds good, I am not a git specialist, but the files of the last commit can be taken from this line

```shell
git diff --name-only HEAD HEAD~1
```

and then added to the fixer as arguments. I will try how to script it.

```shell
composer check-cs -- [files]
```

You are asking because you want saving credits on Travis, right? You stick on Travis for certain reasons and github actions are not an alternative?

grummbeer:
https://github.com/mantisbt/mantisbt/blob/f34773b8fa7c9889ec2463ed4ae046352a13d7d3/tests/rest/RestBase.php#L35-L38

require and require_once with or without parentheses?

dregad:
>> Just an idea for optimization, not sure if it's worth the effort or even feasible, but maybe the CI job could run only against the files modified in the submitted PR, instead of the whole code base?
> 
> Sounds good, I am not a git specialist, but the files of the last commit can be taken from this line
> 
> ```shell
> git diff --name-only HEAD HEAD~1
> ```

The command is right, but I would recommend to switch the commit refs around, i.e. <OLD_REF> <NEW_REF>. 

Also note that `HEAD~1` is no right, as the PR could contain more than one commit. It needs to include all commits / files modified in the build.

I think [TRAVIS_COMMIT_RANGE](https://docs.travis-ci.com/user/environment-variables/) environment variable should be used for that.

> You are asking because you want saving credits on Travis, right? 

Correct.

> You stick on Travis for certain reasons and github actions are not an alternative?

The only reason for sticking with Travis, is lack of time to migrate to Github Actions, which is another one of those things I'd like to do at but have not had time to look into.

> require and require_once with or without parentheses?

That's another legacy thing... Historically even though it's technically not necessary, we've always added the parentheses, like `require_once( xxxx );`.. This is less of an issue today (since version 1.3 actually), as we've replaced most of the native require* / include* calls by custom require_api()/require_lib() functions.

I'm fine with standardizing on removal of the parentheses.

grummbeer:
first try was 7f1a1a3686b0d2a29bb02dcf9924ea8024280205

> I think [TRAVIS_COMMIT_RANGE](https://docs.travis-ci.com/user/environment-variables/) environment variable should be used for that.

Thanks. works. d66226c3f996d16d3dbbb8d5a7c833c6bee6e65b.

"echo" … one more language construct which to fix (no parentheses) is not that simple. 

`echo ($foo === $bar) ? 'this' : 'that'` is now `echo($foo === $bar) ? 'this' : 'that'`. No space after "echo" … there are discussions about that, will see how to do this right. But this is just one of many things i have stumbled accross so far.

> github actions

Dont wondering, i will remove travis.yaml temporarily and add a github workflow for testing.

PR #1992, opened 2024-04-15
Additional buttons proposal

paspiz85:
I want add extra buttons in header. Those buttons are added via global configuration and can require an access level.
Default is a button to display link to Mantis GitHub project for developer access level but I can change it to an empty array (no buttons at all).
Please tell me if it is usefull or can be do it better.
Thank you

PR #1993, opened 2024-04-18
Allow changing navbar buttons

paspiz85:
I want to customize navbar buttons by adding extra buttons using a plugin.

I need to add an event to manage these extra buttons. It could also be useful to modify existing buttons.

Fixes [#34406](https://mantisbt.org/bugs/view.php?id=34406)


dregad:
This looks much better than the code you initially proposed in https://github.com/mantisbt/mantisbt/pull/1992#issuecomment-2062124253, where I didn't like the approach of BEGIN/END events to add HTML.

At first glance, this seems much more in line with what I expected. Need to test, will do as time allows and get back to you.

PR #1994, opened 2024-04-18
Attachment to new tab

lucmuller:
This pull request implements the code review suggested changes in #1767
I would be glad to see them merged as having the attachement opened in a new tab would be helpful !

Fixes [#27551](https://mantisbt.org/bugs/view.php?id=27551)

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2388,13 +2388,6 @@ $g_max_dropdown_length = 40;
  */
 $g_wrap_in_preformatted_text = ON;
 
-/**
- * Controls whether attachment links redirect users t a new tab or not
- * If turned off, a click on a link will not redirect to a new tab
- * @global integer $g_attachment_to_new_tab
- */
-$g_attachment_to_new_tab = OFF;
-
 #############################################
 # MantisBT Authentication and LDAP Settings #
 #############################################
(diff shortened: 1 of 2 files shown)


dregad:
Thanks for your contribution. I'll review when I have some time, and get back to you.

vboctor:
Thanks, @lucmuller, for your PR.

This functionality comes up every few years with different implementations. At one point, we had links open in the current tab with an icon next to them to have the user choose to open in a new tab. We then removed it since it was considered redundant to the browser's native support for opening in the current window vs. the new tab vs. the new window. I'm not sure why we would add this again.

dregad:
@vboctor just making sure you're aware, that this is not about URL links (what you seem to be referring to), but links to attachments.

For context, please read https://mantisbt.org/forums/viewtopic.php?f=2&t=25561

lucmuller:
The point of my PR is only to respond to code reviews in #1767.

About links, there's already a configuration to handle them :
$g_html_make_links =  LINKS_NEW_WINDOW | LINKS_NOOPENER;

This PR would allow a new configuration to handle filedownload in the same way :
$g_attachment_to_new_tab = TRUE;

dregad:
Merged in d1cc59114e6ff3498d1f874557b8a667e19b86c7

Thanks @lucmuller @MaloMn for your contribution !

PR #1996, opened 2024-04-19
Admin Approval for new users

CasN:
Addresses issue 34408.
When a user signs up, the account remains disabled till an administrator enables the account in user management. These users can be found by searching for a definable string(currently set to "=NEW ACCOUNT=")


dregad:
See my note in https://mantisbt.org/bugs/view.php?id=34408#c68845

PR #1998, opened 2024-04-22
Configuration to require fixed_in_version

paspiz85:
I would need a mechanism to ensure that 'fixed_in_version' are always filled in before the bug reaches the Resolved state. The reason is to populate the changelog for a specific project.

Resolving issues having a target version should set the fixed in version by default.

Fixes [#21734](https://mantisbt.org/bugs/view.php?id=21734)

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2300,7 +2300,13 @@ $g_bug_submit_status = NEW_;
 $g_bug_assigned_status = ASSIGNED;
 
 /**
- * Required fixed_in_version in RESOLVED state.
+ * In RESOLVED state fixed_in_version default from target_version.
+ * @global integer $g_bug_fixed_in_version_from_target_version
+ */
+$g_bug_fixed_in_version_from_target_version = OFF;
+
+/**
+ * In RESOLVED state fixed_in_version is required.
  * @global integer $g_bug_fixed_in_version_required
  */
 $g_bug_fixed_in_version_required = OFF;
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -2300,6 +2300,12 @@ $g_bug_submit_status = NEW_;
 $g_bug_assigned_status = ASSIGNED;
 
 /**
+ * Required fixed_in_version in RESOLVED state.
+ * @global integer $g_bug_fixed_in_version_required
+ */
+$g_bug_fixed_in_version_required = OFF;
+
+/**
  * Status to assign to the bug when reopened.
  * @global integer $g_bug_reopen_status
  */
--- a/core/bug_api.php
+++ b/core/bug_api.php
@@ -508,6 +508,16 @@ class BugData {
 			category_ensure_exists( $this->category_id );
 		}
 
+		if ( $this->status == RESOLVED ) {
+			if ( is_blank( $this->fixed_in_version ) && !is_blank( $this->target_version ) && config_get( 'bug_fixed_in_version_from_target_version', null, null, $this->project_id ) != OFF ) {
+				$this->fixed_in_version = $this->target_version;
+			}
+			if ( is_blank( $this->fixed_in_version ) && config_get( 'bug_fixed_in_version_required', null, null, $this->project_id ) != OFF ) {
+				error_parameters( lang_get( 'fixed_in_version' ) );
+				trigger_error( ERROR_EMPTY_FIELD, ERROR );
+			}
+		}
+
 		if( !is_blank( $this->duplicate_id ) && ( $this->duplicate_id != 0 ) && ( $this->id == $this->duplicate_id ) ) {
 			trigger_error( ERROR_BUG_DUPLICATE_SELF, ERROR );
 			# never returns
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -324,15 +324,27 @@ layout_page_begin();
 			&& !bug_is_readonly( $f_bug_id )
 			&& access_has_bug_level( config_get( 'update_bug_threshold' ), $f_bug_id )
 		) {
+			$t_fixed_in_version_required = false;
+			$t_fixed_in_version_value = $t_bug->fixed_in_version;
+			if ( $f_new_status == RESOLVED ) {
+				$t_fixed_in_version_required = config_get( 'bug_fixed_in_version_required', null, null, $t_bug->project_id ) != OFF;
+				if ( is_blank( $t_fixed_in_version_value ) && !is_blank( $t_bug->target_version ) && config_get( 'bug_fixed_in_version_required', null, null, $t_bug->project_id ) != OFF ) {
+					$t_fixed_in_version_value = $t_bug->target_version;
+				}
+			}
 ?>
 			<!-- Fixed in Version -->
 			<tr>
 				<th class="category">
+					<?php echo $t_fixed_in_version_required ? '<span class="required">*</span> ' : ''; ?>
 					<?php echo lang_get( 'fixed_in_version' ) ?>
 				</th>
 				<td>
-					<select name="fixed_in_version" class="input-sm">
-						<?php print_version_option_list( $t_bug->fixed_in_version, $t_bug->project_id, VERSION_ALL ) ?>
+					<select name="fixed_in_version" class="input-sm" <?php
+						echo helper_get_tab_index();
+						echo $t_fixed_in_version_required ? ' required' : '';
+					?>>
+						<?php print_version_option_list( $t_fixed_in_version_value, $t_bug->project_id, VERSION_ALL ) ?>
 					</select>
 				</td>
 			</tr>


core/bug_api.php
```diff
@@ -508,6 +508,16 @@ public function validate( $p_update_extended = true ) {
...
 
+		if ( $this->status == RESOLVED ) {
+			if ( is_blank( $this->fixed_in_version ) && !is_blank( $this->target_version ) && config_get( 'bug_fixed_in_version_from_target_version', null, null, $this->project_id ) != OFF ) {
+				$this->fixed_in_version = $this->target_version;
+			}
+			if ( is_blank( $this->fixed_in_version ) && config_get( 'bug_fixed_in_version_required', null, null, $this->project_id ) != OFF ) {
+				error_parameters( lang_get( 'fixed_in_version' ) );
+				trigger_error( ERROR_EMPTY_FIELD, ERROR );
+			}
+		}
```
dregad, line 519: There is some duplicated logic here vs changes in bug_change_status_page.php, see if it would be possible avoid that by moving that to an API function somehow.

config_defaults_inc.php
```diff
@@ -2299,6 +2299,18 @@
  */
 $g_bug_assigned_status = ASSIGNED;
 
+/**
+ * In RESOLVED state fixed_in_version default from target_version.
+ * @global integer $g_bug_fixed_in_version_from_target_version
```
dregad, line 2304: Please use `int` instead of `integer`

dregad (commented):
What happens if the selected target version is not (no longer) valid for the given project at the time the Issue is resolved ? For example, target version was set to "v1" 2 years ago, "v1" was released without fixing the Issue, and it is finally resolved while "v2" is the current "next" version ?
- set fixed_in_version to v1
- don't set fixed_in_version 
- throw error
- ...

Consider usage via REST or SOAP API, as well as from group action - we need to ensure consistent behavior across the board.

Please break long lines between 80 and max 120 chars.

paspiz85:
@dregad I fixed logic to require fixed_in_version, but haven't a good solution to initialize with target_version then I discarded that logic (it wasn't a required functionality)

dregad (changes requested):
With this reduced scope, this PR no longer qualifies as a fix for [#21734](https://mantisbt.org/bugs/view.php?id=21734). It also does not match your own [#34407](https://mantisbt.org/bugs/view.php?id=34407). Please reference an appropriate issue in the tracker.

I performed a quick test, but it does not work as expected.

If you edit an issue and set status to resolved, you get _A necessary field "Fixed in Version" was empty_ regardless of selected Resolution. Did you read my earlier review note https://github.com/mantisbt/mantisbt/pull/1998#discussion_r1574463015 
> Additionally, you need to consider that a bug may be in RESOLVED status without having FIXED resolution (e.g. duplicate, no change required, etc) so it's necessary to consider bug_resolution_fixed_threshold

I don't understand the need for the new `fixed_in_version_required_status` hidden input; if it's just to pass value of bug_resolution_fixed_threshold to the javascript, then it's probably better handled with a data attribute. 

Additionally, note that technically more than one Resolution could be considered "fixed" (read PHPDoc for $g_bug_resolution_fixed_threshold) so I'm not sure that referring to a single value is appropriate.

The javascript does not work in edit page.

I did not test this, but again please make sure we have consistent behavior with SOAP/REST API.

core/bug_api.php
```diff
@@ -508,6 +508,16 @@ public function validate( $p_update_extended = true ) {
 			category_ensure_exists( $this->category_id );
 		}
 
+		if ( is_blank( $this->fixed_in_version ) && config_get( 'bug_fixed_in_version_required', null, null, $this->project_id ) != OFF ) {
+			$t_resolved = config_get( 'bug_resolved_status_threshold', null, null, $this->project_id );
+			$t_closed = config_get( 'bug_closed_status_threshold', null, null, $this->project_id );
+			$t_resolution_fixed = config_get( 'bug_resolution_fixed_threshold', null, null, $this->project_id );
+			if ( $this->status >= $t_resolved && ( $this->status < $t_closed || !$t_bug_resolution_is_fixed ) ) {
```
dregad, line 515: $t_bug_resolution_is_fixed variable is not defined - did you mean $t_resolution_fixed ?.

paspiz85:
I confused issues and agree: this PR can fix #34407 but only for changelog part.
I think that this PR can be closed because no longer qualifies as a fix for #21734.
Do I need to split issue #34407 in order to continue my work?

Regarding the changes:
- I understood your review note #1998 and now I fixed my code to use (I hope correctly) bug_resolution_fixed_threshold
- Included edit page (in this case I need to also check status changing)
- SOAP/REST API don't use validation bug_api then require an extra check. I will add in function mc_issue_update (mc_issue_api.php)

paspiz85:
Without modifications, via SOAP I receive:
`<SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/">
   <SOAP-ENV:Body>
      <SOAP-ENV:Fault>
         <faultcode>SOAP-ENV:Server</faultcode>
         <faultstring>Error Type: APPLICATION ERROR #11,
Error Description: Il campo obbligatorio "Assegnata alla versione" era vuoto.  Ricontrolla le informazioni che hai inserito.</faultstring>
      </SOAP-ENV:Fault>
   </SOAP-ENV:Body>
</SOAP-ENV:Envelope>`
This message is due to mc_error_handler that handle E_USER_ERROR as Server SOAP fault.

dregad:
Sorry I've been busy. Haven't had time to review changes submitted since my last review yet, will do as time allows, but just wanted to give you feedback on your questions

> I think that this PR can be closed because no longer qualifies as a fix for #21734.

Not necessary to close this PR - just make sure it references the correct issue(s), i.e. edit https://github.com/mantisbt/mantisbt/pull/1998#issue-2255916985

> Do I need to split issue [#34407](https://mantisbt.org/bugs/view.php?id=34407) in order to continue my work?

Since this is now only covering the Changelog (fixed_in_version), then I would propose to simply edit the summary and description of #34407 (which I would have to do for you) to reflect the new, reduced scope. 

I'm not sure if a separate issue for Roadmap (target_version) is needed, or already covered (more or less) by other, existing issues. I let you decide; if yes, then you can clone [#34407](https://mantisbt.org/bugs/view.php?id=34407) and edit to submit a new Issue.

paspiz85:
I updated issue in PR first comment (now issue is https://mantisbt.org/bugs/view.php?id=34429).
I also checked SOAP/REST services and added a logic to intercept validation and return the correct message

core/bug_api.php
dregad: > the main check is server-side 

Yes I understand that, but you are not addressing my point about code duplication. 

It's not blocking, but if you are not going to do something about it, then at least a comment should be added here as well as in common.js, to warn future developers that any changes to the logic must be applied to the other location.

dregad (changes requested):
Hello @paspiz85 

Apologies for the long delay in responding to you, I only found today some time to test this.

There are still some things that are not quite right.

Firstly, keep in mind that _$g_bug_resolved_status_threshold_ is a **threshold**, meaning that an issue is considered as resolved when its status is equal **or higher**; I already mentioned this in https://github.com/mantisbt/mantisbt/pull/1998#discussion_r1574463015. 

This means that the _fixed_in_version_ field is not only available in Resolved state, but also in later statuses (e.g. CLOSED with default settings). I would therefore expect the same validation rules to apply whenever an issue is considered resolved, i.e. if going straight from ASSIGNED to CLOSED, the version should also be required. Currently with your patch it is not.

Secondly, in the same comment referenced above, I also mentioned
> a bug may be in RESOLVED status without having FIXED resolution (e.g. duplicate, no change required, etc) so it's necessary to consider bug_resolution_fixed_threshold

Currently this does not work as expected. With status = RESOLVED, the mandatory check is active regardless of the selected resolution. Please read the PHPDoc for *$g_bug_resolution_fixed_threshold* and the related *$g_bug_resolution_not_fixed_threshold* configs. This means that fixed_in_version should be required when 
*fixed_threshold <= resolution < not_fixed_threshold*.

To fix this in the most consistent and backwards-compatible possible way, I would recommend to rely on *helper_call_custom_function()* to call `changelog_include_issue` https://github.com/mantisbt/mantisbt/blob/cd1f29156539d21f52df7c5c05aed246348b7d75/core/custom_function_api.php#L71-L77.

image

Also a small visual nitpick, the spacing between the mandatory mark and the label is not right on bug_update_page.php (look at Summary field for reference):
image

Note that I only tested the GUI (JavaScript), assuming that the same logic was used for back-end (including API) updates.

dregad:
> I can use changelog_include_issue but only server-side (in Javascript I need to check resolution value agains the two threshold manually), I would prefer have same checks without using changelog_include_issue

The problem is, if you do that then anybody who defined an override for that custom function will have inconsistent behavior between the client and server validation, which IMO is not acceptable.

Client-side validation is nice and more user-friendly, but unfortunately MantisBT was not designed for that. Maybe an AJAX ?

paspiz85:
I understand the problems, and with these functions, it’s very important to preserve server behavior.

I fixed the server behavior and removed all client-side behavior to ensure logical consistency.

Adding some client-side logic to create a more user-friendly interface could be useful, but unfortunately, I don't have enough time to set up an AJAX-based solution.

dregad (changes requested):
Hello @paspiz85 

I just tested this, but unfortunately it does not work as expected, and I am sorry but it's a bit of my fault as I pushed you to use existing custom function logic.

The problem is that *custom_function_default_changelog_include_issue()* receives an *issue ID* and uses it to retrieve information from the database, but we call it from BugData::validate() when the data has not yet been saved. Consequently, the test does not give the expected result, because we check the Issue's stored Resolution (e.g. OPEN == 10) instead of the new one (e.g. FIXED).

My first thought was to move the default comparison logic in custom_function_default_changelog_include_issue() to a helper function receiving a BugData object as parameter, or maybe even better a new BugData method.

The problem with this approach is that we cannot execute user-custom logic in custom_function_override_changelog_include_issue(), and I can't think of a good way to achieve that without breaking backwards-compatibility. 

If we accept the BC-break, we could change the parameter to the custom function from int to BugData. That would of course affect anyone who actually implemented an override, forcing them to adapt their code, but I'm sure most people stick to the default so the impact is probably not so wide.

Any better ideas ?

PR #2002, opened 2024-05-14
Don't offer columns without having access rights to them

atrol:
Fixes [#34454](https://mantisbt.org/bugs/view.php?id=34454)

### DIFF
--- a/core/columns_api.php
+++ b/core/columns_api.php
@@ -72,14 +72,28 @@ require_api( 'string_api.php' );
  */
 function columns_filter_disabled( array $p_columns ) {
 	$t_columns = array();
-	$t_enable_profiles = ( config_get( 'enable_profiles' ) == ON );
+	$t_enable_profiles = null;
+	$t_enable_due_date = null;
 
 	foreach ( $p_columns as $t_column ) {
 		switch( $t_column ) {
 			case 'os':
 			case 'os_build':
 			case 'platform':
+				if( $t_enable_profiles == null ) {
+					$t_enable_profiles = config_get( 'enable_profiles' ) == ON;
+				}
 				if( ! $t_enable_profiles ) {
+					continue 2;
+				}
+				break;
+
+			case 'due_date':
+			case 'overdue':
+				if ( $t_enable_due_date == null ) {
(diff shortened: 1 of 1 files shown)


core/columns_api.php
```diff
@@ -73,6 +73,7 @@
 function columns_filter_disabled( array $p_columns ) {
 	$t_columns = array();
 	$t_enable_profiles = ( config_get( 'enable_profiles' ) == ON );
+	$t_enable_due_date = access_has_project_level( config_get( 'due_date_view_threshold' ) );
```
vboctor, line 76: Should we only make the call to `access_has_project_level()` when $p_columns include the corresponding fields?

dregad: I'm with @vboctor here, better move this to the switch case, I believe it's better to risk running the access check twice (considering that due date and overdue are not so frequently used, and likely not simultaneously as they are redundant). 

And I guess the same reasoning applies to os/platform as well, although in this case it may be worth using a static var to cache the config value as the fields are often used together.

atrol:
@vboctor I optimized now the same way for the profiles and the due date columns.
It will not help that much like you might have expected, as we have the due date column in our default setting of `$g_bug_view_page_fields`

@dregad using static vars to cache the config settings will not work on pages where multiple issues from different projects are listed.
Am I missing something?

dregad:
No you're right, I did not consider this scenario. Ignore me.

dregad:
> we have the due date column in our default setting of `$g_bug_view_page_fields`

This is kind of silly, considering that we also have `$g_due_date_view_threshold = NOBODY;`... Shouldn't we remove due_date from $g_bug_view_page_fields ?

vboctor:
> > we have the due date column in our default setting of `$g_bug_view_page_fields`

> This is kind of silly, considering that we also have `$g_due_date_view_threshold = NOBODY;`... Shouldn't we remove due_date from $g_bug_view_page_fields ?

@dregad Removing `due_date` will affect instances that have `$g_due_date_view_threshold = DEVELOPER` (for example). I also don't think it is silly, since it is indicating that when due_date is accessible, it should be viewable in the View Issues page by default. This is like saying it should be visible on Report/Update/View pages when due date is enabled.

dregad:
OK I rest my case

PR #2005, opened 2024-05-19
Remove icons from view state and overdue columns

atrol:
Workaround for regression [#34442](https://mantisbt.org/bugs/view.php?id=34442)


dregad:
I started working on a fix for #34442 earlier this week, so there is no point in merging this. I will submit a pull request shortly.

PR #2006, opened 2024-05-20
Fix display of columns using icons as label in View Issues page

dregad:
Fixes [#34442](https://mantisbt.org/bugs/view.php?id=34442)

Supersedes PR #2005

### DIFF
--- a/core/columns_api.php
+++ b/core/columns_api.php
@@ -1062,7 +1062,7 @@ function print_column_title_due_date( $p_sort, $p_dir, $p_columns_target = COLUM
 function print_column_title_overdue( $p_sort, $p_dir, $p_columns_target = COLUMNS_TARGET_VIEW_PAGE ) {
 	echo '<th class="column-overdue">';
 	$t_overdue_text = lang_get( 'overdue' );
-	print_view_bug_sort_link( $t_overdue_text, 'due_date', $p_sort, $p_dir, $p_columns_target, 'fa-times-circle-o' );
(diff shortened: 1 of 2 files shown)


core/columns_api.php
```diff
@@ -1063,8 +1062,7 @@ function print_column_title_due_date( $p_sort, $p_dir, $p_columns_target = COLUM
 function print_column_title_overdue( $p_sort, $p_dir, $p_columns_target = COLUMNS_TARGET_VIEW_PAGE ) {
 	echo '<th class="column-overdue">';
 	$t_overdue_text = lang_get( 'overdue' );
-	$t_overdue_icon = ' ' . icon_get( 'fa-times-circle-o', '', $t_overdue_text );
-	print_view_bug_sort_link( $t_overdue_icon, 'due_date', $p_sort, $p_dir, $p_columns_target );
+	print_view_bug_sort_link( $t_overdue_text, 'due_date', $p_sort, $p_dir, $p_columns_target, 'clock-o' );
```
atrol, line 1065: Any specific reason that you use `clock-o` but not `fa-clock-o`?
Being aware that you introducded in `icon_get`
```
	# Add 'fa-' prefix if missing
	if( substr( $p_icon, 0, 3 ) != 'fa-' ) {
		$p_icon = 'fa-' . $p_icon;
	}
```

dregad: > Any specific reason that you use clock-o but not fa-clock-o?
> Being aware that you introducded in icon_get

Sometimes I just forget things...
And this is the exact reason why I added that check 😉

PR #2011, opened 2024-06-16
Refactoring and cleaning up includes

dregad:
- Replace dirname( __FILE__ ) by __DIR__ (3)
- Replace unnecessary DIRECTORY_SEPARATOR by '/'

Fixes [#34468](https://mantisbt.org/bugs/view.php?id=34468)

Also, set current working dir in core.php as a follow-up to [#33906](https://mantisbt.org/bugs/view.php?id=33906), and related required adjustment to mantisconnect.php.

### DIFF
--- a/core.php
+++ b/core.php
@@ -66,7 +66,7 @@ if( file_exists( 'mantis_offline.php' ) && !isset( $_GET['mbadmin'] ) ) {
 $g_request_time = microtime( true );
 
 # Load supplied constants
-require_once( dirname( __FILE__ ) . DIRECTORY_SEPARATOR . 'core' . DIRECTORY_SEPARATOR . 'constant_inc.php' );
+require_once( __DIR__ . DIRECTORY_SEPARATOR . 'core' . DIRECTORY_SEPARATOR . 'constant_inc.php' );
 
 # Enforce our minimum PHP requirements
 if( version_compare( PHP_VERSION, PHP_MIN_VERSION, '<' ) ) {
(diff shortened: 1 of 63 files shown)


core.php
```diff
@@ -103,6 +103,11 @@
 	require_once( $g_config_path . 'config_inc.php' );
 }
 
+# Ensure we always have a consistent current working directory.
+# IIS with FastCGI seems to set CWD to directory where php.exe resides (see #33906).
+global $g_absolute_path;
+chdir( $g_absolute_path );
```
atrol, line 109: Looks at first sight like a workaround for issues that might be caused by a PHP 8.2 bug on Windows + IIS + FastCGI.
We shouldn't rely at all on CWD, e.g. a plugin could change it.
As a side note: This would be the first time we use chdir in code.

grummbeer:
`dirname(__DIR__);` looks like a double whopper.

https://www.php.net/manual/en/language.constants.magic.php#constant.dir

> The directory of the file. If used inside an include, the directory of the included file is returned. This is equivalent to dirname(__FILE__). This directory name does not have a trailing slash unless it is the root directory.

Therefore the correct replacemant is just `__DIR__`, because `dirname(__DIR__)` means a extra - unecessary and slower - function call.

```php
require_once( dirname( __FILE__ ) . '/mc_core.php' );
```

become (best without brackets)

```php
require_once __DIR__ . '/mc_core.php';
```

dregad:
> `dirname(__DIR__);` looks like a double whopper.

It's not. We're looking for the parent of the script's directory. 

`dirname(__DIR__);` == `dirname(__FILE__, 2);`

core.php
dregad: > might be caused by a PHP 8.2 bug on Windows + IIS + FastCGI

No idea if that's a PHP bug or not, I didn't search for it. It is a corner case anyway and not worth my time to analyze all the way through.

> We shouldn't rely at all on CWD, e.g. a plugin could change it.

We actually don't rely on CWD anymore, since the original fixes for [#33906](https://mantisbt.org/bugs/view.php?id=33906) I committed in March.

Setting CWD in core.php is actually not strictly necessary with the current code base as we now only use absolute path references. Consider that a "belt and braces" approach, trying to reduce the risk of this error occurring again. 

I thought it made sense for the MantisBT root to consistently be set as CWD, but if you disagree and don't want the chdir(), I'm fine with reverting that commit.

atrol: I prefer to remove the unneeded chdir().

dregad: OK

atrol (approved):

PR #2019, opened 2024-07-28
Also returns obsolete project versions through REST API

ctabin:
Hello,

It should be possible to access *obsolete* versions through API otherwise it is impossible to dynamically manage them because there is no other way to get them (other than going manually with the id).

Ideally a query parameter would be the best.

Fixes [#34578](https://mantisbt.org/bugs/view.php?id=34578)

### DIFF
--- a/api/rest/restcore/projects_rest.php
+++ b/api/rest/restcore/projects_rest.php
@@ -275,11 +275,17 @@ function rest_project_version_get( \Slim\Http\Request $p_request, \Slim\Http\Res
 	}
 
 	$t_version_id = $p_args['version_id'] ?? $p_request->getParam( 'version_id' );
+	$t_released = (bool)$_request->getQueryParam( 'released', true );
+	$t_obsolete = (bool)$_request->getQueryParam( 'obsolete', false );
+	$t_inherited = (bool)$_request->getQueryParam( 'inherited', false );
 
 	$t_data = array(
 		'query' => array(
 			'project_id' => $t_project_id,
-			'version_id' => $t_version_id
+			'version_id' => $t_version_id,
+			'r_released' => $t_released,
+			'r_obsolete' => $t_obsolete,
+			'r_inherited' => $t_inherited
 		)
 	);
 
--- a/core/commands/VersionGetCommand.php
+++ b/core/commands/VersionGetCommand.php
@@ -40,6 +40,27 @@ class VersionGetCommand extends Command {
 	 * @var integer|null
 	 */
 	private $version_id;
+	
+	/**
+	 * The released versions will be returned.
+	 *
+	 * @var boolean
+	 */
+	private $r_released;
+	
+	/**
+	 * The obsolete versions will be returned.
+	 *
+	 * @var boolean
+	 */
+	private $r_obsolete;
+	
+	/**
+	 * The inherited versions will be returned.
+	 *
+	 * @var boolean
+	 */
+	private $r_inherited;
 
 	/**
 	 * $p_data['query'] is expected to contain:
@@ -100,9 +121,9 @@ class VersionGetCommand extends Command {
 		if( is_null( $this->version_id ) ) {
 			$t_versions = version_get_all_rows(
 				$this->project_id,
-				/* released */ $this->query( 'released' ),
-				/* obsolete */ $this->query( 'obsolete', false ),
-				/* inherit */ $this->query( 'inherit', false ) );
+				/* released */ $this->query( 'released', $r_released ),
+				/* obsolete */ $this->query( 'obsolete', $r_obsolete ),
+				/* inherit */ $this->query( 'inherit', $r_inherited ) );
 
 			$t_versions = array_map( 'VersionGetCommand::VersionRowToArray', $t_versions );
 		} else {


dregad:
@ctabin are you still working on this ?

ctabin:
@dregad Please find an improved version.

ctabin:
@dregad I might need some help here since the pipeline fails and I don't see why...

api/rest/restcore/projects_rest.php
```diff
@@ -275,11 +275,17 @@ function rest_project_version_get( \Slim\Http\Request $p_request, \Slim\Http\Res
 	}
 
 	$t_version_id = $p_args['version_id'] ?? $p_request->getParam( 'version_id' );
+	$t_released = (bool)$_request->getQueryParam( 'released', true );
```
dregad, line 278: This is wrong. As documented in version_get_all_rows() PHPDoc, *released* is ternary, see 

https://github.com/mantisbt/mantisbt/blob/65cf0d5bc2560f789a218e9e282604a3171d97df/core/version_api.php#L553-L557

so you can't just cast it to bool, because this excludes VERSION_ALL (null).

And I think the default should be `null`, not `true`, matching current API behavior.

core/commands/VersionGetCommand.php
```diff
@@ -40,6 +40,27 @@ class VersionGetCommand extends Command {
...
+	 * @var boolean
+	 */
+	private $r_obsolete;
+	
+	/**
+	 * The inherited versions will be returned.
+	 *
+	 * @var boolean
+	 */
+	private $r_inherited;
```
dregad, line 63: I don't understand the purpose of these new properties, they are never assigned.

api/rest/restcore/projects_rest.php
```diff
@@ -275,11 +275,17 @@ function rest_project_version_get( \Slim\Http\Request $p_request, \Slim\Http\Res
...
+	$t_inherited = (bool)$_request->getQueryParam( 'inherited', false );
 
 	$t_data = array(
 		'query' => array(
 			'project_id' => $t_project_id,
-			'version_id' => $t_version_id
+			'version_id' => $t_version_id,
+			'r_released' => $t_released,
+			'r_obsolete' => $t_obsolete,
+			'r_inherited' => $t_inherited
```
dregad, line 288: Why the `r_` prefix ? 

And the key does not match what you're retrieving in VersionGetCommand::process()
https://github.com/mantisbt/mantisbt/pull/2019/files#diff-93a44681526fcc6496dc1bea69b2d7e596fcb7485be30e49287408df5ee6b1deR124-R126

api/rest/restcore/projects_rest.php
```diff
@@ -275,11 +275,17 @@ function rest_project_version_get( \Slim\Http\Request $p_request, \Slim\Http\Res
 	}
 
 	$t_version_id = $p_args['version_id'] ?? $p_request->getParam( 'version_id' );
+	$t_released = (bool)$_request->getQueryParam( 'released', true );
+	$t_obsolete = (bool)$_request->getQueryParam( 'obsolete', false );
+	$t_inherited = (bool)$_request->getQueryParam( 'inherited', false );
```
dregad, line 280: *inherit* is also ternary, and should default to `null` as well.

core/commands/VersionGetCommand.php
```diff
@@ -100,9 +121,9 @@ protected function process() {
 		if( is_null( $this->version_id ) ) {
 			$t_versions = version_get_all_rows(
 				$this->project_id,
-				/* released */ $this->query( 'released' ),
-				/* obsolete */ $this->query( 'obsolete', false ),
-				/* inherit */ $this->query( 'inherit', false ) );
+				/* released */ $this->query( 'released', $r_released ),
+				/* obsolete */ $this->query( 'obsolete', $r_obsolete ),
+				/* inherit */ $this->query( 'inherit', $r_inherited ) );
```
dregad, line 126: Command::query()'s 2nd parameter is supposed to be the default value, which would have been set in rest_project_version_get() when retrieving the request's parameters so this change is probably not necessary.

dregad (changes requested):
Did you test this ? It does not work, the API endpoint returns HTTP 500 error.

```
Slim Application Error

Type: Error
Message: Call to a member function getQueryParam() on null
File:.../mantisbt/api/rest/restcore/projects_rest.php
Line: 278
```

This is caused by referencing non-existing variable  `$_request` instead of `$p_request`.

If I change that, I get several PHP warnings
```
Warning: Undefined variable $r_released in ./mantisbt/core/commands/VersionGetCommand.php on line 124
Warning: Undefined variable $r_obsolete in ./mantisbt/core/commands/VersionGetCommand.php on line 125
Warning: Undefined variable $r_inherited in .vmantisbt/core/commands/VersionGetCommand.php on line 126
```

dregad:
@ctabin are you planning to address the review comments ?

ctabin:
@dregad Unfortunately I didn't have time to work on this PR... Maybe it would be better if you take it over directly.

dregad:
Well I'm sorry to say, but I I am busy too. This was *your* request... If you have no interest in finalizing the PR (or at the very least answer my questions), then I'm just going to close it without merge, because I can't figure out what your intention was without your feedback.

PR #2020, opened 2024-08-06
Fix project_hierarchy_cache() to cache both full and enabled-only hierarchies

aaribaud-fgs:
Currently, project_hierarchy_cache() caches either all projects or enabled project only, whichever was requested last.

The project edit page calls project_hierarchy_cache() for each subproject in turn, once for the whole hierarchy, and once for the enabled projects hierarchy, which causes a cache reload on each call.

This commit caches both hierarchies so that alternating calls will not cause cache reloads any more.

Fixes [#34526](https://mantisbt.org/bugs/view.php?id=34526)

### DIFF
--- a/core/project_hierarchy_api.php
+++ b/core/project_hierarchy_api.php
@@ -31,9 +31,15 @@
 require_api( 'constant_inc.php' );
 require_api( 'database_api.php' );
 
+# We may need to build caches for either all projects or enabled projects or both
+$g_cache_project_hierarchy_enabled = null;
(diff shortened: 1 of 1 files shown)


dregad:
Merge pending feedback from Author, see https://mantisbt.org/bugs/view.php?id=34526#c69089

dregad (approved):

dregad:
Merged in 5ff35d16e85c1aa58cc80fea260cd0f3cfc39910

PR #2023, opened 2024-09-02
Allow display of additional info for ERROR_GENERIC  messages

dregad:
Fixes [#34613](https://mantisbt.org/bugs/view.php?id=34613)

Also fixes unexpected display of `<br>` HTML tag in error messages on CLI [#34612](https://mantisbt.org/bugs/view.php?id=34612).


dregad:
Merge this before #2024

PR #2025, opened 2024-09-04
Update account_prefs_inc.php

pokrad:
Invalid class for:
<select id="timezone" name="timezone" class="input-sm">


pokrad:
No problem, I was not aware of the process.. sorry.

 

Regards !

PR #2027, opened 2024-09-06
Fix disabled projects are not listed on page manage_proj_page.php

dregad:
Fixes [#34618](https://mantisbt.org/bugs/view.php?id=34618) (regression introduced by PR #2020)

@aaribaud could you please review this, to ensure my fix is not breaking the improved cache you introduced; my testing shows that it does not, but a second pair of eyes would be nice... Merci d'avance

### DIFF
--- a/core/project_hierarchy_api.php
+++ b/core/project_hierarchy_api.php
@@ -279,11 +279,7 @@ function project_hierarchy_get_subprojects( $p_project_id, $p_show_disabled = fa
 
 	project_hierarchy_cache( $p_show_disabled );
 
-	if( isset( $g_cache_project_hierarchy[$p_project_id] ) ) {
-		return $g_cache_project_hierarchy[$p_project_id];
-	} else {
-		return array();
-	}
+	return $g_cache_project_hierarchy[$p_project_id] ?? [];
 }
 
 /**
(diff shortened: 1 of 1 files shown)


atrol (approved):
Tested that the regression is fixed.
No performance tests.

aaribaud:
* Changes from lines 282..286 into single line 282, and from line 302 to line 298, are obvious no-ops so won't have any impact anyway.
* Changes on lines 159..171 remove `isnull()` calls and factor out two `return` statements into one, but otherwise keep the original functionality, so no impact there.
* change on lines 211..216 amounts to removing the `isnull()` call, so no impact there either.

... Which makes me wonder how these four changes, which all appear to not affect functionality, can have any impact on [#35418](https://mantisbt.org/bugs/view.php?id=34618). What am I missing?

dregad:
Thanks for the feedback @aaribaud 

> wonder how these four changes, which all appear to not affect functionality, can have any impact on [#35418](https://mantisbt.org/bugs/view.php?id=34618)

It does, because `false !== null`, therefore condition `is_null( $p_show_disabled )` is **always false**. 

This caused the *cache all* globals to be populated regardless of $p_show_disabled parameter's value, and since the first call to project_hierarchy_cache() happens to be with $p_show_disabled = false (to populate the project selection list) then the *all* cache incorrectly contains the *enabled* hierarchy when manage_proj_page.php wants to get the list of all projects including disabled ones.

So the key change is the removal of `is_null()`, resulting in a true boolean test, which yields the expected result.

Factoring out the return statement is indeed a noop, and just makes the code clearer (IMHO).

PR #2028, opened 2024-09-20
Return HTTP response code when error occurs

dregad:
Until now, our error handler displayed an error page, but actually
returned HTTP status code 200 (success), which was causing problems in
certain usage scenarios.

The REST and SOAP APIs do define a mapping from Mantis error to the
corresponding HTTP response code ApiObjectFactory::faultFromException().

This moves the mapping logic to a new Error API function
error_map_mantis_error_to_http_code(), and the error handler calls
http_response_code() with its return value so a proper HTTP status code
is returned to the user agent.

faultFromException() method was modified to use the new API function to
avoid code duplication.

Fixes [#34634](https://mantisbt.org/bugs/view.php?id=34634)

Includes a couple of code cleanup commits.


dregad:
@c-schmitz testing and feedback would be appreciated.

c-schmitz:
Great, will give it a whirl next week!

vboctor (approved):

dregad:
@c-schmitz did you get a chance to test ?

c-schmitz:
@dregad I cannot see this being fixed in 2.27.0

PR #2029, opened 2024-09-20
Fix broken absolute URL on proj_doc_page.php

dregad:
Fixes [#32557](https://mantisbt.org/bugs/view.php?id=32557)


vboctor (approved):

vboctor:
Does this change have an impact on developer manual relating to some plugin event?

dregad:
I'll double check, but I don't think so.

Plugin behavior is unchanged (they must provide cooked html links),  and the array format was not documented until now.

PR #2030, opened 2024-09-21
Fix various REST & SOAP API issues

dregad:
- [#26672](https://mantisbt.org/bugs/view.php?id=26672): mc_issue_add fails with "Object of class SoapFault could not be converted to int"
- [#34683](https://mantisbt.org/bugs/view.php?id=34683): REST POST /issues allows creation of Issue when invalid Category is specified
- [#34684](https://mantisbt.org/bugs/view.php?id=34684): SOAP API throwing deprecation warning on PHP 8.1

Plus some code cleanup.


tests/rest/RestIssueTest.php
```diff
@@ -547,6 +547,21 @@ public function testCreateIssueNoCategory() {
 		}
 	}
 
+	public function testCreateIssueWithCategoryInvalid() {
+		$t_issue_to_add = $this->getIssueToAdd();
+		$t_issue_to_add['category'] = 'Non-existent category';
+
+		$t_response = $this->builder()->post( '/issues', $t_issue_to_add )->send();
+
+		$this->assertEquals( HTTP_STATUS_NOT_FOUND, $t_response->getStatusCode() );
```
vboctor, line 556: I wonder if in such case we should be returning NOT FOUND or BAD REQUEST.

- NOT FOUND - the argument is that format is valid, just one of the foreign keys in not found.
- BAD REQUEST - the argument is that 404 should be returned if we are attempting to get an element that doesn't exist, but in this case we are creating an element with an invalid value, hence, bad request.

vboctor (approved):
Looks good. I provided a comment on the status code, but we can go either way.

dregad:
I thought about that too actually, I agree that bad request would make more sense in this context. 

The problem is that the http response code is determined based on the Mantis Error code which is thrown in the Exception, I.e. "category not found". 

Not sure how I could change that behavior, do you have a suggestion?

PR #2031, opened 2024-10-03
Add release marker for database schema

atrol:
Fixes [#34813](https://mantisbt.org/bugs/view.php?id=34813)

### DIFF
--- a/admin/schema.php
+++ b/admin/schema.php
@@ -902,6 +902,7 @@ $g_upgrade[212] = array( 'AlterColumnSQL', array( db_get_table( 'category' ), "
 	status					I		UNSIGNED NOTNULL DEFAULT '1' ") );
 $g_upgrade[213] = array( 'UpdateFunction', 'category_status_default' );
 
+# Release marker: 2.27.0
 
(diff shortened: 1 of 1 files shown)


vboctor (approved):
Do we really need a bug or even a PR for adding a comment line? I suggest deleting the bug.

dregad:
> Do we really need a bug or even a PR for adding a comment line? I suggest deleting the bug.

I agree on principle - I would personally have committed the fix directly without a PR. Regarding the Issue, it was created by a user so we should not delete it.

dregad (approved):

vboctor:
@dregad @atrol I missed the fact that a user reported this. Makes sense.

atrol:
@dregad @vboctor right, PR is a bit overhead, I will push such small changes without asking next time.

Having linked issues to commits should be a must in my opinion, even if there is none created by external users.
I am also working on some products where I configured branch policies for master and release branches.
No one is allowed to push to these branches without having a linked work item.
Most of the team members like it, especially the ones that have to deal with fixing regressions and writing test cases for such fixes.

PR #2032, opened 2024-10-06
WIP: REST API OpenAPI Spec

vboctor:
Fixes #[34823](https://mantisbt.org/bugs/view.php?id=34823)

TODO: Cleanup schemas
TODO: Cleanup Auth
TODO: Remove swagger.json


api/rest/mantisbt_openapi.yaml
```diff
@@ -0,0 +1,1733 @@
+openapi: 3.0.0
```
dregad, line 1: Latest OpenAPI spec version is 3.1.0.

api/rest/mantisbt_openapi.yaml
```diff
@@ -0,0 +1,1733 @@
...
+info:
+  title: Mantis Bug Tracker REST API
+  version: 2.27.0
+  license:
+    name: GPL-2.0
```
dregad, line 6: *name* key's value should be `GNU General Public License v2.0 or later`.

SPDX-License expression should be specified under *identifier* key (then the _url_ key should be removed). Correct value for us is `GPL-2.0-or-later`.

api/rest/mantisbt_openapi.yaml
```diff
@@ -0,0 +1,1733 @@
+openapi: 3.0.0
+info:
+  title: Mantis Bug Tracker REST API
+  version: 2.27.0
```
dregad, line 4: Suggest adding a *description* key, see example 
https://github.com/dregad/mantisbt/blob/7199f739cf7cf9b116a918522d1c03f45efe25e3/api/rest/openapi.yaml#L9-L20

The above was originally generated by Postman export IIRC. This is Markdown text.

Also, a *contact* key would be nice
```yaml
  contact:
    name: MantisBT Team
    email: mantisbt-dev@lists.sourceforge.net
    url: https://mantisbt.org
```

api/rest/mantisbt_openapi.yaml
```diff
@@ -0,0 +1,1733 @@
...
+      description: ''
+      tags:
+        - 'Filters'
+      responses:
+        '200':
+          description: 'Success'
+        '4XX':
+          description: 'Bad Request'
+      parameters:
+        - name: filter_id
```
dregad, line 22: Most parameters are reused all over the API, so they should be defined globally under *components* to avoid duplication

api/rest/mantisbt_openapi.yaml
```diff
@@ -0,0 +1,1733 @@
...
+          in: path
+          schema:
+            type: integer
+            minimum: 1
+          description: ''
+          required: true
+      requestBody:
+        $ref: '#/components/requestBodies/update_user'
+servers:
+  - url: https://mantisbt.org/bugs/
```
dregad, line 1001: Add server entry to describe the user's instance,

https://github.com/dregad/mantisbt/blob/7199f739cf7cf9b116a918522d1c03f45efe25e3/api/rest/openapi.yaml#L29-L47

dregad (commented):
Some early comments and suggestions.

api/rest/mantisbt_openapi.yaml
vboctor: I will update to `3.0.3` since that version seems to have better tooling support than `3.1.0`. We can update later, as needed.

PR #2034, opened 2024-10-06
Fix error when clearing default profile

atrol:
Fixes [#34826](https://mantisbt.org/bugs/view.php?id=34826)

### DIFF
--- a/account_prof_menu_page.php
+++ b/account_prof_menu_page.php
@@ -201,7 +201,7 @@ if( $g_global_profiles ) {
 					print_form_button(
 						'account_prof_update.php',
 						lang_get( $t_is_default_profile ? 'clear_default' : 'make_default' ),
-						array_merge( $t_param, array( 'action' => 'make_default' ) ),
+						array_merge( $t_param, array( 'action' => 'change_default' ) ),
 						$t_security_token
 					);
 					echo '</div>';
--- a/account_prof_update.php
+++ b/account_prof_update.php
@@ -61,9 +61,10 @@ $f_redirect_page = gpc_get_string( 'redirect', 'account_prof_menu_page.php' );
 
 if( $f_action != 'add' ) {
 	$f_profile_id = gpc_get_int( 'profile_id' );
-	profile_ensure_can_update( $f_profile_id );
-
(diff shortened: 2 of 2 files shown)


atrol:
This some kind of hidden way to clear the default profile by using profile_id 0 was probably one of the reasons why this regression was introduced.
You were not aware when changing the code, I was not aware when reviewing the code, you and me were not aware that testing of this workflow is needed.
Isn't the code now better readable and easier to understand, especially in account_prof_menu_page.php?

dregad:
> was probably one of the reasons why this regression was introduced

Not at all, it's just because I failed to consider the set/clear default case when moving code out of the switch and into the ` if( $f_action != 'add' ) {` block.

There is in fact another regression, if an unprivileged user tries to set a global category as default they get *access denied*.

> Isn't the code now better readable and easier to understand

I don't think so IMHO. I actually find this bit `if( $f_action != 'make_default' && $f_action != 'clear_default' ) {` makes things worse, and having the same call to current_user_set_pref() differing only by using 0 or $f_profile_id (when we know that the variable's value is 0), feels like code duplication.

What we should probably do to make code clearer and self-documenting, is to define a proper constant e.g. `PROFILE_DEFAULT_UNSET` to replace the use of magic constant 0. And maybe add a comment like `# No need to check if user can update the Profile when setting default` before the if in my patch.

atrol:
What confuses me in the current approach, is to name an action `make_default` but to have two different actions behind it, one of them clearing the default.

@dregad as you don't like my approach to separate it (I am also not 100% convinced about the readability of the code)), what do you think about to use something like `set_profile_preference` instead of `make_default` for the action name ?

dregad:
> what do you think about to use something like set_profile_preference instead of make_default for the action name ?

The term `set_profile_preference` just feels a bit strange, as we are actually changing the default profile column (`current_user_set_pref( 'default_profile', $f_profile_id );`).

Even though I understand your argument about *make_default* being somewhat inappropriate when we're effectively clearing it, I honestly don't see much benefit from changing it, but I guess it would not really hurt either. 

If you insist on changing the action string, I would rather go for `set_default` or `change_default`.

atrol:
> If you insist on changing the action string, I would rather go for set_default or change_default.

Changed to `change_default`

dregad (approved):

PR #2037, opened 2024-10-12
MantisGraph plugin: upgrade chart.js library to 3.9.1

dregad:
Fixes [#34847](https://mantisbt.org/bugs/view.php?id=34847), [#34848](https://mantisbt.org/bugs/view.php?id=34848)


atrol:
Doesn't work with `$g_cdn_enabled = ON;`

dregad:
> Doesn't work with `$g_cdn_enabled = ON;`

Fixed, sorry I forgot to test this.

atrol:
The "Issue Trends" line chart is no longer interpolated.
Is it intended or a side effect of the upgrade?
I am not sure if interpolation should be reintroduced.
It looks nice, but the intermediate values are questionable, calculation stresses client CPU / battery, it's not needed if lots of input values are provided. 

Your branch
dregad
Master branch
master

dregad:
> The "Issue Trends" line chart is no longer interpolated.
Is it intended or a side effect of the upgrade?

This was not intentional. I failed to notice that regression, as I did most of my testing with a mantisbt.org dump, which has a lot of data points on this graph and that causes the bézier effect to become invisible. 

I just checked, and it turns out this setting's default changed from 2.x (`lineTension: 0.4`) to 3.x (`tension: 0`) but that was not documented in the [v3 migration guide](https://www.chartjs.org/docs/latest/migration/v3-migration.html) which is why I missed it. 

> I am not sure if interpolation should be reintroduced.
It looks nice, but the intermediate values are questionable, calculation stresses client CPU / battery, it's not needed if lots of input values are provided.

I agree that the soft curves look nicer, and agree as well with the fact that on the downside it could alter perception of the data. I did not actually measure it, but I don't think the extra CPU consumption is significant.

About adding it back or not, we can also use a lower *tension* setting, to have the nice-looking soft lines with a limited alteration of the data. I think that a value of `0.2` could be a good compromise. If you want to play with this, just add the following in the `options` section in MantisGraph.js (line 92)
```
  datasets: {
      line: {
          tension: 0.2,
      }
  },
```

See screenshots below to illustrate the differences

> Check also the changed background colors.

Now *that* was intentional. The transparent filler colors overlapping each other have always bothered me. 
See commit message fa96f4399fc9c2f84bba815fcc230585087af525 for a more detailed explanation. 

image
image
image

atrol:
> Now that was intentional. The transparent filler colors overlapping each other have always bothered me.

It's a bit inconsistent, as the legend shows red line and light red fill area for "Still Open".
Also a bit confusing that there is no fill area for "Still Open" when deactivating "Reported" and "Resolved".

dregad:
You did not comment on the adjusted interpolation variants. Should I go with `0.2` ?

> the legend shows red line and light red fill area for "Still Open".

I could not find a way to customize the legend's color boxes.

> Also a bit confusing that there is no fill area for "Still Open" when deactivating "Reported" and "Resolved".

I personally find it *more* confusing when there is a fill color for "still open", which causes the "reported" blue color to turn purple, not to mention that the visualisation becomes misleading because "still open" is the  delta from the other 2 sets. 

Keep in mind that what we're showing here, is effectively a stacked line graph: reported + still open = total reported. So the "still open" red line is effectively showing the same data as the green area. One solution could be to remove the red line completely.

dregad:
Thinking of stacked lines gave me a better idea... If we *actually* stack the *resolved + still open* datasets, then we can "hide" the *resolved* line if it is drawn on the same y scale, as it will overlap the cumulated *resolved + still open* but still be present.

* All 3 datasets are shown
  image
* With *Resolved* hidden
  image

I believe this is an elegant solution to the problem. What do you think ?

atrol:
> You did not comment on the adjusted interpolation variants. Should I go with 0.2 ?

Prefer 0.3

> I believe this is an elegant solution to the problem. What do you think ?

Looks good.

atrol (approved):

PR #2038, opened 2024-10-14
Allow adding a note when moving an issue

dregad:
Fixes [#6803](https://mantisbt.org/bugs/view.php?id=6803)


atrol:
A bug note gets added even if no note text was entered.

dregad:
Duh. Can't believe I overlooked such a basic check 😯 Thanks for testing, I'll fix that

atrol (approved):

PR #2043, opened 2024-10-27
On Windows, GraphViz tools need '.exe' extension

atrol:
Fixes [#34917](https://mantisbt.org/bugs/view.php?id=34917)


vboctor (approved):

dregad (approved):
See 32af908f57b73dfa40ec146b8d7fac39d1bc1cdc fixing the same bug in graph_api.php

PR #2046, opened 2024-11-09
mc_issue_update: Don't delete due_date 

tgulacsi:
If don't have access level to modify it, then don't modify it (current code does erase the due date instead of leaving it as is).

Fixes [#34959](https://mantisbt.org/bugs/view.php?id=34959)


dregad (approved):

dregad:

PR #2047, opened 2024-11-10
Update Version date conversion with non-US format

dregad:
VersionAddCommand and VersionUpdateCommand now accept a Unix timestamp as payload for timestamp (date_order).

manage_proj_ver_update.php calls date_strtotime() on the $f_timestamp parameter to build the VersionUpdateCommand payload.

Fixes [#34928](https://mantisbt.org/bugs/view.php?id=34928)

### DIFF
--- a/api/soap/mc_api.php
+++ b/api/soap/mc_api.php
@@ -208,11 +208,13 @@ class ApiObjectFactory {
 	}
 
 	/**
-	 * Convert a timestamp to a soap DateTime variable
-	 * @param integer $p_value Integer value to return as date time string.
-	 * @return datetime in expected API format.
+	 * Convert a timestamp to a soap DateTime variable or ISO-8601 date.
+	 *
+	 * @param int $p_value Unix Timestamp.
+	 *
+	 * @return SoapVar|string datetime in expected API format.
 	 */
-	static function datetime($p_value ) {
+	static function datetime( $p_value ) {
 		$t_string_value = self::datetimeString( $p_value );
 
 		if( ApiObjectFactory::$soap ) {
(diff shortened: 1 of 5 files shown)


vboctor (changes requested):
Let's make sure we accept and generate a single format across our APIs to avoid confusion. Hence, we should have all methods that parse or format datetime values in one place. Have a look at: `ApiObjectFactory::datetime()` and `ApiObjectFactory::datetimeString()`.

dregad:
Thanks for the review @vboctor. I'm not sure I understand your point though.

Are you saying it's a problem that the Commands are now accepting a Unix timestamp as input for the Project Version's date_order ? I think it makes sense to allow that (at least for internal use) because this is how we process dates in Core (as a side note, we had a related discussion on what dates format to accept from API in #1864). It seemed silly and wasteful to perform multiple conversions from localized date format (original user input from the GUI) to ISO-8601 format (expected by REST API / Command) to finally get to a timestamp (internal storage). 

Or do you want to move the updated date conversion logic from VersionAddCommand::validate() and VersionUpdateCommand::process() to a new helper method in ApiObjectFactory ?

Or is it something else ?

Please clarify.

vboctor:
I have two points:

1. All APIs should use the same parsing and generation logic -- hence, the parsing logic should be moved to ApiObjectFactory as a helper method that can be used by any API now or in the future that parses dates.

2. We don't typically have many cases where we accept dates, and in this case, it is associated with adding a version or updating a version, which is a very low-frequency action. Hence, I'm OK with conversion to standard ISO format (which is more readable for debugging) and back. This is much lower overhead than conversion to JSON and back anyway for cases like Web or SOAP using the commands.

Hence, I'm of the opinion of sticking with the standard ISO format.

dregad:
@vboctor I rebased the branch, and added a commit that addresses your review comments:
- revert the change to VersionAddCommand and VersionUpdateCommand accepting a Unix timestamp as payload for timestamp (date_order).
- The datetime parsing and generation logic has been standardized in 2 newdate API functions, used in ApiObjectFactory datetime*() methods

dregad:
After over 6 months I assume everybody is OK with the change.

PR #2051, opened 2024-11-26
Change the image type to svg for relationship and workflow graphs.

raspopov:
All modern browsers support svg images, so why not use that? Graph in svg format is smaller and much sharper. And text can even be copied to the clipboard.

Fixes [#35038](https://mantisbt.org/bugs/view.php?id=35038).


dregad:
Hello @raspopov, 

I think switching to SVG makes sense, thanks for your contribution. 

Did you actually test this ? I just tried it locally, but I'm not getting any output in the Mantis GUI; if I open the generated SVG file directly, there is a malformed XML error. This is on Linux though; I don't have a Windows environment to test with.

dregad:
The XML error is caused by GraphViz not encoding the `&` in the nodes' _URL_ attribute.

This is a [known issue](https://gitlab.com/graphviz/graphviz/-/issues/1687) in the library that has actually been [fixed in release 2.42.4](https://gitlab.com/graphviz/graphviz/-/blob/main/CHANGELOG.md#2424---2020-04-05), but my box is running an older version of dot (as installed by apt package manager: 2.42.2-6ubuntu0.1, reported as 2.43.0 by `dot -V`) which does not have the fix. 

I'm reluctant to merge this for now, because it would introduce a regression for all users running on Ubuntu < 24.10 which [are still bundling 2.42.x](https://packages.ubuntu.com/search?keywords=graphviz&searchon=names)), forcing them to manually install a newer version of GraphViz.

This can be reconsidered when the next Ubuntu LTS including the fix (26.04) is released.

raspopov:
> Did you actually test this ? 

Yes, I tested this on Windows with Graphviz versions 3.0 and 12.2.0 (the latest).

It might be worth adding a new configuration option with a default ‘png’ type. Or even default type selection based on Graphviz version check.

dregad:
Thanks for the feedback. 

> default type selection based on Graphviz version check.

Interesting idea. The drawback is that an additional `dot -V` system call would be required to retrieve the version number every time a graph needs to be rendered. Seems a bit costly.

> adding a new configuration option with a default ‘png’ type. 

That would work, but is it really worth adding a new config for this ?

dregad:
This was merged in commit 7c45c3f562d1fdf2d8885cc87a63ff5a13aa553f a couple weeks ago, I forgot to close the PR.

PR #2054, opened 2024-12-04
Request for improvement of plugin_file_path() function.

raspopov:
This PR improved plugin_file_path() to support automatically detecting the current plugin (when $p_base_name = null) and retrieving the plugin's "files" directory path (when $p_filename = null). This will allow plugins to dynamically provide files from their directory. Currently, their ‘files’ directory is hard-coded but not easy accessible.

Fixes: [35082](https://mantisbt.org/bugs/view.php?id=35082).

### DIFF
--- a/core/plugin_api.php
+++ b/core/plugin_api.php
@@ -185,21 +185,17 @@ function plugin_route_group( $p_base_name = null ) {
 }
 
 /**
- * Return a path to a plugin file.
- * @param string $p_filename  File name.
- * @param string $p_base_name Plugin base name.
- * @return mixed File path or false if FNF
+ * Return a path to a plugin file(s).
+ * @param string $p_filename  File name (if it is not set, the function will return a directory).
+ * @param string $p_base_name Plugin base name (defaults to the current plugin).
+ * @return mixed File path or false if file cannot be found
  */
-function plugin_file_path( $p_filename = null, $p_base_name = null ) {
-	$t_file_path = config_get_global( 'plugin_path' );
-	if( is_null( $p_base_name ) ) {
-		$t_file_path .= plugin_get_current() . DIRECTORY_SEPARATOR;
-	} else {
-		$t_file_path .= $p_base_name . DIRECTORY_SEPARATOR;
-	}
-	$t_file_path .= 'files' . DIRECTORY_SEPARATOR;
+function plugin_file_path( $p_filename = '', $p_base_name = '' ) {
+	$t_file_path = config_get_global( 'plugin_path' )
+		. ( $p_base_name ? $p_base_name : plugin_get_current() ) . DIRECTORY_SEPARATOR
+		. 'files' . DIRECTORY_SEPARATOR;
 
-	if( is_null( $p_filename ) ) {
+	if( !$p_filename ) {
 		return $t_file_path;
 	}
 	$t_file_path .= $p_filename;
--- a/core/plugin_api.php
+++ b/core/plugin_api.php
@@ -190,10 +190,19 @@ function plugin_route_group( $p_base_name = null ) {
  * @param string $p_base_name Plugin base name.
  * @return mixed File path or false if FNF
  */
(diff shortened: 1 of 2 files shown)


core/plugin_api.php
```diff
@@ -190,10 +190,19 @@ function plugin_route_group( $p_base_name = null ) {
...
+function plugin_file_path( $p_filename = null, $p_base_name = null ) {
 	$t_file_path = config_get_global( 'plugin_path' );
-	$t_file_path .= $p_base_name . DIRECTORY_SEPARATOR;
-	$t_file_path .= 'files' . DIRECTORY_SEPARATOR . $p_filename;
+	if( is_null( $p_base_name ) ) {
+		$t_file_path .= plugin_get_current() . DIRECTORY_SEPARATOR;
+	} else {
+		$t_file_path .= $p_base_name . DIRECTORY_SEPARATOR;
+	}
+	$t_file_path .= 'files' . DIRECTORY_SEPARATOR;
```
dregad, line 200: Use a short ternary `$p_base_name ?: plugin_get_current()` instead of an il...else block.

Also, I'd rather have a single assignment with several concats, on  multiple lines if needed to avoid long lines, than a series of $var .= xxx as in the original code - much easier to read IMHO.

```php
$t_file_path = config_get_global( 'plugin_path' )
    . ( $p_base_name ?: plugin_get_current() ) . DIRECTORY_SEPARATOR
    . 'files' . DIRECTORY_SEPARATOR;

**Review by dregad — CHANGES_REQUESTED** [review]
PHPDoc block needs to be updated.

**Comment by raspopov** [comment]
Thank you for taking the time to review the code. The recommendations have been implemented.

File: core/plugin_api.php [hunk 2]
```diff
@@ -185,21 +185,17 @@ function plugin_route_group( $p_base_name = null ) {
 }
 
 /**
- * Return a path to a plugin file.
- * @param string $p_filename  File name.
- * @param string $p_base_name Plugin base name.
- * @return mixed File path or false if FNF
+ * Return a path to a plugin file(s).
```
dregad, line 188: I find the `(s)` somewhat misleading, as the function only returns the path for a single file.

I suppose you added that to reflect the new ability to return the directory; if that's the case, then I believe it would be better to rephrase, like `Returns the path to a plugin file, or the plugin's files directory.`

core/plugin_api.php
```diff
@@ -185,21 +185,17 @@ function plugin_route_group( $p_base_name = null ) {
...
+ * @param string $p_filename  File name (if it is not set, the function will return a directory).
+ * @param string $p_base_name Plugin base name (defaults to the current plugin).
+ * @return mixed File path or false if file cannot be found
```
dregad, line 191: Change *mixed* to `string|false`

core/plugin_api.php
```diff
@@ -185,21 +185,17 @@ function plugin_route_group( $p_base_name = null ) {
 }
 
 /**
- * Return a path to a plugin file.
- * @param string $p_filename  File name.
- * @param string $p_base_name Plugin base name.
- * @return mixed File path or false if FNF
+ * Return a path to a plugin file(s).
+ * @param string $p_filename  File name (if it is not set, the function will return a directory).
```
dregad, line 189: *a directory* is vague. I suggest to rephrase: 
`File name; if not set, the function will return the plugin's files directory.`

dregad (changes requested):
A few nitpicks

raspopov:
Perfection. 😎

dregad:
> Is it possible to view a Travis build artifacts i.e. created docs

I don't think so, the output is not persisted as far as I know.

Publican is an old, unmaintained and somewhat painful toolchain to use, but you can use Docker to build the documentation locally (see https://github.com/mantisbt/mantisbt/blob/master/docbook/Dockerfile), this is what do on my dev box.

dregad (approved):
Thanks a lot ! Will merge later tonight.

PR #2055, opened 2024-12-14
Add SQLite 3 support

raspopov:
Fixes [#10573](https://mantisbt.org/bugs/view.php?id=10573).

Requires ADODb ~[PR 1053](https://github.com/ADOdb/ADOdb/pull/1053)~ and ~[PR 1055](https://github.com/ADOdb/ADOdb/pull/1055)~ to work.

Specify the following settings when creating a database:
- Type of Database: SQLite 3
- Hostname (for Database Server): full path to database (for example `c:\Projects\bugtracker.db`)

Screenshot:

firefox_2024-12-14_16-14-20

### DIFF
--- a/admin/check/check_database_inc.php
+++ b/admin/check/check_database_inc.php
@@ -207,6 +207,8 @@ if( db_is_mysql() ) {
 	$t_db_min_version = DB_MIN_VERSION_MSSQL;
 } elseif( db_is_oracle() ) {
 	$t_db_min_version = DB_MIN_VERSION_ORACLE;
+} elseif( db_is_sqlite3() ) {
+	$t_db_min_version = DB_MIN_VERSION_SQLITE3;
 } else {
 	$t_db_min_version = 0;
 }
@@ -491,4 +493,45 @@ if( db_is_mysql() ) {
 			}
 		}
 	}
+}
+
+if( db_is_sqlite3() ) {
+	check_print_info_row(
+		'Database file size',
+		check_format_number( @filesize( $t_database_hostname ) )
+	);
+
+	$t_result = db_query( 'PRAGMA compile_options' );
+	$t_icu = false;
+	while( $t_row = db_fetch_array( $t_result ) ) {
+		if( $t_row['compile_options'] === 'ENABLE_ICU' ) {
+			$t_icu = true;
+		}
+	}
+	check_print_test_warn_row(
+		'Database International Components for Unicode (ICU) extension is available',
+		$t_icu,
+		[ false => 'The operations are case insensitive for ASCII symbols only. '
+			. 'Use a different version of SQLite compiled with the SQLITE_ENABLE_ICU option.' ]
+	);
+
+	$t_collation = db_result( db_query( 'PRAGMA encoding' ) );
+	check_print_test_row(
+		'Database default collation is UTF-8',
+		( $t_collation === 'UTF-8' ),
+		[ false => 'Database is using ' . htmlentities( $t_collation )
+			. ' collation where UTF-8 collation is required.' ]
+	);
+
+	$t_result = db_query( 'PRAGMA integrity_check' );
+	$t_integrity_errors = '';
+	while( $t_row = db_fetch_array( $t_result ) ) {
+		$t_integrity_errors .= ( $t_integrity_errors ? '<br>' : '' )
+			. htmlentities( $t_row['integrity_check'] );
+	}
+	check_print_test_row(
+		'Database integrity',
+		( $t_integrity_errors === 'ok' ),
+		'Integrity check result: ' . $t_integrity_errors
+	);
 }
--- a/admin/install.php
+++ b/admin/install.php
@@ -556,6 +556,11 @@ if( 2 == $t_install_state ) {
 			$t_warning = "Unable to determine '$f_db_type' version.";
 		} else {
 			switch( $f_db_type ) {
+				case 'sqlite3':
+					if( version_compare( $t_version_info['version'], DB_MIN_VERSION_SQLITE3, '<' ) ) {
+						$t_error = 'SQLite (' . DB_MIN_VERSION_SQLITE3 . ') or later is required for installation';
+					}
+					break;
 				case 'mysqli':
 					if( version_compare( $t_version_info['version'], DB_MIN_VERSION_MYSQL, '<' ) ) {
 						$t_error = 'MySQL ' . DB_MIN_VERSION_MYSQL . ' or later is required for installation';
@@ -659,6 +664,7 @@ if( !$g_database_upgrade ) {
 				'mssqlnative' => 'Microsoft SQL Server Native Driver',
 				'pgsql'       => 'PostgreSQL',
 				'oci8'        => 'Oracle',
+				'sqlite3'     => 'SQLite3',
 			);
 
 			foreach( $t_db_list as $t_db => $t_db_descr ) {
@@ -966,6 +972,7 @@ if( 3 == $t_install_state ) {
 
 		# fake out database access routines used by config_get
 		config_set_global( 'db_type', $f_db_type );
+		$g_db_functional_type = db_get_type( $f_db_type );
 
 		# Initialize table prefixes as specified by user
 		config_set_global( 'db_table_prefix', $f_db_table_prefix );
@@ -974,6 +981,8 @@ if( 3 == $t_install_state ) {
 		# database_api references this
 		require_once( __DIR__ . '/schema.php' );
 		$g_db = ADONewConnection( $f_db_type );
+		$g_db->SetFetchMode( ADODB_FETCH_ASSOC );
+
 		$t_result = @$g_db->Connect( $f_hostname, $f_admin_username, $f_admin_password, $f_database_name );
 		if( !$f_log_queries ) {
 			$g_db_connected = true;
--- a/admin/install.php
+++ b/admin/install.php
@@ -40,6 +40,7 @@ const MANTIS_MAINTENANCE_MODE = true;
 require_once( dirname( __DIR__ ) . '/core.php' );
 require_api( 'install_helper_functions_api.php' );
 require_api( 'crypto_api.php' );
+require_api( 'database_api.php' );
 $g_error_send_page_header = false; # bypass page headers in error handler
 
 $g_failed = false;
@@ -275,6 +276,22 @@ if( CONFIGURED_PASSWORD == $f_admin_password ) {
 $f_log_queries    = gpc_get_bool( 'log_queries', false );
 $f_db_exists      = gpc_get_bool( 'db_exists', false );
 
+$f_db_setup       = config_get_global( 'db_setup', '' );
+if( empty( $f_db_setup ) ) {
+	if( $f_db_type == 'mysqli' ) {
+		$f_db_setup =
+			  'SET NAMES UTF8';				# charset for the connection
+	}
+	elseif( $f_db_type == 'sqlite3' ) {
+		$f_db_setup =
+			  'PRAGMA journal_mode=WAL;'	# concurrent readers + writers
+			. 'PRAGMA synchronous=NORMAL;'	# performance, safe with WAL
+			. 'PRAGMA busy_timeout=10000;'	# 10s retry on locked DB
+			. 'PRAGMA foreign_keys=ON;'		# enforce referential integrity
+			. 'PRAGMA cache_size=-8000';	# 8 MB page cache
+	}
+}
+
 if( $t_config_exists ) {
 	if( 0 == $t_install_state ) {
 		print_test( 'Config File Exists - Upgrade', true );
@@ -289,6 +306,10 @@ if( $t_config_exists ) {
 		if( $f_db_type == 'oci8' ) {
 			$t_db_conn_exists = $t_db_conn_exists || ( $f_database_name == '' && $f_db_username !== '' && $f_hostname !== '' );
 		}
+		# SQLite only needs hostname (= path to db file)
+		elseif( $f_db_type == 'sqlite3' ) {
+			$t_db_conn_exists = $f_hostname !== '';
+		}
 		print_test( 'Checking Database connection settings exist',
 			$t_db_conn_exists,
 			true,
@@ -297,11 +318,16 @@ if( $t_config_exists ) {
 		print_db_support_tests( $f_db_type );
 	}
 
-	/** @var ADOConnection $g_db */
-	$g_db = ADONewConnection( $f_db_type );
-	$t_result = @$g_db->Connect( $f_hostname, $f_db_username, $f_db_password, $f_database_name );
-	if( $g_db->IsConnected() ) {
-		$g_db_connected = true;
+	try {
+		/** @var ADOConnection $g_db */
+		$g_db = ADONewConnection( $f_db_type );
+		$t_result = @$g_db->Connect( $f_hostname, $f_db_username, $f_db_password, $f_database_name );
+		if( $g_db->IsConnected() ) {
+			$g_db_connected = true;
+		}
+	}
+	catch( Error $e ) {
+		print_test( 'Retrieving Database Version', false, true, string_attribute( $e->getMessage() ) );
 	}
 
 	$t_cur_version = config_get( 'database_version', -1, ALL_USERS, ALL_PROJECTS);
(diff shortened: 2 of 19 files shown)


admin/install.php
```diff
@@ -1121,6 +1124,30 @@ function_exists( 'mysqli_stmt_get_result' ),
...
+						$t_pos = strrpos( $t_request, 'PRIMARY' );
+						$t_request = substr( $t_request, 0, $t_pos ) . substr( $t_request, $t_pos + 7 );
+					}
+					$t_request = str_replace( 'PRIMARY', 'PRIMARY KEY', $t_request );
+
+					# Remove NOT NULL of BLOB, use db_update_blob() instead
+					$t_request = preg_replace( '/B\s+NOTNULL/', 'B', $t_request );
+
+					# Replace VARCHAR(x) with TEXT
+					$t_request = preg_replace( '/C\([0-9]+\)/', 'X', $t_request );
```
dregad, line 1146: Is this really needed ? I believe SQLite will just swallow whatever type it's given and use TEXT internally as appropriate.

core/database_api.php
```diff
@@ -852,10 +867,6 @@ function db_get_table_list() {
 function db_update_blob( $p_table, $p_column, $p_val, $p_where = null ) {
 	global $g_db, $g_db_log_queries, $g_queries_array;
 
-	if( !db_is_oracle() ) {
-		return false;
-	}
-
```
dregad, line 858: Why remove this safety check ? The function should return false for DBs not using BLOBs.

Check PHPDoc, update accordingly.

core/file_api.php
```diff
@@ -1081,7 +1081,7 @@ function file_add( $p_bug_id, array $p_file, $p_table = 'bug', $p_title = '', $p
 
 	$t_file_info['id'] = db_insert_id( $t_file_table );
 
-	if( db_is_oracle() ) {
+	if( db_is_oracle() || db_is_sqlite3() ) {
```
dregad, line 1084: Note that *db_update_blob()* is also called from mc_file_api.php.

Maybe we should introduce a new, more functionally-oriented check, like db_uses_blob() returning true for oci8 and sqlite, instead of checking both db_is_oracle() / db_is_sqlite3() every time.

dregad (changes requested):
Some comments (code review only - did not actually test).

core/database_api.php
raspopov: The db_update_blob() function is called everywhere only after if( db_is_oracle() ), there is no point in doing double check inside the function, especially since it is now also needed in the case of db_is_sqlite().

I've changed the description accordingly.

admin/install.php
raspopov: The "B NOTNULL" i.e. "BLOB NOT NULL" has to be changed to just "BLOB" because SQLite handles blob separately in the two SQL calls, and in the first call it's actually NULL.

The "C(x)" i.e. VARCHAR(x) needs to be changed to "TEXT" to mitigate the missed AlterColumnSQL which is commonly used to expand the column size.

admin/install.php
```diff
@@ -1121,6 +1124,30 @@ function_exists( 'mysqli_stmt_get_result' ),
...
+					# Rely on auto-detection
+					$t_request = str_replace( [ 'UNSIGNED', 'AUTOINCREMENT' ], '', $t_request );
+
+					# Remove the second PRIMARY
+					$t_request = str_replace( 'PRIMARY KEY', 'PRIMARY', $t_request );
+					while( substr_count( $t_request, 'PRIMARY' ) > 1 ) {
+						$t_pos = strrpos( $t_request, 'PRIMARY' );
+						$t_request = substr( $t_request, 0, $t_pos ) . substr( $t_request, $t_pos + 7 );
+					}
+					$t_request = str_replace( 'PRIMARY', 'PRIMARY KEY', $t_request );
```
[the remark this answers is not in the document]
raspopov: > Why the initial PRIMARY KEY -> PRIMARY replacement ?

Ah, the 'PRIMARY KEY' was left over from another version of the code, where it was called after converting to classic SQL. Removed.

> Isn't this going to create incorrect primary keys for intersection tables (e.g. bug_tag) and maybe more problematic, config table ?

Yes, you are correct, this should only be done for AUTOINCREMENT. SQLite will not apply the AUTOINCREMENT attribute to a primary key column if the key includes more than one column.  I will fix the code after additional testing.

admin/install.php
```diff
@@ -1144,6 +1171,15 @@ function_exists( 'mysqli_stmt_get_result' ),
...
 
+					case 'AlterColumnSQL':
+						# SQLite: AlterColumnSQL can be safely ignored till it changing only a type sizes
+						if( $f_db_type == 'sqlite3' ) {
+							$t_target = null;
+							$t_sqlarray = array();
+							break;
+						}
+						# no break;
+
```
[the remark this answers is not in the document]
raspopov: The first AlterColumnSQL() call shows error and stops execution of schema upgrade... I added by myself to ADODB for debugging, looks like I fixed my own debug code. 😑Removed.

core/file_api.php
raspopov: > Note that db_update_blob() is also called from mc_file_api.php.

Added.

admin/install.php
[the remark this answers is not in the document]
raspopov: UNSIGNED is meaningless to SQLite, its just INTEGER inside. NOTNULL is just a check. But DEFAULT requires fix for 'api_token' and 'category' tables.

core/file_api.php
raspopov: I liked the clever approach with db_uses_blob(), added it.

admin/install.php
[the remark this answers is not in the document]
raspopov: Done.

admin/install.php
raspopov: Actually VARCHAR doesn't need to be touched, SQLite doesn't take advantage of the restrictions as I thought. And BLOB NOT NULL I redesigned as BLOB NOT NULL DEFAULT ‘’ and put it directly into schema.php.

admin/schema.php
```diff
@@ -471,7 +473,7 @@
 $g_upgrade[61] = array( 'CreateTableSQL', array( db_get_table( 'tag' ), "
 	id						I		UNSIGNED NOTNULL PRIMARY AUTOINCREMENT,
 	user_id					I		UNSIGNED NOTNULL DEFAULT '0',
-	name					C(100)	NOTNULL PRIMARY DEFAULT \" '' \",
+	name					C(100)	NOTNULL UNIQUE DEFAULT \" '' \",
```
dregad, line 476: You can't just do that.

This will introduce a schema discrepancy between existing installations, which will have the composite (id, name) primary key, while new installations will have 2 indexes primary (id) and unique(name).  Moreover, the the latter will be a duplicate of _idx_tag_name_ created at schema 180 which should probably be dropped.

Additional schema steps must be added to ensure the schema is identical in both scenarios.

Since this composite primary key issue can be considered as a bug independent from the SQLite3 support, a separate Issue should be created to track it and give it visibility in the change log. It can still be fixed as part of this PR if you want, with proper reference in relevant commit(s).

admin/schema.php
```diff
@@ -849,8 +851,8 @@
 	user_id					I		NOTNULL DEFAULT '0',
 	name					C(128)	NOTNULL,
 	hash					C(128)	NOTNULL,
-	date_created			I		UNSIGNED NOTNULL DEFAULT '0',
-	date_used				I		UNSIGNED NOTNULL DEFAULT '0'",
+	date_created			I		UNSIGNED NOTNULL DEFAULT '1',
+	date_used				I		UNSIGNED NOTNULL DEFAULT '1'",
```
dregad, line 855: OK this does not really hurt as we alter the columns to set the default to 1 later on, but adding a comment at step 209 explaining the history and reason for the change will help our future selves understand why we're issuing an AlterColumnSQL which does nothing without having to delve through git history.

admin/schema.php
```diff
@@ -513,7 +515,7 @@
 	project_id				I		UNSIGNED NOTNULL DEFAULT '0',
 	user_id					I		UNSIGNED NOTNULL DEFAULT '0',
 	name					C(128)	NOTNULL DEFAULT \" '' \",
-	status					I		UNSIGNED NOTNULL DEFAULT '0' ",
+	status					I		UNSIGNED NOTNULL DEFAULT '1' ",
```
dregad, line 518: Same comment as for *api_token* table below, bug for schema step 212.

core/classes/DbQuery.class.php
```diff
@@ -952,6 +952,7 @@ public function sql_like( $p_alias, $p_pattern, $p_escape = null, $p_force_ci =
 					$t_operator = 'ILIKE';
 					break;
 				case DB_TYPE_ORACLE:
+				case DB_TYPE_SQLITE3:
```
dregad, line 955: Update above comment

dregad (changes requested):

core/classes/DbQuery.class.php
raspopov: To be precise for SQLite we need to use external functions for non-ASCII characters... 🤔
I'd rather not make this change, but add a comment and TODO to use [mb_strtoupper()](https://www.php.net/manual/en/function.mb-strtoupper.php).

admin/install.php
```diff
@@ -1202,6 +1206,19 @@ function_exists( 'mysqli_stmt_get_result' ),
...
+							foreach( $t_sqlarray as & $t_request ) {
+								if( preg_match( '/\(\s+([a-z_]+).*AUTOINCREMENT/', $t_request, $t_matches ) ) {
+									$t_request = preg_replace( '/,\s+PRIMARY KEY \(' . $t_matches[ 1 ] . '\)/',
+										'', $t_request );
+									$t_request = preg_replace( '/INTEGER (UNSIGNED |)NOT NULL AUTOINCREMENT/',
+										'INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT', $t_request );
+								}
+							}
+						}
+
```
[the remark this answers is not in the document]
raspopov: Did I understand correctly you are suggesting to fix ADODB2_sqlite::_CreateSuffix() function?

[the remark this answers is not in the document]
dregad: I did not actually execute your PR's code, but if my understanding is correct the SQL generated by ADOdb is not valid (i.e. ADODB2_sqlite::executeSQLArray() throws errors when executing it). If that is indeed the case, then IMO it's an upstream bug that should be fixed. 

The problem could be in ADODB2_sqlite::_CreateSuffix() indeed, but also possibly in in ADODB_DataDict::_genFields() which may need some refactoring to handle the sqlite3 specifics for AUTOINCREMENT, PRIMARY KEY, UNSIGNED, etc

core/classes/DbQuery.class.php
dregad: > use [mb_strtoupper()](https://www.php.net/manual/en/function.mb-strtoupper.php).

Makes sense

admin/schema.php
raspopov: Comment added.

admin/schema.php
raspopov: Comment added.

core/classes/DbQuery.class.php
raspopov: I checked and found that the case-insensitive variant of sql_like() is only used in sql_ilike(), which is only used to look up e-mails. So basic SQLite functionality is enoguh.

admin/schema.php
raspopov: I removed UNIQUE.

Checking the use of the ‘tag’ table showed:
- Creating a new tag with manual uniqueness check of the name by tag_is_unique().
- Updating tag data by id with manual uniqueness check by tag_is_unique().
- Deleting a tag by id.
- Selecting a specific tag by id list.
- Selecting a specific tag by name (two functions).
- Selecting all tags.
- Counting all tags.
- Selecting all tags that are not yet assigned to a bug by id.

I don't see any complex queries on the table, and even uniqueness is not checked by database tools. I think that only two indexes on `id` and `name` columns are enough for efficient work, and they are already there.

I tested the scheme also on MySQL 9.0.1 and PostgreSQL 16.1 - everything works perfectly. 😊

And I have a two news for ADODb - bad news and good news. The good news is that everything works, even with UNIQUE. The bad news is that it should NOT work with UNIQUE. Unfortunately the UNIQUE attribute seems to be removed and ignored by ADODb from the SQL query, although it should result in either setting the appropriate table column attribute or creating a column index with the column name, depending on the database type. It's just ignored in all three tested bases. ☹️

vboctor (changes requested):
Thanks @raspopov for your contribution. +1 for sqlite support.

1. Please include validation of sqlite in the build, so it is continuously validated similar to MySQL, etc.
2. Do we need to do the change of defaults in the schema? I'm not sure why this is necessary and I'm not sure it should be part of a PR that supports a new db driver.

raspopov:
@vboctor

> Do we need to do the change of defaults in the schema?

That's the thing, there are actually no schema changes here, only changes to the schema update history. SQLite does not have the ability to change schema on the fly as easily as other databases. So for example default values have to be set once and in advance. We are fortunate that the changes are so minor. A full-fledged change in SQLite would require deleting and recreating the entire table.

vboctor:
The default settings for Sqlite is not optimized for a multi-user setup. We should add the following settings in `db_connect()` method similar to the UTF8 setting for mysql.

```php
if( db_is_sqlite3() ) {
    db_query( 'PRAGMA journal_mode=WAL' );    # concurrent readers + writers
    db_query( 'PRAGMA busy_timeout=10000' );  # 10s retry on locked DB
    db_query( 'PRAGMA foreign_keys=ON' );     # enforce referential integrity
    db_query( 'PRAGMA cache_size=-8000' );    # 8 MB page cache
}
```

I suggest adding global configs (to config_defaults_inc.php and docs) that are retrieved via `config_get_global()` and the defaults above. sqlite_journal_mode = 'WAL' or '' (empt for default sqlite - i.e. skip query - support any sqlite journal mode), sqlite_cache_size (in MB - multiply by `-1024`), and sqlite_timeout (in ms). Foregin keys honoring should always be ON.

admin/check/check_database_inc.php
```diff
@@ -492,3 +494,44 @@ function_exists( 'mysqli_stmt_get_result' ),
...
+
+	$t_result = db_query( 'PRAGMA compile_options' );
+	$t_icu = false;
+	while( $t_row = db_fetch_array( $t_result ) ) {
+		if( $t_row['compile_options'] === 'ENABLE_ICU' ) {
+			$t_icu = true;
+		}
+	}
+	check_print_test_warn_row(
+		'Database International Components for Unicode (ICU) extension is available',
```
vboctor: In `DbQuery` has special handling for Oracle and Pgsql for case sensitivity. It may make sense to all add for Sqlite in case this is not available.

vboctor (changes requested):

config_defaults_inc.php
```diff
@@ -46,6 +46,11 @@
  *
  * If you need to supply a port to connect to, set hostname as 'localhost:3306'.
  *
+ * For SQLite, you must specify the full path to the database file,
+ * using $g_database_name as the name, i.e. '/var/database/bugtracker.db' for
```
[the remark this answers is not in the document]
raspopov: > A simple js can help that reacts the db type drop down can help.

Thanks for the interesting idea; I will try to implement it.

>  We should also default the path to $g_path + data/mantisbt.db

That's a bit too risky; it's better not to store the database where a web server can access it as part of the website.

admin/check/check_database_inc.php
```diff
@@ -492,3 +494,44 @@ function_exists( 'mysqli_stmt_get_result' ),
...
+		$t_icu,
+		[ false => 'The operations are case insensitive for ASCII symbols only. '
+			. 'Use a different version of SQLite compiled with the SQLITE_ENABLE_ICU option.' ]
+	);
+
+	$t_collation = db_result( db_query( 'PRAGMA encoding' ) );
```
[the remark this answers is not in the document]
vboctor: What about setting the encoding during db creation to UTF-8? Since once the db is created, it can't be changed.

vboctor:
@raspopov @dregad 

Just checking in on whether this can be merged. From my perspective, here are the blockers:

1. Dependency on ADODB dev-master - not sure if a proper release includes what is needed by this PR.
2. Setting encoding to UTF-8 during db creation.

PR #2056, opened 2024-12-18
fix #35179 Could not use plugins filters with "Permalink"

elpoutro:
fixes: [#35179](https://mantisbt.org/bugs/view.php?id=35179)
Before modification, filters created through plugins broke "Permalink" functionality with an error.
The functionality as never work properly and was fixed on my side since I pushed it in 2018. 
Sorry for the correction delay.

### DIFF
--- a/search.php
+++ b/search.php
@@ -149,19 +149,19 @@ $t_plugin_filters = filter_get_plugin_filters();
 foreach( $t_plugin_filters as $t_field_name => $t_filter_object ) {
     switch( $t_filter_object->type ) {
         case FILTER_TYPE_STRING:
-            $t_my_filter[$t_field_name] = gpc_get_string( $t_field_name, $t_meta_filter_any_array[$t_field_name] );
+            $t_my_filter[$t_field_name] = gpc_get_string( $t_field_name, $t_meta_filter_any_array );
             break;
         case FILTER_TYPE_INT:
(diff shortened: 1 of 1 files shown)


dregad (changes requested):
@elpoutro thanks for your contribution.

I don't think your fix is correct: $t_meta_filter_any_array is meant as a default value for _gpc_get_XXX_array()_ functions only, it does not make sense for single-value _gpc_get_XXX()_ calls.

In fact this code will probably trigger an ERROR_GPC_ARRAY_UNEXPECTED (202) if the plugin's field is not provided in the permalink.

You need to set an appropriate default value corresponding to the gpc variable's type; `''`, `0` , `false`, or maybe `META_FILTER_ANY` for STRING, INT and BOOLEAN.

dregad:
I confirm that ERROR_GPC_ARRAY_UNEXPECTED does get triggered:
http://example.com/mantis/search.php?project_id=1
image

FYI, while I was testing this I noticed a couple of minor bugs in the [FilterBugList](https://github.com/mantisbt-plugins/FilterBugList) plugin, which I fixed right away. Version 2.1.1 has been released.

dregad:
@elpoutro please address the code review comments

dregad:
Replaced by #2108

PR #2057, opened 2024-12-19
Fix a <style> tag that was incorrectly placed

raspopov:
Fix a `<style>` tag that was incorrectly placed inside a `<body>` tag instead of a `<head>` tag.

Fix for: [#35180](https://mantisbt.org/bugs/view.php?id=35180)

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -289,7 +289,15 @@ function layout_head_css() {
 
 	echo "\n";
 
-	# Print user font preference
+	# Set font preference
+	layout_user_font_preference();
+}
+
+/**
+ * Print user font preference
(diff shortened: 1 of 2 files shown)


dregad (changes requested):
It makes sense to move the user font to layout_head_css(), but I'd rather keep it as a separate function.

dregad (approved):
Thanks. I'll merge this in 2.27 branch, and keep [#35180](https://mantisbt.org/bugs/view.php?id=35180) open as discussed.

dregad:
Merged in 39e340e4c2806428271e018c0e39816a5f1ed158

PR #2058, opened 2024-12-20
Add the 'lang' attribute to the '<html>' tag.

raspopov:
Add new function `lang_get_current_lang()`.

Fix for: [#35180](https://mantisbt.org/bugs/view.php?id=35180).

### DIFF
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -1065,7 +1065,10 @@ $g_language_choices_arr = array(
 
 /**
  * Browser language mapping for 'auto' language selection.
+ * An ISO 639-1 language codes, the most general one must be set the last.
  *
+ * @see lang_get_current_lang()
+ * @see lang_get_current_datetime_locale()
  * @global array $g_language_auto_map
  */
 $g_language_auto_map = array(
--- a/config_defaults_inc.php
+++ b/config_defaults_inc.php
@@ -1065,7 +1065,10 @@ $g_language_choices_arr = array(
 
 /**
  * Browser language mapping for 'auto' language selection.
- * An ISO 639-1 language codes, the most general one must be set the last.
+ *
+ * The mapping between a comma-separated list of BCP 47 language codes, usually
+ * ending with the most generic one (i.e. the primary language without extended
+ * subtag), and the corresponding MantisBT language code.
  *
  * @see lang_get_current_lang()
  * @see lang_get_current_datetime_locale()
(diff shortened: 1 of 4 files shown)


core.php
```diff
@@ -189,6 +189,12 @@
 # Register global shutdown function
 shutdown_functions_register();
 
+# Push default language to speed calls to lang_get
+if( !defined( 'LANG_LOAD_DISABLED' ) ) {
+	require_api( 'lang_api.php' );
+	lang_push( lang_get_default() );
+}
+
```
dregad, line 197: Did you check that this is not introducing any unwanted side effects in the core's loading sequence ? I'm guessing it does not, since plugin_api.php requires lang_api.php, but just in case.

config_defaults_inc.php
```diff
@@ -1065,7 +1065,10 @@
 
 /**
  * Browser language mapping for 'auto' language selection.
+ * An ISO 639-1 language codes, the most general one must be set the last.
```
dregad, line 1068: What do you mean by *the most general one must be set the last* ?

dregad (changes requested):
Issue [#35180](https://mantisbt.org/bugs/view.php?id=35180) is about MantisBT producing valid HTML.

Changing the core's loading sequence is a different work item, that should be tracked separately. Please open a new issue, and consider splitting the commit.

config_defaults_inc.php
raspopov: Language code without country code, i.e. "en" is on the right in "en-us, en-gb, en-au, en".

core.php
raspopov: Moved to: #2060.

raspopov: I have checked this, including in `admin\test_langs.php`.

config_defaults_inc.php
dregad: OK, so if I understand correctly, the comment's purpose is to explain the array key's structure, is this correct ?

> ISO 639-1 language codes

639-1 is 2-char codes - I think this is incorrect. If I'm not mistaken, these are actually BCP 47 language codes, which also allow three-letter codes from ISO 639-3 and 639-5 (see https://en.wikipedia.org/wiki/IETF_language_tag).

> the most general one must be set the last

This is not always true, see *chinese_traditional*, *portuguese_brazil* for example.

If the above is correct, may I suggest the following alternative description

`This configuration defines the mapping between a comma-delimited list of related BCP 47 language codes, usually ending with the most generic one (i.e. the primary language without extended subtag), and the corresponding MantisBT language code.`

raspopov: You're quite right, clarified.

dregad (approved):

dregad:
Thanks. Merged in 4e49e9312ac02a00617348075dac2c14da8d4f7c.

PR #2059, opened 2024-12-21
Optimize lang_load() function by replacing preg_replace()

raspopov:
with str_starts_with(), and substr().

For 'auto' language and non-English browsers it was about 2500 preg_replace() calls per page load, making this function in the top 10 of CPU eaters according to xDebug2 profiler. No more.

Fixes: [#35198](https://mantisbt.org/bugs/view.php?id=35198).

### DIFF
--- a/core/lang_api.php
+++ b/core/lang_api.php
@@ -86,7 +86,7 @@ function lang_load( $p_lang, $p_dir = null ) {
 	$t_vars = get_defined_vars();
 
 	foreach( array_keys( $t_vars ) as $t_var ) {
-		if( str_starts_with( $t_var, 's_' ) ) {
+		if( strncmp( $t_var, 's_', 2 ) === 0 ) {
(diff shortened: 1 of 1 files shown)


raspopov:
Ah, str_starts_with() requires PHP 8...

dregad (changes requested):
> Ah, str_starts_with() requires PHP 8...

FYI we'll increase the minimum PHP version from 7.4 to 8.1 when Ubuntu 20.04 LTS reaches end of standard support (April 2025).

If we're going after performance, then we strpos() is slightly faster than strncmp() - although probably not by a factor likely to make a noticeable difference with only a few thousand iterations.

Please open an issue in the tracker.

raspopov:
Opened.

dregad (approved):

dregad:
Merged in 9e2b936c7d5bfbad273c757caaa965b6ea183620, thanks for your contribution

PR #2063, opened 2024-12-24
Fix plugin enumeration on the language test page.

raspopov:
Fixes: [35200](https://mantisbt.org/bugs/view.php?id=35200).

### DIFF
--- a/admin/test_langs.php
+++ b/admin/test_langs.php
@@ -25,7 +25,7 @@
 define( 'PLUGINS_DISABLED', true );
 define( 'LANG_LOAD_DISABLED', true );
 
-$t_mantis_dir = dirname( __DIR__ ) . '/';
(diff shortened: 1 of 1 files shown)


admin/test_langs.php
```diff
@@ -25,7 +25,7 @@
 define( 'PLUGINS_DISABLED', true );
 define( 'LANG_LOAD_DISABLED', true );
 
-$t_mantis_dir = dirname( __DIR__ ) . '/';
+$t_mantis_dir = dirname( __DIR__ ) . DIRECTORY_SEPARATOR;
```
dregad, line 28: Just for the record, using DIRECTORY_SEPARATOR is not needed in this context, PHP will take care of converting `/` as appropriate.

dregad:
Fixed in 1f556e79502e833556a46a17e97de3877e2fcf96.

PR #2068, opened 2025-01-06
Fix Parsedown 1.8 compatibility

raspopov:
Quick fix to run [Parsedown 1.8.0 i.e. master branch](https://github.com/erusev/parsedown).

Fixes [35227](https://mantisbt.org/bugs/view.php?id=35227).

### DIFF
--- a/plugins/MantisCoreFormatting/core/MantisMarkdown.php
+++ b/plugins/MantisCoreFormatting/core/MantisMarkdown.php
@@ -88,6 +88,9 @@ class MantisMarkdown extends Parsedown
 		$this->setSafeMode( true );
 		# Only turn URLs into links if config says so
 		$this->setUrlsLinked( (bool) $this->config_process_urls );
+
+		$this->InlineTypes['@'] []= 'EmailText';
+		$this->inlineMarkerList .= '@';
 	}
 
 	public static function getInstance( ?int $p_process_urls = OFF, ?int $p_process_buglinks = OFF ): self {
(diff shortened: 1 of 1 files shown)


plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -88,6 +88,9 @@ public function __construct( ?int $p_process_urls = OFF, ?int $p_process_buglink
 		$this->setSafeMode( true );
 		# Only turn URLs into links if config says so
 		$this->setUrlsLinked( (bool) $this->config_process_urls );
+
+		$this->InlineTypes['@'] []= 'EmailText';
+		$this->inlineMarkerList .= '@';
```
dregad, line 93: I was thrown off by the unusual space before `[]=`, was that intentional ? 
I would expect to see this written as `$this->InlineTypes['@'][] = ` instead.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
```diff
@@ -164,17 +167,32 @@ protected function element( array $Element ): string {
...
+	 *
+	 * @return array|void The email element data or nothing
 	 */
-	protected function unmarkedText( $text ): string {
-		if( ON == $this->config_process_urls && false !== strpos( $text, '@' ) ) {
-			$text = string_insert_hrefs( $text );
+	protected function inlineEmailText( $p_excerpt ) {
+		$t_host = '[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?';
+		$t_email = '/([a-zA-Z0-9.!#$%&\'*+\/=?^_`{|}~-]++@'
+			. $t_host . '(?:\.' . $t_host . ')*)/i';
```
dregad, line 181: This (almost) matches our standard pattern in email_regex_simple()

https://github.com/mantisbt/mantisbt/blob/64a6c746c971d6f7ecc3063f5bafe9cb357cca93/core/email_api.php#L114

- is the possessive `++` quantifier (before `@`) intended ? If so, why do we need it ?
- I believe the `/i` is useless, since the patterns always have upper- and lowercase letters

I'm wondering if we can we somehow reuse the existing pattern, to avoid duplicating ? Looks like we would only need to add a capturing group, and in fact I'm not even sure it is actually needed since it's covering the whole pattern so we could probably use $t_matches[0]. What do you think ?

string_insert_hrefs(), which was previously used, has additional logic to match an optional `mailto:` prefix, and also to exclude what may look like an e-mail address but is actually part of an URL, does the new code take this into account ?

https://github.com/mantisbt/mantisbt/blob/64a6c746c971d6f7ecc3063f5bafe9cb357cca93/core/string_api.php#L532-L541

dregad (commented):
Defining a new inline type to process e-mails is smart, I like it 👍

plugins/MantisCoreFormatting/core/MantisMarkdown.php
raspopov: > I was thrown off by the unusual space before []=, was that intentional ?

Yes, I took this code from the Parsedown extension example and I liked that the framing with spaces `[]=` emphasises this as a "unary operation of adding a new element to the array", so I didn't remove the space. 😎 But I can change it if you want.

plugins/MantisCoreFormatting/core/MantisMarkdown.php
raspopov: > This (almost) matches our standard pattern in email_regex_simple()

PCRE is a voodoo magic. I just copied this code from the standard Parsedown handle for email to be a maximum consistency with it: https://github.com/erusev/parsedown/blob/c9dc49f68f7bd14e5aa107d86a08ce59203f02ee/Parsedown.php#L1277

> string_insert_hrefs(), which was previously used, has additional logic to match an optional mailto: prefix

Unfortunately `mailto:` cannot be used, because Parsedown divides the string also by the inline character ":" earlier.
But I can't imagine who would write `mailto:` before an email, so I think it's an acceptable loss of functionality.

raspopov: > is the possessive ++ quantifier (before @) intended ? If so, why do we need it ?

I think Parsedown has a thing for performance, that's why the fastest regular expression variants are used everywhere.

raspopov: > I believe the /i is useless, since the patterns always have upper- and lowercase letters

There are two options here, either remove the ‘i’ or remove all the ‘A-Z’ and the second option is much shorter.

dregad (approved):
Looking good, thank you !

dregad:
Merged in 36f0cca6d08aeed2c78dbc4317e1ddbc4e3c9f8e

PR #2069, opened 2025-01-09
PHPUnit 11.5 compatibility fix

raspopov:
Provider functions should be static.

Some GetLinkAttributesTest.php providers has mistyped parameter names (9cb002ec8ec401f8f7fca455f16834eb2f5d4036).

Fixes [#35223](https://mantisbt.org/bugs/view.php?id=35223).

### DIFF
--- a/plugins/MantisCoreFormatting/tests/MarkdownTest.php
+++ b/plugins/MantisCoreFormatting/tests/MarkdownTest.php
@@ -267,7 +267,7 @@ EOD;
 		$this->assertSame( '<p>' . $t_sample . '</p>', $this->parser->convert( $t_sample, true ) );
 	}
 
-	public function provideHeaders(): Generator {
+	static public function provideHeaders(): Generator {
 		# Valid headers
 		yield  'valid: # foo' => ['# foo', '<h1>foo</h1>'];
 		yield  'valid: ## foo' => ['## foo', '<h2>foo</h2>'];
@@ -303,7 +303,7 @@ EOD;
 	 * tested by the Parsedown tests. But they ensure that the Parsedown process is
 	 * not affected in any way, no matter what the value of "process_urls" is.
 	 */
-	public function provideEmails(): Generator {
+	static public function provideEmails(): Generator {
 		yield 'process_urls = ON; lorem <user@exmaple.com> ipsum' => [
 			'lorem <user@exmaple.com> ipsum',
 			ON,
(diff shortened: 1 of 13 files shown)


dregad:
This can't be merged until MantisBT minimum PHP version is increased to 8.2 (7.4 at the moment).

raspopov:
I have tested PHPUnit versions 11 (>= PHP 8.2), 10 (>= PHP 8.1) and 9 (>= PHP 7.3) and can confirm that they all work identically. If you could possibly provide a description of the specific problem it is experiencing, I would be very grateful to learn what I might be missing. 🤔

dregad:
Sorry, you're absolutely right. I checked and confirm that adding the static keyword to data providers has no impact on running the unit tests (i.e. it is backwards-compatible) with PHPUnit 9.x, so there is actually no reason to delay merging this.

Please disregard my earlier note https://github.com/mantisbt/mantisbt/pull/2069#issuecomment-2580893392.

That being said, I would prefer to have `public static` instead of `static public`, in accordance with [PHP-FIG PER Coding Style 2.0](https://www.php-fig.org/per/coding-style/#46-modifier-keywords).

About the other commit, I did not realize until now that PHPUnit would use the array key as a parameter name. This does not affect us now, but will do when we upgrade to 10.x or later. Thanks for the fix.

raspopov:
Ah, I see what's wrong, I forgot to mention that I disabled the SOAP and REST tests because they don't affect the parts of the tests I need. Just to be on the safe side, I ran a full (and very slow even on SQLite) PHPUnit 9 test and it worked without error. 😎

> in accordance with PHP-FIG PER Coding Style 2.0

Thanks for pointing out a useful and very specific PHP style guide, although it should be noted that the MantisBT style (as described in the wiki) contradicts this guide in some details. 🧐I've re-formatted the code, but I also noticed that MantisBT still has two `static public` unrelated to this request.

dregad:
> > in accordance with PHP-FIG PER Coding Style 2.0
> 
> Thanks for pointing out a useful and very specific PHP style guide, although it should be noted that the MantisBT style (as described in the wiki) contradicts this guide in some details. 🧐

Just to make sure I did not send the wrong message across - the [_official_ coding guidelines](https://mantisbt.org/wiki/doku.php/mantisbt:coding_guidelines) still are those outlined in the Wiki, but they are somewhat dated and not covering recent features of PHP, so if something is not defined (like here) then it makes sense to apply PER CS 2.0.

I am generally in favor of switching to move standard guidelines (PSR-2 / PSR-12 and now PER CS), but Victor has a different opinion. We've had this discussion several times in the past (latest coming to mind https://mantisbt.org/bugs/view.php?id=34400#c68854), but so far we remain on status quo. 

> MantisBT still has two `static public` unrelated to this request.

Thanks for the heads up, I'll fix those.

dregad:
Thanks @raspopov

PR #2071, opened 2025-01-13
Enhance REST API authentication to fall back to default checks

raspopov:
If the initial authentication token check fails.

This will help in situations where the request contains an "Authorisation:" header that does not contain a token because another authentication method is being used, such as cookies or anonymous.

Fixes [#35233](https://mantisbt.org/bugs/view.php?id=35233).

### DIFF
--- a/api/rest/restcore/AuthMiddleware.php
+++ b/api/rest/restcore/AuthMiddleware.php
@@ -64,7 +64,7 @@ class AuthMiddleware {
 				$t_password = '';
 				$t_login_method = LOGIN_METHOD_ANONYMOUS;
 				if( !auth_anonymous_enabled() || empty( $t_username ) ) {
-					return $response->withStatus( HTTP_STATUS_UNAUTHORIZED, 'API token required' );
+					return $response->withStatus( HTTP_STATUS_UNAUTHORIZED, 'Valid API token required' );
 				}
 			}
 		}
--- a/api/rest/restcore/AuthMiddleware.php
+++ b/api/rest/restcore/AuthMiddleware.php
@@ -30,10 +30,30 @@ require_api( 'user_api.php' );
  */
 class AuthMiddleware {
 	public function __invoke( \Slim\Http\Request $request, \Slim\Http\Response $response, callable $next ) {
-		$t_authorization_header = $request->getHeaderLine( HEADER_AUTHORIZATION );
 
-		if( empty( $t_authorization_header ) ) {
-			# Since authorization header is empty, check if user is authenticated by checking the cookie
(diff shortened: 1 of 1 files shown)


vboctor (changes requested):
Thanks @raspopov for the contribution.

I think the updated version with work with the exception that it will say "API Token Required" when an invalid token is provided.

I suggest the inspecting authorization header types - e.g. `bearer`, `basic`, `NTLM`, etc. As part of that we should support `bearer` in front of API tokens as the suggested format, while still support the case when `bearer` is not provided (treat it as default).

This will deliver the following benefits:

1. Assess if the header is related to basic auth vs. bearer auth and handle it accordingly and show correct error messages.
2. Support for `bearer api-token` format, which is how we should have done it in the first place. That's on me :)

raspopov:
Maybe just clarify the return text as "Valid API Token Required"? 😏

dregad:
> we should support bearer in front of API tokens as the suggested format

I agree. And I would even go one step further: deprecate use of Authorization header without `bearer` and remove it in the future, as it's not compliant with the [standard](https://httpwg.org/specs/rfc9110.html#field.authorization).

>> it will say "API Token Required" when an invalid token is provided.
>
> Maybe just clarify the return text as "Valid API Token Required"? 

Good suggestion, I think this would indeed clarify the response.

vboctor:
@raspopov Your suggested string is a partial fix to the issue introduced here. Here are my thoughts:

1. It would be good to move towards the standard and support `bearer` prefix.

2. For example, If basic auth is used, there is no need to check the database to see if the basic auth string is a valid API token; it will never be.

@dregad I agree that we should deprecate the calls without the `bearer` prefix, but other than documenting this, I'm not sure how we want to surface such deprecation. We can log it, but that penalizes the server (by spamming the logs) for clients using deprecated behavior, etc. Blocking it just seems to harsh and will break the ecosystem a bit, for low value. Adding a deprecation response header is an option as well.

raspopov:
Please note that the purpose of this PR is not to improve Bearer authentication, but to fix a blocking error when using non-Bearer authentication if the Authorisation header is present not for MantisBT but for the web server.

vboctor (approved):

dregad:
> Please note that the purpose of this PR is not to improve Bearer authentication, but to fix a blocking error when using non-Bearer authentication if the Authorisation header is present not for MantisBT but for the web server.

Fair point.

I have opened [#35286](https://mantisbt.org/bugs/view.php?id=35286) to so we don't lose track of this, as I believe it is really something that needs to be done.

I'll merge this PR shortly.

PR #2074, opened 2025-01-15
Fix PHPUnit 10/11 compatibility of the name() method using

raspopov:
Fix PHPUnit compatibility by updating to use the new name() method instead of the deprecated getName() method.

Fixes [#35223](https://mantisbt.org/bugs/view.php?id=35223).

Alternative to 7ccca0f9f991d692cd33156b48e2b90fc8bc0a45.

### DIFF
--- a/tests/rest/RestBase.php
+++ b/tests/rest/RestBase.php
@@ -189,18 +189,6 @@ abstract class RestBase extends PHPUnit\Framework\TestCase {
 	}
 
 	/**
-	 * Generate a Test Case reference (TestClass::TestCase).
-	 *
-	 * This can be used to associate the test data (e.g. Issue Summary, Project
-	 * Name, etc.) with a test case.
-	 * 
-	 * @return string
-	 */
-	protected function getTestCaseReference() {
-		return $this->toString();
-	}
-
-	/**
 	 * Returns a minimal data structure for tests to create a new Issue.
 	 *
 	 * The Issue Summary is set to TestClass::TestCase with an optional
@@ -211,7 +199,7 @@ abstract class RestBase extends PHPUnit\Framework\TestCase {
 	 * @return array
 	 */
 	protected function getIssueToAdd( $p_suffix = '' ) {
-		$t_summary = $this->getTestCaseReference();
+		$t_summary = $this->toString();
 		if( $p_suffix ) {
 			$t_summary .= '-' . $p_suffix;
 		}
--- a/tests/rest/RestBase.php
+++ b/tests/rest/RestBase.php
@@ -24,6 +24,7 @@
  */
 
 use GuzzleHttp\Exception\GuzzleException;
+use PHPUnit\Framework\TestCase;
 
 # Includes
 require_once dirname( __DIR__ ) . '/TestConfig.php';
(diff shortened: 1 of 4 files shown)


tests/soap/ProjectTest.php
```diff
@@ -201,7 +201,8 @@ protected function tearDown(): void {
 	 */
 	protected function getNewProjectName() {
 		do {
-			$t_name = $this->getName() . '_' . rand();
+			$t_name = ( method_exists( $this, 'getName' ) ?
```
dregad, line 204: When there are version-specific code branches, I prefer to check based on version number as I did in my original, WIP commit. The reason is to facilitate future cleanup activities, to identify code that can be removed when we bump the minimum requirements and the check no longer applies.

A comment is also useful, e.g.
https://github.com/mantisbt/mantisbt/commit/7ccca0f9f991d692cd33156b48e2b90fc8bc0a45#diff-b52215be84fb8db024faf46fcfa124b282a4b141f1327588fc78660f3c36d980R201

dregad (changes requested):
Thanks for following up on my earlier attempt to fix this.

tests/soap/ProjectTest.php
raspopov: I think I've found a better way. Without depending on versions.

tests/rest/RestBase.php
```diff
@@ -43,7 +42,7 @@
  * @requires extension curl
  * @group REST
  */
-abstract class RestBase extends TestCase {
+abstract class RestBase extends PHPUnit\Framework\TestCase {
```
dregad, line 45: Why are you removing the import in favor of a qualified reference ? I'd rather not change that.

tests/rest/RestBase.php
```diff
@@ -198,7 +197,7 @@ protected function getCategory() {
 	 * @return string
 	 */
 	protected function getTestCaseReference() {
-		return static::class . '::' . $this->getName();
+		return $this->toString();
```
dregad, line 200: That actually makes our own getTestCaseReference() method useless and redundant, might as well get rid of it completely and replace it by toString() wherever it's used...

dregad (changes requested):
> I think I've found a better way. Without depending on versions.

👍 Very nice, this toString() method is exactly what we need ! I was not even aware it existed 😳

tests/rest/RestBase.php
raspopov: Well, I just gave it the look and feel used in other MantisBT tests, plus the imported name is only used once and it is very close to the beginning of the file and immediately visible to the developer.

tests/rest/RestBase.php
raspopov: You are quite right, indeed we can simplify the code and it will even become similar to the ‘soap’ test suite.

tests/rest/RestBase.php
dregad: Sorry for picking nits, but I insist... you're just reverting a change that I made in 2023. 

https://github.com/mantisbt/mantisbt/commit/829ab70c25cbfac36efccb4a80eac9bb46f38ac5#diff-b52215be84fb8db024faf46fcfa124b282a4b141f1327588fc78660f3c36d980L43-R46

> I just gave it the look and feel used in other MantisBT tests

In that case it's the *other* test scripts that should abandon the fully qualified refs; I usually fix those things when I need to update a file, which explains the discrepancy - but there should not be that many of them.

dregad (approved):

tests/rest/RestBase.php
dregad: Thanks

> I have seen in many places in MantisBT the use of fully qualified names, e.g. function parameter types...

I'm slowly getting rid of them, whenever I notice an occurrence.

PR #2078, opened 2025-01-18
Added firebird database connection feature

manolopygithub:
Mantisbt support MySQL, Oracle, Postgres, etc database connections but not to Firebird; so, after a lot of hard work and patience I finally added firebird database connection feature.

The minnor version of firebird has to be 4.0, that is because the versions before 4.0 have limits the tables names length to 32 characters.

The database file must be created before the instalation, I recommend using the aliases file, create an alias and then use it in the instalation

### DIFF
--- a/adm_config_report.php
+++ b/adm_config_report.php
@@ -213,9 +213,16 @@ while( $t_row = db_fetch_array( $t_result ) ) {
 }
 
 # Build config query
+//Added by MAB - Feb/2024
+if (!db_is_firebird ())
 $t_sql = 'SELECT config_id, user_id, project_id, type, access_reqd,'
 		. ' CASE type WHEN :complex THEN null ELSE value END AS value'
 		. ' FROM {config} WHERE 1=1';
+else
+  $t_sql = 'SELECT config_id, user_id, project_id, type, access_reqd,'
+		. ' CASE type WHEN :complex THEN null ELSE "value" END AS "value"'
+		. ' FROM {config} WHERE 1=1';	
+		
 if( $t_filter_user_value != META_FILTER_NONE ) {
 	$t_sql .= ' AND user_id = :user_id';
 }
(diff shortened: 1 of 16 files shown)


vboctor:
Thanks, @manolopygithub, for your contribution.

Before we review the code, we should assess whether it is worth adding support for Firebird and its associated code bloat, ongoing validation, and ongoing maintenance. I believe that it won't add enough value to the project.

vboctor:
Given the frequent usage of `db_is_firebird()` I can see two issues:

- Firebird DB is not well abstracted by ADODB.
- Firebird DB will likely be frequently broken or slow down development to fix CI failures.

I haven't researched the market demand or popularity of Firebird, but we will need to make the case for that.

manolopygithub:
Thank you for your response Victor, I understand, it is ok, I just wanted
to contribute a little bit to the project.

Additionally I also added some fixes to ADODB project, in order to use it
with Firebird.

Best regards!

dregad:
Thanks for your contribution. I appreciate your efforts, but I'm afraid I have to second @vboctor on this. 

I'm concerned about the extra burden it will put on our small team, to have to support another RDBMS going forward. To be honest, I'm not convinced there is any real demand for Firebird... In nearly 15 years since I joined the project, this is only the [second time](https://mantisbt.org/bugs/view.php?id=13036) I ever see Firebird mentioned.

Aside from that, just a few general comments on your PR
- make sure your code follows our [coding guidelines](https://mantisbt.org/wiki/doku.php/mantisbt:coding_guidelines)
- PRs (and commit messages) must reference at least one issue in our bugtracker
- Adding comments like *Added by MAB* is totally unnecessary and should be removed - this information is automatically managed by Git

raspopov:
It would appear that the escaping of reserved keywords should be a matter for ADODB. 🤔

PR #2083, opened 2025-01-26
Project name trimming on creation and update

vboctor:
Fixes [#35260](https://mantisbt.org/bugs/view.php?id=35260)

### DIFF
--- a/core/project_api.php
+++ b/core/project_api.php
@@ -348,7 +348,9 @@ function project_create( $p_name, $p_description, $p_status, $p_view_state = VS_
 		trigger_error( ERROR_PROJECT_NAME_INVALID, ERROR );
 	}
 
-	project_ensure_name_unique( $p_name );
+	$t_name = trim( $p_name );
+
+	project_ensure_name_unique( $t_name );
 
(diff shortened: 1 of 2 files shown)


dregad (approved):

dregad:
That was fast 😉

atrol:
@vboctor @dregad 
Changing input data in this function in a some kind hidden way looks not that clean to me.
Isn't it like trying to store a primary key value "A" in a database, but the database changes the value to "B" behind the scenes?
Wouldn't the clean approach be to throw an exception in backend if the provided value is not valid (e.g. contains leading / trailing white spaces)?
In addition, not necessarily needed, implement some JavaScript input validation on client side to provide early feedback to end users and to avoid round trips to the server.

dregad:
Dropping whitespace is not entirely new in our codebase, also happens for Versions, and Tags for example.

Personally it does not bother me, if I enter `MyProject ` (with trailing space) to have the system store it as `MyProject` - it' even a good thing because this trailing space is meaningless. Just like if I enter a number like `0123.90`, it will be stored as `123.9` - really not worth throwing an exception IMO.

PR #2085, opened 2025-01-29
Fix unnecessary scroll bar when scaling a page

raspopov:
Fixes: [#35180](https://mantisbt.org/bugs/view.php?id=35180#c69766).

### DIFF
--- a/css/default.css
+++ b/css/default.css
@@ -24,6 +24,8 @@ td.category .small, th.category .small	{ font-weight: normal;  }
 }
 tr.spacer			{ background-color: #ffffff !important; color: #000000; height: 5px; }
 
+/* Issue Summary at bug_report_page.php, bug_update_page.php */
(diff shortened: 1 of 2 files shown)


css/default.css
```diff
@@ -24,6 +24,8 @@ td.category .small, th.category .small	{ font-weight: normal;  }
 }
 tr.spacer			{ background-color: #ffffff !important; color: #000000; height: 5px; }
 
+/* Issue Summary at bug_report_page.php, bug_update_page.php */
+input[type="text"]#summary { width: 100%; }
```
dregad, line 28: Since we identify the field by id, it seems redundant to have `[type="text"]` selector (even `input` is not necessary).

dregad: OK it's a bit of overhead for CSS processing, but I guess that's negligible.

dregad (approved):

PR #2082, opened 2025-01-25
Add namespaces to PHPUnit test suite

dregad:
Fixes [#35259](https://mantisbt.org/bugs/view.php?id=35259), [#35258](https://mantisbt.org/bugs/view.php?id=35258)

See also #2074

PR #2087, opened 2025-02-01
PHP 8.4 compatibility

dregad:
This fixes all the PHP 8.4 compatibility issues that I know about, with the exception of those from Parsedown ([#35217](https://mantisbt.org/bugs/view.php?id=35217)).

- [#35215](https://mantisbt.org/bugs/view.php?id=35215) PHP 8.4: Implicitly nullable parameter types are deprecated  
- [#35214](https://mantisbt.org/bugs/view.php?id=35214) PHP 8.4: fputcsv() empty $escape parameter is deprecated  
- [#35213](https://mantisbt.org/bugs/view.php?id=35213) PHP 8.4: E_STRICT is deprecated  
- [#35283](https://mantisbt.org/bugs/view.php?id=35283) PHP 8.4: SOAP API throws SoapFault: Internal Service Error  
- [#35284](https://mantisbt.org/bugs/view.php?id=35284) Allow REST API to run on PHP 8.4 ignoring E_DEPRECATED notices

### DIFF
--- a/api/rest/index.php
+++ b/api/rest/index.php
@@ -57,6 +57,12 @@ if( ON == config_get_global( 'show_detailed_errors' ) ) {
 # For debugging purposes, uncomment this line to avoid truncated error messages
 # $t_config['settings']['addContentLengthHeader'] = false;
 
+# Disable E_DEPRECATED warnings for PHP 8.4
+# Necessary to catch errors in method signatures triggered before our custom
+# error handler is in place.
+if( version_compare( PHP_VERSION, '8.4', '>=' ) ) {
+	$t_old_error_reporting = error_reporting( error_reporting() & ~E_DEPRECATED );
+}
 
 if( version_compare( Slim\App::VERSION, '4.0', '<' )
 	&& version_compare( PHP_VERSION, '8.1', '>=' )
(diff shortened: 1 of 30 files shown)


api/rest/index.php
```diff
@@ -163,6 +175,7 @@ function deprecated_errors_handler(
 # Wrap the whole API initialization and execution in a try/catch block, to
 # ensure we capture every error that could be occurring.
 try {
+	//var_dump(error_reporting());die;
```
vboctor, line 178: Remove this line.

vboctor (changes requested):
I assume you checked the callers of the methods that used to default parameters to null and you replaced them to `[]` or equivalent to make sure the callers are not passing nulls, since this was part of the supported contract before.

dregad:
Thanks for the review.

> I assume you checked the callers of the methods that used to default parameters to null and you replaced them to [] or equivalent to make sure the callers are not passing nulls

Yes I did. Hopefully I did not miss any, which seems confirmed by the fact that no errors were thrown while running the unit tests.

api/rest/index.php
dregad: Done

raspopov:
Would you mind fixing the bootstrap.php file (or the class files themselves) so that the autoloading of classes works correctly after this change?

dregad:
@raspopov I will gladly fix any issues, but I'm not sure I understand what the problem you're referring to is. Was it introduced by this PR ? If not, then please open an Issue in the bug tracker and clearly explain what is wrong with autoloading and bootstrap.php.

raspopov:
> @raspopov I will gladly fix any issues, but I'm not sure I understand what the problem you're referring to is.

I apologize for being a bit misleading. 😳 It's just that I read about tests and tried to test it and it didn't work. In fact, it's because of https://github.com/mantisbt/mantisbt/pull/2082#pullrequestreview-2588366946.

PR #2088, opened 2025-02-02
Support for email sender providers

vboctor:
Fixes [35288](https://mantisbt.org/bugs/view.php?id=35288)

See [TestEmailSender](https://github.com/mantisbt-plugins/TestEmailSender) as an example.


dregad:
If I understand correctly, someone wanting to implement another provider would have to create the class and store it in `core/classes/EmailSender.ProviderName.class.php`

I don't think that's acceptable as it amounts to modifying MantisBT source code. IMO a mechanism is needed to do this without touching MantisBT core. A plugin hook ?

vboctor:
@dregad I have added the implementation to allow send providers as plugins. A sample implementation of a provider can be found [here](https://gist.github.com/vboctor/86fddf85479d47bc314025dbea7e8081).

vboctor:
Feedback is welcome on naming of the event, class, etc.

vboctor:
@dregad ping on this so I can do any final touches and merge.

dregad (changes requested):
For naming, I personally prefer **EmailSender** to EmailGateway. *Gateway* can mean many things, and the term *email gateway* generally references a security infrastructure to protect mail servers from spam, malware and other email-based attacks, which is not what we're doing.

I only performed minimal testing, making sure that e-mails are being sent by PHPMailer without regression - which seems OK AFAICT. 

I didn't check the plugin event to register a 3rd party provider. If you intend to write such plugin (e.g. for SendGrid), it would be good to release it in the mantisbt-plugins org as an example, and reference it in the Admin Guide.

I believe it would be useful to add at the beginning of Admin Guide section 5.8, a high-level description of what email providers are, mentioning the bundled default PhpMailer and the possibility of adding custom ones (referencing the event in the Dev Guide).

vboctor:
@dregad

1. I addressed your comments
2. I created [TestEmailSender](https://github.com/mantisbt-plugins/TestEmailSender).
3. I did some work to accommodate the requirement for #2092 - where implementations of EmailSender plugins can re-use PhpMailer implementation, but be able to add provider specific headers.
4. I did some more testing and applied fixes.

vboctor:
@dregad I'm planning to merge this over the weekend, if you want to have another look.

dregad (approved):
Apologies, I missed your last post.

Looking good, but only performed minimal testing with default PhpMailer provider. As far as I can tell, notifications are working as before, did not notice any regressions.

I  did not look into the EmailSender plugins at all.

PR #2091, opened 2025-02-09
Fix print_form_button regression

atrol:
Issue [#35215](https://mantisbt.org/bugs/view.php?id=35215)

### DIFF
--- a/core/print_api.php
+++ b/core/print_api.php
@@ -1474,7 +1474,7 @@ function print_manage_project_sort_link( $p_page, $p_string, $p_field, $p_dir, $
  *
  * @return void
  */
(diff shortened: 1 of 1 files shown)


dregad:
As mentioned in https://mantisbt.org/bugs/view.php?id=35215#c69844

If possible, I'd rather fix the calls to print_form_button() to provide an empty array, rather than changing the function signature to allow null. I'll have a closer look.

atrol:
> If possible, I'd rather fix the calls to print_form_button() to provide an empty array

If you want to go this way, use https://github.com/mantisbt/mantisbt/pull/2093

dregad:
I merged #2093, making this PR obsolete.

PR #2096, opened 2025-02-11
Fix HTML errors and warnings

raspopov:
Fixes [#35180](https://mantisbt.org/bugs/view.php?id=35180).

Fix HTML issues reported by Nu Html Checker:

- Error: Stray end tag `i` in filter_print_view_type_toggle()
- Error: Element `option` without attribute `label` must not be empty in  filter_draw_selection_area()
- Error: Row 7 of a row group established by a `tbody` element has no  cells beginning on it
- Warning: A table row was 0 columns wide, which is less than the column  count established by the first row (8) in filter_form_draw_inputs()
- Warning: The `type` attribute is unnecessary for JavaScript resources  in  html_javascript_link(), html_javascript_cdn_link() and  html_head_javascript()
- Error: Illegal character in query: `[` is not allowed in my_view_inc.php
- Error: End tag for `body` seen, but there were unclosed elements in lost_pwd_page.php
- Error: Unclosed element `div`
- Warning: Consider avoiding `viewport` values that prevent users from resizing documents in layout_head_meta()
- Error: An `img` element must have an `alt` attribute, except under certain conditions in layout_login_page_logo()
- Error: Element `div` not allowed as child of element `ul` in this context in layout_breadcrumbs()

Add a new method TableGridLayout::render_spacer() to print a HTML code for a spacer row of current table.

Add new function string_build_query() to string_api.php, a unified call of http_build_query().

Moved the extra code from layout_page_begin() and layout_page_end() to new paired functions layout_main_content_row_begin() and layout_main_content_row_end().

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -1295,7 +1295,7 @@ function layout_scroll_up_button() {
 function layout_login_page_logo() {
 	?>
 	<div class="login-logo">
-		<img src="<?php echo helper_mantis_url( config_get_global( 'logo_image' ) ); ?>">
+		<img src="<?php echo helper_mantis_url( config_get_global( 'logo_image' ) ); ?>" alt="<?php echo string_html_specialchars( config_get( 'window_title' ) ); ?>">
 	</div>
 	<?php
 }
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -161,9 +161,9 @@ function layout_page_begin( $p_active_sidebar_page = null ) {
 
 	layout_breadcrumbs();
 
-	layout_main_content_row_begin();
-
 	layout_page_content_begin();
+
+	layout_main_content_row_begin();
 
 	if( auth_is_user_authenticated() ) {
 		if( ON == config_get( 'show_project_menu_bar' ) ) {
@@ -185,8 +185,8 @@ function layout_page_end() {
 
 	event_signal( 'EVENT_LAYOUT_CONTENT_END' );
 
-	layout_page_content_end();
 	layout_main_content_row_end();
+	layout_page_content_end();
 	layout_main_content_end();
 
 	layout_footer();
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -1015,7 +1015,7 @@ function layout_main_content_end() {
  * @return void
  */
 function layout_page_content_begin() {
-	echo '  <div class="page-content">' , "\n";
+	echo '<div class="page-content">' , "\n";
 }
 
 /**
@@ -1031,7 +1031,6 @@ function layout_page_content_end() {
 	echo '</div>' , "\n";
 }
 
-
 /**
  * Render breadcrumbs bar.
  * @return void
@@ -1045,6 +1044,8 @@ function layout_breadcrumbs() {
 
 	# Login information
 	echo '<ul class="breadcrumb">' , "\n";
+	echo '  <li>';
+	print_icon( 'fa-user', 'home-icon active' );
 	if( current_user_is_anonymous() ) {
 		$t_return_page = $_SERVER['SCRIPT_NAME'];
 		if( isset( $_SERVER['QUERY_STRING'] ) && !is_blank( $_SERVER['QUERY_STRING'] )) {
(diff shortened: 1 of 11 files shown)


core/string_api.php
```diff
@@ -237,6 +237,15 @@ function string_url( $p_string ) {
 	return rawurlencode( $p_string );
 }
 
+/**
+ * Build a URL-encoded query string from an array of key value pairs.
+ * @param  array $p_params Query string parameters.
+ * @return string
+ */
+function string_build_query( array $p_params ): string {
```
dregad, line 245: For consistency, shouldn't this new API function replace other, existing usages of http_build_query() ?

Otherwise we'll end up with some URLs RFC1738-encoded (default for $encoding_type parameter). Maybe not an issue, did not check.

core/layout_api.php
```diff
@@ -1295,7 +1295,7 @@ function layout_scroll_up_button() {
 function layout_login_page_logo() {
 	?>
 	<div class="login-logo">
-		<img src="<?php echo helper_mantis_url( config_get_global( 'logo_image' ) ); ?>">
+		<img src="<?php echo helper_mantis_url( config_get_global( 'logo_image' ) ); ?>" alt="<?php echo string_html_specialchars( config_get( 'window_title' ) ); ?>">
```
dregad, line 1286: Adding *alt* is good, however it should normally describe what the image is about. Using window_title seems inappropriate for this (yes I'm aware this is what is already being done in html_print_logo(), since https://mantisbt.org/bugs/view.php?id=14418).

dregad (changes requested):
Thanks @raspopov $

Regarding commit messages: could you please read https://mantisbt.org/wiki/doku.php/mantisbt:git_commit_messages and follow the guidelines ?

Particularly in this PR
- missing blank line between summary and description
- long lines not wrapped

raspopov:
@dregad,

> Regarding commit messages

Yeah, those commits were a bit sloppy. Thanks for the guide. I rebased the comits with only changes to the descriptions.

core/layout_api.php
raspopov: > Adding alt is good, however it should normally describe what the image is about.

This is an interesting question, I was guided by the “Branding” section in the `config_inc.php.sample` file:
```
# --- Branding ---
# $g_window_title	= 'MantisBT';
# $g_logo_image		= 'images/mantis_logo.png';
# $g_favicon_image	= 'images/favicon.ico';
```
This shows that `logo_image` is proposed for the graphical representation of the brand and `window_title` for the textual title of the brand. In my opinion, this is a perfect candidate for the `alt` tag, which is an alternative text description for the image.

**But** that's not the most interesting part. 🧐 My further investigation of the site branding code revealed that `html_top_banner()` and `html_print_logo()` functions are no longer called, respectively the `top_include_page` option is not used, the `EVENT_LAYOUT_PAGE_HEADER` event is no longer generated. The `bottom_include_page` option is not used anywhere else either. And the `logo_url` option is now only used in `layout_navbar()` when displaying the navbar at the top of the page, although it should wrap the logo.

core/string_api.php
dregad: That's unfortunately something that must be expected with a 25-year-old code base 😉

core/layout_api.php
dregad: Ah yes you're absolutely right - it is actually a known regression issue that was introduced in MantisBT 2.0, and never got fixed. See 
- https://mantisbt.org/bugs/view.php?id=21113
- https://mantisbt.org/bugs/view.php?id=22098

Discussion in PR #860 never went anywhere.

dregad: And to reply on topic: 
In theory _alt_ should offer a description of what the image is about, so it should ideally be *{window_title} logo*, *Company logo* or something like that. But this is really a small detail and outside the scope of this PR. 

I'm fine with using $g_window_title here. (discussion about this code not being used notwithstanding 😉 )

core/string_api.php
raspopov: I think that if this PR is accepted, a separate issue should be opened for string_build_query() using.

core/layout_api.php
raspopov: @dregad I certainly can't guarantee it, but I'll try to do something about it after this PR.
Look here please: https://mantisbt.org/bugs/view.php?id=22098#c69877

dregad (approved):

core/string_api.php
dregad: Opened https://mantisbt.org/bugs/view.php?id=35424 to track this.

PR #2100, opened 2025-02-17
Fix project selection on the edit project page

raspopov:
Also the code is cleaned up and simplified.

Fixes [#6264](https://mantisbt.org/bugs/view.php?id=6264).

### DIFF
--- a/set_project.php
+++ b/set_project.php
@@ -61,6 +61,8 @@ $t_project = explode( ';', $f_project_id );
 $t_top     = $t_project[0];
 $t_bottom  = $t_project[count( $t_project ) - 1];
 
+$t_redirect_url = config_get_global( 'default_home_page' );
+
 if( ALL_PROJECTS != $t_bottom ) {
 	project_ensure_exists( $t_bottom );
 }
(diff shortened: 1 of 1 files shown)


set_project.php
```diff
@@ -77,45 +79,45 @@
...
-			$t_redirect_url = $t_path;
-		} else if( $t_referrer_page == 'plugin.php' ) {
-			$t_redirect_url = $t_referrer_page . $t_param; # redirect to same plugin page
-		} else {
-			$t_redirect_url = $t_home_page;
+		switch( $t_referrer_page ) {
+			case 'view_all_bug_page.php':		
+				$t_source_filter_id = filter_db_get_project_current( $t_bottom );
+				$t_redirect_url = 'view_all_set.php?'
+					. http_build_query( ( $t_source_filter_id !== null ) ?
```
dregad, line 93: Use new string_build_query() ?

Also, when ternary operator has to span multiple lines, I find it easier to follow when it's written like this:

```php
condition
	? true 
	: false
```

Or use an if...else with a $t_query_param variable

set_project.php
```diff
@@ -77,45 +79,45 @@
...
+						[ 'type' => FILTER_ACTION_LOAD,
+						  'source_query_id' => $t_source_filter_id ] :
+						[ 'type' => FILTER_ACTION_GENERALIZE ] );
+				break;
+
+			case 'manage_proj_edit_page.php':
+			case 'manage_proj_page.php':
+				$t_redirect_url = ( ALL_PROJECTS != $t_bottom ) ?
+					'manage_proj_edit_page.php?'
+						. http_build_query( [ 'project_id' => $t_bottom ] ) :
```
dregad, line 103: ditto

dregad (changes requested):
Tested OK, thanks for your contribution.

What about using the new string_build_query() in anticipation of [#35424](https://mantisbt.org/bugs/view.php?id=35424) since #2096 has now been merged ?

set_project.php
raspopov: Sure.

dregad (approved):

PR #2102, opened 2025-02-20
Restore included pages functionality

raspopov:
- Restored working of **top_include_page** and **bottom_include_page** options.
- Restored generation of EVENT_LAYOUT_PAGE_HEADER event.
- Added dynamic footer resizing.

The **top_include_page** file is included within the `navbar` element, which is fixed at the top of the page; the **bottom_include_page** file is included within an existing footer or as a footer itself (on login pages), which is also fixed at the bottom of the page. Because of this, the include template must contain, for example, a `div` element with a solid background.

**Note:** There is some side effect due to the not yet fixed layout of the admin pages, the footer can touch the content of the page.

Fixes [#35402](https://mantisbt.org/bugs/view.php?id=35402), [#22250](https://mantisbt.org/bugs/view.php?id=22250), [#22098](https://mantisbt.org/bugs/view.php?id=22098), [#21113](https://mantisbt.org/bugs/view.php?id=21113).


dregad:
Nice work.

About the modified _ace.css_, _ace-skins.css_ and _ace.js_ files: these are part of ACE template, and until today the general practice has been to leave those untouched, with the exception of [#24520](https://mantisbt.org/bugs/view.php?id=24520). 

I would suggest, if possible, to move these changes to *ace-mantis.css* and *common.js*

> **Note:** There is some side effect due to the not yet fixed layout of the admin pages, the footer can touch the content of the page.

Not sure I understand what you mean by that. Can you please clarify what this side effect is ?

raspopov:
> Not sure I understand what you mean by that. Can you please clarify what this side effect is ?

Layout of admin pages created by pair of **layout_admin_page_begin()** and **layout_admin_page_end()** is different from login and main pages, i.e. missing parts:
```
<div class="main-content">
  <div class="page-content">
    <div class="row">
```
And since this commit deleted the following now unnecessary lines:
```
echo '<div class="clearfix"></div>' . "\n";
echo '<div class="space-20"></div>' . "\n";
```
Which gave a small margin to the footer even for the wrong page layout, now gone.
I could include a fix right in this PR, but it seems better to separate the different issues.

raspopov:
> I would suggest, if possible, to move these changes to _ace-mantis.css_ and _common.js_

I had to move `ace-mantis.css` to the last place, it should have been done earlier instead of writing the `!important` attribute everywhere...

dregad:
> I had to move ace-mantis.css to the last place

That's fine.

> it should have been done earlier instead of writing the !important attribute everywhere...

Absolutely right. 
ACE makes extensive use of !important too... Messy...

dregad (changes requested):
It's looking pretty good. I did not test the EVENT_LAYOUT_PAGE_* events though.

The PHPDoc for $g_top_include_page and $g_bottom_include_page, as well as the relevant sections in the Admin Guide (docbook/Admin_Guide/en-US/config/html.xml) need to be updated, including 
- the sample HTML for top include, which was written in 2016 for MantisBT 1.3.0, and 
- the bit about the logo not being shown if there is a top-include file, which is wrong as the logo is no longer shown since 2.0.

Additionally, these 2 config's default value (`%absolute_path%`) makes absolutely no sense IMO, they should be just `''`.

Lastly,I would rather have this change targeted at 2.28 (master), if you don't mind rebasing your branch (otherwise I'll take care of it when I merge).

dregad:
As a side note, in the commit message you wrote

> `Fixes #35402, Fixes #22250, Fixes #22098.`

FYI, unlike GitHub, the source integration will accept a comma-delimited list of referenced issues, so you can shorten this to `Fixes #35402,  #22250, #22098`

css/ace-mantis.css
```diff
@@ -756,3 +756,16 @@ ul.changelog > li, ul.roadmap > li {
...
+    padding-left: 0;
+    padding-right: 0;
+}
+/* Move the button higher than the bottom included page */
+.btn-scroll-up {
+    z-index: 1099;
+}
+/* Fix the background of the navbar */
+.skin-3 .navbar.navbar-collapse {
+	background-color: #777;
```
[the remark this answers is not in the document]
raspopov: This is incredible, I interacted and tested this interface for hours, but never saw that the color was slightly different, all my attention was on the layouts...

docbook/Admin_Guide/en-US/config/html.xml
```diff
@@ -98,48 +98,31 @@
...
-				<para>If a file is supplied, the logo specified by
-					<literal>$g_logo_image</literal> (see
-					<xref linkend="admin.config.display" />) will not be shown,
-					and the include file will have to handle display of the logo.
-					To do so you can use the <literal>html_print_logo()</literal>
-					API function, which will display the logo with an URL link if
-					one has been specified in <literal>$g_logo_url</literal>
-				</para>
-				<para>Example top include PHP file with logo and centered page title:
+				<para>Example include file with logo and centered page title:
```
dregad, line 107: The new example does not actually display a logo anymore. The `html_print_logo();` call should be added back.

dregad:
Out of curiosity, can you explain why you started adding `Signed-off-by: ` to your commits ?

dregad (changes requested):

raspopov:
> Out of curiosity, can you explain why you started adding `Signed-off-by: ` to your commits ?

According paragraph 4 of https://mantisbt.org/wiki/doku.php/mantisbt:git_commit_messages in case there's a re-base (but I often forget).

docbook/Admin_Guide/en-US/config/html.xml
raspopov: Indeed.

dregad:
Ah I see. I guess the intent behind that was not very clear, maybe wording should be improved. Sorry about that.

The idea was not for the commit's author to sign it off. What I meant is that a **MantisBT core dev** applying/merging a change from a contributor should sign it off, to indicate that it has been reviewed. 

This made sense 15 years ago when most contributions we received were from diffs pasted in Issues, or if we were lucky, git-formatted patches. With the switch to GitHub, this became kind of pointless, as attribution is given by commit author, code review is traceable through Pull Requests, and the committer tells us who applied the change.

In other words, you don't need to sign off your own commits.

dregad (approved):
Looks good to me.

I'm keeping this PR open for a bit before merging, as the Git client I'm using (SmartGit) showed a strange behavior when retrieving this PR yesterday (it was also unexpectedly pulling your all your _archive/*_ tags), and I'm waiting for their support to get back to me.

dregad:
Merged in 2ebef4eaa0970dd72048da6b87bfd98ac9db36b9

PR #2103, opened 2025-02-22
Increase navbar user name width

raspopov:
The max-width increased from 100px to 270px, this is the maximum size at which the navbar menu does not collapse on narrow screens.

Fixes [#23593](https://mantisbt.org/bugs/view.php?id=23593).

### DIFF
--- a/css/ace-mantis.css
+++ b/css/ace-mantis.css
@@ -242,6 +242,7 @@ a:link:after, a:visited:after {
 
 .user-info {
     line-height: 33px !important;
+    max-width: 270px !important;
 }
 
 .page-content {


dregad:
Not sure about this change. How did you come up with 270px ? When $g_window_title is long, this causes wrapping - navbar displays on 2 lines. 

For example:

### 1024px
image
image

### 800px
image
image

Of course, with a long enough title, this behavior would happen anyway with the current max width of 100px but it's less likely to occur.

I suppose in this case the expected behavior could be to display just the avatar, a click on which will display a 2nd line with the user menu (similar to what happens on narrow screens) but that's probably opening a big can of worms.

image

raspopov:
> When $g_window_title is long, this causes wrapping - navbar displays on 2 lines.

This is fine, the layout is carried over a few lines (where else can it go?) but the text stays solid inside the intact panels. But if the width is more than 270px at the size of a smartphone, the name panel bursts and the frames and down arrow fall off.

If we want a different behavior for the whole header, not just the username field, it's a matter of opening another issue and redesigning the item flow.

By the way, in the original ACE theme, user information can take up multiple rows within its element.

dregad:
> if the width is more than 270px at the size of a smartphone, the name panel bursts and the frames and down arrow fall off.

You mean this ?

image

> If we want a different behavior for the whole header, not just the username field, it's a matter of opening another issue and redesigning the item flow.

Let's not go there, not worth the effort IMHO.

raspopov:
> You mean this ?

Yes.

dregad (approved):

PR #2104, opened 2025-02-22
Fix inconsistent display in navbar user menu

dregad:
When using avatar ($g_show_avatar = ON), the username is displayed in a span with user-info class, but when OFF the span and class were not present.

With this commit, the span and class are always used.

Fixes [#35425](https://mantisbt.org/bugs/view.php?id=35425)


raspopov:
That's smart. And I didn't even look (in #2103) at the code that generates the markup, only at the markup itself...

PR #2105, opened 2025-02-23
Fix filters with date custom fields on PHP 8.0

dregad:
Fixes [#35291](https://mantisbt.org/bugs/view.php?id=35291)

Includes a fix for a related PHP 8.1 deprecation warning [#35428](https://mantisbt.org/bugs/view.php?id=35428).

### DIFF
--- a/core/gpc_api.php
+++ b/core/gpc_api.php
@@ -131,7 +131,7 @@ function gpc_get_int( $p_var_name, $p_default = null ) {
 		error_parameters( $p_var_name );
 		trigger_error( ERROR_GPC_ARRAY_UNEXPECTED, ERROR );
 	}
-	$t_val = str_replace( ' ', '', trim( $t_result ) );
+	$t_val = str_replace( ' ', '', trim( (string)$t_result ) );
 	if( !preg_match( '/^-?([0-9])*$/', $t_val ) ) {
 		error_parameters( $p_var_name );
 		trigger_error( ERROR_GPC_NOT_NUMBER, ERROR );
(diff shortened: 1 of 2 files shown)


core/gpc_api.php
```diff
@@ -131,7 +131,7 @@ function gpc_get_int( $p_var_name, $p_default = null ) {
 		error_parameters( $p_var_name );
 		trigger_error( ERROR_GPC_ARRAY_UNEXPECTED, ERROR );
 	}
-	$t_val = str_replace( ' ', '', trim( $t_result ) );
+	$t_val = str_replace( ' ', '', trim( (string)$t_result ) );
```
atrol, line 134: I am wondering why we have this `str_replace` call.
Isn't this hiding possible bugs?
If a string like `1      2` is provided, we will return int value 12

atrol (approved):

core/gpc_api.php
dregad: I also thought that logic was a bit strange, but I was focused on fixing the PHP 8 compatibility.

This code was introduced in 64e17ff48916b789fdf0f52240fb42d50e5b211f, as a fix for [#5092](https://mantisbt.org/bugs/view.php?id=5092). The purpose was for gpc_get_int() to throw an error if not given an integer value.

While trim() is fine, I agree that removing spaces in the middle of the string has potential for introducing bugs.

I think we should probably use `filter_var( $t_result, FILTER_VALIDATE_INT );` here - that function did not exist back in 2005, as it was introduced in PHP 5.2 (nov 2006).

I have opened [#35525](https://mantisbt.org/bugs/view.php?id=35525) to track this.

dregad: PR #2112

PR #2108, opened 2025-02-26
Fix Permalinks for Plugin Filter fields

dregad:
Fixes 
- Undefined array key system warning [#35179](https://mantisbt.org/bugs/view.php?id=35179) (replaces PR #2056)
- Incorrect styling of Plugin Filter Fields [#35471](https://mantisbt.org/bugs/view.php?id=35471)

### DIFF
--- a/core/filter_form_api.php
+++ b/core/filter_form_api.php
@@ -1853,13 +1853,13 @@ function print_filter_plugin_field( $p_field_name, $p_filter_object, array $p_fi
 
 	switch( $p_filter_object->type ) {
 		case FILTER_TYPE_STRING:
-			echo '<input class="input-xs" name="', string_attribute( $p_field_name ), '"',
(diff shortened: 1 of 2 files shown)


dregad:
> @dregad shouldn't this be fixed in 2.28.0 as part of [this issue](https://mantisbt.org/bugs/view.php?id=35215)?

You're probably right. Yesterday I replied from my phone without a detailed analysis and did not realize that the user was on 2.27.3. 

I was only trying to say that reporting at https://mantisbt.org/bugs is the right thing to do, as opposed to post a comment on a PR to discuss issues.

@fjungclaus if the problem does persist on 2.28.0, please report a new issue, and include your plugin's source (or at least the relevant portion of it so we can reproduce the bug).

fjungclaus:
Hi everyone, thanks for your replies. I'll try version 1.28 first... and will open an issue if the problem persists ...

PR #2110, opened 2025-02-26
Fix HTML errors and warnings 2

raspopov:
Fixes [#35503](https://mantisbt.org/bugs/view.php?id=35503).

Add a new function print_table_spacer() to print table spacer with specified columns.

Restored pagination when printing in both HTML and DOC formats.

Fix Nu Html Checker issues in dropzone (print_api.php):

- Error: Element `img` is missing required attribute `src`.
- Error: An `img` element must have an `alt` attribute, except under certain conditions.

In bug_view_inc.php:

- Error: Element `tbody` not allowed as child of element `table` in this context.
- Error: Duplicate ID history.
- Error: Row 1 of a row group established by a `thead` element has no cells beginning on it.
- Error: Element `option` without attribute `label` must not be empty.
- Warning: A table row was 6 columns wide and exceeded the column count established by the first row (0).

In bug_update_page.php, bugnote_view_inc.php, common.js:

- Error: Attribute `url` not allowed on element `a` at this point.
- Error: The width attribute on the `td` element is obsolete.
- Warning: A table row was 0 columns wide, which is less than the column count established by the first row (6).
- Error: Element `option` without attribute `label` must not be empty.
- Error: Duplicate ID bugnotes.
- Error: The first child `option` element of a `select` element with a `required` attribute, and without a `multiple` attribute, and without  a `size` attribute whose value is greater than 1, must have either an  empty `value` attribute, or must have no text content.

In print_all_bug_page.php:

- Error: Stray end tag `td`.
- Error: Row 2 of a row group established by a `tbody` element has no cells beginning on it.
- Warning: A table row was 10 columns wide and exceeded the column count established by the first row (5).

In print_all_bug_page_word.php:

- Error: The `width` attribute on the `td` element is obsolete.
- Error: Row 2 of a row group established by a `tbody` element has no cells beginning on it.

In bug_change_status_page.php:

- Error: Start tag `input` seen in `table`.
- Error: Element `option` without attribute `label` must not be empty.
- Error: End tag `div` seen, but there were open elements.
- Error: Unclosed element `fieldset`.

In bug_report_page.php:

- Error: Element `option` without attribute `label` must not be empty.
- Error: The value of the `for` attribute of the `label` element must be  the ID of a non-hidden form control.

In manage_user_create_page.php:

- Error: Start tag `fieldset` seen in `table`.
- Error: Unclosed element `div`.

### DIFF
--- a/bug_change_status_page.php
+++ b/bug_change_status_page.php
@@ -131,6 +131,9 @@ layout_page_begin();
 	<form id="bug-change-status-form" name="bug_change_status_form" method="post" action="bug_update.php">
 
 	<fieldset>
+		<input type="hidden" name="bug_id" value="<?php echo $f_bug_id ?>" />
+		<input type="hidden" name="status" value="<?php echo $f_new_status ?>" />
+		<input type="hidden" name="last_updated" value="<?php echo $t_bug->last_updated ?>" />
 
 	<?php echo form_security_field( 'bug_update' ) ?>
 	<div class="widget-box widget-color-blue2">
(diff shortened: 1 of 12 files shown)


dregad:
I'd like to freeze the scope of 2.27 branch so I can release 2.27.1, so please target this to master..

dregad (approved):

PR #2111, opened 2025-02-28
Add support of sitemap generation

raspopov:
Fixes [#35523](https://mantisbt.org/bugs/view.php?id=35523).

Discussion on [forum](https://mantisbt.org/forums/viewtopic.php?p=81300).

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -71,6 +71,8 @@ function layout_page_header( $p_page_title = '', $p_redirect_url = null, $p_page
  * @return void
  */
 function layout_page_header_begin( $p_page_title = '' ) {
+	$t_path = config_get_global( 'path' );
+
 	html_begin();
 	html_head_begin();
 	html_content_type();
@@ -89,6 +91,13 @@ function layout_page_header_begin( $p_page_title = '' ) {
 	$t_favicon_image = config_get_global( 'favicon_image' );
 	if( !is_blank( $t_favicon_image ) ) {
 		echo "\t", '<link rel="shortcut icon" href="', helper_mantis_url( $t_favicon_image ), '" type="image/x-icon" />', "\n";
+	}
+
+	if( layout_sitemap_enabled() ) {
+		echo "\t",
+			'<link rel="sitemap" type="application/xml" title="Sitemap" ',
+			'href="', $t_path, 'sitemap.php', '">',
+			"\n";
 	}
 
 	# Advertise the availability of the browser search plug-ins.
@@ -253,6 +262,18 @@ function layout_admin_page_end() {
 function layout_is_rtl() {
 	if( lang_get( 'directionality' ) == 'rtl' ) {
 		return true;
+	}
+	return false;
+}
+
+/**
+ * Check if Sitemap enabled
+ * @return bool
+ */
+function layout_sitemap_enabled() {
+	if( auth_anonymous_enabled() ) {
+		$t_anonymous_user_id = user_get_id_by_name( auth_anonymous_account() );
+		return $t_anonymous_user_id && user_is_enabled( $t_anonymous_user_id );
 	}
 	return false;
 }
@@ -689,11 +710,11 @@ function layout_navbar_user_avatar( $p_img_class = 'nav' ) {
 }
 
 /**
- * Print sidebar
- * @param string $p_active_sidebar_page page where the displayed page lives under
- * @return void
+ * Get accessible sidebar items
+ *
+ * @return array containing sidebar items
  */
-function layout_print_sidebar( $p_active_sidebar_page = null ) {
+function layout_get_sidebar_items() {
 	if( auth_is_user_authenticated() ) {
 		$t_current_project = helper_get_current_project();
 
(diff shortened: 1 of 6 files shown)


sitemap.php
```diff
@@ -0,0 +1,102 @@
...
+require_api( 'config_api.php' );
+require_api( 'database_api.php' );
+require_api( 'filter_api.php' );
+require_api( 'filter_constants_inc.php' );
+
+# Access level to get a sitemap
+$t_threshold = config_get( 'view_bug_threshold' );
+
+# Maximum number of URLs
+$t_max_urls = 50000;
```
vboctor, line 47: Use `$g_`

sitemap.php
```diff
@@ -0,0 +1,102 @@
...
+
+$t_path = config_get_global( 'path' );
```
vboctor, line 49: Use `$g_`

core/layout_api.php
```diff
@@ -91,18 +93,26 @@ function layout_page_header_begin( $p_page_title = '' ) {
 		echo "\t", '<link rel="shortcut icon" href="', helper_mantis_url( $t_favicon_image ), '" type="image/x-icon" />', "\n";
 	}
 
-	# Advertise the availability of the browser search plug-ins.
-	$t_title = htmlspecialchars( config_get( 'search_title' ) );
-	$t_searches = array( 'text', 'id' );
-	foreach( $t_searches as $t_type ) {
+	if( auth_anonymous_enabled() ) {
```
vboctor, line 96: Should you use `current_user_is_anonymous()` instead? Is the sitemap useful for signed in users?

core/layout_api.php
```diff
@@ -91,18 +93,26 @@ function layout_page_header_begin( $p_page_title = '' ) {
...
-			'<link rel="search" type="application/opensearchdescription+xml" ',
-				'title="' . sprintf( lang_get( "opensearch_{$t_type}_description" ), $t_title ) . '" ',
-				'href="' . string_sanitize_url( 'browser_search_plugin.php?type=' . $t_type, true ) .
-				'"/>',
+			'<link rel="sitemap" type="application/xml" title="Sitemap" ',
+			'href="', $t_path, 'sitemap.php', '">',
 			"\n";
 	}
 
+	# Advertise the availability of the browser search plug-ins.
```
vboctor, line 103: Is this a different feature?

sitemap.php
```diff
@@ -0,0 +1,102 @@
...
+		'<loc>', htmlspecialchars( $t_path . $p_url, ENT_XML1 | ENT_QUOTES ), '</loc>', "\n",
+		'</url>', "\n";
+}
+
+header( 'Content-Type: application/xml; charset=UTF-8' );
+
+echo '<?xml version="1.0" encoding="UTF-8"?>', "\n",
+	'<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">', "\n";
+
+if( auth_anonymous_enabled() ) {
```
vboctor, line 68: Should you use `current_user_is_anonymous()` instead?

vboctor (changes requested):
I think we should try to reduce the amount of work that is done from anonymous end points when we can. How about we have this script generate a sitemap cache file that is re-used and only refreshed if older than an hour.

sitemap.php
vboctor, line 47: If the generation of the sitemap is limited to once an hour, then we can remove the max limit which is arbitrary and not configurable.

raspopov: Thanks for your interest. But this variable is not a global setting, just a constant for the code below, put in the file header to make it easier to find. I don't think the standard will change, and 50000 will never change.

core/layout_api.php
raspopov: >  Is the sitemap useful for signed in users?

Yes, sitemap is for search engines only. They'll never log in anywhere.

sitemap.php
raspopov: Search engines load sitemaps, if they load them at all, at best once before a site's re-indexing, once a day or less often.

sitemap.php
raspopov: The current_user_is_anonymous() function not only uses the auth_anonymous_enabled() function, but also checks cookies in the database, which seems unnecessary in this situation.

raspopov:
> I think we should try to reduce the amount of work that is done from anonymous end points when we can.

I understand that caching sounds like a good idea, but then we’d have to constantly overwrite the sitemap.xml file somewhere on the site... 🤔

Search engines load sitemaps, if they load them at all, at best once before a site's re-indexing, once a day or less often. 

Even if MantisBT returns the maximum number of links—that is, 50,000—that would amount to about 150 KB of compressed content, since the links to the issues are fairly short.

core/layout_api.php
raspopov: Good point, I'm trying to stick small changes together again... 😔 Moved to #2126.

raspopov:
If #2127 is accepted, I will be able to add more links to the sitemap.

raspopov:
I updated the code by removing unnecessary dependencies and incorporating the relevant sections from #2127 and #2183. I strengthened the check to ensure that the sitemap is fully disabled when anonymous users are not permitted access. Accessing links to the issue now requires both 'view_bug_threshold' and 'limit_view_unless_threshold' permissions. A disabled sitemap will now return a 404 error.

raspopov:
This is a confusing situation with the tests... I expected that because of #2230, the tests would fail everywhere, reporting a lost link to the news. However, for some reason, the MySQL tests passed, while the PostgreSQL tests failed—but in a different way... 🤔

raspopov:
I've added more detailed error messages to the tests; this is useful for PostgreSQL.

As for MySQL, the code `SELECT * FROM mantis_news_table WHERE project_id=‘$c_project_id’` actually evaluates to `SELECT * FROM mantis_news_table WHERE project_id=0` due to the rules for converting strings to numeric values. And, as it luck would have it, this happens to be correct in this particular situation.

PR #2115, opened 2025-03-06
Fix Debug Log broken HTML

raspopov:
Added collapse support.

Also implemented some workaround for inline error messages that are often hidden under the top bar.

Code deduplication has been performed.

Fixes [#22098](https://mantisbt.org/bugs/view.php?id=22098#c69943).

### DIFF
--- a/core/error_api.php
+++ b/core/error_api.php
@@ -403,7 +403,7 @@ function error_handler( $p_type, $p_error, $p_file, $p_line ) {
 					if( defined( 'DELAY_INLINE_ERROR_REPORTING')) {
 						error_log_delayed( $t_error_type . ': ' . $t_error_description );
 					} else {
-						echo '<div class="alert alert-warning">', $t_error_type, ': ', $t_error_description, '</div>';
+						echo '<div class="alert alert-warning" style="position: relative; top: 4em;">', $t_error_type, ': ', $t_error_description, '</div>';
 					}
 				}
 				$g_error_handled = true;
--- a/admin/email_queue.php
+++ b/admin/email_queue.php
@@ -174,7 +174,5 @@ if( count( $t_ids ) > 0 ) {
 		</form>
 	</div>
 	</div></div></div>
-
-</div>
 <?php
 layout_admin_page_end();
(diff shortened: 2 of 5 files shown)


core/error_api.php
```diff
@@ -403,7 +403,7 @@ function error_handler( $p_type, $p_error, $p_file, $p_line ) {
 					if( defined( 'DELAY_INLINE_ERROR_REPORTING')) {
 						error_log_delayed( $t_error_type . ': ' . $t_error_description );
 					} else {
-						echo '<div class="alert alert-warning">', $t_error_type, ': ', $t_error_description, '</div>';
+						echo '<div class="alert alert-warning" style="position: relative; top: 4em;">', $t_error_type, ': ', $t_error_description, '</div>';
```
dregad, line 406: This is fine when the inline warning appears at the top the page, but the `position: relative` messes up display when it's at the bottom as the alert(s) are shifted down and overlap the bottom elements. For example:

image
image

dregad:
This PR is growing out of its own scope. 

Since I had already fixed the invalid HTML issue on my dev box following [my note](https://mantisbt.org/bugs/view.php?id=22098#c69945), I'll cherry-pick the nice improvements (log events count and collapsible section) on top of my branch.

Unrelated changes will be committed separately (inline errors behind navbar, code cleanup in access_has_global_level(), admin pages layout).

raspopov:
I agree. PR has started to follow a "geographical" direction of development rather than an idea.

core/error_api.php
raspopov: Indeed. There are other ways we can try to solve the problem:
- Add the style `z-index: 10000`, then the element will be on top and overlap the others.
- Define a default constant: `define( 'DELAY_INLINE_ERROR_REPORTING', true );` so that all errors go to the delayed messages panel.
- Add a mechanism to keep track of when the page title is printed, and elements before the `navbar` should be printed with `position: relative`, and after - without.

dregad:
Replaced by 
- #2119 (broken HTML, collapsible section, events count)
- #2120

Other changes (admin pages layout, duplicate code in access_has_global_level()) have been merged directly in master.

PR #2119, opened 2025-03-08
Refactor and improve log_print_to_page()

dregad:
- fixes the invalid HTML structure [#22098](https://mantisbt.org/bugs/view.php?id=22098).
- refactored the function (use HEREDOC, call helper_log_to_page(), process log events with a single loop)
- highlighted duplicate SQL queries in DB log entries (as it was before MantisBT 2.0)
- localized string in the section header
- Improvements by @raspopov (cherry-picked from #2115):
  - make the Debug log section collapsible 
  - display the number of events logged

Fixes [#35551](https://mantisbt.org/bugs/view.php?id=35551)
Supersedes PR #2115

### DIFF
--- a/core/logging_api.php
+++ b/core/logging_api.php
@@ -25,12 +25,14 @@
  * @copyright Copyright 2002  MantisBT Team - mantisbt-dev@lists.sourceforge.net
  * @link http://www.mantisbt.org
  *
+ * @uses collapse_api.php
  * @uses config_api.php
  * @uses constant_inc.php
  * @uses event_api.php
  * @uses utility_api.php
  */
 
+require_api( 'collapse_api.php' );
 require_api( 'config_api.php' );
 require_api( 'constant_inc.php' );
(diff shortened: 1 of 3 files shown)


dregad:
@raspopov review and feedback would be appreciated

raspopov:
@dregad Good. Nu Html Checker shows that there are no HTML errors.

I think it's still worth adding the line “No log data” or similar when there are zero events, since the unroll works but nothing happens, it's confusing. 
```
<tr>
    <td colspan="4">No log data.</td>
</tr>
```

dregad:
Thanks for your feedback.

> I think it's still worth adding the line “No log data” or similar when there are zero events, since the unroll works but nothing happens, it's confusing.

Agreed, will do

> Collapse does not work. Id attribute is missing.

Added.

dregad:
I added the "no data" message as a div instead of a table row.

image

PR #2120, opened 2025-03-08
Fix inline error message showing behind navbar

dregad:
<del>At this point, this PR is just a cherry-pick of @raspopov's proposed fix from PR #2115, with the shortcomings discussed there and in the bugtracker.</del>

Improves the delayed printing mechanism implemented in [#35207](https://mantisbt.org/bugs/view.php?id= 35207), replacing the DELAY_INLINE_ERROR_REPORTING constant by a global variable ($g_errors_delay_reporting), allowing to switch delayed reporting on and off as needed.

Reporting is delayed by default, and Layout API (layout_page_begin()) switches it off when the page header has been displayed.

Fixes [#35552](https://mantisbt.org/bugs/view.php?id=35552)

Also fixes delayed warnings not displayed on login page [#35583](https://mantisbt.org/bugs/view.php?id=35583).

### DIFF
--- a/core/error_api.php
+++ b/core/error_api.php
@@ -42,7 +42,25 @@ require_api( 'html_api.php' );
 require_api( 'lang_api.php' );
 
 $g_error_parameters = array();
+
+/**
+ * Determine if inline warnings should be printed immediately (true)
+ * or at page bottom (false).
+ *
+ * True initially, layout API sets it to false when the page header has been
+ * printed. Call {@see error_delay_reporting()} to change this value.
+ *
+ * @see error_log_delayed(), error_print_delayed()
+ * @global bool $g_error_delay_reporting
+ */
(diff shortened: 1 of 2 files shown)


dregad:
@raspopov In the end I went for a different approach than your original contribution. Your testing and feedback on this would be appreciated.

raspopov:
Looks okay.

But there is a small problem with calling an error with `DISPLAY_ERROR_INLINE` after `error_print_delayed();`, maybe we should disallow inline errors after this function.

dregad:
Ah yes good point, if an inline error is triggered  when delayed reporting is on after that point it won't be displayed. The easy fix would be to call `error_delay_reporting( false );` at the end of error_print_delayed().

I'm not sure would be worth the effort to implement something more complicated than that, what do you think ?

PR #2121, opened 2025-03-11
Fix layout of Access Denied message page

raspopov:
The page layout for the anonymous user has been restored, and the admin layout has been replaced with the normal one.

Fixes [#35561](https://mantisbt.org/bugs/view.php?id=35561).


dregad (approved):

dregad:
Thank you

PR #2122, opened 2025-03-11
Fix error message page layout

raspopov:
For the anonymous user, the layout was broken and the footer overlapped the content of the page.

Fixes [#35562](https://mantisbt.org/bugs/view.php?id=35562).


dregad (approved):

dregad:
Thanks !

PR #2123, opened 2025-03-16
Fix error message page layout 2

raspopov:
This is the alternative fix in place of ec7784aa042839714c661ace706ae15a398df0b2.

Admin page does not have infinite login loop if anonymous account is disabled.

Fixes [#35562](https://mantisbt.org/bugs/view.php?id=35562#c69995).


dregad (approved):
Oops 😉 

Maybe add a comment in the code, to explain this so our future selves don't make the same mistake again ?

raspopov:
> Maybe add a comment in the code, to explain this so our future selves don't make the same mistake again ?

That's a good idea. Done.

PR #2124, opened 2025-03-18
Fix URLs in the summary field

raspopov:
~~Added a new function string_wrap() and test for it.~~

The new string_display_line_without_links() function was added to simply remove any anchors.

Fixes [#35539](https://mantisbt.org/bugs/view.php?id=35539).


dregad (commented):
I don't really like this.

See my [note in the tracker](https://mantisbt.org/bugs/view.php?id=35539#c70776).

dregad:
Thanks for the update, will test later.

dregad:
Merged in https://github.com/mantisbt/mantisbt/commit/2b74b361dfa8a20767f271a47f42e6748a075172

PR #2126, opened 2025-03-19
Add ability to disable OpenSearch

raspopov:
To disable the OpenSearch tag on every page, simply set $g_search_title to an empty string.

Fixes [#35627](https://mantisbt.org/bugs/view.php?id=35627).

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -71,6 +71,8 @@ function layout_page_header( $p_page_title = '', $p_redirect_url = null, $p_page
  * @return void
  */
 function layout_page_header_begin( $p_page_title = '' ) {
+	$t_path = config_get_global( 'path' );
+
 	html_begin();
 	html_head_begin();
 	html_content_type();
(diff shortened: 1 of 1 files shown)


dregad (changes requested):
What is the reason for removing the call to string_sanitize_url() ?

The new functionality needs to be documented

- config_defaults_inc.php
  https://github.com/mantisbt/mantisbt/blob/c7f7c84f7e0cf987345f2d825f8c70c98fb0228d/config_defaults_inc.php#L1264-L1269
- Admin Guide
  https://github.com/mantisbt/mantisbt/blob/c7f7c84f7e0cf987345f2d825f8c70c98fb0228d/docbook/Admin_Guide/en-US/config/display.xml#L54-L61

raspopov:
> What is the reason for removing the call to string_sanitize_url() ?

All parts of the URL are completely controlled, using string_sanitize_url() is quite unnecessary. It is also desirable for the URL to be absolute.

PR #2127, opened 2025-03-19
Add new Menu API

raspopov:
- Mostly part of `layout_print_sidebar()` and whole `layout_options_for_sidebar()` combined and moved to `menu_api.php` as `menu_sidebar_options()`
- `layout_manage_menu_link()` moved to `menu_api.php` as `menu_manage_link()`
- `layout_plugin_menu_options_for_sidebar()` and `layout_config_menu_options_for_sidebar()` moved to `menu_api.php` as `menu_config_options()`
- A new key `access_level_any` has been added to the sidebar elements array, which works similarly to the existing `access_level`, but not for the current project, but for any project available to the user.

Fixes [#35629](https://mantisbt.org/bugs/view.php?id=35629).

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -37,11 +37,13 @@
  * @link http://www.mantisbt.org
  *
  * @uses access_api.php
+ * @uses menu_api.php
  * @uses utility_api.php
  */
 
 
 require_api( 'access_api.php' );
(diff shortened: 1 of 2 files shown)


raspopov:
These changes weren't really necessary other than for the Sitemap PR; the necessary changes were made directly in #2111.

PR #2128, opened 2025-03-20
Fix dropzone extra page load

raspopov:
Regression in 78cf25f2a5b43b5b8ac7d1c283bd1d4281c3c91b.

Fixes [#35644](https://mantisbt.org/bugs/view.php?id=35644).


dregad (approved):

dregad:
> Regression in 78cf25f2a5b43b5b8ac7d1c283bd1d4281c3c91b.

Be careful when you reference commits - this SHA is from your original branch, but as I rebased it before merge, it does not actually exist in the mantisbt repo - the real commit is 10971ed15a2f8476130c996f54dda6dc068828f9.

PR #2129, opened 2025-03-20
Fix collapsible widgets

raspopov:
One new fix and two regression fixes in 78cf25f2a5b43b5b8ac7d1c283bd1d4281c3c91b.

Fixes [#35645](https://mantisbt.org/bugs/view.php?id=35645).


dregad (approved):

dregad (approved):
Looking good, although I must admit I did not test every single collapse scenario.

PR #2131, opened 2025-03-21
Show GitHub Actions build status in README.md

atrol:
Fixes [#35647](https://mantisbt.org/bugs/view.php?id=35647)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Mantis Bug Tracker (MantisBT)
 =============================
 
-[![Build Status](https://img.shields.io/travis/com/mantisbt/mantisbt/master?logo=travis)](https://app.travis-ci.com/mantisbt/mantisbt)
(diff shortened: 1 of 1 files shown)


dregad (approved):

dregad:
Previously the badge was clickable, opening the latest build on TravisCI site. With this change it is no longer the case, i.e. the badge is just an image. I think it would be nice to restore the link.

Problem is, AFAIK Github does not have a way to link to the latest build, but we could at least point to https://github.com/mantisbt/mantisbt/actions/workflows/mantisbt.yml to show the list of workflow runs for unit tests.

atrol:
> Previously the badge was clickable, opening the latest build on TravisCI site. With this change it is no longer the case, i.e. the badge is just an image. I think it would be nice to restore the link.

Added proposed link

PR #2133, opened 2025-03-27
Updated mention_api.php for issue 0025698

MarcoLodini:
Edited pattern to allow usage of email addresses in tags.

Fixes [0025698](https://mantisbt.org/bugs/view.php?id=25698).

### DIFF
--- a/core/mention_api.php
+++ b/core/mention_api.php
@@ -65,21 +65,19 @@ function mention_get_candidates( $p_text ) {
 	static $s_pattern = null;
 	if( $s_pattern === null ) {
 		$t_quoted_tag = preg_quote( mentions_tag() );
-		$s_pattern = '/(?:'
-			# Negative lookbehind to ensure we have whitespace or start of
-			# string before the tag - ensures we don't match a tag in the
-			# middle of a word (e.g. e-mail address)
-			. '(?<=^|[^\w])'
-			# Negative lookbehind  to ensure we don't match multiple tags
-			. '(?<!' . $t_quoted_tag . ')' . $t_quoted_tag
-			. ')'
-			# any word char, dash or period, must end with word char
-			. '([\w\-.]*[\w])'
-			# Lookforward to ensure next char is not a valid mention char or
-			# the end of the string, or the mention tag
-			. '(?=[^\w@]|$)'
-			. '(?!$t_quoted_tag)'
-			. '/';
+		$s_pattern = '/'
+			# Positive lookbehind ensuring the tag symbol is either at the start of the string or preceded by a non-word character
+			. '(?:(?<=^|[^\w])'
+			# Negative lookbehind ensuring the tag symbol is not preceded by another one
+			. '(?<!' . $t_quoted_tag . ')'
+			# Matches the literal tag character that starts our pattern
+			. $t_quoted_tag . ')'
+			# First capture group; ensures also that the username ends with a word character
+			. '([\w\-.]*[\w]'
+			# Email domain portion; optionally captures it if present, allowing usage of email addresses
+			. '(?:@[\w\-.]+(?:\.[a-z]{2,})?)?)'
+			# Positive lookahead for ending boundary
+			. '(?=\s|[^\w.]|$)'
 	}
 
 	preg_match_all( $s_pattern, $p_text, $t_mentions );


dregad (changes requested):
Hello @MarcoLodini. 

Thanks for your contribution, but unfortunately this breaks the unit tests.

Please fix your updated regex so that it does not introduce any regression. If you believe that the test suite needs to be updated, then please explain why, and include the necessary changes as part of this PR.

Also, kindly avoid long lines, the updated comments should generally be wrapped around 80 chars as in the original code, but in any case no more than 120.

MarcoLodini:
Hi @dregad
Thank you for the information; I have updated the mention files to keep the comments under the specified number of cols (should be 80), and updated the test suite.

The changes in the test suite involve expecting the `@` symbol to be treated as part of the username, hence looking up the user with the matching username.

With this in mind, `vboctor@localhost` and `vboctor@example.com` are both valid examples, regardless of them being actual email addresses or not.

Let me know is something is amiss or requires clarifications.
Thanks!

MarcoLodini:
Ok, should be fixed now

dregad (changes requested):
Still failing unit tests, see review comments.

The regex for email is not accepting `+` and other special characters allowed by the default value for *$g_user_login_valid_regex* https://github.com/mantisbt/mantisbt/blob/ee85879e9b970a170a344c8b59f38334549af3d7/config_defaults_inc.php#L3781 

Ideally, user mentions parsing should match whatever is configured for this, or at the very least be as close as possible to the default value (and before you ask, yes I'm aware that we currently allow space in there, this has to be change at some point).

Do we need additional test cases for email address mentions used in context (i.e. with text before/after, wrapped in special chars `<user@example.com>`, etc) ? Also what happens with a URL including basic auth `https://user:password@www.example.com/` ?

tests/Mantis/MentionParsingTest.php
```diff
@@ -144,11 +144,23 @@ public static function provider() {
...
-				array()
+				array('vboctor@example.com')
 			),
 			'MentionUsernameThatIsLocalhost' => array(
 				"@vboctor@localhost",
-				array()
+				array('vboctor@localhost')
+			),
+			'MentionUsernameThatIsAnEmailAddressWithInvalidChars' => array(
+				"@vboctor@example.com/path",
```
vboctor, line 154: I'm not we should treat this as a proper mention. The `/` is not a punctuation.

vboctor (commented):
Thanks @MarcoLodini for your contribution.

I haven't reviewed the regex, but based on the test cases, I provided a case that seems odd.

However, my higher level comment independent of this PR, is that we should be more descriptive about what makes a proper username. At the moment, we are super flexible with what is a valid username, which is problematic when it comes to features like @ mention, show email address threshold, etc.

1. Is email address a valid username? Often services will allow signing up with an email address, but will have a username that is still distinct based on first + last name or a user specified username.

2. Should we allow spaces? If so, would @ mentions work? What about punctuation characters, etc?

vboctor:
Here an email from a user that got surprised that their email address was exposed after signing up to MantisBT Official Bugtracker and requested a change of email address. I hid the email address and name for privacy reasons.

image

dregad:
@vboctor 
> we should be more descriptive about what makes a proper username. At the moment, we are super flexible about what makes a valid username, which is problematic when it comes to features like @ mention, show email address threshold, etc.

Agreed. This is why I said we should remove the space in https://github.com/mantisbt/mantisbt/pull/2133#pullrequestreview-2797605895

> 1. Is the email address a valid username? Often, services will allow signing up with an email address but will have a username that is still distinct based on first + last name or a user-specified username.

Using an email address as a username is a feature you introduced back in 2026, see https://mantisbt.org/bugs/view.php?id=20543 / PR #712. I personally don't believe it's such a good thing to do for internet-facing trackers, for privacy reasons, and based on https://github.com/mantisbt/mantisbt/pull/2133#issuecomment-2833657318, looks like I'm not the only one...

> 2. Should we allow spaces? If so, would @ mentions work? What about punctuation characters, etc?

No, No, and it depends on which punctuation chars.

> 3. Does it make sense to have show email address threshold DEVELOPER while usernames are email addresses for some users that decided to sign up this way?

Probably not, but as mentioned before, using email address as user id should probably be restricted as mentioned above.

> 4. Do users that signup with an email address in a public instances like mantisbt.org/bugs, expect that when they report a bug, there email address would be exposed?

Apparently not...


@MarcoLodini all of the above is a higher-level discussion, and you do not need to take this into account for this PR. But you do need to address the other review comments, and make sure your branch passes the unit tests.

tests/Mantis/MentionParsingTest.php
```diff
@@ -128,11 +128,11 @@ public static function provider() {
 			),
 			'MentionWithEmailAddress' => array(
 				'xxx@example.com',
-				array()
+				array('xxx@example.com')
```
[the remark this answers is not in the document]
MarcoLodini: @dregad I was just thinking that email addresses being accepted would be a "side-effect" of allowing @ to be used in the first place: that's why I modified the test suite to include `xxx@example.com` as a valid example, but that would mean I could also accept `myname@whynot` as a valid username, much like allowing `user@domain.com`.

I would consider all of these options to be alright, and this PR would be more of a "allowing @ as a character in a username mention" more than "allowing specifically email addresses to work".

What is your opinion on this? Shall we keep it strictly for emails or can we expand the scope?
Thanks!

tests/Mantis/MentionParsingTest.php
MarcoLodini: Hi @vboctor , I agree with you; I wanted to test whether `/` was considered by the regex as part of the username, but I would expect the test to pass with `vboctor@example.com` being returned in the array, without `/path`.
Is that correct?
Thank you!

MarcoLodini: As a similar example for the logic behind the test, I would refer to `MentionWithMultipleSlashSeparated`

PR #2136, opened 2025-04-21
Fix REST API not allowing to unset the Category when updating an Issue

dregad:
REST API was not allowing to unset the Category when updating an Issue
with $g_allow_no_category = ON with a payload `{"category": {"id": 0}}`.

Regression in mci_get_category_id() function, introduced by commit
656f93eebaf033f4b5bf34f17487db6346f12d8a (Issue [#34683](https://mantisbt.org/bugs/view.php?id=34683)).

Fixes [#35668](https://mantisbt.org/bugs/view.php?id=35668)


dregad:
This will be back-ported to master-2.27 branch.

tests/core/MantisTestCase.php
```diff
@@ -64,4 +64,35 @@ function getTestName(): string {
...
+
+	/**
+	 * Restores a configuration to its initial state.
+	 *
+	 * @param string $p_config Configuration option name.
+	 * @param mixed  $p_value  Configuration option value. If false, the config
+	 *                         will be deleted.
+	 */
+	public function restoreConfig( string $p_config, $p_value ) {
+		if( $p_value === false ) {
```
vboctor, line 91: Isn't it the case that some configs may be enabled by default? Hence, false doesn't mean deletion?

tests/rest/RestIssueUpdateCategory.php
```diff
@@ -0,0 +1,199 @@
...
+ * @subpackage UnitTests
+ * @copyright  Copyright 2025 MantisBT Team - mantisbt-dev@lists.sourceforge.net
+ * @link       https://mantisbt.org
+ */
+
+namespace Mantis\tests\rest;
+
+use Psr\Http\Message\ResponseInterface;
+
+class RestIssueUpdateCategory extends RestBase
```
vboctor, line 30: Add phpdoc for member attributes and methods.

tests/rest/RestIssueUpdateCategory.php
```diff
@@ -0,0 +1,199 @@
...
+			$t_name = 'cat-' . rand();
+		} while( !category_is_unique( $this->getProjectId(), $t_name ) );
+		return $t_name;
+	}
+
+	private function updateCategory( int $p_issue_id, $p_category ): ResponseInterface {
+		$t_payload = ['category' => $p_category];
+		return $this->builder()->patch( "/issues/$p_issue_id", $t_payload )->send();
+	}
+
```
vboctor, line 71: remove extra blank line.

vboctor (approved):

tests/core/MantisTestCase.php
dregad: Technically it is possible if a config value is defined as boolean false. The logic relies on our use ON/OFF constants in the config files, corresponding to *integer* values 0/1.

IMO considering the limited use cases for this, it was not worth implementing something more complex for a fully general usage.

PR #2137, opened 2025-04-23
Fix scroll offset for bugnotes when using anchor with fixed navbar.

matejbrodziansky:
Fixes issue where bugnote anchors are hidden behind fixed navbar by adding scroll-margin-top.

[#21675](https://mantisbt.org/bugs/view.php?id=21675)

Snímka obrazovky 2025-04-23 o 13 23 41

### DIFF
--- a/css/default.css
+++ b/css/default.css
@@ -52,6 +52,9 @@ span.dependency_upgrade		{ color: orange; }
 
 tr.bugnote .bugnote-note { background-color: #e8e8e8; color: #000000; width: 75%; vertical-align: top; }
 
+tr.bugnote {
+  scroll-margin-top: 50px;
+}
 .bugnote-private { background-color: #fcf8e3 !important;}
 
 .nowrap


dregad (changes requested):
Thanks for your contribution. 

Please open an Issue in our bugtracker at https://mantisbt.org/bugs describing the problem you're trying to fix, and reference it in the PR description and the commit message (`Fixes #xxx`).

atrol:
@dregad I suspect we can use  [Incorrect positioning of "View Issue Details" when recalled from "Direct link to note"](https://mantisbt.org/bugs/view.php?id=21675) for it.

matejbrodziansky:
@atrol yes, that's the problem. :)

dregad:
That's a nice CSS trick, I did not know about this `scroll-margin-top` property.

This is something we should consider applying to other places where we use anchor tags, e.g. *Jump to Notes* and *Jump to History* on View Issue page

dregad:
Superseded by #2140

PR #2146, opened 2025-08-24
Ensured relationship reverse is used for destination issue

vboctor:
Fixes [#36278](https://mantisbt.org/bugs/view.php?id=36278)


dregad:
Target 2.27.2 ?

PR #2147, opened 2025-08-25
Fix / improve test email admin functionality

vboctor:
Fixes [#35288](https://mantisbt.org/bugs/view.php?id=35288)


vboctor:
I didn't do it in EmailData definition since the initialization logic can load configs and I felt it is better not to have it in the EmailData object.

PR #2162, opened 2025-11-15
Reorder actions on View Issues page

atrol:
Fixes [#34649](https://mantisbt.org/bugs/view.php?id=34649)


dregad (approved):
Totally forgot about this... Thanks for preparing the PR

PR #2165, opened 2025-12-16
Add reproducibility filter

DavidSribouavong:
PR for the issue 
https://mantisbt.org/bugs/view.php?id=6523

### DIFF
--- a/core/classes/BugFilterQuery.class.php
+++ b/core/classes/BugFilterQuery.class.php
@@ -452,7 +452,10 @@ class BugFilterQuery extends DbQuery {
 				case FILTER_PROPERTY_SEVERITY:
 					$this->build_prop_severity();
 					break;
-				case FILTER_PROPERTY_RESOLUTION:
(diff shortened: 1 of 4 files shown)


dregad (changes requested):
Thanks for your contribution. 

I tested and functionally this is OK. 

Please make the following changes:
1. Use tabs for code indentation
2. The bugtracker issue must be referenced in the commit message (add `Fixes #6523`, amend commit)

Minor layout issue: with the default setting for $g_filter_custom_fields_per_row (8 columns) the first line of filter items wraps, causing the last one to be displayed alone on row 2. Maybe we should reorganize filter items, or change the default value to 9. I have to think about it.

dregad:
> This discussion is diverging from the PR topic though. Maybe we should open a separate Issue to follow up on this ?

I have created [#37131](https://mantisbt.org/bugs/view.php?id=37131) to track the reorganization of filters.

dregad:
Merged in 1d99e3e8d9d9bc5f3984e2f74f17f8153e81abc1

PR #2176, opened 2026-02-02
Optimize the verification of uploaded files.

raspopov:
- Verification code is collected in one place inside the `file_ensure_uploaded()` function.
- Added checks to the `IssueFileAddCommand` and `IssueNoteAddCommand` commands.
- Extended checking for web server upload errors.
- Tests for the `file_ensure_uploaded()` function have been added.

Fixes [#36861](https://mantisbt.org/bugs/view.php?id=36861).

### DIFF
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -1221,13 +1221,13 @@ function file_ensure_uploaded( array $p_file ) {
 
 	if( empty( $t_tmp_file ) ) {
 		throw new ClientException(
-			'File path is empty',
+			"File '$t_file_name' path is empty",
 			ERROR_FILE_NO_UPLOAD_FAILURE );
 	}
 
 	if( !is_readable( $t_tmp_file ) ) {
 		throw new ClientException(
-			'File is not readable',
+			"File '$t_file_name' is not readable",
 			ERROR_UPLOAD_FAILURE );
 	}
 
--- a/core/file_api.php
+++ b/core/file_api.php
@@ -904,10 +904,6 @@ function file_is_name_unique( $p_name, $p_bug_id, $p_table = 'bug' ) {
 function file_add( $p_bug_id, array $p_file, $p_table = 'bug', $p_title = '', $p_desc = '', $p_user_id = null, $p_date_added = 0, $p_skip_bug_update = false, $p_bugnote_id = 0 ) {
 	$t_file_info = array();
 
-	if( !isset( $p_file['error'] ) ) {
-		$p_file['error'] = UPLOAD_ERR_OK;
-	}
-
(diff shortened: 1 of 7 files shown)


core/file_api.php
```diff
@@ -1205,28 +1171,78 @@ function file_ensure_valid_upload_path( $p_upload_path ) {
...
+			"File '$t_file_name' type not allowed",
+			ERROR_FILE_NOT_ALLOWED
+		);
+	}
+
+	$t_tmp_file = $p_file['tmp_name'] ?? null;
+
+	if( empty( $t_tmp_file ) ) {
+		throw new ClientException(
+			'File path is empty',
```
vboctor: Include `$t_file_name`.

core/file_api.php
```diff
@@ -1205,28 +1171,78 @@ function file_ensure_valid_upload_path( $p_upload_path ) {
...
-			throw new ClientException(
-				sprintf( "File '%s' too big", $p_file['name'] ),
-				ERROR_FILE_TOO_BIG );
 
-		case UPLOAD_ERR_PARTIAL:
-		case UPLOAD_ERR_NO_FILE:
-			throw new ClientException(
-				sprintf( "File '%s' upload failure", $p_file['name'] ),
-				ERROR_FILE_NO_UPLOAD_FAILURE );
+	antispam_check();
```
vboctor: I'm wondering if this may cause an issue.

- You run validate, and all is good.
- You go ahead with the processing, create an issue, and then add the files, and midway, you get an anti-spam error.

vboctor (changes requested):
Thanks @raspopov - nice refactoring.

proj_doc_update.php
```diff
@@ -80,10 +80,6 @@
 	file_ensure_uploaded( $f_file );
 
 	$t_file_name = $f_file['name'];
-	if( strlen( $t_file_name ) > DB_FIELD_SIZE_FILENAME ) {
```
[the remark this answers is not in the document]
raspopov: I moved the `DB_FIELD_SIZE_FILENAME` check above inside `file_ensure_uploaded( $f_file );`.

core/file_api.php
raspopov: Yes, that's exactly how the old code worked. In the new code, this check is performed before modifying the database. If the first check passes, the second one will also pass since there are almost no changes in history between them. If the second check fails, it means the user has uploaded something to the database rapidly and in bulk — probably a spammer.

dregad (approved):
LGTM

dregad:
Merged in b7741aa561bc1632ec24231e5b17dcca3c1e67ab

PR #2184, opened 2026-02-09
Fix the synchronization of mail and real name with LDAP.

raspopov:
- The update of the email address and real name to the database from LDAP was added when a user is created.
- Corrected the issue of erroneously clearing the email and real name from the database when updating a non-existent user in LDAP.

Fixes [#36873](https://mantisbt.org/bugs/view.php?id=36873).

Likely needs PR  #1308, as there is a shared file. 🤔


dregad:
Merged in https://github.com/mantisbt/mantisbt/pull/2184/commits/fed92e344cb6db6ce9b462a6900802c53e799f4b

PR #2188, opened 2026-02-17
Let bugnotes be moved to a different issue

raspopov:
The transfer is performed by relinking the records in the database to the new issue. Rights are verified as if the bugnote had been deleted from the old issue and added to the new one. The change history was added. SOAP and REST access are not implemented. The design of the editing form has been slightly modified.

Fixes [#5688](https://mantisbt.org/bugs/view.php?id=5688).


vboctor (changes requested):
For this to be done properly, it adds a bunch of complexity:

- Access validation for notes, attachments, time tracking, etc.
- APIs to enable such functionality.
- UI to enable such action and errors to explain why it doesn't work
- Tests to validate it works as expected.
- The access of the author vs. the person moving.
- History events explaining the move -- clarifies why person X who has no access to add notes to project Y has a note there.

As mentioned on the [issue](https://mantisbt.org/bugs/view.php?id=5688#c41907) - I believe this should be closed as won't fix.

PR #2190, opened 2026-03-01
Add link to issue notes in timeline.

raspopov:
Fixed wrong format of bugnote anchor link (mistakenly left-side padded by zeroes) in `string_get_bugnote_view_url()` function.

Also added bug and bugnote link tag symbols.

Fixes [#23220](https://mantisbt.org/bugs/view.php?id=23220).

### DIFF
--- a/core/classes/IssueNoteCreatedTimelineEvent.class.php
+++ b/core/classes/IssueNoteCreatedTimelineEvent.class.php
@@ -50,7 +50,7 @@ class IssueNoteCreatedTimelineEvent extends TimelineEvent {
 	 */
 	public function html() {
 		$t_html = $this->html_start( 'fa-comment-o' );
-		$t_html .= '<div class="action">' . sprintf( lang_get( 'timeline_issue_note_created' ), prepare_user_name( $this->user_id ), string_get_bug_view_link( $this->issue_id ) ) . '</div>';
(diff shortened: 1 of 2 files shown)


dregad (changes requested):
While having the ability to jump straight to a bugnote from timeline is a nice and useful feature, I'm not a big fan of the way it looks.

image

<p>

I propose a couple of variants:

1. Add the link to bugnote on *commented* - looks good but probably painful to implement with localisation:  
  [tadm](http://localhost/mantis/view_user_page.php?id=2) [commented](http://localhost/mantis/view.php?id=42370#c70326) on issue [0042370](http://localhost/mantis/view.php?id=42370)
2. Reword the string to have a separate link for the bugnote   
  [tadm](http://localhost/mantis/view_user_page.php?id=2) added note [~0070326](http://localhost/mantis/view.php?id=42370#c70326) to issue [0042370](http://localhost/mantis/view.php?id=42370)

The change to string_api.php systematically adding of the `#` / `~` to the links is unrelated to this work item and should be moved to a separate PR. Maybe the other devs will think differently, but I must say I personally don't like this at all - IMO it adds clutter and I don't see the benefits. 

Furthermore, it breaks the hyperlinks when bugnote text is rendered. The following note
```
Bug #42370
Bugnote ~0070326
```
Displays like that with this PR:
image

Compare to original (master branch):
image

PR #2193, opened 2026-03-11
Add a check for the presence of the admin folder.

raspopov:
The check is performed from the browser by attempting to load the "/admin/install.php" and "/admin/check/" pages and checking for a 4xx error code is returned or if a redirect to the login page occurs.

Fixes [#23211](https://mantisbt.org/bugs/view.php?id=23211).

### DIFF
--- a/js/common.js
+++ b/js/common.js
@@ -657,6 +657,27 @@ $(document).ready( function() {
 		bottom_placeholder.css('height', parseInt(bottom.height()) + 'px');
 		footer.css('padding-top', parseInt(footer_content.height()) + 'px');
 		}).triggerHandler('resize.footer');
+
+	/**
+	 * Admin directory on-line check (login_page.php)
+	 */
+	$('#warning_admin_directory_present').each(function() {
+		var spinner = $('<i class="fa fa-spin fa-spinner margin-left-8"></i>');
+		$(this).append(spinner);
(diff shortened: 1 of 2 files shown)


dregad (changes requested):
This does not work - accessing /admin returns an _Access Denied_ page (which does not contain the search string) so the warning is not shown, yet the admin pages are accessible (particularly, install.php and admin/checks which do not have an access_ensure_global_level() check).

raspopov:
> This does not work - accessing /admin returns an _Access Denied_ page (which does not contain the search string) so the warning is not shown

That's the right response. The issue is precisely that when access is effectively blocked, there shouldn't be a warning.

dregad:
I'm afraid you're missing the point.

The reason we recommend admin/ directory to be deleted, is because access to admin/check and install.php can lead to system information disclosure and quite a few nastier things.

In your PR's current state, since the check gets a default *Access Denied* page that does not contain the search string, it happily thinks that the directory is not accessible (when in fact it is), does not show any warning to the admin, but in fact the sensitive pages can be accessed.

raspopov:
Thank you for your patience. I finally realized that need to check the files without `access_ensure_global_level()` as well, i.e., “admin/install.php”, “admin/check”...

PR #2200, opened 2026-04-04
Deleting a user now deletes its filters

vboctor:
Fixes #37004

### DIFF
--- a/admin/schema.php
+++ b/admin/schema.php
@@ -904,6 +904,11 @@ $g_upgrade[213] = array( 'UpdateFunction', 'category_status_default' );
 
 # Release marker: 2.27.0
 
+# Release marker: 2.28.0
+
+# Delete orphaned private filters owned by users that no longer exist
+$g_upgrade[214] = array( 'UpdateFunction', 'delete_orphaned_private_filters' );
+
 # ----------------------------------------------------------------------------
 # End of schema definition, clear local variables
 #
--- a/core/filter_api.php
+++ b/core/filter_api.php
@@ -1781,17 +1781,24 @@ function filter_db_delete_current_filters() {
 }
 
 /**
- * Delete all filters owned by the specified user.
+ * Delete filters owned by the specified user.
  *
  * This function is called when a user is deleted to clean up orphaned filters.
+ * By default, shared (public) filters are preserved since they may be used by other users.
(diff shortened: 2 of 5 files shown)


dregad (commented):
Code looks good, but please read my note on the tracker  https://mantisbt.org/bugs/view.php?id=37004#c70941

vboctor:
@dregad @atrol feedback addressed.

dregad:
All the PostgreSQL builds are failing, and I am not sure why as I see nothing in the PR that could cause a problem.

The logs are not really giving a clear error either.
- PHP 7.4 - 8.3 show an incomplete output at *MantisBT installation* and errors out at *Post-installation steps*
- PHP 8.4 and 8.5 show a database error
    ```
    Database query failed. Error received from database was #0:  for the query:  
    ```

It looks like the DB is not properly setup, but why ???

There is a warning in job execution
> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4.1.0, shivammathur/setup-php@2.29.0.

Maybe that's what is causing the problem. I'll try to update the actions, see what happens.

dregad:
Updating actions fixes the warning, but the pgsql tests still fail.

dregad:
I nailed it - the culprit is the use of (deprecated) *db_prepare_bool()* function. 

It calls *db_is_pgsql()*, which checks the *$g_db_functional_type* global, which is initialised as part of core based on the value of *$g_db_type*. The problem is that at this time we do not yet have a config_inc.php file, so the variable has its default value of `mysqli`.

As a result, the query is initialised to `DELETE FROM {filters} WHERE is_public = 0 AND user_id NOT IN (SELECT id FROM {user})` instead of `WHERE is_public = false` as expected for pgsql, resulting in a query execution error which is not properly caught by the installer: *ERROR:  operator does not exist: boolean = integer*.

dregad:
Using a query parameter instead of db_prepare_bool() fixes the problem.

See the test build in my fork with https://github.com/dregad/mantisbt/actions/runs/24270936335

dregad (approved):

dregad:
Merged in 97292ce6a7b4caf1bfc36111b640887c086e4505.

PR #2204, opened 2026-04-10
Fix API error when summary is too long

vboctor:
Fixes [#37008](https://mantisbt.org/bugs/view.php?id=37008)


tests/rest/RestIssueTest.php
```diff
@@ -143,6 +143,28 @@ public function testCreateIssueWithLongText() {
...
 
+	/**
+	 * Tests that creating an issue with a summary longer than the DB field size
+	 * fails with a field-too-long error.
+	 *
+	 * @return void
+	 */
+	public function testCreateIssueWithTooLongSummary() {
+		$t_issue_to_add = $this->getIssueToAdd( 'summary-too-long' );
+		$t_issue_to_add['summary'] = str_repeat( 'x', DB_FIELD_SIZE_BUG_SUMMARY + 1 );
```
dregad: Just wondering, shouldn't this be tested with multibyte chars as well ?

I'm thinking in particular of legacy MySQL instances that may still use the *utf8mb3* character set (3 bytes), where this might cause issues (just a hunch, did not make any tests).

vboctor: Updated the test to use multi-byte characters and it works.

dregad (approved):
A bit of code duplication in the unit tests, but I guess we can live with that.

dregad:
@vboctor In the interest of keeping the master branch's history as linear as possible, in the future it would be best if you could rebase your feature branch before merging it; alternatively, use the *squash&merge* or *rebase* options as appropriate.

PR #2205, opened 2026-04-11
Fix REST API user self deletion + update web UI to use REST API

vboctor:
Fixes [#37012](https://mantisbt.org/bugs/view.php?id=37012)


dregad (approved):
Looks good. Did not test, just code review.

PR #2206, opened 2026-04-12
MantisCoreFormatting: fix [*] corruption by Parsedown emphasis engine

bcp-software:
When Markdown processing is active, Parsedown corrupts BBCode [*] list item markers. Parsedown's inline scanner triggers inlineLink on every "[", which fails for "[*]" (no URL follows), then backtracks and re-scans the bare "*" as an emphasis delimiter. It pairs that "*" with the "*" from the next "[*]" item in the list, creating an <em> span that crosses the item boundary:

  [*]**item 1**  ->  [<em>]<strong>item 1</strong>
  [*]**item 2**  ->  [</em>]<strong>item 2</strong>

This leaves the bullet markers destroyed and any inline formatting inside the items potentially corrupted.

Fix: override inlineLink() to detect "[*]" at the start of the excerpt and return it immediately as a literal 3-character markup token. The "extent => 3" advances Parsedown past all three characters so the emphasis engine never sees the "*".

This affects any use of BBCode [*] list items alongside Markdown, not just when used with the BBCodePlus plugin.

### DIFF
--- a/plugins/MantisCoreFormatting/core/MantisMarkdown.php
+++ b/plugins/MantisCoreFormatting/core/MantisMarkdown.php
@@ -214,15 +214,42 @@ class MantisMarkdown extends Parsedown
 	}
 
 	/**
-	 * Add link attributes to element data.
+	 * Add link attributes to element data, with BBCode list item protection.
 	 *
 	 * - [link](http://example.com)
+	 *
+	 * Overrides parent to protect the BBCode list item marker [*] from being
+	 * consumed by Parsedown's emphasis engine.
+	 *
+	 * When Parsedown's inline scanner encounters "[", it calls inlineLink.
+	 * For "[*]", the link parse fails (no URL after "]"), and Parsedown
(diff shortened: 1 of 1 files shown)


dregad:
Thanks for your contribution. I see several issues here

1. "out of the blue" pull requests are not acceptable. A PR must target an Issue in our bug tracker at https://mantisbt.org/bugs, and the issue number must be referenced in the commit message (i.e. `Fixes #xxxx`).
2. It does not make sense to have 2 distinct markup processors running at the same time, it is bound to cause problems. Incidentally, this is evidenced by this PR's description, which has messed up formatting due to GitHub's own Markdown processor processing BBCode markup. I suggest disabling Markdown in MantisCoreFormatting options if you chose to use  BBCode.
3. BBCode formatting is a functionality handled by a 3rd-party plugin (I'm assuming you're using [BBCodePlus](https://github.com/mantisbt-plugins/BBCodePlus). It is up to the the plugin to manage this somehow, we should not be altering core for it.

As it stands, I see no reason to merge this change.

PR #2207, opened 2026-04-12
Fix extra require_once() calls from require_api()

raspopov:
To prevent recursive calls to the `require_api()` function, move the loading flag after `require_once()` to before it.

Fixes [#37018](https://mantisbt.org/bugs/view.php?id=37018).


dregad (approved):
Looks good. Ran a few local tests, did not notice any regressions. I assume you did the same

dregad:
Merged in https://github.com/mantisbt/mantisbt/commit/6b0a01986188ac4182f51d2e05415e0cfe746134

PR #2208, opened 2026-04-15
Fix an invalid array index error for unknown issue.

raspopov:
Fixes [#37022](https://mantisbt.org/bugs/view.php?id=37022).


dregad:
Merged in 1cc10355464a61e05b2a7736e455c1fce227facd

PR #2209, opened 2026-04-25
Fix debug log backtrace for Windows

raspopov:
Absolute paths and problematic path separators have been removed, since function and file names are already sufficiently unique, and the location of MantisBT files can be changed via the configuration. The text has also been formatted consistently.

Fixes [#37097](https://mantisbt.org/bugs/view.php?id=37097).


dregad (approved):
Looks good to me.

PR #2210, opened 2026-04-25
Add a cache for tokens retrieved from the database

raspopov:
A cache of tokens retrieved from the database has been added as a new global variable, `$g_cache_token`.

The chosen cache format simplifies the process of obtaining a token and storing it in the cache, but makes it more complicated to verify the token ID; however, in practice, this does not reduce the cache's efficiency due to the optimal order of function calls in MantisBT.

In addition, the code explicitly casts token identifiers to integers throughout.

Fixes [#37098](https://mantisbt.org/bugs/view.php?id=37098).

### DIFF
--- a/tests/Mantis/TokensApiTest.php
+++ b/tests/Mantis/TokensApiTest.php
@@ -71,7 +71,7 @@ final class TokensApiTest extends MantisCoreBase {
 			db_query( 'INSERT INTO {tokens} ( owner, type, value, timestamp, expiry ) VALUES ( '
 				. db_param() . ', ' . db_param() . ', ' . db_param() . ', ' . db_param() . ', ' . db_param() . ' )'
 				, $t_token );
-			$t_token = array_merge( [ db_insert_id( 'tokens' ) ], $t_token );
+			$t_token = array_merge( [ db_insert_id( db_get_table( 'tokens' ) ) ], $t_token );
 		}
 		unset( $t_token );
 	}
--- a/tests/Mantis/TokensApiTest.php
+++ b/tests/Mantis/TokensApiTest.php
@@ -1,0 +1,452 @@
+<?php declare(strict_types=1);
+# MantisBT - A PHP based bugtracking system
+
+# MantisBT is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# MantisBT is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You must have received a copy of the GNU General Public License
+# along with MantisBT.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * Test cases for Tokens API within mantis
+ *
+ * Cache usage is checked by counting the number of database queries using
+ * the db_count_queries() function.
+ *
+ * @package    Tests
+ * @subpackage TokensAPI
+ * @copyright  Copyright 2026  MantisBT Team - mantisbt-dev@lists.sourceforge.net
+ * @link       https://www.mantisbt.org
+ */
+
+namespace Mantis\tests\Mantis;
+
+/**
+ * PHPUnit tests for Tokens API
+ */
+final class TokensApiTest extends MantisCoreBase {
+
+	protected const UNKNOWN = -1; # a guaranteed unknown value for an owner, type, or id.
+	protected const KNOWN_OWNER = 1234;
+	
+	protected const ROW_ID = 0;
+	protected const ROW_OWNER = 1;
+	protected const ROW_TYPE = 2;
+	protected const ROW_VALUE = 3;
+	protected const ROW_TIMESTAMP = 4;
+	protected const ROW_EXPIRY = 5;
+
+	/**
+	 * Test tokens inserted in the database
+	 */
+	protected array $tokens;
+
+	/**
+	 * Prepare tokens for tests
+	 * @return void
+	 */
+	protected function setUp(): void {
+		global $g_cache_token, $g_tokens_purged;
+		$g_cache_token = []; # clear cache
+		$g_tokens_purged = true; # disable auto-purging
+
+		# The order of the array keys must match the order of the rows in the database schema
+		$t_now = db_now();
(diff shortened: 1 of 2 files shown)


dregad (commented):
@raspopov Thanks for your contribution !

Question regarding the foreach loop in token_touch() / token_update():

While I don't think it's technically required in this case, I understand the rationale for calling unset() on a byref foreach loop variable as good practice to avoid unexpected side effects later.

However, this is generally done *after* the foreach loop, so I'm curious why you do it within. Can you explain ?

Other than that, the PR is looking pretty good overall, but I still need to test a bit.

raspopov:
> While I don't think it's technically required in this case, I understand the rationale for calling unset() on a byref foreach loop variable as good practice to avoid unexpected side effects later.

The PHP documentation states: "Reference of a $value and the last array element remain even after the foreach loop. It is recommended to destroy it by unset()." However, I kept running into errors with the remaining references, and since they were very difficult to debug, I decided to limit the lifetime of the references as much as possible and break the connection as soon as it was no longer needed. A type of PHP PTSD. 🫡

dregad:
> I kept running into errors with the remaining references

Possibly caused by using nested loops both using references ? I don't think I've ever done that.

> limit the lifetime of the references as much as possible and break the connection as soon as it was no longer needed. 

This is somewhat academical and maybe not worth the effort, but it still feels strange to unset within the loop. Did you try to unset right after loop, like this?

```php
foreach( $g_cache_token as & $t_cache_type ) {
    foreach( $t_cache_type as & $t_cache_owner ) {
        if( isset( $t_cache_owner['id'] ) && $t_cache_owner['id'] == $c_token_id ) {
            $t_cache_owner['expiry'] = $c_token_expiry;
        }
    }
    unset( $t_cache_owner );
}
unset( $t_cache_type );
```

raspopov:
> Did you try to unset right after loop, like this?

This approach works, too.

Unfortunately, 😔 I no longer remember what caused the error. It occurred in an early version of the patch around a month ago. The global cache variable was receiving invalid keys (with tokens) that were supposed to be deleted, but which were mistakenly restored due to the incorrect use of references after loops. There’s a three-level chain of references here: `global` and two nested loops.

In its current form, we could probably remove the `unset`s; its no longer affect the result. However, I would still like to leave them in as a '⚠️ beware of mines' warning.

dregad:
> I would still like to leave them in as a '⚠️ beware of mines' warning.

I'm cool with that. But if it does not break stuff I'd rather have them out of their respective loops for efficiency (it's a performance optimisation PR after all 😉).

raspopov:
It's not entirely clear what PostgreSQL didn't like... 🤔

tests/Mantis/TokensApiTest.php
```diff
@@ -0,0 +1,452 @@
...
+			[ self::KNOWN_OWNER, TOKEN_UNKNOWN, 'expired_unknown', $t_now, $t_now - 1 ],
+		];
+
+		# Insert test tokens
+		foreach( $this->tokens as & $t_token ) {
+			db_param_push();
+			db_query( 'INSERT INTO {tokens} ( owner, type, value, timestamp, expiry ) VALUES ( '
+				. db_param() . ', ' . db_param() . ', ' . db_param() . ', ' . db_param() . ', ' . db_param() . ' )'
+				, $t_token );
+			$t_token = array_merge( [ db_insert_id( 'tokens' ) ], $t_token );
```
dregad: db_insert_id() expects a valid table name (i.e. it's actual name in the schema) so you need to give it `db_get_table( 'tokens' )` -> `mantis_tokens_table` with default prefix/suffix.

That being said (looking at this code for the first time in nearly 10 years), I'm wondering why we have PostgreSQL-specific logic here - if I'm not mistaken the ADOdb pgsql driver does support *Insert_ID()*... Need to check.

raspopov: > db_insert_id() expects a valid table name

Yes, that's true — it's incredible, but it works perfectly for MySQL and SQLite 3...

raspopov:
This is getting even more interesting... I'm surprised that PostgreSQL returns a string instead of a number...
(The ‘expected’ and ‘actual’ parameters are swapped)
```
1) Mantis\tests\Mantis\TokensApiTest::testTokenGet
The token must exist.
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 [
-    0 => '13',
-    1 => '1234',
-    2 => '1000',
+    0 => 13,
+    1 => 1234,
+    2 => 1000,
     3 => 'normal_user',
-    4 => '1777569374',
-    5 => '1777572974',
+    4 => 1777569374,
+    5 => 1777572974,
 ]
/home/runner/work/mantisbt/mantisbt/tests/Mantis/TokensApiTest.php:129
```
I thought the ADODb layer was precisely what was needed to hide such differences from the application layer...

tests/Mantis/TokensApiTest.php
dregad: > it works perfectly for MySQL and SQLite 3

Because these drivers follow the default path in the `switch( $g_db_functional_type )`, which calls ADOConnection::Insert_ID() without specifying the table name as it's not required for these databases' native methods to retrieve the last inserted record's Id.

dregad:
> I'm surprised that PostgreSQL returns a string instead of a number... 

I agree it's surprising and even somewhat silly IMO, but we have no control over this as this is what PHP's pgsql extension does, see [pg_fetch_array() documentation](https://www.php.net/manual/en/function.pg-fetch-array.php#refsect1-function.pg-fetch-array-returnvalues) :

*Each value in the [returned] array is represented as a string. Database NULL values are returned as null.*

This behavior is due to the Postgres client protocol and the C API used by PHP ([source](https://stackoverflow.com/questions/8930567/php-and-postgresql-standard-types-in-the-query-result-why-is-it-always-strings)).

dregad:
> I thought the ADODb layer was precisely what was needed to hide such differences from the application layer...

Indeed it is, however it's not perfect some features are not supported in all drivers. 

Also, keep in mind that MantisBT is old 😉 - support for PostgreSQL in db_insert_id() goes back to 2004. At the time it was experimental, and the ADOdb postgres64 driver did not properly support this, hence the workaround implemented in database API.

This should be fixed at some point, but it's outside of this PR's scope.

Anyway, are you still working on it, or is it ready to merge ?

raspopov:
> Anyway, are you still working on it, or is it ready to merge ?

It looks like it's ready. The tests did not reveal any issues with the initial patch. The tests provide only basic coverage of the old and new API functionality.

PR #2211, opened 2026-04-25
Docker Compose Development Setup

wolframite:


dregad:
Thanks for your contribution. 

We don't have much Docker experience in the team, which is the reason there has not been any official setup so far. That being said I am not opposed to merging this, but doing so begs the question of future maintenance, i.e. who will take care of it...

Anyway please open an Issue at https://mantisbt.org/bugs for tracking and reference it here.

PR #2214, opened 2026-05-07
Added openssl requirement to installation documentation

fsa-SM:
Up to this point, `openssl` PHP extension was not mentioned as PHP extension required to be able to send e-mails via PHPMailer's SMTP method.  
  
After some debugging, I found out that openssl extension was simply missing from my setup.  
  
This PR adds mention about requirement of `openssl`, if Mantis is going to use SSL/TLS SMTP servers.

### DIFF
--- a/docbook/Admin_Guide/en-US/Installation.xml
+++ b/docbook/Admin_Guide/en-US/Installation.xml
@@ -227,6 +227,12 @@
 									(see <xref linkend="admin.auth.ldap" />).
 								</para></listitem>
 
+								<listitem><para><emphasis>openssl</emphasis> -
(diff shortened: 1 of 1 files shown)


dregad:
Thanks for your contribution. Change looks good.
Please open an Issue at https://mantisbt.org/bugs to track the change and reference the ID here.

fsa-SM:
Hello @dregad,  
  
thank you for your feedback.  
Issue identifier on MantisBT side is **0037112** (https://mantisbt.org/bugs/view.php?id=37112). 
  
Best regards,
SM

dregad (approved):

PR #2215, opened 2026-05-07
Add a category description.

raspopov:
The description appears as a tooltip when selecting a category on the issue creation page.

A "description" column has been added to the category table.
The database schema has been updated to version 215.

Fixes [#7715](https://mantisbt.org/bugs/view.php?id=7715).

category-description

### DIFF
--- a/admin/schema.php
+++ b/admin/schema.php
@@ -909,6 +909,9 @@ $g_upgrade[213] = array( 'UpdateFunction', 'category_status_default' );
 # Delete orphaned private filters owned by users that no longer exist
 $g_upgrade[214] = array( 'UpdateFunction', 'delete_orphaned_private_filters' );
 
+$g_upgrade[215] = array( 'AddColumnSQL', array( db_get_table( 'category' ), "
+	description				XL	$t_notnull DEFAULT ( '' )" ) );
+
 # ----------------------------------------------------------------------------
 # End of schema definition, clear local variables
 #
--- a/core/category_api.php
+++ b/core/category_api.php
@@ -176,12 +176,13 @@ function category_ensure_can_remove( $p_category_id ) {
 
 /**
  * Add a new category to the project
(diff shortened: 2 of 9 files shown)


raspopov:
🤔 It's not entirely clear why the MySQL test refuses to insert a category using the default value of the ‘description’ column (i.e., an empty string). Even though that's what's specified in the schema. Everything works fine for SQLite 3 and PostgreSQL... There is a theory that this is related to the STRICT_TRANS_TABLES option, but, for example, another column named ‘status’ is inserted by default without any issues.

raspopov:
The test was fixed by explicitly specifying an empty value when adding a category. Killing two birds with one stone. 😥

vboctor:
I'm not a fan of the description bubble. I think this will get in the way of repeat users and can become annoying. I'm also not sure how this will work in touch interfaces (e.g. phone).

vboctor (changes requested):
see earlier comment.

raspopov:
Thanks for the helpful tip about tooltips being practically invisible on mobile devices. I've added the ability to display the category description alongside the drop-down menu.

msedge_2026-05-16_22-02-50

raspopov:
A new variant with Bootstrap Popover:
firefox_2026-06-24_06-30-53
It is currently shown on a focused `<select>`.

raspopov:
Now with Bootstrap Tooltip:
firefox_2026-06-24_18-39-22
Activation on focus (for mobile) and on hover.

vboctor:
@raspopov here are my thoughts on this PR:

- I'm not in favor of anything that pops over.
- I'm ok with adding a category description as a field.
- REST API must have full support for the category description.
- SOAP API must at least not overwrite it. That is unlikely to be an issue.
- Update existing tests to take the category description into account.
- Document / enforce max length

PR #2218, opened 2026-05-10
Fix wrong rendering of strings

atrol:
Fixes [#34465](https://mantisbt.org/bugs/view.php?id=34465)

### DIFF
--- a/account_page.php
+++ b/account_page.php
@@ -154,7 +154,7 @@ print_account_menu( 'account_page.php' );
 					<?php echo lang_get( 'username' ) ?>
 				</td>
 				<td>
-					<?php echo string_display_line( $u_username ) ?>
+					<?php echo string_attribute( $u_username ) ?>
 				</td>
 			</tr>
 			<tr>
@@ -174,7 +174,7 @@ print_account_menu( 'account_page.php' );
 					<?php echo lang_get( 'username' ) ?>
 				</td>
 				<td>
-					<?php echo string_display_line( $u_username ) ?>
+					<?php echo string_attribute( $u_username ) ?>
 				</td>
 			</tr>
 			<?php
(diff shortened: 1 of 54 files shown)


dregad (approved):
OK for me, although Idid not review and test each occurrence individually.

Even though it's effectively the same thing today technically speaking (not counting the one extra function call), I was just wondering why we're using string_attribute() - whose semantic purpose is to display string *in a text box* - all over the place, instead of calling string_htmlspecialchars() directly. Any idea ?

atrol:
> I was just wondering why we're using string_attribute()

Exactly my thoughts when working on this, I don't know if there is/was  a special reason. Maybe @vboctor knows.

I did it this way, as string_attribute() is used much more frequently than string_html_specialchars().
Also in plugins, string_attribute() seems to be the preferred function.
Deprecating one of the functions would be the clean way, but would lead to regressions in (unmaintained) plugins after removing the function.

For better performance, my proposal is to add the code from string_html_specialchars() to string_attribute() and add a comment to both functions that the code must be synced.
Let me know if you agree to that.

dregad:
I don't mind keeping both string_htmlspecialchars() and string_attribute() to be honest - even if technically they are doing the same thing, they offer a semantic difference.

The performance cost of the extra function call is marginal as shown below (0.1s for 1 million calls on my dev box), so I'm not sure it's really worth it to duplicate the code.

```
php > $t = microtime(true); for($i=0; $i<1e6; $i++) string_html_specialchars('Escape this'); echo microtime(true) - $t;
0.52043700218201
php > $t = microtime(true); for($i=0; $i<1e6; $i++) string_attribute('Escape this'); echo microtime(true) - $t;
0.62979006767273
```

I think we can merge the PR as it stands now, it is already quite a big improvement vs using string_display_line().

We can always come back to this later and selectively replace string_attribute() by string_htmlspecialchars() wherever the former is not needed.

And by the way, I'm not even sure we still need the preg_replace() call that was added to string_htmlspecialchars() in 2006 by Alexey Chumakov - the purpose was to fix an issue with code pages (see [#6494](https://mantisbt.org/bugs/view.php?id=6494)), which is not applicable anymore since we now use UTF-8.

PR #2219, opened 2026-05-13
Fix style attributes of HTML tags

raspopov:
Inline styles have been replaced with the `class` attribute, or unnecessary ones have been removed.

- **core/http_api.php**: Removed the `unsafe-inline` CSP header.
- **account_sponsor_page.php**: Fixed the layout by adding a widget.
- **adm_config_report.php**: Removed unnecessary styling; horizontal scrolling will now occur in the parent `<div>`.
- **adm_permissions_report.php, manage_config_email_page.php, manage_config_work_threshold_page.php, manage_config_workflow_page.php, manage_plugin_page.php**: The text centering style has been replaced with a class.
- **admin/install.php, bug_sponsorship_list_view_inc.php**: Removed redundant styling.
- **core/filter_api.php**: Styling replaced with the `hide` class; accordingly, the class toggle has been updated in js/bugFilter.js.
- **core/layout_api.php**: The built-in user font style has been replaced with the external file css/status_config.php; the `layout_user_font_preference()` function has been removed.
- **core/print_api.php**: Image sizing has been moved to css/status_config.php too.
- **plugins/MantisGraph/pages/developer_graph.php, plugins/MantisGraph/pages/reporter_graph.php**: Fixed the markup for the "truncated" widget.
- **print_all_bug_page_word.php**: Added a new `page-break` style to css/default.css.
- **roadmap_page.php**: Added the ability to dynamically set the progress bar width in js/common.js using the `data-dynamic-width` attribute.
- **signup_page.php**: Removed the hack intended to initially hide the image loading indicator, securimage.js needs to be updated.

In all other files, the call `require_css( ‘status_config.php’ );` has been removed, since this file is now always loaded within the `layout_head_css()` function.

Fixes [#21908](https://mantisbt.org/bugs/view.php?id=21908).


core/layout_api.php
```diff
@@ -301,21 +301,17 @@ function layout_head_css() {
...
+ *
+ * @deprecated Use css/mantis.php instead.
  */
 function layout_user_font_preference() {
-	$t_font_family = config_get( 'font_family', null, null, ALL_PROJECTS );
-	$t_font_family = string_html_specialchars( $t_font_family );
-	echo '<style>', "\n";
-	echo  '* { font-family: "' . $t_font_family . '"; } ', "\n";
-	echo  'h1, h2, h3, h4, h5 { font-family: "' . $t_font_family . '"; } ', "\n";
-	echo '</style>', "\n";
```
dregad, line 318: We usually call something like this in deprecated functions.

```php
error_parameters( __FUNCTION__ . '()', 'css/mantis.php' );
```

core/layout_api.php
```diff
@@ -301,21 +301,17 @@ function layout_head_css() {
...
-	layout_user_font_preference();
+	$t_font_family = config_get( 'font_family', null, null, ALL_PROJECTS );
+	html_css_link( 'mantis.php', helper_generate_cache_key( [ 'user' ], $t_font_family ) );
 }
 
 /**
  * Print user font preference
  * @return void
+ *
+ * @deprecated Use css/mantis.php instead.
```
dregad: Indicate the deprecation version (2.29.0) in the PHPdoc tag

css/mantis.php
```diff
@@ -0,0 +1,68 @@
...
+require_once( dirname( __DIR__ ) . '/core.php' );
+require_api( 'config_api.php' );
+require_api( 'http_api.php' );
+
+# Send correct MIME Content-Type header for css content.
+header( 'Content-Type: text/css; charset=UTF-8' );
+
+# Don't let Internet Explorer second-guess our content-type, as per
+# http://blogs.msdn.com/b/ie/archive/2008/07/02/ie8-security-part-v-comprehensive-protection.aspx
+header( 'X-Content-Type-Options: nosniff' );
```
dregad: We stopped supporting IE in 2.22.0 (2019) - this is probably not needed anymore.

dregad (commented):
Excellent initiative, thanks @raspopov 

I think there is some overlap with css/status_config.php, would it make sense to combine with the new css/mantis.php ?

Speaking of which, I would suggest to rename it to `mantis_css.php` so it's more obvious what it is (from a dev's standpoint, when you have many files open in your IDE).

I'm not sure you've covered all usages of style in the code base. Off the top of my head there is at least one occurrence in roadmap (to display the progress bar)

core/layout_api.php
raspopov: It might not be necessary to throw an error here, because the code that has been moved to another function will be executed anyway, and no further action is required.

css/mantis.php
raspopov: Although this header was originally designed for Internet Explorer, all modern browsers support it when implementing MIME-sniffing algorithms, and it is recommended for security reasons (to prevent XSS).

core/layout_api.php
dregad: Yes I get that. The point is more to let the caller know that they should be updating their code.

raspopov: There's also the issue of the “trigger_error” construct itself becoming deprecated in PHP 8.4 and later... 🤔

raspopov:
> I think there is some overlap with css/status_config.php, would it make sense to combine with the new css/mantis.php?

I’ve been thinking about this, and I wasn’t really keen on doing it because of the growing number of files involved. But since you brought it up, I guess I’ll just have to bite the bullet. 🫡 Especially since status_config.php is processed using a hack.

core/layout_api.php
dregad: It's only passing E_USER_ERROR to trigger_error() which is being [deprecated](https://wiki.php.net/rfc/deprecations_php_8_4#deprecate_passing_e_user_error_to_trigger_error). I actually have a local WIP branch to fix that, but I got so sidetracked lately with security issues that I nearly forgot about it (https://mantisbt.org/bugs/view.php?id=36812).

Other error types are still allowed, AFAICT.

raspopov:
I guess I'd better do a rebase so I don't go crazy with this Git... 😔

dregad:
Sorry, master is a moving target at the moment as I'm merging a bunch of PRs (preparing for release 2.29.0 in the next few days). Maybe you'll want to delay that rebase for a bit

raspopov:
There's one last issue left—we need to update the file **vendor/dapphp/securimage/securimage.js**:
```diff
@@ -109,2 +109,3 @@
                 this.loadingImage = el;
+                this.loadingImage.style.display = 'none';
             }
@@ -194,3 +195,3 @@
     var newAudioEl = document.createElement('audio');
-    newAudioEl.setAttribute('style', 'display: none;');
+    newAudioEl.style.display = 'none';
     newAudioEl.setAttribute('preload', 'false');
```
securimage.js needs to be CSP-compliant (e.g. no setAttribute for `style`) and hide the loading indicator on its own.

I see an issue about this problem there: https://github.com/dapphp/securimage/issues/74.

PR #2221, opened 2026-05-15
Seed CSRF login_token on anonymous

khushalVekariya:
Fixes [#37135](https://mantisbt.org/bugs/view.php?id=37135)

## Summary
Fixes CSRF validation failures (`ERROR #2800`) on two GET-redirect login
paths that bypass `login_page.php` — where the `login` form security token
is normally minted.

`form_security_token('login')` both returns a nonce **and** seeds it into
`$_SESSION['form_security_tokens']['login']`; `form_security_validate('login')`
reads `login_token` from GET or POST. Generating the token at the redirect
and passing it in the URL therefore makes downstream validation pass, since
`PHPSESSID` (and thus the session store) survives the redirect.

## Changes

- **`login_anon.php`** — anonymous login redirects directly to `login.php`,
  which calls `form_security_validate('login')`. Seed `login_token` into the
  redirect query string so validation succeeds.

### DIFF
--- a/core/authentication_api.php
+++ b/core/authentication_api.php
@@ -52,6 +52,7 @@ require_api( 'crypto_api.php' );
 require_api( 'current_user_api.php' );
 require_api( 'database_api.php' );
 require_api( 'error_api.php' );
+require_api( 'form_api.php' );
 require_api( 'gpc_api.php' );
 require_api( 'helper_api.php' );
 require_api( 'html_api.php' );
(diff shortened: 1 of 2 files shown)


dregad:
@khushalVekariya thanks for your contribution. Please reference an Issue in our [bugtracker](https://mantisbt.org/bugs/), create one if needed.

dregad:
> - login_anon.php 

OK, I was actually hit by this while testing something else yesterday. 

This one is just a minor inconvenience, as a workaround user can simply access pages directly (e.g. click on the desired link in the side bar).

> - core/authentication_api.php (auth_reauthenticate())

I thought this would be fixed by PR #2220... At least it works in my local test environment AFAICT. Can you provide steps to reproduce ?

khushalVekariya:
> OK, I was actually hit by this while testing something else yesterday.
> This one is just a minor inconvenience, as a workaround user can simply access pages directly (e.g. click on the desired link in the side bar).

@dregad  Agreed on the low severity — direct navigation auto-authenticates anonymously, so users still have a workaround. Keeping the fix though, since the "Login Anonymously" link is UI-exposed and should behave correctly.

> I thought this would be fixed by PR #2220... At least it works in my local test environment AFAICT. Can you provide steps to reproduce ?

Reverted.

dregad:
I see you have created [#37135](https://mantisbt.org/bugs/view.php?id=37135), thanks. For next time, please reference it in the PR (when opening it).

> Keeping the fix though, since the "Login Anonymously" link is UI-exposed and should behave correctly.

Of course. I'm not saying it should not be fixed, just that it's not urgent like #2220.

> > I thought this would be fixed by PR #2220... At least it works in my local test environment AFAICT. Can you provide steps to reproduce ?
> 
> Reverted.

Thanks for confirming. I was wondering if there was another use case which I missed while testing.

PR #2222, opened 2026-05-15
Fix the string_sanitize_url() lost slash.

raspopov:
Fixed an issue where the “/” was missing when rebuilding URLs during sanitization.

Fixes [#37133](https://mantisbt.org/bugs/view.php?id=37133).


dregad (approved):
LGTM

vboctor (approved):

PR #2225, opened 2026-05-20
Add display of all plugin dependencies

raspopov:
Fixes [#37144](https://mantisbt.org/bugs/view.php?id=37144).

Before:
before

After:
after

### DIFF
--- a/lang/strings_english.txt
+++ b/lang/strings_english.txt
@@ -1809,3 +1809,4 @@ $s_filter_new = 'New';
 $s_filter_unused_tags = 'Unused';
 $s_filter_unused = 'Unused';
 $s_unknown = 'unknown';
+$s_optional = 'Optional';
--- a/css/default.css
+++ b/css/default.css
@@ -50,6 +50,10 @@ span.dependency_met			{ color: green; }
 span.dependency_unmet		{ color: red; }
 span.dependency_upgrade		{ color: orange; }
 
+.label.multiline {
+	margin-top: .3em;
+}
+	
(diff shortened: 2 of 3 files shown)


raspopov:
firefox_2026-05-24_11-41-34

lang/strings_english.txt
```diff
@@ -1809,3 +1809,4 @@ $s_filter_new = 'New';
 $s_filter_unused_tags = 'Unused';
 $s_filter_unused = 'Unused';
 $s_unknown = 'unknown';
+$s_optional = 'Optional';
```
dregad, line 1812: Use lowercase `o`

dregad (changes requested):
See my note at https://mantisbt.org/bugs/view.php?id=37144#c71156

I don't like the implementation with the $t_deps_names array.

I would suggest introducing a private helper method in AvailablePlugin class to handle the dependencies: 
- parameters: array (_requires_ or _uses_ property) and a bool (true for optional), 
- body: the if() block after the `# Dependencies` comment.

Then the main block in the constructor can call the method twice, one for each dependency type, and the else becomes an if checking whether _dependencies_ property is empty.

lang/strings_english.txt
raspopov: But what if we need that word for, say, a table header?

dregad: I'm afraid we're not very consistent with this. That's a fair point, leave it as-is.

dregad:
Merged in 3298e1ffdc496d1f546c79343c22fe2b153e8dd8

PR #2226, opened 2026-05-20
[MantisGraph] Combine all JS scripts into a single function

raspopov:
The loading of the `MantisGraph.js` file has been moved from the `resources()` function to the `include_chartjs()` function.
The request to load scripts by the Mantis Graph plugin itself is now placed earlier — in the `csp_headers()` function.
The redundant `resources()` function — the `EVENT_LAYOUT_RESOURCES` event handler — has been removed.
Now all scripts are loaded in order and in the same way.

Fixes [#37146](https://mantisbt.org/bugs/view.php?id=37146).

### DIFF
--- a/core/html_api.php
+++ b/core/html_api.php
@@ -130,7 +130,14 @@ function html_rss_link() {
  * @return void
  */
 function html_javascript_link( $p_filename ) {
-	echo "\t", '<script src="', helper_mantis_url( 'js/' . $p_filename ), '"></script>', "\n";
+	$t_filename = $p_filename;
+
+	# If no path is specified, look for JavaScript files in default directory
+	if( $t_filename == basename( $t_filename ) ) {
+		$t_filename = 'js/' . $p_filename;
+	}
+
+	echo "\t", '<script src="', helper_mantis_url( $t_filename ), '"></script>', "\n";
 }
 
 /**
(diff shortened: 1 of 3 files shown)


dregad (commented):
One preliminary remark: with this refactoring, the modified functions' names no longer reflect their actual purpose.

I suggest renaming them as appropriate.

raspopov:
I wonder why I can't finish what I set out to do right away without your input? After all, the whole point of this and the original patch was to tighten the CSP rules... 😔

dregad:
It's OK, that's precisely why we have code reviews 😉 

Thanks for your efforts and contributions, I really appreciate the collaboration too.

dregad (commented):
I tested this PR with the [GaugeSupport](https://github.com/mantisbt-plugins/GaugeSupport) plugin. **The change introduces a regression**, chart.js is not loaded.

To make it work I had to hook EVENT_CORE_HEADERS to call include_chartjs() (in replacement of the previously used EVENT_LAYOUT_RESOURCES), as recommended in the method's PHPDoc.

The plugin's own javascript can then either be loaded from EVENT_CORE_HEADERS (using require_js() function as modified in PR #2185), or in EVENT_LAYOUT_RESOURCES handler with the usual script tag (as before).

Such a change in 3rd-party plugins is acceptable of course, but do you think there is a way to make this backwards-compatible ? 

If not the BC-break should be clearly documented, and if possible to somehow detect that include_chartjs() is called too late, throw an error.

raspopov:
> The plugin's own javascript can then either be loaded from EVENT_CORE_HEADERS (using require_js() function as modified in PR https://github.com/mantisbt/mantisbt/pull/2185), or in EVENT_LAYOUT_RESOURCES handler with the usual script tag (as before).

Unfortunately, I don't know how to do this yet. 🤔
There are several things we would like to accomplish simultaneously:
1) Prevent MantisGraph from adding the CSP header to pages where it isn't used.
2) Allow MantisGraph.js to be used on pages from other plugins as well.
3) Maintain compatibility i.e. ensure that include_chartjs() still works in EVENT_LAYOUT_RESOURCES.

Gotta think about it...

raspopov:
I tried to work around this issue by allowing calls to `require_js()` and `require_css()` relatively late in the process; their last chance is the `EVENT_LAYOUT_RESOURCES` handler.

I had to revert the CSP configuration in MantisGraph to its original state, but I added a TODO so we don’t forget to tighten the policy when the opportunity arises. Now compatibility with GaugeSupport is maintained, and my plugin also gets a free script from MantisGraph.

PR #2227, opened 2026-05-25
Allow bug_link_tag prefix in jump to bug form

dregad:
Fixes [#37134](https://mantisbt.org/bugs/view.php?id=37134)

### DIFF
--- a/core/layout_api.php
+++ b/core/layout_api.php
@@ -1103,18 +1103,30 @@ function layout_breadcrumbs() {
 		}
 	}
 
-	# Bug Jump form
+	layout_bug_jump_form();
+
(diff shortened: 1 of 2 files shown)


atrol:
This will not work for the rare case where `bug_link_tag` of the current project is different from `bug_link_tag` of the project where the bug you are searching for belongs to.
What about simply stripping any non-numeric characters from the entered string?

dregad:
Thanks for the review. Good idea, I was probably over-engineering this...

dregad:
@atrol I updated the code per your suggestion. Let me know what you think.

atrol (approved):

PR #2228, opened 2026-05-30
Drop user_print_pref table and related unused code

dregad:
- Removed unused code
- Removed unused language strings
- Drop table in schema.php
- Updated Entity-Relationship diagram

## Reference
Fixes [#19370](https://mantisbt.org/bugs/view.php?id=19370)


atrol (approved):

vboctor (approved):
Should we delete just the English language and translation cascades to other languages or delete from all languages?

dregad:
> Should we delete just the English language and translation cascades to other languages or delete from all languages?

It's just a simple sed command to remove the 2 strings everywhere (`$s_printing_preferences_title` and `$s_printing_options_link`) and it does not cause issues with TranslateWiki.net so I think it's cleaner to remove them everywhere right away instead of waiting for weeks for the change to eventually cascade.

PR #2230, opened 2026-06-10
Fix SQL in news_get_rows()

raspopov:
Fixed regression in b78f966e013d494ea27d92bba39d797a1266b7b7.

Fixes #[37250](https://mantisbt.org/bugs/view.php?id=37250).

### DIFF
--- a/core/news_api.php
+++ b/core/news_api.php
@@ -203,18 +203,21 @@ function news_get_rows( $p_project_id, $p_global = true ) {
 		$t_projects[] = ALL_PROJECTS;
 	}
 
+	db_param_push();
+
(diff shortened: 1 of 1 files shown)


vboctor (approved):

dregad (changes requested):
Inlining variables is not good practice. I would suggest to take the opportunity of using a query parameter instead.

Either use the legacy db_param() (quick and easy change) or even better refactor to use DbQuery class.

dregad (approved):

dregad:
Merged in 960c146476d25be16cac5c2f195e1d02b012a2ea

PR #2231, opened 2026-06-16
Fix incorrect log message

raspopov:
Fixes [#37256](https://mantisbt.org/bugs/view.php?id=37256).


dregad (approved):
Personally I would simply have changed single to double quotes, but that's fine. Thanks !

dregad:
Merged in e7e19efeccc12d7810c79293b9090d1ba2eb5043

PR #2252, opened 2026-07-30
Accept optional Bearer prefix on API token

khushalVekariya:
Fixes [#37333](https://mantisbt.org/bugs/view.php?id=37333)

## What

Accept an optional `Bearer ` prefix on the API token in the `Authorization`
header, so both of these are equivalent:

    Authorization: <api-token>
    Authorization: Bearer <api-token>

## Why

`Bearer` is the RFC 6750 convention and the default for virtually every HTTP
client, SDK, API tool, and AI agent. Sending it against MantisBT currently
fails authentication with no hint about the cause, which is a confusing
first-contact experience for anyone integrating with the REST API.

## How

`api_token_parse_credentials()` in `core/api_token_api.php` strips the prefix
before the token lookup:

- scheme matched case-insensitively (`Bearer`, `bearer`, `BEARER`)
- separator is `[ \t]` — SP or HTAB only, per RFC 7230, deliberately not `\s`
- surrounding whitespace trimmed
- any other scheme returned unchanged, so `Basic ...` behaves exactly as before
- the bare token form is untouched — this is purely additive

`AuthMiddleware` runs each credential through the helper, and the stripped
token is what reaches both `api_token_get_user()` and `mci_check_login()`.

Nothing new is logged; the token never reaches an error message.


core/api_token_api.php
```diff
@@ -137,6 +137,30 @@ function api_token_name_ensure_unique( $p_token_name, $p_user_id ) {
...
+ *
+ * @param string $p_credentials The Authorization header value.
+ * @return string The token, trimmed and without the 'Bearer ' prefix.
+ * @access public
+ */
+function api_token_parse_credentials( $p_credentials ) {
+	$t_credentials = trim( (string)$p_credentials );
+
+	# [ \t] not \s: RFC 7230 allows only SP or HTAB after a scheme.
+	if( preg_match( '/^Bearer[ \t]+(.*)$/i', $t_credentials, $t_matches ) ) {
```
vboctor: Based on the RFC, only spaces should be accepted and not tabs.

api/rest/mantisbt_openapi.yaml
```diff
@@ -955,6 +960,9 @@ components:
       type: apiKey
```
vboctor: Switch this to type: http, scheme bearer. This will indicate via OpenAPI spec that this is a bearer standard authentication rather than an arbitrary header.

vboctor (changes requested):
Thanks @khushalVekariya 

Since now we are compliant with the RFC, we should return `WWW-Authenticate: Bearer` response header when we return `401 Valid API token required`. This is required by the RFC.

Please also update the postman collection and Api_Rest.xml to reflect the bearer format in the API token.

vboctor (approved):

PR #2254, opened 2026-08-04
Fix project-specific stored query access

vboctor:
Fixes #[37341](https://mantisbt.org/bugs/view.php?id=37341)


dregad (approved):
Look good, did not actually test.

PR #2256, opened 2026-08-05
Code Cleanup: Fix cases where existing commands were not used

vboctor:
[#37362](https://mantisbt.org/bugs/view.php?id=37362)


dregad (changes requested):
No problem with the changes to the code.

Please create an issue listing the fixed occurrences to document things, is so it shows on the change log. The Issue should then be referenced either in the merge commit, or in the branch's individual commits.

PR #2257, opened 2026-08-06
change validation error check to fit the function's return types

przemyslawWiatr:
Fixes [https://mantisbt.org/bugs/view.php?id=37343](https://mantisbt.org/bugs/view.php?id=37343)

Alters the conditions for validation to be counted as failed


dregad:
This change is not correct. Please read https://mantisbt.org/bugs/view.php?id=37343#c71344

PR #2260, opened 2026-08-22
REST API: Move Issue API + command

vboctor:
Fixes [#37352](https://mantisbt.org/bugs/view.php?id=37352)

- Implements IssueMoveCommand and leverages it to expose a new REST API.
- Change Web UI to leverage the new command.
- Add an trigger `EVENT_MOVE_BUG` event after the move.


dregad (commented):
See [my note](https://mantisbt.org/bugs/view.php?id=37352#c71365) on the tracker

PR #2261, opened 2026-08-22
Reject multiline issue summaries when creating via API

vboctor:
Fixes [#26984](https://mantisbt.org/bugs/view.php?id=26984)

This fixes new occurrences of the issue. Due to this being an unlikely issue to happen, I didn't add a db migration step.


dregad (approved):
Considering that a simple Edit / Save operation from MantisBT UI will strip the newline, I'm OK with not providing an upgrade step or admin check.

PR #2262, opened 2026-08-22
REST API: Add delete issue attachment API

vboctor:
Fixes [#34425](https://mantisbt.org/bugs/view.php?id=34425)

- Add command to delete attachment
- Add to REST API
- Use from SOAP API
- Use from Web UI


dregad (approved):
LGTM

PR #2270, opened 2026-08-30
REST API: Link/Unlink/Re-order Project Custom Fields

vboctor:
Fixes [#32449](https://mantisbt.org/bugs/view.php?id=32449)

### DIFF
--- a/manage_proj_custom_field_copy.php
+++ b/manage_proj_custom_field_copy.php
@@ -27,10 +27,10 @@
  * @uses authentication_api.php
  * @uses config_api.php
  * @uses constant_inc.php
+ * @uses custom_field_api.php
  * @uses form_api.php
  * @uses gpc_api.php
  * @uses print_api.php
- * @uses project_api.php
  */
 
 require_once( 'core.php' );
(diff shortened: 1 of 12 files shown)


manage_proj_custom_field_copy.php
```diff
@@ -65,7 +62,22 @@
 	trigger_error( ERROR_NO_COPY_ACTION, ERROR );
 }
 
-project_copy_custom_fields( $t_dst_project_id, $t_src_project_id );
```
dregad, line 68: This was the only usage of project_copy_custom_fields(), remove or deprecate API function ?

dregad (changes requested):
Please resolve merge conflicts.

General note: when you merge a PR with several commits such as this one, instead of using GitHub's rebase+merge option, it would be better first rebase/force-push the branch (if necessary) and then merge with a commit referencing the PR so the changes are visually grouped together in Git history. That makes future troubleshooting easier.

For example, see b83500663e171b59892db4df2eb8688385ce9380 which I committed after you merged #2256 via rebase.

dregad (approved):

PR #2271, opened 2026-08-31
Assert both anonymous login modes in REST auth header tests

khushalVekariya:
[Task](https://mantisbt.org/bugs/view.php?id=37333)

Follow-up to #2252, addressing review feedback on the tests added there.

`testUnauthorizedResponseAdvertisesBearerScheme` was skipped whenever the
instance under test allows anonymous access. Since CI enables it
(`build/ci_post_install.sh` sets `$g_allow_anonymous_login = ON`), the test
never actually ran, leaving the `WWW-Authenticate: Bearer` header untested.

The test now always issues the credential-less request and asserts the
behaviour matching the instance configuration:

| `$g_allow_anonymous_login` | expected |
| --- | --- |
| `ON` | no 401, no `WWW-Authenticate` header |
| `OFF` | 401 with `WWW-Authenticate: Bearer` |

It is renamed to `testAnonymousRequestHonoursAnonymousLoginConfig`, as it is
no longer only about the 401 case. The `ON` arm asserts "not 401" rather than
a fixed 200, so it still holds on instances where the anonymous user cannot
read the endpoint (403).

Note the two scenarios cannot both be exercised in a single run:
`allow_anonymous_login` is listed in `$g_global_settings`, so `config_set()`
will not persist it, and the REST tests drive a separate server process that
would not see an in-process change anyway. Covering both arms by
configuration is the closest equivalent.

Also makes `providerBearerPrefixes()` and `providerMultipleCredentials()`
static, silencing the PHPUnit deprecation notices for non-static data
providers. Happy to drop that part if you have already fixed it locally.

### DIFF
--- a/tests/rest/RestAuthHeaderTest.php
+++ b/tests/rest/RestAuthHeaderTest.php
@@ -77,7 +77,7 @@ class RestAuthHeaderTest extends RestBase {
 	 * Bearer scheme spellings that must all be accepted.
 	 * @return array
 	 */
-	public function providerBearerPrefixes() {
+	public static function providerBearerPrefixes() {
 		return array(
 			'Bearer' => array( 'Bearer ' ),
 			'bearer lowercase' => array( 'bearer ' ),
@@ -109,7 +109,7 @@ class RestAuthHeaderTest extends RestBase {
 	 * Credential separators that must all be tolerated.
 	 * @return array
 	 */
-	public function providerMultipleCredentials() {
+	public static function providerMultipleCredentials() {
 		return array(
 			'comma space' => array( ', ' ),
 			'comma only' => array( ',' ),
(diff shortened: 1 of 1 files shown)


dregad:
Thanks for the PR @khushalVekariya that's much better.

> Note the two scenarios cannot both be exercised in a single run:
allow_anonymous_login is listed in $g_global_settings, so config_set()
will not persist it, and the REST tests drive a separate server process that
would not see an in-process change anyway

Of course it would be best if the test could cover both usage scenarios, but I understand the limitation imposed by $g_allow_anonymous_login being a global setting (i.e. it can only be changed in config_inc.php) so we can't rely on config API like we do in other usage scenarios (e.g. RestUserRest).

@vboctor do you have any idea how we could achieve that ? I believe such mechanism could be useful for other testing scenarios, allowing better coverage for unit tests.

PR #2278, opened 2026-09-09
Fix return value plugin_lang_get_defaulted when no default supplied

SL-Gundam:
## Reference
Fixes [#37379](https://mantisbt.org/bugs/view.php?id=37379)

### DIFF
--- a/core/plugin_api.php
+++ b/core/plugin_api.php
@@ -456,12 +456,10 @@ function plugin_lang_get_defaulted( $p_name, $p_default = null, $p_basename = nu
 	if( !is_null( $p_basename ) ) {
 		plugin_push_current( $p_basename );
 	}
-	if( is_null( $p_default ) ) {
-		$p_default = $p_name;
-	}
+	$t_default = ( ( is_null( $p_default ) ) ? $p_name : $p_default );
 	$t_basename = plugin_get_current();
 	$t_name = 'plugin_' . $t_basename . '_' . $p_name;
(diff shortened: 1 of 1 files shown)


core/plugin_api.php
```diff
@@ -456,9 +456,10 @@ function plugin_lang_get_defaulted( $p_name, $p_default = null, $p_basename = nu
 	if( !is_null( $p_basename ) ) {
 		plugin_push_current( $p_basename );
 	}
+	$t_default = ( ( is_null( $p_default ) ) ? $p_name : $p_default );
 	$t_basename = plugin_get_current();
 	$t_name = 'plugin_' . $t_basename . '_' . $p_name;
-	$t_string = lang_get_defaulted( $t_name, $p_default );
+	$t_string = lang_get_defaulted( $t_name, $t_default );
```
dregad: I would suggest an alternate fix - no need for $t_default variable and ternary operator
```php
$t_string = lang_get_defaulted( $t_name, $p_default ?? $p_name );
```

What do you think ?

dregad (changes requested):

core/plugin_api.php
SL-Gundam: Both work and give the desired result. If this is within the MantisBT coding style, i'm all for it.
Made the change as suggested.

dregad (approved):

dregad:
Thanks for your contribution

PR #2279, opened 2026-09-10
Display memory/file sizes with bigger, dynamic units

dregad:
Fixes [#34949](https://mantisbt.org/bugs/view.php?id=34949)

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image |


atrol:
Why did you remove "Maximum Size"?
grafik

dregad:
Thanks for the review. I only meant to remove it from the tooltip, I must have made a mistake somewhere. I'll check the code.

dregad:
I forgot to use the $p_format parameter in the new get_size_info_span() function. Fixed.

image

core/print_api.php
```diff
@@ -2220,13 +2220,116 @@ function print_timezone_option_list( $p_timezone ) {
...
+ * @param string|null $p_unit Optional unit language string (bytes, KiB, MiB, GiB),
+ *                            or null for automatic selection based on file size.
+ *
+ * @return string[] 0-based array containing formatted file size and unit.
+ *                  If unit > bytes, array index keys 2 & 3 will hold the
+ *                  formatted size in bytes and localized unit.
+ * @throws UnexpectedValueException Invalid file size unit.
+ */
+function get_size_info_array( int $p_size, ?string $p_unit = null ):array {
+	$t_units = ['bytes', 'kib', 'mib', 'gib'];
```
vboctor, line 2243: Why not use KB, MB, and GB which are more commonly used

core/print_api.php
```diff
@@ -2220,13 +2220,116 @@ function print_timezone_option_list( $p_timezone ) {
...
+ * @param int         $p_size File or memory size, in bytes.
+ * @param string|null $p_unit Optional unit language string (bytes, KiB, MiB, GiB),
+ *                            or null for automatic selection based on file size.
+ *
+ * @return string[] 0-based array containing formatted file size and unit.
+ *                  If unit > bytes, array index keys 2 & 3 will hold the
+ *                  formatted size in bytes and localized unit.
+ * @throws UnexpectedValueException Invalid file size unit.
+ */
+function get_size_info_array( int $p_size, ?string $p_unit = null ):array {
```
vboctor, line 2242: Should this be in `core/helper_api.php`?

vboctor, line 2242: I suggest adding some test cases to make sure this method works as expected.

lang/strings_qqq.txt
```diff
@@ -403,7 +403,13 @@ $s_filter_resolved = '{{Identical|Resolved}}';
 $s_advanced_filters = '[[File:MantisBT-filter.png|150px|right]]
 Link on bug report filter screen. See screenshot.';
 $s_attachments = '{{Identical|Attachment}}';
-$s_kib = 'IEEE 1541 unit kibibytes (1024 bytes), see https://en.wikipedia.org/wiki/Kibibyte
+$s_kib = 'IEEE 1541 unit kibibytes (1024 bytes), see https://en.wikipedia.org/wiki/Byte#Multiple-byte_units
```
vboctor, line 406: I'm no tsure why this is not showing a unit and showing some links, explanation, etc.

vboctor (changes requested):

core/print_api.php
dregad: Windows (being the most widely used OS) shows file sizes in multiples of 1024. Linux does too, at least on cli (eg. with `-h` flag for ls) - I no longer have my old Ubuntu dev box to what the GUI does - so we show the same size as most OS (sorry for MacOS users :wink:).

I think we've had this discussion before when I standardized on IEEE units ([#27700](https://mantisbt.org/bugs/view.php?id=27700)).

core/print_api.php
dregad: I put the new functions in print_api.php because the original *get_filesize_info()* function was already there.

I can move the function to helper API, if you feel it would be better. But in that case shouldn't *get_size_info()* and *get_size_info_span()* be moved too ?

lang/strings_qqq.txt
dregad: TranslateWiki's `qqq` "language" is just an indication to help translators do their job, it is never used in our UI.

core/print_api.php
atrol: > Windows (being the most widely used OS) shows file sizes in multiples of 1024

Right, but for the labels in Windows File Explorer and the file property dialog they use KB, MB, ... instead of KiB, MiB, ...
The size of a file of 1.000.000 bytes is shown as 977 KB in file explorer and 976 KB in file properties dialog.

grafik

core/print_api.php
vboctor: The way I think about it is `helper` is for misc utility methods. Where `print` is for functions that output UI/html elements. Ideally, print APIs are wrappers around logic functions that live elsewhere. And also over time, other API files don't have methods that print html/UI elements. i.e. separation of UI from logic.

core/print_api.php
dregad: @atrol I know, and that's just plain wrong. KB = kilobyte = 1000 bytes. That's misleading, and the scientist in me just can't accept that. Using IEEE 1541 units is the only right thing to do IMO.

I never heard a single complain over our use of KiB since I introduced them over 5 years ago. 

And for the record, this PR's purpose is not to discuss the choice of unit, but to introduce the use of dynamic, larger ones.

core/print_api.php
dregad: OK I'm cool with that. I'll move them to helper.

core/print_api.php
seanm: Please please please don't use "KB" and "MB" if you are using base 2. It's just wrong. (It's a shame MS doesn't fix this in Windows, but I guess Hyrum’s Law applies.)

vboctor: You may be scientifically correct, but I don't see many companies advertising file sizes, memory, or SSD storage or phone storage in gib. I personally don't care much how we calculate it - so we can keep as is to change the calculation. But before we add more strings and localize them, I think we should use the user friendly units that are used by industry leaders when addressing end users.

seanm: As @dregad said "this PR's purpose is not to discuss the choice of unit, but to introduce the use of dynamic, larger ones."  Mantis is already using base 2, and using the correct prefixes.  The correct base 2 prefixes *are* the user friendly ones, because being correct is user friendly. More and more software is slowly but surely using base 2 prefixes when they are using base 2, example: https://wiki.ubuntu.com/UnitsPolicy

vboctor: Unfortunately, a side effect of the change is adding two more strings that will be localized in N languages. It doesn't make sense to go back and revisit it later.

seanm: Mantis 2.28.4 is already using the `KiB` string. Adding `MiB` and `GiB` are trivial to translate, as the prefixes are never translated, and only the `B` suffix is maybe translated, and the translation is presumably already known since `KiB` existed already.

vboctor: I suggest using [SI standard](http://en.wikipedia.org/wiki/International_System_of_Units) for base-10 units:

1 kB = 1,000 bytes (Note: small k)
1 MB = 1,000 kB = 1,000,000 bytes
1 GB = 1,000 MB = 1,000,000 kB = 1,000,000,000 bytes
1 TB = 1,000 GB = 1,000,000 MB = 1,000,000,000 kB = 1,000,000,000,000 bytes

seanm: That would indeed be correct if base 10 was being used. But the mantis project decided on using base 2 here: https://mantisbt.org/bugs/view.php?id=27700

That decision could be revisited, but isn't really related to this PR.

PR #2282, opened 2026-09-13
Cache invalidation fixes

vboctor:
Fixes [#37387](https://mantisbt.org/bugs/view.php?id=37387)


dregad (commented):
I have no objections with this overall, but I did not test.

The only thing I would challenge, is patching and introducing tests for sponsorship API, which is [officially deprecated since 1.3.0](https://mantisbt.org/blog/archives/mantisbt/440), although we never actually removed it . I'm pretty sure we also discussed it elsewhere but I can't find a written trace right now. In any case, it's effectively broken from a UI perspective since 2.0 (see https://mantisbt.org/bugs/view.php?id=21138)

@vboctor you did not set a target version - since we have a pretty much stable scope for 2.29.0, I'd rager aim for 2.30.0. In any case, please do not merge to master just now to avoid messing with the release branch I'm working on locally - I'd rather avoid having to rebase it.

PR #2284, opened 2026-09-16
Send 304 from plugin_file.php without a gzip body

brlumen:
Fixes [#37393](https://mantisbt.org/bugs/view.php?id=37393)

`plugin_file_include()` answers a matching `If-Modified-Since` with `304 Not Modified` and no output, but `compress_handler_is_enabled()` has already turned on `zlib.output_compression`, and PHP's zlib output handler emits a gzip stream (20 bytes) even for an empty buffer. The 304 therefore goes out with `Content-Encoding: gzip` and a body, which RFC 9110 forbids. Apache and nginx silently drop it; Caddy (Go `net/http`) aborts the response, so behind Caddy every plugin CSS/JS asset fails with `ERR_HTTP2_PROTOCOL_ERROR` as soon as the browser starts revalidating its cache (`max-age=10800`, i.e. about three hours after the first visit).

Reproduce:

```
curl --http2 -H 'Accept-Encoding: gzip' \
     -H 'If-Modified-Since: <Last-Modified of the file>' \
     https://mantis.example/plugin_file.php?file=Calendar/calendar_filter.js
```

Expected `304` with an empty body; actual `curl: (92) HTTP/2 stream 1 was not closed cleanly`. Without `Accept-Encoding` the same request is fine. Talking to php-fpm directly (`cgi-fcgi -bind -connect …`) shows the gzip bytes after the headers.

Fix: `compress_disable()` only cleared `$g_compression_started`, which does nothing for the `zlib.output_compression` path. It now also switches that off while the headers are still unsent, and `plugin_file_include()` calls it before sending the 304. Turning `zlib.output_compression` off at runtime keeps already-buffered output and flushes it uncompressed, so the existing caller in `error_api.php` is unaffected.

Verified on MantisBT 2.28.4 / PHP 8.4 php-fpm / Caddy: the 304 is now sent with an empty body, a stale `If-Modified-Since` still returns 200 with the file, and regular pages are still gzip-compressed.


dregad (approved):

dregad:
Thanks for your contribution, PR looks good; I'll merge it later in my 2.29.0 release preparation branch.

